PR #1, opened 2019-12-13
Improve the err message when submgr not available

bocekm:


convert2rhel/subscription.py
```diff
@@ -123,14 +123,16 @@ def hide_password(cmd):
...
-    """Install subscription-manager RPM and its dependencies bundled with this
-    tool.
-    """
+    """Install subscription-manager RPM and its dependencies."""
     loggerinst = logging.getLogger(__name__)
     sm_dir = os.path.join(utils.data_dir, "subscription-manager")
-    if not os.path.exists(sm_dir):
-        loggerinst.critical("The {0} directory does not exist. Consider using"
-                            " the --disable-submgr option.".format(sm_dir))
+    if not os.path.exists(sm_dir) or not os.listdir(sm_dir):
```
pirat89: In case it has to be directory, wouldn't be better to check whether it is dir directly?
```suggestion
    if not os.path.isdir(sm_dir) or not os.listdir(sm_dir):
```

That solution could be problematic in case of symlink.

PR #6, opened 2019-12-20
make: install missing deps for srpm build

bocekm:


.copr/Makefile
```diff
@@ -1,8 +1,10 @@
-mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
-mkfile_dir := $(dir $(mkfile_path))
+mkfile_dir := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 repo_root := $(mkfile_dir)/../
+yum_or_dnf := $(shell command -v dnf || command -v yum)
 
 srpm:
+	@echo --- Installing SRPM build requirements ---
+	@$(yum_or_dnf) install -y git python2 python2-setuptools rpmlint rpm-build
```
pirat89: @bocekm I would rather keep that part as separate target - e.g. install_build_deps or something like that. And keep it as deps for srpm target.

bocekm: https://github.com/oamg/convert2rhel/pull/9/files

PR #10, opened 2020-01-13
help/man: variant option for submgr only

bocekm:
- the variant option has no meaning if used with --disable-submgr option
- it is supposed to be used when converting with subscription-manager


bocekm:
@pirat89, how so?

PR #20, opened 2020-02-19
Fix copr build name with multiple git branches available

bocekm:


packaging/epel/create_epel_srpm.sh
```diff
@@ -20,7 +20,8 @@ cp ${SCRIPTPATH}/convert2rhel.spec convert2rhel.spec
 rpmlint convert2rhel.spec
 
 TIMESTAMP=`date +%Y%m%d%H%MZ -u`
-GIT_BRANCH=$(git for-each-ref --format='%(objectname) %(refname:short)' refs/heads | awk "{print \$2}")
+GIT_BRANCH=$(git for-each-ref --format='%(objectname) %(refname:short)' refs/heads | head -n1 | awk "{print \$2}")
+echo GIT_BRANCH: ${GIT_BRANCH}
```
pirat89: @bocekm  I am not sure that you really want this change (or commit msg & titles are wrong?). Basically, you are telling that you have more identical branches in the repository, which doesn't make sense so much in our cases. You have master and some branches - where for every branch you have new specific commit (head) from the others. If you go into the state you are now, basically the GIT_BRANCH looses sense as you go into the situation, that you are searching for branch A - e.g. is used for reference in the PR - but you can get branch B - so you never find what you are looking for using the branch name.

I think you should rather go with the approach, that repository is not messed by branches with identical heads. I can imagine situation where you would like to have branches e.g. master & stable, and in some point in the stable branch you would get two reference then. But in such case, there should be used clear priority e.g. for *stable* branch explicitly. In other cases I would raise an error as it looks valid from my POV.

pirat89: If you think it's really good approach, please describe that more clearly why so we can think about that in other projects as well. But currently I do not see the good here now.

bocekm: Unfortunately I removed the [copr](https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/builds/) build which failed without this fix.
But basically I was requesting copr manually to create a build for my private branch in my fork https://github.com/bocekm/convert2rhel. And the `GIT_BRANCH=$(git for-each-ref --format='%(objectname) %(refname:short)' refs/heads | awk "{print \$2}")` left the `GIT_BRANCH` empty, just because there were multiple lines returned by the `git for-each-ref` command.

This is what the Copr is doing:
```
$ git clone https://github.com/bocekm/convert2rhel.git /tmp/copr-rpmbuild-9p0cjwyk/obtain-sources/convert2rhel --depth 500 --no-single-branch
Cloning into '/tmp/copr-rpmbuild-9p0cjwyk/obtain-sources/convert2rhel'...
...
$ cd /tmp/copr-rpmbuild-9p0cjwyk/obtain-sources/convert2rhel
$ git checkout add_missing_spec_deps
Branch 'add_missing_spec_deps' set up to track remote branch 'add_missing_spec_deps' from 'origin'.
Switched to a new branch 'add_missing_spec_deps'
$ git for-each-ref --format='%(refname:short)' refs/heads
add_missing_spec_deps
master
```

I'm using the `git for-each-ref` command to [avoid getting](https://github.com/oamg/convert2rhel/pull/16) `HEAD` instead of `master`.

pirat89: @bocekm  Thanks for explanation. Now I've got it. It looks that the command actually is not doing what you really want at all, as it will print all local branches all the time. Now you will get just one branch when you use the `head`, but the one that is alphabetically the first in the list, so I guess the command is not doing really what you want. See:

```
$ git for-each-ref --format='%(objectname) %(refname:short)' refs/heads 
541746d2e424edcf7b6e089ed661333a233554a5 __REBASE_HELL_GOTCHA__
8cd7fd0e0bf59c916b327b8d5ae2de913ed81074 add_checkminimumrequirements_actor
c0b35c75622dbfe87a473d1947b830a15eeca4e4 add_deps_to_unit_file
11b7a531f8364b732604fd1293254988dbf6cb6f add_supplementary_to_pes_mapping
608043d0214bf98295a90295de7f14e8d0875100 altarch_geninitrd
bb7615f162ad2c513882fba9ebc8accc3ba33f45 beta_devel
345a7f5b631e821cc39008fe07bf8405b22af236 blacklisting-enabling-crb
55ff838a6bbb8ca0f7d1d1297ae6132870d5bd22 bump_master_release
c377b222c31da898d5d73c3a9f7114aff177ecfe check_s390x
0476163885c0f7460c12146f925f4a0193d4f4d4 coprcli-makefile-fix
eed28aab9a63316f3bdeac66863f0b4d21bac199 cups_filters_migration
2550e72f7c0333a0ab0ae698cd8156d6e9e0653b dm_no_udev
960fd3f167d66192ecd0f632388bb1d08691e1d4 efi_boot_order_fix
6aa10e6600bde0ace6991139db9e6683ff410529 enable77
6222f2c8f5ef2d85eab2fd6b1fe7df22d6788ae0 enable_s390x
df33c3b9f5b87504f57f9d8661507c5ded7c1444 fix-lint-py3
c775bc2e4e81808abc4eaaa25160975bba982fef fix_eth_inhibit
62e75ee958ff6071fe9935be8324e540a187d157 fix_finding_copr_builds
ca824ed27512433421da92e9263ec5d2308b0878 fix_makefile
95316792c57a5c1647abbf4058c8dbcb2e5ea5f4 fix_tmp_get_copr_build
5c4e7b30f24fce547d34b9ff121a9a1c6d794154 fix_tz_resolv
3582538710c17cf670f72c12516e711771ba055e inhibit_znet
365c504c78b49dfa2c95097040b1faf7783cfce4 kurna/fix
4b349b0a58962c18e6af89c6bec194e1cebb71c9 master
672c0abb7787adc0734415ff42417355e244dfa2 new_ipu_path
a70a5832a5dc9327e9b9e5df30c7982b79d23265 no_rhsm_in_nspawn
fef774aa0faf02dc24afaca54d831696e4991d4f ntp-migration
2a74ada9beade3645f690e920b45e4f29f7d2de9 opensshups_tests
25041f52a7ea000db196bf550b5d88a3c572f6aa packaging_chroot
28499d398e4c1c1ea4d79429fa1e668979b713ad pkg_version_deps
76598854d8c47ce00ad8a084e62f2d1a1d283431 python_inform_user
4c0012ffaf7f873617f886201bbc3086ce14a4d3 rename_boot_entry
80604b26d787ede37773051c7bfe9dd545936931 repomap-repoids
05845ca5467983aa14dade65e81b350167f633a8 report_refactoring
1cd67291cdcbb4ffe3b7f6e90201a619d0c94863 repositories_mapping
4722938923c37e758c034c8aa95130eab73a3312 revert_pr301
e020ec52e0888fd5f2ccd94629a6a726ba78974f rhel8-userspace
30edfd3c117117a249d480ea56c2c7a56f1d1c58 rhel8-userspace-fix
3c540ee5883a285077358c36a802e3b0361eeb35 rhsm_experiment
669dc041fc8c750f99759905bd14a14f5f5fc0c4 rm_checkznet
8166451d80ddc050064b8f8f7151984e803055e6 s390x_fixes
bdbdc2ab2414d8ab50d6e059c55d47ae0b295747 selinux
5f89297c351466ec15d6df0b42f5df5021ee92e5 sendmail
297989f422e737768cddbcaecceca8822f0a1f4c spamassassin
8ad02651bbd7c5f5fd3b4b75e6b0f58d50a5b665 targetuserspacegen-logging-cleanup
ccb1426baf4cff106ef65b21831e84d046f96b54 verifydialogs
24715b99b0d9d063cccbc078738da92c85cac468 verifydialogsnicerreport
9a602ed43434c440a51153bca6834e56382d27a6 vim_migration
```

If I am right about what you want (just get the actual branch name), use rather something like this:
```
GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
```

bocekm: @pirat89, ah, I see. You're right, it really does not do what I want :)

Unfortunately the `git rev-parse --abbrev-ref HEAD` is not doing what I want either. I had it there before: https://github.com/oamg/convert2rhel/pull/16/files#diff-38d93dcb3b68f5151d2d1b93a738e00cL23. It was giving me `HEAD` on master.

I fear that the only thing that would save me is proper CI, using Jenkins or Packit. Not mere using Makefile for Copr webhooks.

pirat89: If you want, we can sync about that later and come up with working solution. I am interested about the issue.

PR #28, opened 2020-03-25
Enable CI unit-tests

sturivny:


sturivny:
@bocekm Working on it

bocekm:
@sturivny, please rebase.

sturivny:
@bocekm please review.

.packit.yaml
```diff
@@ -16,11 +16,13 @@ jobs:
     - epel-7-x86_64
   trigger: pull_request
 - job: tests
-  trigger: pull_request
   metadata:
+    owner: "@oamg"
+    project: convert2rhel
```
bocekm, line 21: Is this necessary for tests? It seems like it's just for copr.

convert2rhel/main.py
```diff
@@ -92,7 +92,7 @@ def main():
         # restart system if required
         utils.restart_system()
 
-    except (Exception, SystemExit, KeyboardInterrupt), err:
+    except (Exception, SystemExit, KeyboardInterrupt) as err:
```
bocekm, line 95: Using "," is necessary for python 2.4 compatibility (RHEL 5). I believe we don't need the code to be Py3-compatible as of yet. Maybe later if we go for conversions to RHEL 8.

convert2rhel/pkghandler.py
```diff
@@ -196,7 +195,6 @@ def get_installed_pkg_objects(name=""):
     """Return list with installed package objects. The packages can be
     optionally filtered by name.
     """
-    yum_base = yum.YumBase()
```
bocekm, line 199: Is there any problem with the yum module? Without this line, the `yum_base` will be uninitialized.

sturivny: Returned it back as it was

convert2rhel/main.py
sturivny: Returned it back as it was

.packit.yaml
sturivny: Removed it

sturivny:
@bocekm Thanks for reviewing :)
Updated PR

sturivny:
> It looks good. But could be please undelete the _convert2rhel/unit_tests/pkghandler_test.py_ file?

Done :)

PR #29, opened 2020-05-04
Add coverage checks

sturivny:


bocekm:
Pylint installation fails on the CentOS 6 and 7. Actually I remember I was trying to deal with that a long time ago, October 2016: https://github.com/bocekm/pylint/commits/py26_compatibility. Back then it was python 2.7+/3.x compatible. I asked the upstream if they could make it 2.6 compatible. But they didn't want to.

So we need to find another way. @vojtechsokol, in which environment do we install pylint to check the Leapp code?

sturivny:
@bocekm, @vojtechsokol
Please, check my PR https://github.com/oamg/convert2rhel/pull/30
I think we can use GitHub actions and also we can add badges. 
Like this:
PyLint Runner

The `issue` that it runs it on ubuntu. But we can run docker inside if we need it.
What do you think?

bocekm:
You can switch the coverage on, but there's this error right now in the testing farm results:
`nose.plugins.cover: ERROR: Coverage not available: unable to import coverage module`

Then, change the title of the PR. We'll deal with the pylint check under #30.

vojtechsokol:
@bocekm Sorry for the late reply.
For leapp-repository Travis tests run in centos:7 container with pylint and pytest installed in python2.7 and python3.6 virtual environments. Version of pytest is fixed at 3.6.4 because of some issues that I don't remember anymore.
And as for leapp, Travis tests runs both in default Travis' environment (some Ubuntu) and also in centos:7 and centos:8 containers. Again everything is installed in virtual environment (python 2.7, 3.6, 3.8 in Ubuntu, python 2.7 in centos:7 and python 3.8 in centos:8 containers) pytest is again fixed at 3.6.4 and pytest-pylint at 0.14.1.

bocekm:
Let's go the https://github.com/oamg/convert2rhel/pull/37 way.

PR #31, opened 2020-05-05
fix: Install Red Hat GPG keys

SpyTec:
Fixes `gpgcheck=yes` fatally failing conversion. It now imports GPG keys from c2r.

Unit tests pass and manually tested on CentOS 7.


bocekm:
I'd say you can drop the `RPM-GPG-KEY-redhat-legacy-former` - it was used for signing packages up to RHEL 4, not for RHEL 5+.

convert2rhel/pkghandler.py
```diff
@@ -365,6 +369,21 @@ def preserve_only_rhel_kernel():
...
+    gpg_path = os.path.realpath(os.path.join(os.path.dirname(__file__),
+                                "data/version-independent/gpg-keys/*"))
+    for gpg_key in glob.glob(gpg_path):
+        shutil.copy(gpg_key, _rpm_gpg_dir)
+        output, ret_code = utils.run_subprocess(
+            'rpm --import %s' % os.path.join(_rpm_gpg_dir, gpg_key),
+            print_output=False)
+        if ret_code != 0:
+            loggerinst.critical("Unable to copy GPG key: %s" % output)
+            return
```
sturivny, line 384: This method `install_gpg_keys()` returns None in both cases if ret_code != 0 and if ret_code == 0.
Is it correct behavior?

convert2rhel/pkghandler.py
```diff
@@ -24,11 +24,14 @@
...
 from convert2rhel import utils
 from convert2rhel.toolopts import tool_opts
+import glob
+import shutil
 
 # Limit the number of loops over yum command calls for the case there was
 # an error.
 MAX_YUM_CMD_CALLS = 2
 
+_rpm_gpg_dir = "/etc/pki/rpm-gpg/"
```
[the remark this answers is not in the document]
bocekm: Yeah, currently the place used for paths like this is here: https://github.com/oamg/convert2rhel/blob/master/convert2rhel/utils.py#L46. You can put it there or create a config as Serhii suggests.

convert2rhel/pkghandler.py
bocekm: Actually, calling the critical logger causes exit of the tool: https://github.com/oamg/convert2rhel/blob/master/convert2rhel/logger.py#L100
It's quite unexpected, I know. We can change that. But for now, the `return` here can be removed.

convert2rhel/utils.py
```diff
@@ -57,6 +58,14 @@ def require_root():
         loggerinst.critical("The tool needs to be run under the root user.")
     return
 
+def safe_repr(obj, short=False):
```
bocekm, line 61: Please add a docstring about what's the purpose of the function.
Also, because it's just used for unit tests, I'd move it to https://github.com/oamg/convert2rhel/blob/master/convert2rhel/unit_tests/__init__.py.

convert2rhel/pkghandler.py
```diff
@@ -365,6 +369,21 @@ def preserve_only_rhel_kernel():
     return
 
 
+def install_gpg_keys():
+    loggerinst = logging.getLogger(__name__)
+    loggerinst.info("Importing Red Hat GPG keys ...")
+    gpg_path = os.path.realpath(os.path.join(os.path.dirname(__file__),
+                                "data/version-independent/gpg-keys/*"))
+    for gpg_key in glob.glob(gpg_path):
+        shutil.copy(gpg_key, _rpm_gpg_dir)
```
bocekm, line 378: Right now I'm not sure if copying the keys is a good idea, because the redhat-release package contains the keys as well and installs them to the same location. Let me check if it does not cause a conflict during the installation of redhat-release.

convert2rhel/pkghandler.py
```diff
@@ -365,6 +369,21 @@ def preserve_only_rhel_kernel():
...
+        output, ret_code = utils.run_subprocess(
+            'rpm --import %s' % os.path.join(_rpm_gpg_dir, gpg_key),
+            print_output=False)
+        if ret_code != 0:
+            loggerinst.critical("Unable to copy GPG key: %s" % output)
```
bocekm, line 383:
```suggestion
            loggerinst.critical("Unable to import GPG key: %s" % output)
```
The `ret_code` holds the result of importing, not copying the key.

convert2rhel/pkghandler.py
SpyTec: Would you think it's better to just import the keys right from c2r instead of copying over and importing?

bocekm: Yeah, I think it would be better. Probably, when somebody uninstalls convert2rhel later and thus the keys won't be available, it should not matter because during the conversion the redhat-release package gets installed and it contains these keys.

SpyTec:
Rebased with master and addressed the comments:

- Dropped `legacy-former` GPG key
- Moved `install_gpg_keys()` into main
- Removed copying GPG keys to `/etc/pki/rpm-gpg/` - now we just import it from c2r directory
- Re-ordered imports in alphabetical order

Tested once again in a Vagrant box and successfully converted with `gpgcheck=1`

bocekm:
Interestingly, the packit build is not available anymore.

convert2rhel/pkghandler.py
```diff
@@ -365,6 +366,18 @@ def preserve_only_rhel_kernel():
     return
 
 
+def install_gpg_keys():
+    loggerinst = logging.getLogger(__name__)
+    gpg_path = os.path.realpath(os.path.join(os.path.dirname(__file__),
+                                "data/version-independent/gpg-keys/*"))
```
bocekm, line 372: I wonder if this works, I'll test it out shortly. But if I get it right, the gpg keys are moved to /usr/share/convert2rhel/ folder during the RPM installation (you can use `utils.data_dir` to get this dir here).

convert2rhel/pkghandler.py
```diff
@@ -365,6 +366,18 @@ def preserve_only_rhel_kernel():
...
+    for gpg_key in glob.glob(gpg_path):
```
bocekm, line 373:
```suggestion
    gpg_keys = [os.path.join(gpg_path, x) for x in os.listdir(gpg_path)]
    for gpg_keys:
```
Similarly to:
https://github.com/oamg/convert2rhel/blob/master/convert2rhel/subscription.py#L137

convert2rhel/pkghandler.py
SpyTec: Seems to work if we update tests to have a mocked directory. Commiting a change now

bocekm:
Please rebase and resolve conflicts.

SpyTec:
> Please rebase and resolve conflicts.

Done and done

bocekm:
I'm sorry but the branch needs an additional rebase. I swear I'll re-approve and merge right away.

SpyTec:
> I'm sorry but the branch needs an additional rebase. I swear I'll re-approve and merge right away.

Too many pylint fixes :stuck_out_tongue: 

Rebased

PR #33, opened 2020-05-06
Bump version to 0.10

bocekm:
- fixed rpm dependencies
- blacklisted kmod-kvdo causing a transaction failure on CentOS 7
- convert2rhel exits with 0 on a help message
- added packit configuration for Copr builds and unit testing on a PR

Resolves: OAMG-3165

### DIFF
--- a/packaging/centos_ol/convert2rhel.spec
+++ b/packaging/centos_ol/convert2rhel.spec
@@ -3,7 +3,7 @@
 %{!?python2_sitearch: %global python2_sitearch %(%{__python2} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib(1))")}
 
 Name:           convert2rhel
-Version:        0.9
+Version:        0.10
 Release:        1%{?dist}
 Summary:        Automates the conversion of RHEL derivative distributions to RHEL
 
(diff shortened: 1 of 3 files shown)


bocekm:
@fellipeh, they could probably be merged. But I'm unsure of two things:

1. [Fedora packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_spec_legibility) say:

    > "To` help facilitate legibility, only macros and conditionals for Fedora and EPEL are allowed to be used in Fedora Packages. Use of macros and conditionals for other distributions, including Fedora derivatives, is not permitted in spec files of packages in the main Fedora repositories unless those macros and conditionals are also present in Fedora."

   - Are EPEL repositories among the main Fedora repositories? I would say so. Then I can't use the `{rhel}` macro in the spec for EPEL.

1. [The EPEL:Packaging](https://fedoraproject.org/wiki/EPEL:Packaging#Limited_Arch_Packages) guidelines say:

   > Make sure the package is not shipped for all architectures. The valid architectures are:
   > - EPEL6: i686, x86_64
   > - EPEL7: ppc64le, x86_64.
   > - EPEL8: Because we have had problems with these packages causing either build, compose, or user problems, we are not allowing this for EPEL8.

   - But currently the convert2rhel is possible to build on i386 OL5 and ppc64 (big endian) CentOS7. I don't know how would I handle this arch conflict in one spec.

Maybe @pirat89 has an idea?

PR #40, opened 2020-05-07
Fix packit conf on proposing a downstream release

bocekm:
For the downstream release proposal, packit configuration requires dist_git_branches instead of dist-git-branch: https://packit.dev/docs/configuration/#packit-service-jobs

The incorrect key most probably caused that the proposal https://src.fedoraproject.org/rpms/convert2rhel/pull-request/1# has targeted Fedora rawhide instead of EPEL.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -24,6 +24,6 @@ jobs:
 - job: propose_downstream
   trigger: release
   metadata:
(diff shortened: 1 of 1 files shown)


sturivny:
LGTM

PR #41, opened 2020-05-07
Always log debug info to the log file

fellipeh:
Now debug info is stored inside log file, whatever the user use or not
`--debug` argument.

Fixed when user not pass `--debug` information, the info was passed to
stdout, now these info only passed to the stdout if user add the debug
argument.

### DIFF
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
-'''
+"""
 Customized logging functionality
 
 CRITICAL  (50)    Calls critical() function and sys.exit(1)
@@ -24,12 +24,15 @@ INFO      (20)    Prints info message (no date/time, just plain message)
 TASK      (15)    CUSTOM LABEL - Prints a task header message (using asterisks)
 DEBUG     (10)    Prints debug message (using date/time)
 FILE      (5)     CUSTOM LABEL - Prints only to file handler (using date/time)
-'''
+"""
+from convert2rhel.toolopts import tool_opts
 import logging
-import sys
 import os
+import sys
+from utils import format_msg_with_datetime
 
 LOG_DIR = "/var/log/convert2rhel"
+
 
 class LogLevelTask:
     level = 15
@@ -42,10 +45,10 @@ class LogLevelFile:
 
 
 def initialize_logger(log_name):
-    '''Initialize custom logging levels, handlers, and so on. Call this method
+    """Initialize custom logging levels, handlers, and so on. Call this method
     from your application's main start point.
         log_name = the name for the log file
-    '''
+    """
     # check if already initialized with custom class
     if logging.getLoggerClass() == CustomLogger:
         return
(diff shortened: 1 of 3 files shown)


convert2rhel/logger.py
```diff
@@ -101,10 +103,13 @@ def critical(self, msg, *args, **kwargs):
 
     def debug(self, msg, *args, **kwargs):
         from convert2rhel.toolopts import tool_opts
+        from utils import format_msg_with_datetime
```
sturivny, line 106: Let's stick to the PEP8 style. 
`Imports are always put at the top of the file`
https://www.python.org/dev/peps/pep-0008/#imports

convert2rhel/utils.py
```diff
@@ -46,6 +46,13 @@ class color:
 tmp_dir = "/tmp/convert2rhel/"
 
 
+def format_msg_with_datetime(msg, level):
+    """Returna string with msg formatted according to the level"""
```
sturivny, line 50: typo: `return a string`

convert2rhel/utils.py
```diff
@@ -46,6 +46,13 @@ class color:
...
+    import datetime
```
sturivny, line 51: import ditto

convert2rhel/utils.py
```diff
@@ -206,9 +213,8 @@ def log_traceback(debug):
     if debug:
         # Print the traceback to the user when debug option used
         loggerinst.debug(traceback_str)
-    else:
-        # Print the traceback to the log file in any way
-        loggerinst.file(traceback_str)
+    # Print the traceback to the log file in any way
+    loggerinst.file(traceback_str)
```
ferdnyc, line 217: Combined with the change to `CustomLogger.debug()`, won't this give you double messages in the logfile when `tool_opts.debug` is false but `debug` here is true?

When `loggerinst.debug()` is called, it checks `tool_opts.debug`, and if that's `False` now it will automatically log with `LOG_LEVEL_FILE.level` as a fallback. Then this (now-)unconditional `loggerinst.file(traceback_str)` will log the same message a second time at `LOG_LEVEL_FILE.level`, unless I'm not understanding the code properly.

convert2rhel/utils.py
fellipeh: nice catch!

convert2rhel/utils.py
fellipeh: you're right. fixed

fellipeh:
Rebased and squashed
@bocekm could you please take a look to merge it?

bocekm:
@fellipeh, unit tests seems to have failed:
```
File "/__w/convert2rhel/convert2rhel/convert2rhel/unit_tests/logger_test.py", line 124, in test_log
    self.assertEqual(logging.Handler.handle.called, 6)
AssertionError: 7 != 6
```

bocekm:
Please rebase and resolve the conflict. It looks good. I've tested it and it works fine. After the rebase, I'll approve right away and merge.

fellipeh:
rebased

bocekm:
The conflict resolution didn't end up well:
`NameError: global name 'LOG_LEVEL_FILE' is not defined`

fellipeh:
rebased

bocekm:
Wow, the testing farm is so slow. Waiting for it 45 mins already.

PR #44, opened 2020-05-12
Fix CI pylint check

fellipeh:
pylint check on our CI doesn't use the `.pylintrc` configuration file.
These commit fix that

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -24,7 +24,7 @@ jobs:
         
     - name: Run Pylint Checks
       run: |
-        pylint convert2rhel/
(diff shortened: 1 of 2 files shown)


fellipeh:
> Indeed, the use of the rcfile option was missing, thanks @fellipeh. Because it's enabled now, the py2.6 pylint check is failing with a traceback. Let's try to figure out why's that.
> EDIT: Try moving the comments in https://github.com/oamg/convert2rhel/blob/master/.pylintrc to separate lines.

These PR is suposed to fix only pylintrc argument on o.pylint.yml so github can use the pylintrc file... We still need to fix a lot of lint errors.. :)

PR #45, opened 2020-05-13
Fix: Require --enablerepo with --disable-submgr

SpyTec:
If `--enablerepo` is empty when providing `--disable-submgr`, the tool will fail to function correctly and go past point of no recovery. This PR fixes that by requiring `--enablerepo` if `--disable-submgr` is given

Few more things:

* `--disablerepo` defaults to `*` if `--disable-submgr` is given
* Added `.vscode/` dir to `.gitignore`

It passes the new tests, but I've not tested it manually

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ SRPMS/
 *.pyc
 system_tests/vmdefs/centos*/.vagrant/
 *.copr.conf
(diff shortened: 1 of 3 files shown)


bocekm:
Looks great. Thanks.

PR #47, opened 2020-05-15
Add RHN Classic checker

fellipeh:
If RHN Classic is present,  conert2rhel will raise a CRITICAL message
and abort the process.

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -70,6 +70,7 @@ def main():
         loggerinst.task("Prepare: Backup System")
         redhatrelease.system_release_file.backup()
         redhatrelease.yum_conf.backup()
+        rhelvariant.rhn_reg_file.backup()
 
         # begin conversion process
         process_phase = ConversionPhase.PRE_PONR_CHANGES
@@ -145,6 +146,10 @@ def pre_ponr_conversion():
     loggerinst.task("Convert: Remove blacklisted packages")
     pkghandler.remove_blacklisted_pkgs()
 
+    # checking RHN Classic
+    loggerinst.info("Checking RHN Classic")
+    subscription.unregister_from_rhn_classic()
+
     # install redhat release package
     loggerinst.task("Convert: Install Red Hat release package")
     redhatrelease.install_release_pkg()
@@ -210,6 +215,7 @@ def rollback_changes():
     loggerinst.warn("Abnormal exit! Performing rollback ...")
     utils.changed_pkgs_control.restore_pkgs()
     redhatrelease.system_release_file.restore()
+    rhelvariant.rhn_reg_file.restore()
     redhatrelease.yum_conf.restore()
     subscription.rollback_renamed_repo_files()
     return
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -147,7 +147,7 @@ def pre_ponr_conversion():
     pkghandler.remove_blacklisted_pkgs()
 
     # checking RHN Classic
-    loggerinst.info("Checking RHN Classic")
+    loggerinst.task("Checking RHN Classic")
     subscription.unregister_from_rhn_classic()
 
     # install redhat release package
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -70,7 +70,7 @@ def main():
         loggerinst.task("Prepare: Backup System")
         redhatrelease.system_release_file.backup()
         redhatrelease.yum_conf.backup()
-        rhelvariant.rhn_reg_file.backup()
+        subscription.rhn_reg_file.backup()
 
         # begin conversion process
         process_phase = ConversionPhase.PRE_PONR_CHANGES
@@ -215,7 +215,7 @@ def rollback_changes():
     loggerinst.warn("Abnormal exit! Performing rollback ...")
     utils.changed_pkgs_control.restore_pkgs()
     redhatrelease.system_release_file.restore()
-    rhelvariant.rhn_reg_file.restore()
+    subscription.rhn_reg_file.restore()
     redhatrelease.yum_conf.restore()
     subscription.rollback_renamed_repo_files()
     return
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -35,6 +35,7 @@ class PkgWFingerprint(object):
     """Tuple-like storage for the RPM object of a package and a fingerprint
     with which the package was signed.
     """
+
     def __init__(self, pkg_obj, fingerprint):
         self.pkg_obj = pkg_obj
         self.fingerprint = fingerprint
@@ -469,8 +470,7 @@ def replace_non_rhel_installed_kernel(version):
         loggerinst.critical("Unable to download %s from RHEL repository" % pkg)
         return
 
-    loggerinst.info(
-        "Replacing %s %s with RHEL kernel with the same NEVRA ... " % (system_info.name, pkg))
+    loggerinst.info("Replacing %s %s with RHEL kernel with the same NEVRA ... " % (system_info.name, pkg))
     output, ret_code = utils.run_subprocess(
         'rpm -i --force --replacepkgs %s*' % os.path.join(utils.TMP_DIR, pkg),
         print_output=False)
--- a/convert2rhel/rhelvariant.py
+++ b/convert2rhel/rhelvariant.py
@@ -59,6 +59,7 @@ def print_supported_variants():
     for index, variant in enumerate(get_supported_variants()):
         index += 1
         loggerinst.info("%s) %s" % (index, variant))
+    return
 
 
 def get_supported_variants():
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -23,6 +23,21 @@ import logging
 from convert2rhel.toolopts import tool_opts
 from convert2rhel import utils
 
+_RHN_REGISTRATION_FILE = "/etc/sysconfig/rhn/systemid"
+rhn_reg_file = utils.RestorableFile(_RHN_REGISTRATION_FILE)  # pylint: disable=C0103
+
+
+def unregister_from_rhn_classic():
+    loggerinst = logging.getLogger(__name__)
+    if os.path.isfile(_RHN_REGISTRATION_FILE):
+        loggerinst.warning("The use of RHN Classic is not allowed during the conversion.\n"
+                           "The convert2rhel is going to unregister from RHN Classic.\n"
+                           "See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/installation_guide/unregister-rhn for details.")
+        utils.ask_to_continue()
+        rhn_reg_file.remove()
+    else:
+        loggerinst.info("RHN Classic not detected.")
+
 
 def subscribe_system():
     """Register and attach a specific subscription to OS."""
--- a/convert2rhel/unit_tests/subscription_test.py
+++ b/convert2rhel/unit_tests/subscription_test.py
@@ -16,7 +16,9 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 # Required imports:
+import os
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
+
 try:
     import unittest2 as unittest  # Python 2.6 support
 except ImportError:
@@ -28,7 +30,6 @@ from convert2rhel.toolopts import tool_opts
 
 
 class TestSubscription(unittest.TestCase):
-
     class GetAvailSubsMocked(unit_tests.MockFunction):
         def __call__(self, *args, **kwargs):
             return [{'name': 'sample',
@@ -91,6 +92,7 @@ class TestSubscription(unittest.TestCase):
     class GetLoggerMocked(unit_tests.MockFunction):
         def __init__(self):
             self.info_msgs = []
+            self.warning_msgs = []
             self.critical_msgs = []
 
         def __call__(self, msg):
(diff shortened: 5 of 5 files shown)


convert2rhel/main.py
```diff
@@ -145,6 +145,10 @@ def pre_ponr_conversion():
     loggerinst.task("Convert: Remove blacklisted packages")
     pkghandler.remove_blacklisted_pkgs()
 
+    # checking RHN Classic
+    loggerinst.task("Convert: Checking RHN Classic")
+    pkghandler.check_rhn_classic()
```
SpyTec, line 150: Since this just checks if the system is on RHN Classic it doesn't make much sense to have this in `pre_ponr_conversion` since it's not doing any tasks. I'd suggest we move this just under `rhelvariant.determine_rhel_variant()`

convert2rhel/pkghandler.py
```diff
@@ -500,3 +501,14 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
...
+
+def check_rhn_classic():
+    loggerinst = logging.getLogger(__name__)
+    if os.path.isfile("/etc/sysconfig/rhn/systemid"):
+        loggerinst.critical("RHN Classic detected! \n"
+                            "Please follow the bellow link for more information how to disable it:\n"
+                            "https://access.redhat.com/documentation/en-us/red_hat_subscription_management/1/html/migrating_from_rhn_classic/index"  # pylint: disable=C0321
+                            "\n\n Convertion will not proceed.")
+
+    return
```
sturivny, line 514: I think `return` here is unnecessary

SpyTec: We should be consistent with everywhere else, code still reaches the return so we should keep it

fellipeh:
> Had one issue with code placement, otherwise code looks good. Could you also add a test as well to cover `check_rhn_classic()` functionality?

@SpyTec tests added on last push.

convert2rhel/rhelvariant.py
```diff
@@ -18,11 +18,16 @@
...
+
 from convert2rhel import utils
 
 _supported_variants = None
 
 
+def check_rhn_classic():
+    return os.path.isfile("/etc/sysconfig/rhn/systemid")
+
+
```
SpyTec, line 30: This can be a private function, since we won't use it elsewhere in the app. That way we don't need to write tests for it and can test in `determine_rhel_variant()` instead

Convention would be `_check_rhn_classic()`

convert2rhel/unit_tests/rhnclassic_test.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
```
SpyTec, line 1: It's better to keep names of test files match the files they're testing in, so rename this to `rhelvariant_test.py`

convert2rhel/unit_tests/rhnclassic_test.py
```diff
@@ -0,0 +1,45 @@
...
+            self.path = path
+            return self.return_value
+
+    @unit_tests.mock(os.path, "isfile", os_is_file_mock(False))
+    def test_returns_successfully_if_rhn_classic_does_not_exist(self):
+        self.assertEqual(rhelvariant.check_rhn_classic(), False)
+
+    @unit_tests.mock(os.path, "isfile", os_is_file_mock(True))
+    def test_throws_if_rhn_classic_exists(self):
+        self.assertEqual(rhelvariant.check_rhn_classic(), True)
```
SpyTec, line 45: Mock LGTM, but lets make `check_rhn_classic()` private, drop these tests, and write tests for `determine_rhel_variant()`. Then we can make sure that we get an exception raised by asserting

`self.assertRaises(SystemExit, rhelvariant.determine_rhel_variant)`

convert2rhel/unit_tests/rhnclassic_test.py
```diff
@@ -0,0 +1,45 @@
...
+from convert2rhel import utils
+from convert2rhel import rhelvariant
+
+try:
+    import unittest2 as unittest  # Python 2.6 support
+except ImportError:
+    import unittest
+
+
+class TestUtils(unittest.TestCase):
```
SpyTec, line 29: If file is renamed:
```suggestion
class TestRhelVariant(unittest.TestCase):
```

convert2rhel/rhelvariant.py
bocekm: I think that in general, we should write unit tests for any function regardless of being private or public. We want to ensure that each function does what it's supposed to.
However, in this case, a unit test wouldn't make much sense - you would be just testing a standard library.

convert2rhel/unit_tests/rhnclassic_test.py
bocekm: I agree, currently the test files are named after the code files. I'd stick to that unless we have a good reason for not to.

convert2rhel/rhelvariant.py
fellipeh: Agree. Changing to private function, and removing the unit test

convert2rhel/rhelvariant.py
```diff
@@ -33,6 +38,15 @@ def determine_rhel_variant():
...
+    # checking RHN Classic
+    loggerinst.info("Checking RHN Classic")
+    if _check_rhn_classic():
+        loggerinst.critical("RHN Classic detected! \n"
+                            "Please follow the bellow link for more information how to disable it:\n"
+                            "https://access.redhat.com/documentation/en-us/red_hat_subscription_management/1/html/migrating_from_rhn_classic/index"  # pylint: disable=C0321
+                            "\n\n Convertion will not proceed.")
+
     return
```
bocekm, line 52:
```suggestion
rhn_reg_file = utils.RestorableFile(_RHN_REGISTRATION_FILE)  # pylint: disable=C0103
```

Also, add `rhelvariant.rhn_reg_file.backup()` to https://github.com/oamg/convert2rhel/blob/master/convert2rhel/main.py#L72.
And add `rhelvariant.rhn_reg_file.restore()` to https://github.com/oamg/convert2rhel/blob/master/convert2rhel/main.py#L212.

In another PR, we need to refactor how we do these backups and restores.

convert2rhel/rhelvariant.py
```diff
@@ -33,6 +38,15 @@ def determine_rhel_variant():
...
         tool_opts.variant = _user_to_choose_rhel_variant()
     loggerinst.info("Variant: %s" % tool_opts.variant)
+
+    # checking RHN Classic
+    loggerinst.info("Checking RHN Classic")
+    if _check_rhn_classic():
+        loggerinst.critical("RHN Classic detected! \n"
+                            "Please follow the bellow link for more information how to disable it:\n"
+                            "https://access.redhat.com/documentation/en-us/red_hat_subscription_management/1/html/migrating_from_rhn_classic/index"  # pylint: disable=C0321
+                            "\n\n Convertion will not proceed.")
```
bocekm, line 48:
```suggestion
        loggerinst.warning("The use of RHN Classic is not allowed for the conversion.\n"
                           "Unregistering from RHN Classic by removing %s.\n"
                           "See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/installation_guide/unregister-rhn for details." % _RHN_REGISTRATION_FILE)
        rhn_reg_file.remove()
                           
```

SpyTec:
> I thought about it more and I think it would be better if we just remove the particular file with a warning (causing unregister from the Oracle's RHN Classic) and continue with the conversion. WDYT?

If we went that route I'd rather prompt the user about removing it.

bocekm:
> > I thought about it more and I think it would be better if we just remove the particular file with a warning (causing unregister from the Oracle's RHN Classic) and continue with the conversion. WDYT?
> 
> If we went that route I'd rather prompt the user about removing it.

Good point. I've added the prompt to my proposal: https://github.com/oamg/convert2rhel/pull/47#discussion_r430448084

fellipeh:
@bocekm ready to go with these PR?

convert2rhel/rhelvariant.py
```diff
@@ -18,9 +18,16 @@
...
 import logging
+
 from convert2rhel import utils
 
 _SUPPORTED_VARIANTS = None
+_RHN_REGISTRATION_FILE = "/etc/sysconfig/rhn/systemid"
+rhn_reg_file = utils.RestorableFile(_RHN_REGISTRATION_FILE)  # pylint: disable=C0103
+
+
+def _check_rhn_classic():
```
bocekm, line 29:
```suggestion
def _is_rhn_classic_used():
```

convert2rhel/rhelvariant.py
```diff
@@ -33,6 +40,17 @@ def determine_rhel_variant():
     if not tool_opts.variant:
         tool_opts.variant = _user_to_choose_rhel_variant()
     loggerinst.info("Variant: %s" % tool_opts.variant)
+
+    # checking RHN Classic
```
bocekm, line 44: Actually, the detection of RHN Classic has nothing to do with determining RHEL variant. Please move the code from `rhelvariant.py` to `subscription.py`. Add this piece of code to a separate function, like `unregister_from_rhn_classic`. And call that from `main.py`, for example right before calling `pkghandler.remove_blacklisted_pkgs()`.

convert2rhel/rhelvariant.py
```diff
@@ -33,6 +40,17 @@ def determine_rhel_variant():
...
+    loggerinst.info("Checking RHN Classic")
+
+    if _check_rhn_classic():
+        loggerinst.warning("The use of RHN Classic is not allowed during the conversion.\n"
+                           "The convert2rhel is going to unregister from RHN Classic.\n"
+                           "See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/installation_guide/unregister-rhn for details.")
+        utils.ask_to_continue()
+        rhn_reg_file.remove()
```
bocekm, line 52:
```suggestion
        rhn_reg_file.remove()
    else:
        loggerinst.info("RHN Classic not detected.")
```

bocekm:
@fellipeh, do you need any help with finishing this PR?

fellipeh:
Sorry @bocekm last friday I finished my setup after fedora32 install, and unable to run the test at time.. I'll send the changes right now!

fellipeh:
Made all changes, and rebased

convert2rhel/rhelvariant.py
bocekm: > And call that from main.py, for example right before calling pkghandler.remove_blacklisted_pkgs().

Not resolved ^

convert2rhel/rhelvariant.py
```diff
@@ -35,6 +35,16 @@ def determine_rhel_variant():
         tool_opts.variant = _user_to_choose_rhel_variant()
     loggerinst.info("Variant: %s" % tool_opts.variant)
 
+    # checking RHN Classic
+    loggerinst.info("Checking RHN Classic")
+
+    from convert2rhel import subscription
+    if subscription.is_rhn_classic_used():
```
bocekm, line 42:
```suggestion
    if subscription.is_rhn_classic_used():
```
Please move this condition to `unregister_from_rhn_classic()`.

convert2rhel/subscription.py
```diff
@@ -23,6 +23,22 @@
...
 
+_RHN_REGISTRATION_FILE = "/etc/sysconfig/rhn/systemid"
+rhn_reg_file = utils.RestorableFile(_RHN_REGISTRATION_FILE)  # pylint: disable=C0103
+
+
+def is_rhn_classic_used():
+    return os.path.isfile(_RHN_REGISTRATION_FILE)
+
+
+def unregister_from_rhn_classic():
```
bocekm, line 34: Create a unit test(s) for this function, as I wrote in the previous review.

fellipeh:
> @bocekm when it comes to `utils.RestorableFile()`, I checked the code and it will give a warning if the file doesn't exist. Most of the time the RHN Classic file will not exist, do we want to add an argument for changing the warning to a debug if file doesn't exist?

I believe we do not need that, since if file doesn't exist will return `RHN Classic not detected`  https://github.com/oamg/convert2rhel/pull/47/files#diff-26e87aec779179309df7bc4c2645ad8aR39

SpyTec:
> I believe we do not need that, since if file doesn't exist will return `RHN Classic not detected` [#47 (files)](https://github.com/oamg/convert2rhel/pull/47/files#diff-26e87aec779179309df7bc4c2645ad8aR39)

I meant specifically with `backup()`, since it is called before we will check and possibly remove https://github.com/oamg/convert2rhel/pull/47/files#diff-d8778fdc183f4f4465c18f9e6c1fa12bR73

https://github.com/oamg/convert2rhel/blob/1e3c8637f5db0814f16c8089ffeebb35310566ee/convert2rhel/utils.py#L393-L395

convert2rhel/main.py
```diff
@@ -70,6 +70,7 @@ def main():
         loggerinst.task("Prepare: Backup System")
         redhatrelease.system_release_file.backup()
         redhatrelease.yum_conf.backup()
+        rhelvariant.rhn_reg_file.backup()
```
bocekm, line 73:
```suggestion
        subscription.rhn_reg_file.backup()
```

convert2rhel/main.py
```diff
@@ -210,6 +215,7 @@ def rollback_changes():
     loggerinst.warn("Abnormal exit! Performing rollback ...")
     utils.changed_pkgs_control.restore_pkgs()
     redhatrelease.system_release_file.restore()
+    rhelvariant.rhn_reg_file.restore()
```
bocekm, line 218:
```suggestion
    subscription.rhn_reg_file.restore()
```

bocekm:
@SpyTec, showing the message isn't bad - at least people see transparently what the tool does. I guess we agree on that. But yeah, the `warning` log level is probably too high. I'd probably not go all the way down to `debug`, I think `info` is ok. But let's not change it in this PR.

I'll create a PR for that and two other things:

1. creating the temp dir shouldn't be happening in quite unrelated function https://github.com/oamg/convert2rhel/blob/master/convert2rhel/systeminfo.py#L66 but rather in the spec file so that the dir is owned by the rpm
2. Mentioning the dir in `WARNING - Can't find /etc/sysconfig/rhn/systemid or /tmp/convert2rhel/` is quite strange.

convert2rhel/main.py
```diff
@@ -145,6 +146,10 @@ def pre_ponr_conversion():
     loggerinst.task("Convert: Remove blacklisted packages")
     pkghandler.remove_blacklisted_pkgs()
 
+    # checking RHN Classic
+    loggerinst.info("Checking RHN Classic")
```
bocekm, line 150:
```suggestion
    loggerinst.task("Checking RHN Classic")
```

PR #48, opened 2020-05-19
Add linting to the makefile

fellipeh:
Now we can run `make tests` and the images will be mounted before the tests run

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,24 @@
 IMAGE = convert2rhel
 
-.PHONY: all
-all: clean images tests
+all: clean images tests lint
 
-.PHONY: clean
 clean:
 	@rm -rf build/ dist/ *.egg-info .pytest_cache/
 	@find . -name '__pycache__' -exec rm -fr {} +
 	@find . -name '*.pyc' -exec rm -f {} +
 	@find . -name '*.pyo' -exec rm -f {} +
 
-.PHONY: images
 images:
 	@docker build -f Dockerfiles/centos5 -t $(IMAGE)/centos5 .
 	@docker build -f Dockerfiles/centos6 -t $(IMAGE)/centos6 .
 	@docker build -f Dockerfiles/centos7 -t $(IMAGE)/centos7 .
 
-.PHONY: tests
 tests:
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos5 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos6 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos7 ./run_unit_tests.sh
+
+lint:
+	pylint --rcfile=.pylintrc convert2rhel/
+
+.PHONY: all clean images tests


Makefile
```diff
@@ -1,23 +1,21 @@
...
 	@docker build -f Dockerfiles/centos7 -t $(IMAGE)/centos7 .
 
-.PHONY: tests
-tests:
+tests: images
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos5 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos6 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos7 ./run_unit_tests.sh
+
+.PHONY: all clean images tests
```
sturivny, line 21: Please, add new line :)

sturivny: At the end of the file

fellipeh:
> If `make all` or `make images` is called just once is there a reason to run it again? If not, I'm not sure whether this change makes sense as it would just increase the time it takes to start running the tests, even if the container build is cached it still takes up some 5-10 seconds of time.

Now, if you want to run the tests, you need to run 2 commands: 1) `make images`  2) `make tests`  it's normal behaviour to run tests just in one comment `make tests` but for now we need to build the images before the tests, that's the reason to call images on tests, I believe if you already have built your images it's not take more then 3 sec these part.

SpyTec:
> Now, if you want to run the tests, you need to run 2 commands: 1) `make images` 2) `make tests` it's normal behaviour to run tests just in one comment `make tests` but for now we need to build the images before the tests, that's the reason to call images on tests, I believe if you already have built your images it's not take more then 3 sec these part.

I still doesn't make much sense to have `tests` coupled with `images` as they are completely unrelated. Unless we change the images it only needs to be run once and that's it. Even if it was only 3 seconds it still is an unnecessary amount of time.

You can already run everything if you so want by doing `make` or `make all`, which will always run `images` and `tests`, so it comes down to one command. But `tests` themselves will be vastly faster if we already have images and do not need to run Dockerfiles

bocekm:
I'm on the fence here. I understand that it makes running the tests longer which isn't nice.

The issue was that without running `make images` first, the `make tests` failed with quite strange error and it was not immediately obvious what was the problem.

Maybe what could we do instead of this change is add a check to the `make tests` that would detect that the images are not available and exit with a message "Run `make images` first". WDYT?

Anyway I think we'll need to revisit the way we run unit tests later. Because if I'm a newbie, I'd intuitively run the script `run_unit_tests.sh` instead of running `make` or `make all`, but on Fedora that would not only fail miserably but also could be "dangerous" if the tests are not written safely and change the OS :)

SpyTec:
> The issue was that without running `make images` first, the `make tests` failed with quite strange error and it was not immediately obvious what was the problem.
> 
> Maybe what could we do instead of this change is add a check to the `make tests` that would detect that the images are not available and exit with a message "Run `make images` first". WDYT?

If it is possible, that would definitely be worth pursuing

> Anyway I think we'll need to revisit the way we run unit tests later. Because if I'm a newbie, I'd intuitively run the script `run_unit_tests.sh` instead of running `make` or `make all`, but on Fedora that would not only fail miserably but also could be "dangerous" if the tests are not written safely and change the OS :)

To reduce the chance of this happening (which is what I intuitively did as well when I cloned the repo) would be to put `run_unit_tests.sh` in a different folder so it is not in the root. As well as creating a `CONTRIBUTING.md` for getting started

fellipeh:
Removed images running from tests, and added `make lint` to show lint results

Makefile
```diff
@@ -1,23 +1,24 @@
...
 	@docker build -f Dockerfiles/centos7 -t $(IMAGE)/centos7 .
 
-.PHONY: tests
 tests:
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos5 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos6 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos7 ./run_unit_tests.sh
+
+lint:
+	pylint --rcfile=.pylintrc convert2rhel/
```
ZhukovGreen, line 22: I think .pylintc should be picked up automatically by pylint (http://pylint.pycqa.org/en/latest/user_guide/run.html?highlight=pylintrc#command-line-options)

```
Pylint searches for a configuration file in the following order and uses the first one it finds:

    pylintrc in the current working directory
    .pylintrc in the current working directory
    pyproject.toml in the current working directory, providing it has at least one tool.pylint. section.
    setup.cfg in the current working directory, providing it has at least one pylint. section
    If the current working directory is in a Python module, Pylint searches up the hierarchy of Python modules until it finds a pylintrc file. This allows you to specify coding standards on a module-by-module basis. Of course, a directory is judged to be a Python module if it contains an __init__.py file.
    The file named by environment variable PYLINTRC
    if you have a home directory which isn't /root:
        .pylintrc in your home directory
        .config/pylintrc in your home directory
    /etc/pylintrc
```

sturivny: @ZhukovGreen  
Good idea.
But now a lot of pylint checks are failed.
I think we should fix it in the first place.

bocekm: What version of python does the pylint considers? When I run it on my Fedora, does it follow rules for python 3.8.2? If so, running it in the docker images would make more sense, to have it use python 2.6 and 2.7 instead.

SpyTec: I'm pretty sure pylint follows the rules for Python 3 if that's what's it installed for. 

Though if we do setup a virtualenv, be it plain virtualenv, poetry, or pipenv, we should be able to run linting without going into a container. (#55)

bocekm: Yeah, that could work too. You can get inspired by @ZhukovGreen's https://github.com/oamg/leapp-repository/pull/501. The experience (makefile calls) should be ideally the same since both project come from the OAMG.

bocekm: @ZhukovGreen, @SpyTec, as I'm thinking about it, isn't using virtualenv tied to the python version(s) you have installed on your system? So if I don't have py2.6 on my Fedora, I wouldn't be able to lint the code for py2.6 rules using virtualenv? Or is it solvable with poetry/pipenv (I don't know either)?

ZhukovGreen: @bocekm Hi Michal, poetry or pipenv doesn't help in the sense of providing your system with the python2. For that purpose is awesome tool https://github.com/pyenv/pyenv:
```
curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
git clone https://github.com/pyenv/pyenv-virtualenv.git ~/.pyenv/plugins/pyenv-virtualenv
pyenv install 3.8
pyenv install 2.7
pyenv global 3.8-dev    # makes pyenv py38 venv global to your system, so you can isolate your real system python
``` 
```

bocekm: Now the question is whether it's not easier to just use the docker images prepared by `make images`.
Or dropping the docker and use just pyenv could be also the way. But I don't see a benefit in mixing docker and pyenv for running unit tests and linting - both have very similar needs in terms of the environment.

SpyTec: I think it makes sense to run both tests and builds within containers. Builds because you then test whether we can build in the right OS. Tests because a missed mock or badly configured mock wouldn't affect your own system - e.g. unregistering the system, or uninstalling packages.

Linting on the other hand can be done outside a container. AFAIK the only thing we'd gain from running it in a container would be consistency for all developers in case someone runs the wrong python version

I would vote for setting up a venv with python 2.6 just for linting, keeping tests and builds within containers

bocekm: The build process is already [handled by packit](https://github.com/oamg/convert2rhel/blob/master/.packit.yaml#L10) which builds a copr build for every commit pushed to GitHub.

These new ideas are nice, but since the original PR was about something completely else, I feel we hijacked it a bit from @fellipeh. Let's close this one. @SpyTec, @sturivny, or @ZhukovGreen can pick up that task and create a new PR.

ZhukovGreen: I see you test your code on 3 Centos. Then I can't understand how you might use pyenv as a replacement of those 3 containers, because 1 container (or your local system) can't give you 3 Centos versions. But what you can achieve with the pyenv is to have different python versions within each of these Centos containers and then test your code against those pythons. You can do the same on your Fedora OS. In all these cases you can take advantage of cool docker caching mechanism:) Anyway look slike the discussion flows out of the initial comment topic. resolving...

PR #50, opened 2020-05-20
Pylint fix C0103 invalid Class name

sturivny:
Class names should follow the UpperCaseCamelCase

### DIFF
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -32,15 +32,15 @@ from convert2rhel import utils
 
 class TestCert(unittest.TestCase):
 
-    class glob_mocked(unit_tests.MockFunction):
+    class GlobMocked(unit_tests.MockFunction):
         def __call__(self, *args, **kwargs):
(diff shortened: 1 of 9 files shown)


SpyTec:
I noticed in the pylint that there are a few more cases of C0103, specifically with variables. Do we want to tackle this in a separate PR?

sturivny:
@SpyTec 
> I noticed in the pylint that there are a few more cases of C0103, specifically with variables. Do we want to tackle this in a separate PR?

I don't know. For me, it looks good. 
I'd rather add an exception to the pylint config file.

bocekm:
Some of the C0103 messages are because the function name is longer than 30 chars. I'm not a fan of this rule. I prefer longer but descriptive function name rather than function name that doesn't clearly tell me what the function does.
For these messages I'd the ignore comment.

I guess that you can change constants to uppercase and classes to PascalCase.

sturivny:
Sure, will do that in other commits. 
There is the first one https://github.com/oamg/convert2rhel/pull/53

PR #52, opened 2020-05-20
Set max number of characters on a single line 120

sturivny:


bocekm:
@sturivny, please create a new PR for this. The change has been accidentally reverted by https://github.com/oamg/convert2rhel/pull/53/files.

PR #53, opened 2020-05-21
Pylint checks for length of function names to 50

sturivny:


.pylintrc
```diff
@@ -1,4 +1,10 @@
+[BASIC]
+
+# Regular expression matching correct function names
+function-rgx=[a-z_][a-z0-9_]{2,50}$
```
SpyTec, line 4: So we can't have function names that are 1 characters long? Aww

LGTM! :)

sturivny: Yeah, I was also thinking about 2 characters, something like `def _y():` or `def ya():` =D

sturivny:
Rebased

PR #54, opened 2020-05-25
Pylint fix warnings: [C0103 invalid-name]

sturivny:
As discussed in https://github.com/oamg/convert2rhel/pull/50#issuecomment-631512340
Disabled pylint checks for the:
```
[C0103 invalid-name] Constant name "tool_opts" doesn't conform to UPPER_CASE naming style File: convert2rhel/toolopts.py, line 226, in print_non_interactive_opts
[C0103 invalid-name] Constant name "tool_opts" doesn't conform to UPPER_CASE naming style File: convert2rhel/toolopts.py, line 255, in 
[C0103 invalid-name] Constant name "system_info" doesn't conform to UPPER_CASE naming style File: convert2rhel/systeminfo.py, line 161, in 
```
by adding comment `# pylint: disable=invalid-name  # noqa`

PR #57, opened 2020-05-25
Revert renaming global variables as constants

bocekm:
- these objects are not constants but mutable objects
- the change being reverted has been done under https://github.com/oamg/convert2rhel/pull/54

Sorry, @sturivny, I missed three other globals that shouldn't be considered constants.


sturivny:
NP, LGTM

PR #58, opened 2020-05-25
Add check RHEL kernel

fellipeh:
If there is no RHEL kernel, the convert process will halt with CRITICAL message.

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -16,7 +16,6 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from itertools import imap
-
 import logging
 import os
 import re
@@ -353,6 +352,14 @@ def replace_non_red_hat_packages():
 def preserve_only_rhel_kernel():
     loggerinst = logging.getLogger(__name__)
     needs_update = install_rhel_kernel()
+
+    loggerinst.info("Verifying that RHEL kernel has been installed")
+    if not is_rhel_kernel_installed():
+        loggerinst.critical(
+            "No RHEL kernel installed. Verify that the repository used for installing kernel contains RHEL packages.")
+    else:
+        loggerinst.info("RHEL kernel has been installed.")
+
     non_rhel_kernel_pkgs = remove_non_rhel_kernels()
     if non_rhel_kernel_pkgs:
         install_additional_rhel_kernel_pkgs(non_rhel_kernel_pkgs)
@@ -371,6 +378,7 @@ def install_gpg_keys():
             print_output=False)
         if ret_code != 0:
             loggerinst.critical("Unable to import GPG key: %s", output)
+
 
 
 def install_rhel_kernel():
@@ -507,3 +515,8 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
         if name != "kernel":
             loggerinst.info("Installing RHEL %s" % name)
             call_yum_cmd("install %s" % name)
+
+
+def is_rhel_kernel_installed():
+    installed_rhel_kernels = get_installed_pkgs_by_fingerprint(system_info.fingerprints_rhel, name="kernel")
+    return len(installed_rhel_kernels) > 0
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -28,6 +28,14 @@ from convert2rhel.systeminfo import system_info
 
 
 class TestPkgHandler(unit_tests.ExtendedTestCase):
+    class GetInstalledPkgsWithFingerprintMocked(unit_tests.MockFunction):
+        def __init__(self, data=None):
+            self.data = data
+            self.called = 0
+
+        def __call__(self, *args, **kwargs):
+            self.called += 1
+            return self.data
 
     class CallYumCmdMocked(unit_tests.MockFunction):
         def __init__(self):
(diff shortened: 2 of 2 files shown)


convert2rhel/pkghandler.py
```diff
@@ -356,6 +357,11 @@ def replace_non_red_hat_packages():
 def preserve_only_rhel_kernel():
     loggerinst = logging.getLogger(__name__)
```
ZhukovGreen, line 358: Please instantiate the logger outside of the function scope, you probably don't want it re-instantiate each time you call this function. Pls see the following log example, when you have logger outside
```
----------------------------- live log collection ------------------------------
| 10:43:49 | my_pkg | INFO | __init__.py | Logger configured
| 10:43:49 | my_pkg.log | INFO | log.py | Logger obtained
collected 1 item

test_logger.py::test_foo 
-------------------------------- live log call ---------------------------------
| 10:43:49 | my_pkg.log | INFO | log.py | Foo called
| 10:43:49 | my_pkg.log | INFO | log.py | Foo called
PASSED
```

ZhukovGreen: /cc @SpyTec

SpyTec: Outside the scope for this and #60, should be fixed in a separate PR.

As you [linked in #60](https://github.com/oamg/convert2rhel/pull/60#discussion_r432329984), Logging Cookbook mentions 

> Multiple calls to logging.getLogger('someLogger') return a reference to the same logger object.

If the method call returns a singleton I'm not sure I understand the reason for instantiating it outside method scope

ZhukovGreen: @SpyTec because it is an additional overhead which still returns the same object

bocekm: See https://github.com/oamg/convert2rhel/pull/60#discussion_r432385516. I agree with changing the way we work with the logger now. But let's do that under a separate PR for all the code.

fellipeh:
Fixed conflict and rebased

run_unit_tests.sh
```diff
@@ -27,7 +27,7 @@
 
 # -- TEST OUTPUT --
 # To let the tests print to stdout, uncomment the following line:
-#TEST_OUTPUT="--nocapture"
+TEST_OUTPUT="--nocapture"
```
SpyTec, line 30: Committed by mistake? Otherwise seems out of scope

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -28,6 +28,12 @@
 
 
 class TestPkgHandler(unit_tests.ExtendedTestCase):
+    class GetInstalledPkgsWithFingerprintMocked(unit_tests.MockFunction):
+        def __init__(self, data=None):
+            self.data = data
+
+        def __call__(self, *args, **kwargs):
+            return self.data
```
SpyTec, line 36:
```suggestion
    class GetInstalledPkgsWithFingerprintMocked(unit_tests.MockFunction):
        def __init__(self, data=None):
            self.data = data
            self.called = 0

        def __call__(self, *args, **kwargs):
            self.called += 1
            return self.data
```

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -380,12 +388,13 @@ def __call__(self):
     @unit_tests.mock(pkghandler, "install_gpg_keys",
                      DumbCallableObject())
     @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked())
+    @unit_tests.mock(pkghandler, "get_installed_pkgs_by_fingerprint",
+                     GetInstalledPkgsWithFingerprintMocked(data=['kernel']))
     def test_preserve_only_rhel_kernel(self):
         pkghandler.preserve_only_rhel_kernel()
 
         self.assertEqual(utils.run_subprocess.cmd, "yum update -y kernel")
```
SpyTec, line 396:
```suggestion
    def test_preserve_only_rhel_kernel(self):
        pkghandler.preserve_only_rhel_kernel()

        self.assertEqual(utils.run_subprocess.cmd, "yum update -y kernel")
        self.assertEqual(pkghandler.get_installed_pkgs_by_fingerprint.called, 1)
```

SpyTec, line 396: Create another test for `preserve_only_rhel_kernel()` to verify that branch for `loggerinst.critical` works as well

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -577,6 +586,16 @@ def test_install_additional_rhel_kernel_pkgs(self):
         pkghandler.install_additional_rhel_kernel_pkgs(removed_pkgs)
         self.assertEqual(pkghandler.call_yum_cmd.called, 2)
 
+    @unit_tests.mock(pkghandler, "get_installed_pkgs_by_fingerprint",
+                     GetInstalledPkgsWithFingerprintMocked(data=['kernel']))
+    def test_check_exist_rhel_kernel_by_fingerprint(self):
```
SpyTec, line 591:
```suggestion
    def test_check_installed_rhel_kernel_returns_true(self):
```

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -577,6 +586,16 @@ def test_install_additional_rhel_kernel_pkgs(self):
...
+        self.assertEqual(pkghandler.is_rhel_kernel_installed(), True)
+
+    @unit_tests.mock(pkghandler, "get_installed_pkgs_by_fingerprint",
+                     GetInstalledPkgsWithFingerprintMocked(data=[]))
+    def test_check_exist_rhel_kernel_by_fingerprint(self):
```
SpyTec, line 596:
```suggestion
    def test_check_installed_rhel_kernel_returns_false(self):
```

convert2rhel/unit_tests/pkghandler_test.py
SpyTec: Could probably rewrite that to be more clear. There isn't any tests right now covering what happens inside `preserve_only_rhel_kernel()` if `is_rhel_kernel_installed()` returns False

Can be checked with a new test mocking out `is_rhel_kernel_installed` to return False and assert `self.assertRaises(SystemExit, preserve_only_rhel_kernel)`

fellipeh:
Code rebased and squashed

PR #59, opened 2020-05-26
[WIP] fix/pylint messages

fellipeh:
Fixing pylint errors

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -27,3 +27,14 @@ disable=
 [REPORTS]
 
 msg-template='[{msg_id} {symbol}] {msg} File: {path}, line {line}, in {obj}'
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=120
+
+[LOGGING]
+
+# Format style used to check logging format string. `old` means using %
+# formatting, `new` is for `{}` formatting,and `fstr` is for f-strings.
+logging-format-style=old
--- a/.pylintrc
+++ b/.pylintrc
@@ -23,7 +23,7 @@ disable=
  old-style-class,
 # "R" Refactor recommendations
  too-few-public-methods,
-# we need, some time, to disable pylint locally :)
+# do not warn about pylint checks being disabled through comments in the code
  locally-disabled,
 
 [REPORTS]
--- a/.pylintrc
+++ b/.pylintrc
@@ -23,6 +23,8 @@ disable=
  old-style-class,
 # "R" Refactor recommendations
  too-few-public-methods,
+# we need, some time, to disable pylint locally :)
+ locally-disabled,
 
 [REPORTS]
 
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -236,22 +236,23 @@ def print_pkg_info(pkgs):
     for pkg in pkgs:
         if not pkg.vendor:
             pkg.vendor = "N/A"
-    max_nvra_length = max(imap(len, map(lambda pkg: get_pkg_nvra(pkg), pkgs)))  # pylint disable: bad-builtin, deprecated-lambda
+    max_nvra_length = max(imap(len, map(lambda pkg: get_pkg_nvra(pkg),
+                                        pkgs)))  # pylint: disable=line-too-long, bad-builtin, deprecated-lambda
     max_vendor_length = max(max(imap(len, map(lambda pkg: pkg.vendor, pkgs))),
-                            len("Vendor")) # pylint disable: bad-builtin, deprecated-lambda
+                            len("Vendor"))  # pylint: disable=bad-builtin, deprecated-lambda
     loggerinst = logging.getLogger(__name__)
     result = "%-*s  %-*s  %s" % (max_nvra_length, "Package", max_vendor_length,
                                  "Vendor", "Repository") + "\n"
     loggerinst.info("%-*s  %-*s  %s"
                     % (max_nvra_length, "Package", max_vendor_length, "Vendor",
-                       "Repository"))  # pylint disable: logging-not-lazy
+                       "Repository"))  # pylint: disable=logging-not-lazy
     result += "%-*s  %-*s  %s" % (max_nvra_length, "-" * len("Package"),
                                   max_vendor_length, "-" * len("Vendor"),
                                   "-" * len("Repository")) + "\n"
     loggerinst.info("%-*s  %-*s  %s"
                     % (max_nvra_length, "-" * len("Package"),
                        max_vendor_length, "-" * len("Vendor"),
-                       "-" * len("Repository")))  # pylint disable: logging-not-lazy
+                       "-" * len("Repository")))  # pylint: disable=logging-not-lazy
     for pkg in pkgs:
         try:
             from_repo = pkg.yumdb_info.from_repo
@@ -266,7 +267,7 @@ def print_pkg_info(pkgs):
         loggerinst.info("%-*s  %-*s  %s"
                         % (max_nvra_length, get_pkg_nvra(pkg),
                            max_vendor_length, pkg.vendor,
-                           from_repo))  # pylint disable: logging-not-lazy
+                           from_repo))  # pylint: disable=logging-not-lazy
     loggerinst.info("")
     return result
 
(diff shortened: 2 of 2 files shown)


convert2rhel/pkghandler.py
```diff
@@ -49,7 +50,7 @@ def call_yum_cmd_w_downgrades(cmd, fingerprints):
     for _ in xrange(MAX_YUM_CMD_CALLS):
         output, ret_code = call_yum_cmd(cmd, "%s" % (" ".join(
             get_installed_pkgs_by_fingerprint(fingerprints))))
-        loggerinst.info("Received return code: %s\n" % str(ret_code))
+        loggerinst.info("Received return code: {}\n".format(str(ret_code)))
```
sturivny, line 53: As we support python2.6  it requires numbers in the placeholders:
`'{0} {1}\n'.format(numb, foo)` or: `'{numb} {foo}\n'.format(numb=numb, foo=foo)`

fellipeh: You're right! fixed!

.pylintrc
```diff
@@ -1,10 +1,16 @@
...
+# In order to make a check more strict add proper regex http://pylint-messages.wikidot.com/messages:c0103
+argument-rgx=.*
+attr-rgx=.*
+class-rgx=.*
+const-rgx=.*
+function-rgx=.*
+method-rgx=.*
+module-rgx=.*
+variable-rgx=.*
+inlinevar-rgx=.*
```
SpyTec, line 11: Doesn't this mean we can name functions, etc. whatever we want? Matching as anything doesn't make much sense to me

fellipeh: Just following Leapp guide lines, but we can go back, what you think about it @bocekm ?

convert2rhel/utils.py
```diff
@@ -246,13 +248,13 @@ def backup_and_track_removed_pkg(self, pkg):
     def _remove_installed_pkgs(self):
         """For each package installed during conversion remove it."""
         loggerinst = logging.getLogger(__name__)
-        loggerinst.task("Rollback: Removing installed packages")
+        loggerinst.task("Rollback: Removing installed packages") # pylint: disable=W1201
```
bocekm, line 251: I wonder why is this needed here. There are no arguments passed. http://pylint-messages.wikidot.com/messages:w1201.
Actually, what pylint reported on the latest master on this line is E1103, not W1201:
https://github.com/oamg/convert2rhel/runs/709656220#step:5:44

The same applies to the change few lines below.

convert2rhel/utils.py
```diff
@@ -208,7 +208,7 @@ def log_traceback(debug):
         loggerinst.debug(traceback_str)
     else:
         # Print the traceback to the log file in any way
-        loggerinst.file(traceback_str)
+        loggerinst.file(traceback_str)  # pylint: disable=W1201
```
bocekm, line 211: I wonder why is it needed for the `.file` but not `.debug`. Also, the traceback str should be just string with no arguments.
Actually, what pylint reported on the latest master on this line is E1103, not W1201:
https://github.com/oamg/convert2rhel/runs/709656220#step:5:41

.pylintrc
bocekm: I think these benevolent regexps are not needed. We've already fixed the annoying one https://github.com/oamg/convert2rhel/pull/53, the rest seems to be fine. We can discuss particular ones if necessary instead of "disabling" all.
By the way, @sturivny has resolved the invalid name "violations" here https://github.com/oamg/convert2rhel/pull/54 already.

convert2rhel/utils.py
fellipeh: I'll rebase, several files are changed since I send these PR.

convert2rhel/utils.py
fellipeh: I'll rebase, several files are changed since I send these PR.

.pylintrc
```diff
@@ -23,7 +29,23 @@ disable=
...
+
+[LOGGING]
+
+# Format style used to check logging format string. `old` means using %
+# formatting, `new` is for `{}` formatting,and `fstr` is for f-strings.
+logging-format-style=old
+
+[DESIGN]
+max-args=11
+max-attributes=21
```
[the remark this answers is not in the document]
fellipeh: No longer need anymore, after latest change, pylint stop warning..

.pylintrc
```diff
@@ -23,7 +29,23 @@ disable=
...
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=119
+
+[LOGGING]
+
+# Format style used to check logging format string. `old` means using %
+# formatting, `new` is for `{}` formatting,and `fstr` is for f-strings.
+logging-format-style=old
```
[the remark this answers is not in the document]
fellipeh: These are related to the, eg: ` loggerinst.info("Received return code: %s\n" % str(ret_code))` 

py26 allow % but py27 allow only ` loggerinst.info("Received return code: %s\n", str(ret_code))`

So, these config make py26_lint and py27_lint happy.

.pylintrc
```diff
@@ -23,7 +23,20 @@ disable=
  old-style-class,
 # "R" Refactor recommendations
  too-few-public-methods,
+# we need, some time, to disable pylint locally :)
+ locally-disabled,
```
bocekm, line 27: What's the purpose of this one?

.pylintrc
```diff
@@ -23,7 +23,20 @@ disable=
...
 
 [REPORTS]
 
 msg-template='[{msg_id} {symbol}] {msg} File: {path}, line {line}, in {obj}'
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=120
```
bocekm, line 36: Again, don't add it when it's already there. As I wrote in https://github.com/oamg/convert2rhel/pull/59#discussion_r431192097.

.pylintrc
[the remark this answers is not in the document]
bocekm: Then I don't understand, why are you doing all the changes in `convert2rhel/pkghandler.py`.

There, you're changing `loggerinst.info("Received return code: %s\n" % str(ret_code))` to `loggerinst.info("Received return code: %s\n", str(ret_code))`.

If you change all the occurrences of `%` to `,`, why do we need new style then?

.pylintrc
SpyTec: Apparently it was [accidentally removed](https://github.com/oamg/convert2rhel/commit/963215b823b872304ec9326972c4d5d45a887751)

bocekm: Aha, I'm sorry then. Thanks @SpyTec.

.pylintrc
fellipeh: On py2.7, if you disable pylint in particular line, pylint raise a locally-disabled warning these config makes pylint happy and doesn't show these warning when you disable inline.

bocekm: Ah, ok. Then please update the comment, it's quite cryptic/misleading. Change it to something like `# do not warn about pylint checks being disabled through comments in the code`.

.pylintrc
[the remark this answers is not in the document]
fellipeh: I believe I give you wrong answer, sorry... the reason of old style is: on py27 force us to use `"abcd {0}".format()`  but py26 doesn't have the `format` on str... so.. we need these old style to pylint27 not warning if we not use `format`.

From the doc:

> logging-format-style is a new option for the logging checker for usage of str.format() style format strings in calls to loggers.
> 
> It accepts two options: --logging-format-style=old for using % style formatting, which is the assumed default, and --logging-format-style=new for using {} style formatting.
>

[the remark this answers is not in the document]
fellipeh: and the Changes from `loggerinst.info("Received return code: %s\n" % str(ret_code))` to `loggerinst.info("Received return code: %s\n", str(ret_code))` is related to another warning massage, these one:

> [W1201 logging-not-lazy] Specify string format arguments as logging function parameters 

not for old style warning.

I'll focus more in fix the lint errors instead of warnings, It's god to have all clean, but I believe we can handle with some warnings for now, right?

.pylintrc
fellipeh: Done!

.pylintrc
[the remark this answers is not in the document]
bocekm: Yes, we can deal with warnings later. Focus on errors first.

Anyway, it seems to me that if we change all the logging calls to use comma instead of `%`, we kill two birds with one stone - 1) we would need the `logging-format-style=old` and 2) we wouldn't be getting the `logging-not-lazy` warnings. Am I right?

fellipeh:
As discuss with @bocekm, Closing these PR to focus more on lint error instead of non-errors messages.. we came back when we have time.

PR #60, opened 2020-05-26
fix: Unsubscribe system on rollback

SpyTec:
This makes it so rollback also unregisters the system.

Making it draft until I've tested whether it works as I expect.


convert2rhel/subscription.py
```diff
@@ -35,6 +35,17 @@ def subscribe_system():
         tool_opts.password = None
 
 
+def unregister_system():
+    """Unregister the system"""
+    loggerinst = logging.getLogger(__name__)
```
ZhukovGreen, line 40: Usually, loggers instantiated outside of the function scope. In your case, it will be always reinstating when unregister_system will be called. This https://docs.python.org/3/howto/logging-cookbook.html is useful

ZhukovGreen: Also useful to note that logging uses a singleton pattern, so each time you still have the same instance of it

bocekm: When you call `getLogger` you don't get a new instance but "a reference to the same logger object" as in the link you posted.
We used to take advantage of the singleton pattern in the past, we were simply calling e.g. `logger.log("warning", "Some warning")` from anywhere. But then John has changed it the way it is now, he was following https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/.
I'm also not in favor of getting the logger through `loggerinst = logging.getLogger(__name__)` in every function that wants to log. I'd be happy to get that changed.

ZhukovGreen: Oh, I am not sure which python version do you use, but since 2.7 disable_existing_loggers is True by default, so using module level instantiation is ok (moreover it is recommended in py2 and py3):
https://docs.python.org/2/howto/logging.html#advanced-logging-tutorial
https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial

```
A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:

This means that logger names track the package/module hierarchy, and it’s intuitively obvious where events are logged just from the logger name.
```

ZhukovGreen: Happy we know more about logging module now:) Never heard about disable exisiting loggers:)

bocekm: Currently, we need to be 2.6 and 2.7 compatible. Soon, we'll need to be also 3.6+ compatible. Getting the logger on the module level is fine with me.

convert2rhel/main.py
```diff
@@ -209,7 +209,7 @@ def rollback_changes():
     utils.changed_pkgs_control.restore_pkgs()
     redhatrelease.system_release_file.restore()
     redhatrelease.yum_conf.restore()
-    subscription.rollback_renamed_repo_files()
+    subscription.rollback()
```
bocekm, line 212: The unregistering should happen before calling `utils.changed_pkgs_control.restore_pkgs()` three lines above, because that removes the installed subscription-manager.

SpyTec: Fixed

SpyTec:
Ready for review. Rebased, ran tests, and manually verified in a Vagrantbox

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -229,4 +232,19 @@ def test_parse_sub_date(self):
         sku = self.FakeSubscription()
         for i in sku.dates_formats:
             self.assertEqual(subscription.parse_sub_attrs(sku(i))["ends"], i)
-            self.assertEqual(len(subscription.logging.getLogger.critical_msgs),0)
+            self.assertEqual(len(subscription.logging.getLogger.critical_msgs), 0)
+
+
+    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked())
+    def test_unregister_system(self):
```
bocekm, line 239: You can add a test that the logging happens for both the return code 0 and non-0.

SpyTec:
Rebased

PR #62, opened 2020-05-28
Pylint fix useless-return import-outside-toplevel bad-whitespace

sturivny:


convert2rhel/rhelvariant.py
```diff
@@ -18,6 +18,8 @@
 import os
 import re
 import logging
+
+from convert2rhel.toolopts import tool_opts
```
bocekm, line 22: Have you verified that this does not cause a circular import loop because of https://github.com/oamg/convert2rhel/blob/master/convert2rhel/toolopts.py#L20?

sturivny: I'll take a look. Thank you for pointing.

SpyTec:
Packit testing shows one test is failing

```
test_get_installed_pkg_objects (convert2rhel.unit_tests.pkghandler_test.TestPkgHandler) ... ERROR
```

sturivny:
@SpyTec Just missed `patterns=None` inthe paramenters =D
```       
 def __call__(self, patterns=None):
            if patterns is None:
                patterns = []
               ...
```

PR #64, opened 2020-06-01
feat: Ability to specify RHSM url

SpyTec:
This makes it easier to test locally using a staging/test RHSM server.

Tested and verified in a Vagrantbox


convert2rhel/toolopts.py
```diff
@@ -127,6 +130,10 @@ def _register_options(self):
                          " conversions from CentOS/OL 6/7, Server, and Client"
                          " for conversions from CentOS/OL 5. If not used, the"
                          " user is asked to choose a variant.")
+        group.add_option("--serverurl", help="Use a custom Red Hat Subscription"
+                         " Manager server URL to register the system with. If"
+                         " not provided subscription-manager defaults will be"
+                         " used")
```
bocekm, line 136:
```suggestion
                         " not provided, the subscription-manager defaults will be"
                         " used.")
```

convert2rhel/toolopts.py
```diff
@@ -202,6 +209,9 @@ def _process_cli_options(self):
                                         rhelvariant.get_supported_variants()))
             tool_opts.variant = parsed_opts.variant
 
+        if parsed_opts.serverurl:
```
bocekm, line 212:
```suggestion
        if parsed_opts.serverurl:
            if parsed_opts.disable_submgr:
                loggerinst.warn("Ignoring the --serverurl option. It has no effect when --disable-submgr is used.")
```

bocekm:
It works as expected.

PR #65, opened 2020-06-01
fix: Make Manifest read README.md properly

SpyTec:
Fixes build errors with missing README

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
 recursive-include convert2rhel/data *
 include man/*
 include LICENSE
-include README
+include README.md


fellipeh:
When I trying to build, getting these error:
```
+ /usr/bin/python2 setup.py build
  File "setup.py", line 35, in <module>
    long_description=read('README.md'),
  File "setup.py", line 27, in read
    return open(os.path.join(os.path.dirname(__file__), fname)).read()
IOError: [Errno 2] No such file or directory: 'README.md'
```

fellipeh:
I believe the problem  is here:
`open(os.path.join(os.path.dirname(__file__), fname)).read()`  the function `os.path.dirname(__file__)` returns nothing:
```bash
(Pdb) os.path.join(os.path.dirname(__file__), fname)
'README.md'
(Pdb) os.path.dirname(__file__)
''
```

You need to change it for these:

```python
def read(fname):
    return open(os.path.join(os.path.dirname(os.path.abspath(__file__)), fname)).read()
```

SpyTec:
@fellipeh I wasn't able to reproduce it at all using different folder structures, and even the same Virtualbox.

Your suggestion seems to be right however, I'll commit a fix in a separate PR

PR #67, opened 2020-06-02
style: Add Black formatter config

SpyTec:
This adds [Black code formatter](https://black.readthedocs.io/en/stable/index.html) configuration and information on how to use it.

Blocked by #59 #55

I've also created #68 to showcase what would be changed currently.

PR #68, opened 2020-06-02
style: Black formatting showcase

SpyTec:
Showcase of what would be changed using black. Refer to #67

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -75,11 +75,12 @@ def main():
         process_phase = ConversionPhase.PRE_PONR_CHANGES
         pre_ponr_conversion()
 
-        loggerinst.warning("The tool allows rollback of any action until this"
-                           " point.")
-        loggerinst.warning("By continuing all further changes on the system"
-                           " will need to be reverted manually by the user,"
-                           " if necessary.")
+        loggerinst.warning("The tool allows rollback of any action until this" " point.")
+        loggerinst.warning(
+            "By continuing all further changes on the system"
+            " will need to be reverted manually by the user,"
+            " if necessary."
+        )
         utils.ask_to_continue()
 
         process_phase = ConversionPhase.POST_PONR_CHANGES
@@ -111,8 +112,7 @@ def main():
             # system rollback without user intervention. If a proper rollback
             # solution is necessary it will need to be future implemented here
             # or with the use of other backup tools.
-            print("Conversion process interrupted and manual user intervention"
-                  " will be necessary.")
+            print("Conversion process interrupted and manual user intervention" " will be necessary.")
 
         return 1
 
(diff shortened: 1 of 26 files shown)


ZhukovGreen:
from my perspective the biggest benefit of the black formating is that we're no longer discussing a code style. It is just one, uncompromised. This saves a lot of time.
+ black works nicely with the pre-commit

SpyTec:
> from my perspective the biggest benefit of the black formating is that we're no longer discussing a code style. It is just one, uncompromised. This saves a lot of time.
> `+ black works nicely with the pre-commit

I agree 100%. I'm sure we will run into differences between pylint and black that need to be resolved though. Especially given we use pylint 1.9 rather than 2+

ZhukovGreen:
> > from my perspective the biggest benefit of the black formating is that we're no longer discussing a code style. It is just one, uncompromised. This saves a lot of time.
> > `+ black works nicely with the pre-commit
> 
> I agree 100%. I'm sure we will run into differences between pylint and black that need to be resolved though. Especially given we use pylint 1.9 rather than 2+

Never tried:)

convert2rhel/main.py
```diff
@@ -75,11 +75,12 @@ def main():
         process_phase = ConversionPhase.PRE_PONR_CHANGES
         pre_ponr_conversion()
 
-        loggerinst.warning("The tool allows rollback of any action until this"
-                           " point.")
-        loggerinst.warning("By continuing all further changes on the system"
-                           " will need to be reverted manually by the user,"
-                           " if necessary.")
+        loggerinst.warning("The tool allows rollback of any action until this" " point.")
```
SpyTec, line 78: Only bug I've seen so far is the lack of handling of multi-line strings, here it doesn't merge the strings but just brings them to the same line

fellipeh:
@SpyTec  Please resolve the conflicting files, if you still want to merge this PR.

SpyTec:
> @SpyTec Please resolve the conflicting files, if you still want to merge this PR.

We'll open a new PR once we've had a larger discussion with OAMG and this is something we'd want to pursue. Can remain open for now

ZhukovGreen:
I assume next week to make a MR which will be including the black formatter as well. I’ll ping you once it is ready and, ideally, we can do absolutely the same for c2r

ZhukovGreen:
Just an update: have to postpone the work on black formatter to the next week

ZhukovGreen:
One more update. Guys, I have to postpone the work on the feature, due to another, more important tasks. Please follow the progress here - OAMG-3458. I won't give any next date term, since already now feeling bad about this. So, I just will try to do this ASAP, since pre-commits hooks, black formatting really a hue boost to most of the devs performance

SpyTec:
> One more update. Guys, I have to postpone the work on the feature, due to another, more important tasks. Please follow the progress here - OAMG-3458. I won't give any next date term, since already now feeling bad about this. So, I just will try to do this ASAP, since pre-commits hooks, black formatting really a huge boost to most of the devs performance

No rush! It is relatively low priority compared to other tasks on our side as well :)

ZhukovGreen:
ref https://github.com/oamg/leapp-repository/issues/565

PR #70, opened 2020-06-02
fix/dependency issues during converting from CentOS7.7-RHEL7.7

fellipeh:
Convert2rhel failed when try to convert from CentOS 7.7 (1908) to RHEL 7.7 with "error: Failed dependencies: system-release is needed by kernel-3.10.0-1062.el7.x86_64"

These PR fixed these fail

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1821726

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -16,6 +16,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from itertools import imap
+
 import logging
 import os
 import re
@@ -457,6 +458,11 @@ def replace_non_rhel_installed_kernel(version):
     utils.ask_to_continue()
 
     pkg = "kernel-%s" % version
+
+    utils.remove_pkgs(['kmod-kvdo', 'libreport'], critical=False)
+    for pkgname in set(system_info.pkg_blacklist):
+
+        utils.install_pkgs([pkgname.replace("centos", "redhat"), pkgname.replace("centos", "rhel")], replace=True)
 
     ret_code = utils.download_pkg(
         pkg=pkg, dest=utils.TMP_DIR, disablerepo=tool_opts.disablerepo,
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -327,12 +327,29 @@ class TestPkgHandler(unit_tests.ExtendedTestCase):
 
     class RemovePkgsMocked(unit_tests.MockFunction):
         def __init__(self):
+            self.called = 0
             self.pkgs = None
             self.should_bkp = False
+            self.critical = True
 
-        def __call__(self, pkgs_to_remove, should_backup=False):
+        def __call__(self, pkgs_to_remove, should_backup=False, critical=True):
+            self.called += 1
             self.pkgs = pkgs_to_remove
             self.should_bkp = should_backup
+            self.critical = critical
+
+    class InstallPkgsMocked(unit_tests.MockFunction):
+        def __init__(self):
+            self.called = 0
+            self.pkgs = []
+            self.replace = False
+            self.critical = True
+
+        def __call__(self, pkgs_to_install, replace=False, critical=True):
+            self.called += 1
+            self.pkgs += pkgs_to_install
+            self.replace = replace
+            self.critical = critical
 
     @unit_tests.mock(system_info, "pkg_blacklist", ["installed_pkg",
                                                     "not_installed_pkg"])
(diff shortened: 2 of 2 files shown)


convert2rhel/pkghandler.py
```diff
@@ -466,6 +468,28 @@ def replace_non_rhel_installed_kernel(version):
...
+        utils.download_pkg(pkg=pkgname.replace("centos", "redhat"), dest=utils.tmp_dir,
+                           disablerepo=tool_opts.disablerepo, enablerepo=tool_opts.enablerepo)
+        ret_code = utils.run_subprocess('rpm -i --force --replacepkgs %s*' % os.path.join(utils.tmp_dir,
+                                                                                          pkgname.replace("centos",
+                                                                                                          "redhat")),
+                                        print_output=False)
+
+    for pkgname1 in set(system_info.pkg_blacklist):
+        pkgname1.replace("centos", "rhel")
+        print(pkgname1)
```
sturivny, line 485: Why don't you use log instead of print?

fellipeh: my mistake.. changed

fellipeh:
Rebased

fellipeh:
Code rebase

convert2rhel/pkghandler.py
```diff
@@ -461,6 +468,25 @@ def replace_non_rhel_installed_kernel(version):
...
+                                        print_output=False)
+
+    for pkgname1 in set(system_info.pkg_blacklist):
+        pkgname1.replace("centos", "rhel")
+        utils.download_pkg(pkg=pkgname1.replace("centos", "rhel"), dest=utils.tmp_dir,
+                           disablerepo=tool_opts.disablerepo, enablerepo=tool_opts.enablerepo)
+        ret_code = utils.run_subprocess('rpm -i --force --replacepkgs %s*' % os.path.join(utils.tmp_dir,
+                                                                                          pkgname1.replace("centos",
+                                                                                                           "redhat")),
+                                        print_output=False)
```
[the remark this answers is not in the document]
fellipeh: > Is it possible to replace rpm -i with utils.install_pkgs() as well? Actually, could we not simply call install directly instead of downloading and installing?

I believe install_pkgs don't have "replace" function like rpm.

convert2rhel/pkghandler.py
```diff
@@ -291,6 +292,7 @@ def list_non_red_hat_pkgs_left():
         print_pkg_info(non_red_hat_pkgs)
     else:
         loggerinst.info("All packages are now signed by Red Hat.")
+    return
```
sturivny, line 295: It is unnecessary `return` at the end of the function. Anyway, the function returns `None`

fellipeh: Yeah..  we can start removing these kind of return of many function we have...

fellipeh:
Squash and rebased

fellipeh:
> @fellipeh Please add some unit tests as well as I mentioned earlier

We already have the unit test for these method: https://github.com/oamg/convert2rhel/blob/master/convert2rhel/unit_tests/pkghandler_test.py#L509

SpyTec:
> We already have the unit test for these method: [/convert2rhel/unit_tests/pkghandler_test.py@`master`#L509](https://github.com/oamg/convert2rhel/blob/master/convert2rhel/unit_tests/pkghandler_test.py?rgh-link-date=2020-06-22T21%3A44%3A37Z#L509)

That test is just for making sure `replace_non_rhel_installed_kernel` is called, not what's happening inside.

Instead, seems better to extend this test:
https://github.com/oamg/convert2rhel/blob/29fd6a298e6d3a0c02db967dcb26202042ef9de6/convert2rhel/unit_tests/pkghandler_test.py#L544-L553

fellipeh:
@SpyTec unit test added

convert2rhel/pkghandler.py
```diff
@@ -459,6 +459,11 @@ def replace_non_rhel_installed_kernel(version):
 
     pkg = "kernel-%s" % version
 
+    utils.remove_pkgs(['kmod-kvdo', 'libreport'], critical=False)
+    for pkgname in set(system_info.pkg_blacklist):
+        utils.install_pkgs(pkgname.replace("centos", "redhat"), replace=True)
+        utils.install_pkgs(pkgname.replace("centos", "rhel"), replace=True)
```
SpyTec, line 465:
```suggestion
        utils.install_pkgs([pkgname.replace("centos", "redhat"), pkgname.replace("centos", "rhel")], replace=True)
```

fellipeh:
Rebased

convert2rhel/pkghandler.py
```diff
@@ -458,6 +459,11 @@ def replace_non_rhel_installed_kernel(version):
 
     pkg = "kernel-%s" % version
 
+    utils.remove_pkgs(['kmod-kvdo', 'libreport'], critical=False)
+    for pkgname in set(system_info.pkg_blacklist):
```
ZhukovGreen, line 463: system_info.pkg_blacklist could have duplications?

fellipeh: To be honest, I don't know.. but if has some dup, set will remove that.. I believe that's ok this approach.

fellipeh:
rebased.

fellipeh:
rebased to update code

fellipeh:
Waiting for discussion on Jira finish, so we can decide the next steps on this PR.

convert2rhel/pkghandler.py
```diff
@@ -458,6 +459,11 @@ def replace_non_rhel_installed_kernel(version):
...
+
+        utils.install_pkgs([pkgname.replace("centos", "redhat"), pkgname.replace("centos", "rhel")], replace=True)
+
     ret_code = utils.download_pkg(
         pkg=pkg, dest=utils.TMP_DIR, disablerepo=tool_opts.disablerepo,
         enablerepo=tool_opts.enablerepo)
```
bocekm, line 469: After the kernel rpm is downloaded, call a new function that would check if there are any missing deps of the downloaded kernel rpms and install them. Something like:
```
install_missing_kernel_deps(kernel_rpm_path)
```

Then the new function could look like:
```
def install_missing_kernel_deps(kernel_rpm_path):
    """Since we use rpm for force-replacing the kernel, we need to make sure that all the required dependencies
    of the downloaded kernel rpm are installed. Otherwise the force replacement would fail.
    """
    yb = yum.YumBase()
    yb.reinstallLocal(pkg=kernel_rpm_path)
    yb.resolveDeps()
    pkgs_in_transaction = yb.tsInfo.getMembers()
    to_install = []
    for pkg in pkgs_in_transaction:
        if pkg.reason == "dep":
            to_install.append(pkg.name)
    if to_install:
        loggerinst.info("Installing missing kernel dependencies before the kernel force-replacement.")
        _, ret_code = call_yum_cmd("install", "%s" % (" ".join(to_install)))
        if ret_code != 0:
            loggerinst.critical("Unable to install kernel dependencies: %s", % (", ".join(to_install)))
```

bocekm (changes requested):

fellipeh:
closing this PR in favor to create a new one without several conflicts with master.

PR #71, opened 2020-06-02
fix/lint errors

fellipeh:
These PR will 
- Fix py26 and py27 lint errors
- Add  `make lint` to run lint on local machine
- Add `make lint-errors` to run lint on local machine and show **only** errors

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,5 +1,20 @@
 name: pylint
 
+env:
+  interpreted_pylint: |
+    function fail_on_fatal_or_error_only() {
+      # We care about fatal (1) and error (2) messages only.
+      # The exit code is a sum of numbers corresponding to the message severity level.
+      # E.g. if there are both error (2) and warning (4) messages, the exit code is 6.
+      local exit_code=0
+      pylint --rcfile=.pylintrc convert2rhel/ || exit_code=$?
+      if ! [ $((exit_code & 3)) -eq 0 ]; then
+        return $exit_code;
+      fi;
+      return 0;
+    }
+    fail_on_fatal_or_error_only
+
 on:
   push:
     branches: [ master ]
@@ -25,16 +40,7 @@ jobs:
         pip install astroid==1.2
 
     - name: Run Pylint Checks
-      run: |
-         function interpreted_pylint() {
-            local error_code=0
-            pylint --rcfile=.pylintrc convert2rhel/ || error_code=$?
-            if ! [ $((error_code & 3)) -eq 0 ]; then
-                return $error_code;
-            fi;
-            return 0;
-         }
-         interpreted_pylint
+      run: ${{ env.interpreted_pylint }}
 
   pylint_check_2_7:
     runs-on: ubuntu-latest
@@ -52,13 +58,4 @@ jobs:
         pip install pylint
 
     - name: Run Pylint Checks
-      run: |
-         function interpreted_pylint() {
-            local error_code=0
-            pylint --rcfile=.pylintrc convert2rhel/ || error_code=$?
-            if ! [ $((error_code & 3)) -eq 0 ]; then
-                return $error_code;
-            fi;
-            return 0;
-         }
-         interpreted_pylint
+      run: ${{ env.interpreted_pylint }}
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -12,7 +12,9 @@ jobs:
     container: centos:centos6.9
 
     steps:
-    - uses: actions/checkout@v1
+    - name: Checkout Code
+      uses: actions/checkout@v1
+
 
     - name: Install Requirements
       run: |
@@ -21,11 +23,19 @@ jobs:
         python get-pip.py
         pip install pylint==1.1.0
         pip install astroid==1.2
-        
+
     - name: Run Pylint Checks
       run: |
-        pylint --rcfile=.pylintrc convert2rhel/
-        
+         function interpreted_pylint() {
+            local error_code=0
+            pylint --rcfile=.pylintrc convert2rhel/ || error_code=$?
+            if ! [ $((error_code & 3)) -eq 0 ]; then
+                return $error_code;
+            fi;
+            return 0;
+         }
+         interpreted_pylint
+
   pylint_check_2_7:
     runs-on: ubuntu-latest
     container: centos:centos7.7.1908
@@ -40,7 +50,15 @@ jobs:
         pip install --upgrade pip
         pip install setuptools
         pip install pylint
-        
+
     - name: Run Pylint Checks
       run: |
-        pylint --rcfile=.pylintrc convert2rhel/
+         function interpreted_pylint() {
+            local error_code=0
+            pylint --rcfile=.pylintrc convert2rhel/ || error_code=$?
+            if ! [ $((error_code & 3)) -eq 0 ]; then
+                return $error_code;
+            fi;
+            return 0;
+         }
+         interpreted_pylint
--- a/.pylintrc
+++ b/.pylintrc
@@ -7,8 +7,9 @@ function-rgx=[a-z_][a-z0-9_]{2,50}$
 
 disable=
 # "E" Error for important programming issues (likely bugs)
+# skip redefined function false positive
+ E0102,
  no-member,
-# "W" Warnings for stylistic problems or minor programming issues
  arguments-differ,
 # we can drop this when we drop py2.4 support
  broad-except,
@@ -23,7 +24,22 @@ disable=
  old-style-class,
 # "R" Refactor recommendations
  too-few-public-methods,
-
+# do not warn about pylint checks being disabled through comments in the code
+ locally-disabled,
+# For some reason, pylint doesn't get right instance members, and some types could not be inferred
+# Eg: [E1103 maybe-no-member] Instance of 'RootLogger' has no 'task' member File: convert2rhel/main.py, line 60, in main
+# we have several logger task, and s good approach to disable here, instead of each line
+ maybe-no-member,
+# Disable these Refactor message since most of our tests use more then 20 public methods, and we are not refactor all
+# of our tests right now
+ too-many-public-methods,
+ too-many-branches,
+ too-many-instance-attributes,
 [REPORTS]
 
 msg-template='[{msg_id} {symbol}] {msg} File: {path}, line {line}, in {obj}'
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=120
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,9 @@ images:
 tests:
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos6 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos7 ./run_unit_tests.sh
+
+lint:
+	@pylint --rcfile=.pylintrc convert2rhel/
+
+lint-errors:
+	@pylint --rcfile=.pylintrc convert2rhel/ --errors-only
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -196,8 +196,8 @@ def get_installed_pkg_objects(name=""):
     yum_base.doConfigSetup(init_plugins=False)
     if name:
         return yum_base.rpmdb.returnPackages(patterns=[name])
-    else:
-        return yum_base.rpmdb.returnPackages()
+
+    return yum_base.rpmdb.returnPackages()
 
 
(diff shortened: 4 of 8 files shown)


.pylintrc
```diff
@@ -23,7 +23,24 @@ disable=
  old-style-class,
 # "R" Refactor recommendations
  too-few-public-methods,
+# do not warn about pylint checks being disabled through comments in the code
+ locally-disabled,
+# For some reason, pylint doesn't get right instance members, and some types could not be inferred
+# Eg: [E1103 maybe-no-member] Instance of 'RootLogger' has no 'task' member File: convert2rhel/main.py, line 60, in main
+# we have several logger task, and good approach to disable here, instead of each line
```
sturivny, line 30: please update to: `we have several logger tasks, and a good `

Makefile
```diff
@@ -21,3 +21,9 @@ tests:
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos5 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos6 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos7 ./run_unit_tests.sh
+
+lint:
+	bash -c "[[ ! -z ./convert2rhel/ ]] && find ./convert2rhel/ -name '*.py' | sort -u | xargs pylint --rcfile=.pylintrc"
```
SpyTec, line 26: What benefits does this provide as opposed to `pylint convert2rhel/`?

fellipeh: You need to pass `pylint --rcfile=.pylintrc convert2rhel`  or just `make lint`

The goal here is to have `make lint` to show all lint messages and `make lint-errors` top show only the errors

SpyTec: Right, but why do we need to use so many commands and pipe into pylint? Why can't this line simply be `pylint convert2rhel/`?

.github/workflows/pylint.yml
```diff
@@ -21,11 +21,19 @@ jobs:
...
+
     - name: Run Pylint Checks
       run: |
-        pylint --rcfile=.pylintrc convert2rhel/
-        
+         function interpreted_pylint() {
+            local es=0
+            pylint --rcfile=.pylintrc convert2rhel/; es=$?
+            case $es in
+                 1|2|3)  return $es ;;
```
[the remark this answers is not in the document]
AloisMahdal: I'm not familiar with bitwise arithmetic so can't check that part but otherwise it looks like it should work.

Just a nitpick, better coding practice is to use `local` when introducing new vars, and ideally just declare them (perhaps with default values and/or descriptive in-line comments) in function header.  Doesn't matter too much in a one-off throw-away function, though. :)

Anyway, if pylint has this compex way of reporting result, it seems worth considering writing the wrapper as separate binary that does all the arithmetic and perhaps shows real summary of what pylint wanted to say.  (Or see if such tool already exists.)  I could help write one in Bash, OTOH you would not need to use Bash at all at that point.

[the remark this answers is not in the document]
bocekm: Thanks for the remarks. It still doesn't work: ` ##[error]Process completed with exit code 16.` In the output I see `shell: sh -e {0}`. I don't know what that means. The man page for sh is so huge and `-e` is mentioned there hundreds of time. I think we'll just drop this effort and try to fix all the messages.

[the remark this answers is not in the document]
bocekm: Or, @fellipeh, please create a task for the infra team so they make sure the checks fail only if there are fatal and error messages. Unfortunately in the mean time the check will not be useful for us at all because it will be failing all the time.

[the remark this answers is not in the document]
AloisMahdal: Not sure about `{0}` (probably irrelevant to shell) but `-e` would explain a lot.  It enables the mode when shell [will exit on first (untreated) non-zero](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) exit status.  It's actually common for pipeline shells to run in this mode.

You can disable it by adding `set +e`.

(Something like this should also work

    pylint --yada-yada .. || es=$?

but you would need to make sure `es` is initialzed to 0.)

fellipeh:
Createda ticket for infra team to loo at these issue.

.github/workflows/pylint.yml
```diff
@@ -21,11 +23,20 @@ jobs:
...
+         function interpreted_pylint() {
+            local es=0
+            pylint --rcfile=.pylintrc convert2rhel/; es=$?
+            errors=$(( es&3 ))
+            case $errors in
+                 1|2|3)  return $es ;;
+            *) return 0 ;;
+            esac
+         }
+         interpreted_pylint
```
SpyTec, line 38: @fellipeh Could we try something like this?
```suggestion
         function interpreted_pylint() {
            local es=0
            pylint --rcfile=.pylintrc convert2rhel/; es=$?
            if ! [ $((es & 3)) -eq 0 ]; then
                return $es;
            fi;
            return 0;
         }
         interpreted_pylint
```

sturivny: @fellipeh please, take a look previous comment from @SpyTec ;)

Makefile
```diff
@@ -19,3 +19,9 @@ images:
 tests:
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos6 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos7 ./run_unit_tests.sh
+
+lint:
+	bash -c "[[ ! -z ./convert2rhel/ ]] && find ./convert2rhel/ -name '*.py' | sort -u | xargs pylint --rcfile=.pylintrc"
+
+lint-errors:
+	bash -c "[[ ! -z ./convert2rhel/ ]] && find ./convert2rhel/ -name '*.py' | sort -u | xargs pylint --rcfile=.pylintrc --errors-only"
```
SpyTec, line 27:
```suggestion
lint:
	pylint --rcfile=.pylintrc ./convert2rhel/

lint-errors:
	pylint --rcfile=.pylintrc --errors-only ./convert2rhel/
```

.github/workflows/pylint.yml
```diff
@@ -21,11 +23,19 @@ jobs:
...
         pip install astroid==1.2
-        
+
     - name: Run Pylint Checks
       run: |
-        pylint --rcfile=.pylintrc convert2rhel/
-        
+         function interpreted_pylint() {
+            local es=0
+            pylint --rcfile=.pylintrc convert2rhel/; es=$?
```
RheaAyase, line 31:
```suggestion
            pylint --rcfile=.pylintrc convert2rhel/ || es=$?
```
This code is executed with bash pipefail which means that if we use `;` here, the whole application will fail on non-zero return code. To ensure that it doesn't fail through we need to use `or` and execute our assignment on the right hand side.

.github/workflows/pylint.yml
```diff
@@ -40,7 +50,15 @@ jobs:
...
         pip install setuptools
         pip install pylint
-        
+
     - name: Run Pylint Checks
       run: |
-        pylint --rcfile=.pylintrc convert2rhel/
+         function interpreted_pylint() {
+            local es=0
+            pylint --rcfile=.pylintrc convert2rhel/; es=$?
```
RheaAyase, line 58:
```suggestion
            pylint --rcfile=.pylintrc convert2rhel/ || es=$?
```
This code is executed with bash pipefail which means that if we use `;` here, the whole application will fail on non-zero return code. To ensure that it doesn't fail through we need to use `or` and execute our assignment on the right hand side.

.github/workflows/pylint.yml
```diff
@@ -21,11 +23,19 @@ jobs:
...
         pip install pylint==1.1.0
         pip install astroid==1.2
-        
+
     - name: Run Pylint Checks
       run: |
-        pylint --rcfile=.pylintrc convert2rhel/
-        
+         function interpreted_pylint() {
+            local es=0
```
RheaAyase, line 30: I'd also suggest using more verbose variable name. From a glance I don't don't know what this variable stands for. (e.g. `error` or `errorCode` etc...)

Makefile
RheaAyase: Indeed that big line doesn't seem to achieve anything better than letting pylint go through the filestructure.

fellipeh:
Thansk @RheaAyase  very much... you are right, and now it's works!

.pylintrc
```diff
@@ -6,9 +6,14 @@ function-rgx=[a-z_][a-z0-9_]{2,50}$
 [MESSAGES CONTROL]
 
 disable=
+# Disabling "W" Warnings for stylistic problems or minor programming issues, as discussed on our stand up
+ W,
```
bocekm, line 10: I'd not disable all the warning messages. First, this PR is about errors. Second, there's many warning messages and when I look at them some of them are not good to dismiss, like:
`Assigning to function call which only returns None`
`Duplicate key %r in dictionary`
`Redefining built-in %r`
`Redefining name %r from outer scope (line %s)`
`Unreachable code`
`Unused import %s`
`Unused variable %r`

.pylintrc
```diff
@@ -23,7 +28,33 @@ disable=
...
  too-few-public-methods,
-
+# do not warn about pylint checks being disabled through comments in the code
+ locally-disabled,
+# For some reason, pylint doesn't get right instance members, and some types could not be inferred
+# Eg: [E1103 maybe-no-member] Instance of 'RootLogger' has no 'task' member File: convert2rhel/main.py, line 60, in main
+# we have several logger task, and s good approach to disable here, instead of each line
+ maybe-no-member,
+# Disabling duplicate code message, since do not make sense to refactor right now all duplicates code we have
+ duplicate-code,
```
bocekm, line 38: It's not an error. I'd not disable it since the pylint checker will not fail on that.
I'm sure there will many many more refactoring/stylistic/convention messages. We can deal with them later, instead of just "hiding" them by disabling the error.
Also, right now we are not stressed by some delivery deadlines so we can find a time to do some refactoring here and there.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -550,7 +557,7 @@ def test_replace_non_rhel_installed_kernel(self):
         self.assertEqual(utils.download_pkg.called, 1)
         self.assertEqual(utils.download_pkg.pkg, "kernel-4.7.4-200.fc24")
         self.assertEqual(utils.run_subprocess.cmd,
-                         "rpm -i --force --replacepkgs %skernel-4.7.4-200.fc24*" % utils.TMP_DIR)
+                         "rpm -i --force --replacepkgs /var/lib/convert2rhel/kernel-4.7.4-200.fc24*")
```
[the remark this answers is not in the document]
fellipeh: Sorry.. I mess up with some merge, as you could see.. very sorry..

.pylintrc
```diff
@@ -6,9 +6,14 @@ function-rgx=[a-z_][a-z0-9_]{2,50}$
 [MESSAGES CONTROL]
 
 disable=
+# Disabling "W" Warnings for stylistic problems or minor programming issues, as discussed on our stand up
+ W,
+# Disabling convention, for programming standard violation messages
+#C,
 # "E" Error for important programming issues (likely bugs)
+# skip redefined function false positive
+ E0102,
```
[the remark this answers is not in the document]
fellipeh: Put into pylinttrc file so we can't go for each line to disable it...

fellipeh:
@bocekm this PR is ready to merge?

.pylintrc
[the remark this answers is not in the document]
bocekm: Ok, please remove the `# pylint: disable=function-redefined` in-line `convert2rhel/unit_tests/subscription_test.py`.

.github/workflows/pylint.yml
```diff
@@ -21,11 +23,19 @@ jobs:
         python get-pip.py
         pip install pylint==1.1.0
         pip install astroid==1.2
-        
+
     - name: Run Pylint Checks
       run: |
```
bocekm, line 28:
```suggestion
      run: ${{ interpreted_pylint }}
```
You can try adding this to the second line, below `name: pylint`:
```
interpreted_pylint: |
    function interpreted_pylint() {
        local error_code=0
        pylint --rcfile=.pylintrc convert2rhel/ || error_code=$?
        if ! [ $((error_code & 3)) -eq 0 ]; then
            return $error_code;
        fi;
        return 0;
     }
     interpreted_pylint
```

.pylintrc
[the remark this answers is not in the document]
bocekm: Could you please remove the comment as I described in the above comment?

.github/workflows/pylint.yml
bocekm: I guess it's not that important..

PR #72, opened 2020-06-05
Fix parsing RHSM output

bocekm:
The output of `subscription-manager list --available` has changed [1] and due to that the convert2rhel was unable to list any available subscription.

[1] Specifically the output of `subscription-manager list --available` where the `System Type:` changed to `Entitlement Type:`. The change happened probably between RHEL 7.7 and 7.8. Now the code is able to parse both the old and new output.

Resolves: OAMG-3337

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,12 +12,10 @@ clean:
 
 .PHONY: images
 images:
-	@docker build -f Dockerfiles/centos5 -t $(IMAGE)/centos5 .
 	@docker build -f Dockerfiles/centos6 -t $(IMAGE)/centos6 .
 	@docker build -f Dockerfiles/centos7 -t $(IMAGE)/centos7 .
 
 .PHONY: tests
 tests:
-	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos5 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos6 ./run_unit_tests.sh
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos7 ./run_unit_tests.sh
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+from collections import namedtuple
 import os
 import re
 import shutil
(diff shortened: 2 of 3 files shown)


SpyTec:
> TODO: Update tests

Is this ready for review?

bocekm:
@SpyTec, yes, it is. I forgot to update the desc, thanks for reminding me.

Makefile
```diff
@@ -12,12 +12,10 @@ clean:
 
 .PHONY: images
 images:
-	@docker build -f Dockerfiles/centos5 -t $(IMAGE)/centos5 .
```
fellipeh, line 15: These is already in PR #75.

Makefile
```diff
@@ -12,12 +12,10 @@ clean:
...
 	@docker build -f Dockerfiles/centos6 -t $(IMAGE)/centos6 .
 	@docker build -f Dockerfiles/centos7 -t $(IMAGE)/centos7 .
 
 .PHONY: tests
 tests:
-	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos5 ./run_unit_tests.sh
```
fellipeh, line 21: These is already in PR #75.

convert2rhel/subscription.py
```diff
@@ -234,63 +235,29 @@ def get_avail_subs():
...
-        return sub_attr.group(1)
+    pool_id = re.search(r"^Pool ID:\s+(.*?)$",
+                        sub_raw_attrs,
+                        re.MULTILINE | re.DOTALL)
+    if pool_id:
+        return pool_id.group(1)
+
     else:
-        raise ValueError(descr)
+        loggerinst.critical("Cannot parse the subscription pool ID from string:\n%s" % sub_raw_attrs)
```
fellipeh, line 260: I think *else* is unnecessary here

```  python
if pool_id:
        return pool_id.group(1)

loggerinst.critical("Cannot parse the subscription pool ID from string:\n%s" % sub_raw_attrs)
```

Makefile
bocekm: Yes, but that one is not merged yet. Without this change, the tests would be failing.

convert2rhel/subscription.py
bocekm: You're right.

Makefile
bocekm: It's not a problem, I'll resolve the conflict either here or in #75, based on whichever PR will be merged first.

PR #74, opened 2020-06-09
Create the temp dir in spec

bocekm:
- to let convert2rhel rpm to be its owner
- use /var/lib/convert2rhel instead of /tmp/convert2rhel as the latter
  is cleared upon booting

### DIFF
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -44,7 +44,7 @@ class Color:
 # Absolute path of a directory holding data for this tool
 DATA_DIR = "/usr/share/convert2rhel/"
 # Directory for temporary data to be stored during runtime
-TMP_DIR = "/tmp/convert2rhel/"
+TMP_DIR = "/var/lib/convert2rhel/"
 
 
 def format_msg_with_datetime(msg, level):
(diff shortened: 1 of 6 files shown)


convert2rhel/utils.py
```diff
@@ -382,17 +382,17 @@ def backup(self):
         """ Save current version of a file """
         loggerinst = logging.getLogger(__name__)
         loggerinst.info("Backing up %s" % self.filepath)
-        if (os.path.isfile(self.filepath) and
-                os.path.isdir(TMP_DIR)):
+        if not os.path.isdir(TMP_DIR):
```
fellipeh, line 385: Maybe check if the dir is writable?  using `os.access('/path/to/folder', os.W_OK)`  W_OK for write and R_OK for read
https://docs.python.org/2.7/library/os.html#os.access

bocekm: Thanks. It actually made me realize the condition for checking the TMP_DIR is not necessary since there's the copying right below and if that raises an IOError (for example for the reason of the TMP_DIR not being accessible) we abort the conversion with printing the reason.

PR #75, opened 2020-06-09
Drop support for conversions to RHEL 5

bocekm:
RHEL 5 is reaching the end of the Extended life cycle support (ELS) which happens on November 30, 2020.
Reasons to drop the support:
- There's little sense to convert to a system that will stop being supported soon
- To make the convert2rhel developer's life easier as the code won't need to be python 2.4 compatible any more
- There's no EPEL for CentOS/OL 5 where we could be shipping the package

I'm removing the "system" tests as well as they are not properly functional at this point and we're not planning to use them in the future. They're going to be replaced by the OAMG Murphy CI.


ZhukovGreen:
@bocekm if the ELS going to happen on Nov 30, it might be a good idea to introduce a deprecation warning now

bocekm:
@ZhukovGreen, the thing is that we do not release the convert2rhel package for conversions to RHEL 5 anywhere. So unless somebody builds the packages by themselves, it's not being used.

And the words from our Product Manager have been clear - to not spend any time with work on RHEL 5. And, if we keep the tool working on CentOS/OL 5 with a deprecation warning, it would mean to have the code both py2.4+ & py3.6+ compatible. Because we are starting the work on conversions to RHEL 8. That would be a lot of extra work.

ZhukovGreen:
@bocekm thx for the clarification

sturivny:
RHEL 5 mentions are still in:
```
convert2rhel/unit_tests/redhatrelease_test.py:33:    supported_rhel_versions = ["5", "6", "7"]`
convert2rhel/main.py:96:        # (RHEL 5) - 2.7 (RHEL 7) compatibility
```
Python 2.4 is still in:
```
convert2rhel/pkghandler.py method: print_pkg_info(pkgs)
KeyError is for Python 2.4, AttributeError is for Python 2.6+ due

convert2rhel/main.py method: main()
Catching the three exception types separately due to python 2.4

convert2rhel/utils.py class DictWListValues(duct)
Python 2.4 replacement for Python 2.5+ collections.defaultdict(list).
it is used in repo.py
 
convert2rhel/unit_tests/__init__.py in method: wrap(func)
# NOTE: finally need to be used in this way for Python 2.4

```

bocekm:
Rebased and resolved conflicts.

bocekm:
> RHEL 5 mentions are still in:
> 
> ```
> convert2rhel/unit_tests/redhatrelease_test.py:33:    supported_rhel_versions = ["5", "6", "7"]`
> convert2rhel/main.py:96:        # (RHEL 5) - 2.7 (RHEL 7) compatibility
> ```
> 

@sturivny, thanks for the review. I've removed the mentions of RHEL 5. But not the mentions of Python 2.4. Under this PR, I didn't mean to remove all the Python 2.4 code constructs and rewrote them to be 2.6/2.7-only compatible. The purpose of this PR is to get rid of the code that exists solely due to the RHEL 5. When we'll start working on the Python 3.6+ compatibility later, we'll need to get rid of the Python 2.4 constructs anyway.

bocekm:
@SpyTec, @sturivny, could you please re-review?

sturivny:
LGTM

PR #78, opened 2020-06-19
feat: SUSE ES configs

SpyTec:


convert2rhel/data/6/x86_64/configs/suse_es-6-x86_64.cfg
```diff
@@ -0,0 +1,10 @@
+[system_info]
+
```
bocekm, line 2:
```suggestion

# SUSE ES packages are not signed
gpg_fingerprints =
```
This would cause an error due to the `gpg_fingerprints` key not being found:
https://github.com/oamg/convert2rhel/blob/29fd6a298e6d3a0c02db967dcb26202042ef9de6/convert2rhel/systeminfo.py#L130

convert2rhel/data/7/x86_64/configs/suse_es-7-x86_64.cfg
```diff
@@ -0,0 +1,12 @@
+[system_info]
+
+# SUSE ES packages are not signed
+gpg_fingerprints =
```
Conan-Kudo, line 4: This seems off, I don't think this is true...

bocekm: From what we know, it is true. Unless SUSE changed it in the recent year. Do you have access to [a RHEL system updated/patched through the SUSE Manager](https://www.suse.com/c/know-suse-supporting-rhat-years/) to verify that the packages provided by SUSE are really signed?

Conan-Kudo: I'm checking into this...

Conan-Kudo: Unfortunately, this is, in fact, true. The package builds are not done by SUSE, so they aren't signed with SUSE's key.

PR #81, opened 2020-06-23
Fix: throws message when not running as root

fellipeh:
Currently logging is instantiated before utils.require_root() is called which throws a permissions exception.  This PR fix these bug throwing a error message and exiting with code 1, before initialization process started.

### DIFF
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -52,13 +52,16 @@ class TestUtils(unittest.TestCase):
             return self.output, self.ret_code
 
     class DummyGetUID(unit_tests.MockFunction):
+        def __init__(self, uid):
+        	self.uid = uid
         def __call__(self, *args, **kargs):
-            return 1
+            return self.uid
 
-    @unit_tests.mock(os, "geteuid", DummyGetUID())
+    @unit_tests.mock(os, "geteuid", DummyGetUID(1000))
     def test_require_root_is_not_root(self):
         self.assertRaises(SystemExit, utils.require_root)
 
+    @unit_tests.mock(os, "geteuid", DummyGetUID(0))
     def test_require_root_is_root(self):
         self.assertEqual(utils.require_root(), None)
 
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -17,6 +17,7 @@
 
 
 import re
+import os
 
 try:
     import unittest2 as unittest  # Python 2.6 support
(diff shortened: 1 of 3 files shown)


convert2rhel/utils.py
```diff
@@ -60,9 +60,9 @@ def get_executable_name():
 
 def require_root():
     if os.geteuid() != 0:
-        loggerinst = logging.getLogger(__name__)
-        loggerinst.critical("The tool needs to be run under the root user.")
-    return
+        print "The tool needs to be run under the root user."
+        print("\nNo changes were made to the system.")
+        sys.exit(1)
```
SpyTec, line 65: Should be reverted as mentioned earlier

fellipeh: Actually that's was intentional.. if you use loggerinst, will raise a exception, because they will try to open the log file without permission, that's the reason to move to print.

SpyTec: Ack, then one simple fix for print:
```suggestion
        print("The tool needs to be run under the root user.")
        print("\nNo changes were made to the system.")
        sys.exit(1)
```

fellipeh: Still not work, when you initialize the logger, they try to open the log file... And as these message does not need to be logged (since it's only a message before to run the tool), the print is a good idea.. 
Fixing forgot `(`..

convert2rhel/unit_tests/utils_test.py
```diff
@@ -50,6 +51,17 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
 
+    class DummyGetUID(unit_tests.MockFunction):
+        def __call__(self, *args, **kargs):
+            return 1
+
+    @unit_tests.mock(os, "geteuid", DummyGetUID())
+    def test_require_root_is_not_root(self):
+        self.assertRaises(SystemExit, utils.require_root)
+
+    def test_require_root_isroot(self):
```
SpyTec, line 62: Add a mock for `geteuid` on this one as well, relying on the UID of whichever is running the tests means it's a flaky test

Also for consistency:
```suggestion
    def test_require_root_is_root(self):
```

convert2rhel/unit_tests/utils_test.py
```diff
@@ -50,6 +51,17 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
             self.called += 1
             return self.output, self.ret_code
 
+    class DummyGetUID(unit_tests.MockFunction):
+        def __call__(self, *args, **kargs):
+            return 1
```
SpyTec, line 56:
```suggestion
    class DummyGetUID(unit_tests.MockFunction):
        def __init__(self, uid):
        	self.uid = uid
        def __call__(self, *args, **kargs):
            return self.uid
```

convert2rhel/unit_tests/utils_test.py
```diff
@@ -50,6 +51,17 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
+
+    @unit_tests.mock(os, "geteuid", DummyGetUID())
```
SpyTec, line 58:
```suggestion
    @unit_tests.mock(os, "geteuid", DummyGetUID(1000))
```

convert2rhel/unit_tests/utils_test.py
```diff
@@ -50,6 +51,17 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
+    def test_require_root_is_not_root(self):
+        self.assertRaises(SystemExit, utils.require_root)
+
+    def test_require_root_is_root(self):
```
SpyTec, line 62:
```suggestion
    @unit_tests.mock(os, "geteuid", DummyGetUID(0))
    def test_require_root_is_root(self):
```

PR #82, opened 2020-06-23
Conversions to RHEL 8

### LINKED ISSUE #24: RFE: centos8 -> rhel8
While centos8 is new, there may still be those who'd like to switch to the real thing.

SpyTec:
Allow conversions of CentOS 8 to RHEL 8. That includes making the code Python 3 compatible.

Done:
- The code is both python 2 and 3 compatible.
- The unit testing environment includes CentOS 8.
- Pylint is executed with python 3 as well.
- Pylint checks are updated to run on latest CentOS minor versions.
- Dnf is used on CentOS 8, yum on CentOS/OL 6 and 7.
- The rpm package is possible to build on CentOS 8.
- The script for building the RPM package locally on CentOS/OL is updated to support also CentOS/OL 8.
- The python2 shebang is removed from setup.py - the file is both py2 & py3 compatible.
- Spec file updates:
  - Spec linting errors on CentOS/OL 6 are avoided by using defattr.
  - %epel was changed to %rhel in the spec file as it's possible to use %rhel even in a spec file for EPEL.
  - buildsys-macros is removed as that was needed only for RHEL5 and its derivates which are not supported by convert2rhel anymore.
  - The epel-rpm-macros dependency is removed - it's not necessary for building a package for EPEL.
  - The EPEL and CentOS/OL spec files are merged into one, there is no need to differentiate them anymore.
  - The submgr-related deps are identified - these should be removed in the subsequent PRs. We should leave yum to handle those.
  - The sed and gnupg2 deps are removed as they are not used in the code anymore.
- When printing package information in a table, we used to print a package vendor. However, when using dnf now, dnf does not provide the vendor information (https://bugzilla.redhat.com/show_bug.cgi?id=1876561). So instead, we print a package packager now.
- There's a workaround implemented to get a pkg fingerprint on CentOS 8. It's due to the fact that it's not currently possible to get a pkg header through the dnf API (https://bugzilla.redhat.com/show_bug.cgi?id=1876606).
- Unused parameters are removed from the run_unit_tests.sh script, like coverage, output coloring and clearing logging handlers.
- The `run_unit_tests.sh` script is used in the Testing Farm instead of executing `nosetests` command directly (which would not work on CentOS 8 where the binary is `nosetest-3`).
- Multiple default RHSM repoids got allowed in a config. Previously it was possible to have just one default RHSM repo in config file. It was a problem for RHEL 8 where there are two main repos that are supposed to be enabled (BaseOS and AppStream).

To do in a subsequent PR(s):
- Move yum/dnf-related code to a new module handling package management
- Handle patching yum.conf correctly on CentOS 8 (it's actually dnf.conf there) - this is not critical to do now as working with /etc/yum.conf still works under CentOS 8 - dnf keeps it there for backwards compatibility

Resolves #24.


bocekm:
You were a Maintainer, you're an Admin now.

From what I know, yum calls on CentOS 8 are redirected to dnf and dnf should accept them and be backwards compatible. What exactly have you bumped into that led you to thinking that it's not? We can discuss that with the dnf team.

However it's true that there's no yum python module available on CentOS 8. Thus, on the place where we import and use it, we need to add a condition so that the yum module is used on CentOS 6 and 7, and dnf on CentOS 8.

SpyTec:
> You were a Maintainer, you're an Admin now.
Thanks!

> From what I know, yum calls on CentOS 8 are redirected to dnf and dnf should accept them and be backwards compatible. What exactly have you bumped into that led you to thinking that it's not? We can discuss that with the dnf team.
>
> However it's true that there's no yum python module available on CentOS 8. Thus, on the place where we import and use it, we need to add a condition so that the yum module is used on CentOS 6 and 7, and dnf on CentOS 8.

So, first thing I noticed was we mock out `yum.rpmsack.RPMDBPackageSack.returnPackages()` within a test. Since `rpmsack` isn't part of DNF we need to figure out a different way to handle it

As for the code itself,

* `YumBase` was renamed to `Base` in DNF, so we need to handle it differently depending on which we have imported
* [`doConfigSetup()` was removed](https://github.com/rpm-software-management/dnf/commit/a6304bcca9c967c267c0894dcc2823ba5654c5aa), so we need to use some alternative solution there. At a glance it seems we can get around this without having to do special handling for Yum and DNF separately
* `rpmdb` in Base seems to be renamed to `rpm`
* `returnPackages` doesn't seem to be in DNF either

That's what I found out so far. There might very well be alternative solutions within dnf that is also in yum, but not looked into it

bocekm:
In dnf python module, we can use this sack:
```
import dnf
b = dnf.Base()
b.fill_sack(load_system_repo=True, load_available_repos=False)
for pkg in b.sack.query():
    print(pkg, pkg._from_repo)
```
I was doing some investigation around that because of leapp: https://bugzilla.redhat.com/show_bug.cgi?id=1789840.

convert2rhel/systeminfo.py
```diff
@@ -115,9 +123,8 @@ def _get_cfg_section(self, section_name):
                                  " conversion to RHEL.")
 
         options_list = cfg_parser.options(section_name)
-        return dict(zip(options_list,
-                        map(lambda opt: cfg_parser.get(section_name, opt),
-                            options_list)))
+        return dict(list(zip(options_list,
```
ZhukovGreen, line 126: It should work without list, i.e. dict(zip...

SpyTec: Good catch, most of the code changes are from a python 2/3 compatibility tool. I looked it up though and in Py2 zip returns a list, whilst in Py3 it returns an iterator

I can't check whether it works since the tests for Py3 doesn't work yet, but I'll leave it as `list(zip())`

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -241,9 +247,19 @@ def __call__(self, patterns=None):
...
-    def test_get_installed_pkg_objects(self):
+    try:
+        @unit_tests.mock(yum.rpmsack.RPMDBPackageSack, "returnPackages",
+                         ReturnPackagesMocked())
+        def test_get_installed_pkg_objects_yum(self):
+            self.get_installed_pkg_objects()
+    except AttributeError:
+        # TODO: find out what we need to mock
+        @unit_tests.mock(yum.query.Query, "__repr__",
+                         ReturnPackagesMocked())
```
SpyTec, line 258: Mocking out anything from `dnf.query.Query` throws an error. DNF internally uses hawkey/libdnf to handle packages but unsure how to fix this

```py
======================================================================
ERROR: test_get_installed_pkg_objects_dnf (convert2rhel.unit_tests.pkghandler_test.TestPkgHandler)
----------------------------------------------------------------------
  File "/data/convert2rhel/unit_tests/__init__.py", line 68, in wrapped_fn
    setattr(class_or_module, orig_obj, mock_obj)
TypeError: can't set attributes of built-in/extension type '_hawkey.Query'
```

SpyTec: Thanks, managed to find a solution!

That test seems to pass now, on to squash different bugs

convert2rhel/pkghandler.py
```diff
@@ -15,15 +15,23 @@
...
+
 import logging
 import os
 import re
-import yum
+
+try:
+    import dnf as yum  # as yum to avoid errors accessing dnf if it's not imported
+except ImportError:
+    import yum
```
Conan-Kudo, line 29: I strongly recommend you encapsulate this in a wrapper module. This code looks kind of gnarly as it is, and the yum/dnf details make it harder to read.

SpyTec:
@bocekm is there a way to specify which fmf file to use in the tests? At the moment all three tests grab from `unittests.fmf`

bocekm:
@SpyTec, I think that [this packit conf section](https://github.com/oamg/convert2rhel/blob/master/.packit.yaml#L29) makes sure the testing farm runs all tests defined under the `plans/` folder. What do you mean by `all three tests`? Do you mean running the unit tests for CentOS 6, 7 and 8? What exactly do you want to achieve? Run the unit tests on CentOS 8 using different .fmf file than [the only one we have now](https://github.com/oamg/convert2rhel/blob/master/plans/unittests.fmf)?

SpyTec:
> Run the unit tests on CentOS 8 using different .fmf file than the only one we have now?

@bocekm That's correct, since the installation of things are slightly different in CentOS 8 than 6 and 7

bocekm:
I don't really know. I can see some kind of a filter you can use in .fmf, there's an example `filter: "tier: 1 & distro: rhel-8"` at https://packit.dev/testing-farm/.
@thrix, can you tell us how to configure packit to execute some tests defined in the fmf files with EPEL 6 and 7 targets and other tests with EPEL 8?

convert2rhel/toolopts.py
```diff
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+from builtins import object
```
bocekm, line 18: When building a pkg on both RHEL 6 and 7.
```
  File "setup.py", line 22, in <module>
    from man.build_manpage import build_manpage
  File "/builddir/build/BUILD/convert2rhel-v0.11/man/__init__.py", line 18, in <module>
    from convert2rhel import toolopts
  File "/builddir/build/BUILD/convert2rhel-v0.11/convert2rhel/toolopts.py", line 18, in <module>
    from builtins import object
ImportError: No module named builtins
```
https://download.copr.fedorainfracloud.org/results/@oamg/convert2rhel/epel-6-x86_64/01554887-convert2rhel/build.log.gz

SpyTec:
I've gotten the tests to work now for 6,7, and 8. Seemingly all tests pass for 8, which is a good sign. Though all the logger calls are being outputted to stdout for 8, @bocekm is there anything in the code that I haven't spotted that changes this for us within tests or is it handled automatically by unittest package in 6 and 7?

convert2rhel/pkghandler.py
SpyTec: Thanks, fixed

SpyTec:
> I've gotten the local container tests to work now for 6,7, and 8. Seemingly all tests pass for 8, which is a good sign. Though all the logger calls are being outputted to stdout for 8, @bocekm is there anything in the code that I haven't spotted that changes this for us within tests or is it handled automatically by unittest package in 6 and 7?

Found out there was a `-b` argument to the unittest module. Likewise, I added `-v` so it matches our other tests in being more verbose

Found two issues I need to investigate.
These tests take way too long, probably faulty mocks
`test_call_yum_cmd_w_downgrades_continuous_fail (convert2rhel.unit_tests.pkghandler_test.TestPkgHandler)`
`test_call_yum_cmd_w_downgrades_one_fail (convert2rhel.unit_tests.pkghandler_test.TestPkgHandler)`

SpyTec:
There does not seem to be a tangible reason why those two tests are slow. Initial testing seem to indicate unittests itself is hanging up on something, as it gets stuck before we make any calls and before any mocks in the decorator get called. I didn't attach a debugger to find the root cause

convert2rhel/pkghandler.py
```diff
@@ -175,30 +179,54 @@ def get_installed_pkgs_w_fingerprints(name=""):
...
     return yum_base.rpmdb.returnPackages()
 
+def _get_installed_pkg_objects_dnf(name):
+    dnf_base = pkgmanager.Base()
+    dnf_base.fill_sack(load_system_repo=True, load_available_repos=False)
+    query = dnf_base.sack.query()
+    installed = query.installed()
+    if name:
+        installed.filter(name=name)
+    return list(installed)
```
Conan-Kudo, line 229: These `*_(yum|dnf)` methods should be moved into the pkgmanager module. What I expected to see from my feedback is that all the gunk you're doing to do either YUM or DNF would be shunted into classes for YUM and DNF variants of the function, and then the pkghandler module would instantiate the YUM or DNF variant depending on import availability (preferring DNF, obviously).

bocekm:
Thanks, @Conan-Kudo. I've updated the spec per your suggestions: https://github.com/oamg/convert2rhel/pull/82/commits/c40f44e5fa01506c11986db0d875455ce3432db7.

packaging/epel/convert2rhel.spec
```diff
@@ -1,6 +1,12 @@
+%if 0%{?epel} && 0%{?epel} <= 7
```
Conan-Kudo, line 1: Why are you using `%epel`? There are very few circumstances where using `%epel` makes sense over `%rhel`.

bocekm: Hi @Conan-Kudo, the reason is the following paragraph in [the Fedora Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_spec_legibility):
> To help facilitate legibility, only macros and conditionals for Fedora and EPEL are allowed to be used in Fedora Packages. Use of macros and conditionals for other distributions, including Fedora derivatives, is not permitted in spec files of packages in the main Fedora repositories unless those macros and conditionals are also present in Fedora.

If it's not that hot with this rule, I can definitely change it to `%rhel` (and maybe even merge the two spec files into one).

Conan-Kudo: `%rhel` is considered an EPEL conditional. All the macros exposed by RHEL and CentOS are permitted.

bocekm: Ok, I'll change this. Thanks for pointing that out.

bocekm:
I've reported the issue with tests failing with EPEL 8 using packit/testing farm: https://gitlab.com/testing-farm/cruncher/-/issues/8. It seems to be an infrastructure issue.

packaging/epel/convert2rhel.spec
```diff
@@ -11,37 +17,51 @@ License:        GPLv3+
 URL:            https://github.com/oamg/convert2rhel
 Source0:        %{url}/archive/v%{version}/%{name}-%{version}.tar.gz
 BuildArch:      noarch
-BuildRequires:  python-devel
-BuildRequires:  python-setuptools
+
 BuildRequires:  epel-rpm-macros
```
bocekm, line 21: @Conan-Kudo, is this dependency necessary? If I drop it, I could merge both spec files we have into one.

I know that it includes the license related macro, but I can define it in the spec myself (`%{!?_licensedir:%global license %%doc}`).

We have one that's used to build a package on CentOS/Oracle Linux, where `epel-rpm-macros` is not available, and then another spec, now almost the same, for building a package for EPEL.

Conan-Kudo: You shouldn't need it.

bocekm (dismissed):

convert2rhel/systeminfo.py
```diff
@@ -15,7 +15,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-import ConfigParser
+from future import standard_library
```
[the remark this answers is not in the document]
Conan-Kudo: `python3-future`/`python2-future` includes that module.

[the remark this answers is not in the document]
bocekm: I see, it's available in EPEL. Thanks.

convert2rhel/toolopts.py
bocekm: I've removed all the `builtins` imports. @SpyTec, were you adding them for any specific reason?
The thing is that the module is not available on CentOS/RHEL 6 and 7.

SpyTec: In our case I don't think we need it, there is a reason for it being there though https://stackoverflow.com/a/36517807

Wouldn't builtins be available through pip?

bocekm: An rpm package can only depend on other rpms. If it's not packaged in CentOS/OL, we can't import/depend on it.

SpyTec: Ah okay. I think we should be fine. If we need anything we can get around it by creating our own methods for it - luckily not too difficult

Conan-Kudo:
@bocekm `m2crypto` is available in EPEL 8 as `python3-m2crypto`.

bocekm:
@Conan-Kudo, I see, thanks. However, to be honest, I'm not sure what has been the reason for having m2crypto as a dependency. I would say was a dependency of a subscription-manager. And in RHEL, m2crypto is not available (not even as as python3-m2crypto), so if that's the case, subscription-manager will definitely not require m2crypto. But I'll dig deeper in that.

Conan-Kudo:
@bocekm If you don't need `m2crypto` drop the dependency entirely.

bocekm:
`m2crypto` is definitely a dependency of a subscription-manager on CentOS/OL/RHEL 7. And because `convert2rhel` installs the `subscription-manager` packages using `rpm -i` we need to have their deps installed already with convert2rhel.

I'll check exactly what deps does the CentOS/OL/RHEL 8 need.

Conan-Kudo:
@bocekm That makes no sense. Why not just depsolve and install the packages manually instead?

bocekm:
Yeah, we'll try and consider using `yum remove "subscription-manager-*" "*rhsm*"` and then `yum install </path/to/the/rpm/>`. That way yum would draw all the necessary deps.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -241,14 +249,68 @@ def __call__(self, patterns=None):
...
+            self.installed_called = 0
+            self.stop_iteration = False
+            self.pkg_obj = None
+
+        def __call__(self, *args):
+            self._setup_pkg()
+            return self
+
+        def __iter__(self):
+            return self
```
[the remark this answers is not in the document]
bocekm: @SpyTec, do you remember based on what did you create this `QueryMocked` object? dnf uses this library https://github.com/rpm-software-management/hawkey/blob/master/src/query.c which is installed on the CentOS system as a compiled shared object.
Is the __iter__ even necessary?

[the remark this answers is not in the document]
SpyTec: I know `__iter__` was there due to the test failing, since you need both `__iter__` and `__next__` for an iterator. I'm not sure if that is still the case but I would presume it does. 

Since this is in a test file I don't see a reason not to add a pylint exception though

bocekm:
Currently it's possible to convert CentOS 8 to RHEL 8 already. I've succeeded with the following command:
`convert2rhel --debug -y --no-rpm-va --disable-submgr --enablerepo rhel-8-for-x86_64-baseos-rpms --enablerepo rhel-8-for-x86_64-appstream-rpms`

bocekm:
I've reported the testing farm issue `Command '['git', 'clone', 'https://github.com/oamg/convert2rhel.git', 'source']' failed with exit code 128` at https://gitlab.com/testing-farm/cruncher/-/issues/9.

bocekm:
I've succeeded in converting CentOS 8 to RHEL 8 using RHSM as well:
`convert2rhel --debug -y --no-rpm-va --enablerepo rhel-8-for-x86_64-baseos-rpms --enablerepo rhel-8-for-x86_64-appstream-rpms --pool 8a99f9a96df7f938016e03207bd74547`

bocekm:
Squashed and rebased.

plans/unittests.fmf
```diff
@@ -2,10 +2,8 @@ summary:
...
     script:
-    - pwd; cd ~/git-source && pwd && nosetests -v
+    - pwd; cd ~/git-source && pwd && ./run_unit_tests.sh
 prepare:
     how: shell
     script:
-    - yum --enablerepo=extras install -y epel-release
-    - yum install -y python-pip
-    - pip install nose
+    - yum install -y python-nose
```
bocekm, line 9: @thrix, I want to run the unit tests with epel 6, 7, and 8. But with epel 8, there's `python3-nose` available, instead of `python-nose`. 

Do you suggest to create two fmf files, each using `relevancy`? E.g.
```
relevancy:
- "distro = rhel-6, rhel-7: True"
```
Or to rather use a shell condition? E.g.:
```
if [ command -v dnf ]; then dnf install python3-nose; else yum install -y python-nose; fi
```

convert2rhel/pkghandler.py
bocekm: I totally agree and I've started working on that but if you don't mind, I'd actually submit that as another PR. This one is huge enough on its own now and this additional work requires quite a lot of refactoring (I'd move all the yum/dnf-related code to the pkgmanager.py module, and there's quite a lot of it).

Conan-Kudo: I'm fine with it being a followup PR, but I think you should not make a release until after that change is done.

Conan-Kudo:
@bocekm Note for the future: `nose` is dead and you should look to replace with `pytest`.

convert2rhel/pkghandler.py
bocekm: Agreed.

plans/unittests.fmf
bocekm: Ah, the testing farm run `ln -fs $(which yum) /usr/bin/dnf` so I can't use the `command -v dnf` test.

bocekm:
Rebased to resolve a conflict.

plans/unittests.fmf
bocekm: Ok, this doesn't work as well: `if command -v python3 &> /dev/null ; then dnf install python3-nose; else yum install -y python-nose; fi`

bocekm: So, the `relevancy` is being ignored completely. I'm out of ideas.

bocekm: Related RFE: https://github.com/psss/tmt/issues/101

bocekm:
TBD: Rebase to master and handle the dnf version lock so that [clearing of the package version locks](https://github.com/oamg/convert2rhel/pull/92) still works.

fellipeh (approved):
LGTM

Just need to fix the conflict

SpyTec:
Seems good to merge. I've managed to get it working on 6,7,8 on RHSM and 6,7 without RHSM. I'm still unable to get it working without RHSM on 8.

fellipeh (approved):
LGTM, still need to fix the conflicts

bocekm:
Rebased and resolved conflicts.

bocekm:
@fellipeh, please re-review. Thanks.

fellipeh (approved):
LGTM

PR #83, opened 2020-06-24
fix:English locale before all c2r process

fellipeh:
Some regexp are hardcoded in English, the locale must be set to LANG=C before running the tool.

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -39,9 +39,6 @@ class ConversionPhase:
 
 
 def main():
-    # setting locale to LANG=C before start convert2rhel process
-    utils.set_env_locale()
-
     """Perform all steps for the entire conversion process."""
 
     # the tool will not run if not executed under the root user
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -39,6 +39,9 @@ class ConversionPhase:
 
 
 def main():
+    # setting locale to LANG=C before start convert2rhel process
+    utils.set_env_locale()
+
     """Perform all steps for the entire conversion process."""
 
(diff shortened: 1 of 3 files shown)


bocekm:
The PR is related to OAMG-3129, even though that one was in New, not in To Do, so we haven't had a change to discuss it.

One of the regexps that might be affected by the set locale is https://github.com/oamg/convert2rhel/blob/master/convert2rhel/subscription.py#L242.

I think it's not an issue if we set the locale just for the convert2rhel process. But we need to make sure that it's really just for the process.

SpyTec:
Right, I'd prefer to focus on problems than solutions. Lets discuss future things to avoid confusion and potentially going at it in the wrong direction.

As for per process locale, the [Python locale library](https://docs.python.org/2.6/library/locale.html) seems to be for the current process. @fellipeh I'd suggest to give that a go instead and experiment to see whether the locale is being set outside the process as well

fellipeh:
@bocekm  You are right, I see these ticket on my jira board, when you change something, and I think... it's easy to fix, takes less then a minute.. so, why not!?   Sorry about that...

@SpyTec   I tried that before, don't know the reason, but python locale lib appears to have no effect at all..

```python
>>>import locale
>>>locale.setlocale(locale.LC_ALL, 'C')
>>>locale.getlocale()
(None, None)
```
According to the doc: `Returns the current setting for the given locale category as sequence containing language code, encoding. category may be one of the LC_* values except LC_ALL`

So, it's not set during the execution for some reason.. That's the reason I set the os enviroment.

SpyTec (approved):

SpyTec:
LGTM

PR #84, opened 2020-06-30
Fix: Do not rename .repo files

fellipeh:
Currently, when using RHSM, convert2rhel renames all non-RHSM repofiles to make sure only the RHSM ones are used. That causes a problem when somebody passes --enablerepo together with using RHSM, and this repo is in some other repofile than /etc/yum.repos.d/redhat.repo. Because of the repofile renaming, this repository is then unexpectedly not used during the conversion.

Some unit tests need to be changed to reflect `--disable=*` from yum command

A/C:
- Add `--disable=*` as default for yum command (if user pass `--disablerepo` arg would have priority over that)
- When using RHSM, use `--enable=XXX` for each of enabled repository from these command: `subscription-manager repos --list-enabled`

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -69,31 +69,20 @@ def call_yum_cmd_w_downgrades(cmd, fingerprints):
     return
 
 
-def call_yum_cmd(command, args="", enablerepo=None, disablerepo=None,
-                 print_output=True):
+def call_yum_cmd(command, args="", print_output=True):
     """Call yum command and optionally print its output."""
     loggerinst = logging.getLogger(__name__)
 
-    cmd = "yum %s -y" % (command)
-    # disablerepo parameter must be added before the enablerepo parameter
+    cmd = "yum %s -y" % command
 
-    if disablerepo is None:
-        disablerepo = []
-    if disablerepo:
-        repos = disablerepo
-    else:
-        repos = tool_opts.disablerepo
-    for repo in repos:
-        cmd += " --disablerepo=%s " % repo
+    # The --disablerepo yum option must be added before --enablerepo,
+    #   otherwise the enabled repo gets disabled if --disablerepo="*" is used
+    for repo in tool_opts.disablerepo:
+        cmd += " --disablerepo=%s" % repo
 
-    if enablerepo is None:
-        enablerepo = []
-    if enablerepo:
-        repos = enablerepo
-    else:
-        repos = tool_opts.enablerepo
-    for repo in repos:
-        cmd += " --enablerepo=%s " % repo
+    for repo in tool_opts.enablerepo:
+        cmd += " --enablerepo=%s" % repo
+
     if args:
         cmd += " " + args
 
@@ -101,8 +90,7 @@ def call_yum_cmd(command, args="", enablerepo=None, disablerepo=None,
     # handle when yum returns non-zero code when there is nothing to do
     nothing_to_do_error_exists = stdout.endswith("Error: Nothing to do\n")
     if returncode == 1 and nothing_to_do_error_exists:
-        loggerinst.info("Return code 1 however nothing to do. Returning code 0"
-                        " ... ")
+        loggerinst.debug("Yum has nothing to do. Ignoring.")
         returncode = 0
     return stdout, returncode
 
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -26,6 +26,7 @@ from convert2rhel import pkghandler
 from convert2rhel import utils
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
 from convert2rhel.systeminfo import system_info
+from convert2rhel.toolopts import tool_opts
 
 
 class TestPkgHandler(unit_tests.ExtendedTestCase):
@@ -144,16 +145,28 @@ class TestPkgHandler(unit_tests.ExtendedTestCase):
         self.assertRaises(SystemExit, pkghandler.clear_yum_versionlock)
         self.assertEqual(pkghandler.call_yum_cmd.called, 0)
 
+    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked())
+    def test_call_yum_cmd(self):
+        pkghandler.call_yum_cmd("install")
+
+        self.assertEqual(utils.run_subprocess.cmd, "yum install -y")
 
     @unit_tests.mock(pkghandler, "call_yum_cmd", CallYumCmdMocked())
     def test_call_yum_cmd_w_downgrades_continuous_fail(self):
         pkghandler.call_yum_cmd.return_code = 1
 
-        self.assertRaises(SystemExit, pkghandler.call_yum_cmd_w_downgrades,
-                          "test_cmd", ["fingerprint"])
-        self.assertEqual(pkghandler.call_yum_cmd.called,
-                         pkghandler.MAX_YUM_CMD_CALLS)
+        self.assertRaises(SystemExit, pkghandler.call_yum_cmd_w_downgrades, "test_cmd", ["fingerprint"])
+        self.assertEqual(pkghandler.call_yum_cmd.called, pkghandler.MAX_YUM_CMD_CALLS)
 
+    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked())
+    @unit_tests.mock(tool_opts, "disablerepo", ['*'])
+    @unit_tests.mock(tool_opts, "enablerepo", ['rhel-7-extras-rpm'])
+    def test_call_yum_cmd_with_disablerepo(self):
+        pkghandler.call_yum_cmd("install")
+
+        self.assertEqual(utils.run_subprocess.cmd,
+                         "yum install -y --disablerepo=* --enablerepo=rhel-7-extras-rpm")
+        
     @unit_tests.mock(pkghandler, "get_installed_pkgs_by_fingerprint",
                      GetInstalledPkgsByFingerprintMocked())
     @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked())
@@ -167,12 +180,6 @@ class TestPkgHandler(unit_tests.ExtendedTestCase):
         pkghandler.call_yum_cmd.fail_once = True
 
         pkghandler.call_yum_cmd_w_downgrades("test_cmd", ["fingerprint"])
-
-    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked())
-    def test_call_yum_cmd(self):
-        pkghandler.call_yum_cmd("install")
-
-        self.assertEqual(utils.run_subprocess.cmd, "yum install -y")
 
     def test_get_problematic_pkgs(self):
         error_pkgs = pkghandler.get_problematic_pkgs("", [])
@@ -560,8 +567,7 @@ class TestPkgHandler(unit_tests.ExtendedTestCase):
 
         pkghandler.handle_no_newer_rhel_kernel_available()
 
-        self.assertEqual(utils.run_subprocess.cmd,
-                         "yum install -y kernel-4.7.2-201.fc24")
+        self.assertEqual(utils.run_subprocess.cmd, "yum install -y kernel-4.7.2-201.fc24")
 
     class ReplaceNonRhelInstalledKernelMocked(unit_tests.MockFunction):
         def __init__(self):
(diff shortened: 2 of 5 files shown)


convert2rhel/pkghandler.py
```diff
@@ -72,29 +73,42 @@ def call_yum_cmd(command, args="", enablerepo=None, disablerepo=None,
...
     if enablerepo is None:
         enablerepo = []
     if enablerepo:
         repos = enablerepo
     else:
         repos = tool_opts.enablerepo
+
+    if not tool_opts.disable_submgr:
+        if not system_info.submgr_enabled_pkgs:  # double check to get all RHSM enabled repo if not get yet
+            subscription.get_enabled_repos()
```
SpyTec, line 99: This seems like a useless call. `subscription.get_enabled_repos()` just returns a list of repo IDs. Is this to make sure we get `ret_code` 0?

We shouldn't call this more than once if it's the latter, one call should be enough to not slow down c2r

convert2rhel/subscription.py
```diff
@@ -23,11 +23,21 @@
...
 from convert2rhel import utils
+from convert2rhel.systeminfo import system_info
 
 _RHN_REGISTRATION_FILE = "/etc/sysconfig/rhn/systemid"
 rhn_reg_file = utils.RestorableFile(_RHN_REGISTRATION_FILE)  # pylint: disable=C0103
 
 
+def get_enabled_repos():
+    loggerinst = logging.getLogger(__name__)
+    stdout, returncode = utils.run_subprocess(cmd='subscription-manager repos --list-enabled')
```
SpyTec, line 34:
```suggestion
    stdout, returncode = utils.run_subprocess(cmd='subscription-manager repos --list-enabled', print_output=False)
```

convert2rhel/systeminfo.py
```diff
@@ -53,6 +53,9 @@ def __init__(self):
         self.cfg_content = None
         self.system_release_file_content = None
         self.logger = None
+        # list of all enabled packages for Subscription Manager
+        # get using "subscription-manager repos --list-enabled" command
+        self.submgr_enabled_pkgs = []
```
SpyTec, line 58: This is a list of repos, not packages
```suggestion
        # list of all enabled repositories for Subscription Manager
        self.submgr_enabled_repos = []
```

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -144,6 +144,17 @@ def __call__(self, *args, **kwargs):
...
 
+    @unit_tests.mock(subscription.logging, "getLogger", GetLoggerMocked())
+    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked([("Repo ID:   rhel-7-server-rpms", 0)]))
+    def test_get_enabled_repos(self):
+        pkgs = subscription.get_enabled_repos()
+        self.assertEqual(pkgs[0], 'rhel-7-server-rpms')
+
+    @unit_tests.mock(subscription.logging, "getLogger", GetLoggerMocked())
+    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked([("Repo ID:   rhel-7-server-rpms", 1)]))
+    def test_get_enabled_repos_error(self):
```
SpyTec, line 155:
```suggestion
    def test_get_enabled_repos_raises_if_submgr_fails(self):
```

convert2rhel/pkghandler.py
```diff
@@ -72,29 +73,42 @@ def call_yum_cmd(command, args="", enablerepo=None, disablerepo=None,
...
+    cmd = "yum %s -y" % command
     # disablerepo parameter must be added before the enablerepo parameter
 
     if disablerepo is None:
         disablerepo = []
     if disablerepo:
         repos = disablerepo
-    else:
+    elif tool_opts.disablerepo:
         repos = tool_opts.disablerepo
```
SpyTec, line 84: Do we have a test for this branch?

convert2rhel/subscription.py
```diff
@@ -23,11 +23,21 @@
...
+    if returncode == 0:
+        return re.compile(r'Repo ID:\s*([^\s]+)').findall(stdout)
+
+    loggerinst.critical("Not possible to get get enabled repo through subscription-manager:\n%s" % stdout)
```
SpyTec, line 38:
```suggestion
    loggerinst.critical("subscription-manager failed to return enabled repos:\n%s" % stdout)
```

convert2rhel/pkghandler.py
```diff
@@ -72,29 +73,42 @@ def call_yum_cmd(command, args="", enablerepo=None, disablerepo=None,
...
         repos = enablerepo
     else:
         repos = tool_opts.enablerepo
+
+    if not tool_opts.disable_submgr:
+        if not system_info.submgr_enabled_pkgs:  # double check to get all RHSM enabled repo if not get yet
+            subscription.get_enabled_repos()
+
+        for repo in system_info.submgr_enabled_pkgs:
+            repos.append(repo)
```
SpyTec, line 102: Add tests to make sure that the repos from submgr is added to the command. As well as a test that makes sure that `--disablerepo` can disable repos from submgr

SpyTec (dismissed):

convert2rhel/pkghandler.py
fellipeh: These code https://github.com/oamg/convert2rhel/pull/84/files#diff-46f43c5ca8816e809009b34648b60673R150   test if a repo was added in the command, because if the repo is enabled, is was added by the function..

convert2rhel/pkghandler.py
fellipeh: Added here: https://github.com/oamg/convert2rhel/pull/84/files#diff-cdec7561cf20f5b5b314293008180b42R83

convert2rhel/pkghandler.py
SpyTec: That test is just for the `get_enabled_repos()` function itself, which is great. But we need tests for `call_yum_command()` that the repos from submgr is added to the resulting command and as well as a test that ensures RHSM repos isn't added to `--enablerepo` if the user passed it in `--disablerepo`

fellipeh: @SpyTec already added on my lastest commit the unit test for `call_yum_command`
https://github.com/oamg/convert2rhel/pull/84/files#diff-cdec7561cf20f5b5b314293008180b42R83

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -79,13 +80,26 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
+    def test_call_yum_cmd_with_disablerepo(self):
+        pkghandler.call_yum_cmd("install")
+
+        self.assertEqual(utils.run_subprocess.cmd, "yum install -y --disablerepo=rhel-7-extras-rpm")
+
+    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked())
+    def test_call_yum_cmd_without_disablerepo(self):
+        pkghandler.call_yum_cmd("install")
+
+        self.assertEqual(utils.run_subprocess.cmd, "yum install -y --disablerepo=*")
```
SpyTec, line 94: Thinking about it, we can leave this to `test_call_yum_cmd` which covers this already. `test_call_yum_cmd_with_disablerepo` can remain
```suggestion
```

convert2rhel/pkghandler.py
SpyTec: Ah, I see. So that one covers `elif tool_opts.disablerepo:` but it doesn't cover 
```py
if not tool_opts.disable_submgr:
    for repo in system_info.submgr_enabled_repos:
        repos.append(repo)
```
For that two tests should be created
1. if `disable_submgr` is true, `system_info.submgr_enabled_repos` should not be appended to repos var.
1. if `disable_submgr` is false, `system_info.submgr_enabled_repos` should be appended to repos var.

Then add a different test for when the user has `--disablerepo` for a RHSM repo e.g. `--disablerepo=rhel-7-server-rpms`, we should not add this repo to `--enablerepo`. Otherwise we'll end up with a command that looks like `yum install -y --enablerepo=rhel-7-server-rpms --disablerepo=rhel-7-server-rpms` which is rather ambiguous and should be avoided.

bocekm: I agree. The cases @SpyTec mentions in the previous comment is not covered yet.

fellipeh: It's already created on my last commit, I believe :)

convert2rhel/pkghandler.py
```diff
@@ -72,26 +73,34 @@ def call_yum_cmd(command, args="", enablerepo=None, disablerepo=None,
     """Call yum command and optionally print its output."""
     loggerinst = logging.getLogger(__name__)
 
-    cmd = "yum %s -y" % (command)
+    cmd = "yum %s -y" % command
     # disablerepo parameter must be added before the enablerepo parameter
 
     if disablerepo is None:
```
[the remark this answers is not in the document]
bocekm: Please mark it as resolved after you actually resolve it.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -79,13 +80,36 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
+        pkghandler.call_yum_cmd("install")
+
+        self.assertEqual(utils.run_subprocess.cmd, "yum install -y --disablerepo=rhel-7-extras-rpm")
+
+    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked())
+    @unit_tests.mock(tool_opts, "disablerepo", ['rhel-7-extras-rpm'])
+    def test_call_yum_cmd_with_disablerepo(self):
+        pkghandler.call_yum_cmd("install")
+
+        self.assertEqual(utils.run_subprocess.cmd, "yum install -y --disablerepo=rhel-7-extras-rpm")
```
bocekm, line 104: Where in these three tests do you test the proper addition of `submgr_enabled_repos` as described by @SpyTec:

> Ah, I see. So that one covers `elif tool_opts.disablerepo:` but it doesn't cover 
> ```py
> if not tool_opts.disable_submgr:
>     for repo in system_info.submgr_enabled_repos:
>         repos.append(repo)
> ```
> For that two tests should be created
> 1. if `disable_submgr` is true, `system_info.submgr_enabled_repos` should not be appended to repos var.
> 1. if `disable_submgr` is false, `system_info.submgr_enabled_repos` should be appended to repos var.
> 
> Then add a different test for when the user has `--disablerepo` for a RHSM repo e.g. `--disablerepo=rhel-7-server-rpms`, we should not add this repo to `--enablerepo`. Otherwise we'll end up with a command that looks like `yum install -y --enablerepo=rhel-7-server-rpms --disablerepo=rhel-7-server-rpms` which is rather ambiguous and should be avoided.

SpyTec: To add to my comment, there is missing functionality in the code right now where it should not add to `--disablerepo` if it's already added as `--enablerepo`. The tests I mentioned in my comment will catch this, i.e. will fail currently. You could utilize [TDD](https://en.wikipedia.org/wiki/Test-driven_development) approach and write the tests first and see that it is complete once the tests pass

SpyTec: @fellipeh how's the progress on this?

SpyTec: @fellipeh this is still not resolved

fellipeh:
@SpyTec @bocekm  this PR is ready to re-review.

SpyTec:
@fellipeh there's 47 commits here

bocekm:
Yep, something went wrong before you force pushed to the branch.

bocekm:
@SpyTec, @fellipeh, I've taken the liberty of updating the PR with couple of changes, instead of adding bunch of comments. Please take a look.

It looks good to me now.
There's some infrastructure problem with running pylint with Python 2.6. Right off the bat I don't see a way to fix that.
And, there's a problem with running the unit tests with Python 2.6 as well - nose package is not available through pip. I've created a PR for fixing that: https://github.com/oamg/convert2rhel/pull/108.

fellipeh:
LGTM

bocekm:
Rebased to include the #108 fix.

SpyTec (commented):
Might be a future PR but we should make sure to warn the user when they provide the same repo in `disablerepo` and `enablerepo`

fellipeh:
@bocekm @SpyTec  I believe we can merge this PR, even its fail on packit (since this is a packit error)

bocekm:
I've just removed unused imports. The rest is ok.

bocekm (approved):

PR #90, opened 2020-07-15
new: Print modified rpm files

fellipeh:
Created new function `generate_modified_rpms` to check if there's any modified rpm during the conversion process, if yes:
- write a new log file with the modified rpms
- print on standard stdout the modified rpms
- return all modified rpms as list

### DIFF
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -16,6 +16,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import ConfigParser
+import difflib
 import os
 import re
 import logging
@@ -23,6 +24,8 @@ import logging
 from convert2rhel import utils
 from convert2rhel.toolopts import tool_opts
 from convert2rhel import logger
+
+DEFAULT_RPM_VA_LOG_FILENAME = 'rpm_va.log'
 
 
 class SystemInfo(object):
@@ -140,9 +143,10 @@ class SystemInfo(object):
     def _get_pkg_blacklist(self):
         return self._get_cfg_opt("pkg_blacklist").split()
 
-    def _generate_rpm_va(self):
-        """Verify the installed files on the system with the rpm metadata
-        for debug and support purposes."""
+    def _generate_rpm_va(self, log_filename=DEFAULT_RPM_VA_LOG_FILENAME):
+        """Let the rpm command to list all those rpm files that have been modified after the the rpm installation.
+        Such a list is useful for debug and support purposes. It's being saved to the default
+        log folder as log_filename."""
         if tool_opts.no_rpm_va:
             self.logger.info("Skipping execution of 'rpm -Va'.")
             return
@@ -151,10 +155,28 @@ class SystemInfo(object):
                          " minutes. It can be disabled by using the"
                          " --no-rpm-va option.")
         rpm_va, _ = utils.run_subprocess("rpm -Va", print_output=False)
-        output_file = os.path.join(logger.LOG_DIR, "rpm_va.log")
+        output_file = os.path.join(logger.LOG_DIR, log_filename)
         utils.store_content_to_file(output_file, rpm_va)
-        self.logger.info("The 'rpm -Va' output has been stored in the %s file"
-                         % output_file)
+        self.logger.info("The 'rpm -Va' output has been stored in the %s file" % output_file)
+
+    def log_modified_rpms_diff(self):
+        """Generate a log and print stdout message with modified rpms after the convert process."""
+        self._generate_rpm_va(log_filename='rpm_va_after_conversion.log')
+
+        pre_rpm_va_log_path = os.path.join(logger.LOG_DIR, DEFAULT_RPM_VA_LOG_FILENAME)
+        if not os.path.exists(pre_rpm_va_log_path):
+            self.logger.info("Skipping the comparison of the 'rpm -Va' output from before and after the conversion.")
+            return
+        pre_rpm_va = utils.get_file_content(pre_rpm_va_log_path, True)
+        post_rpm_va_log_path = os.path.join(logger.LOG_DIR, "rpm_va_after_conversion.log")
+        post_rpm_va = utils.get_file_content(post_rpm_va_log_path, True)
+        modified_rpm_files_diff = "\n".join(
+            difflib.unified_diff(pre_rpm_va, post_rpm_va, fromfile=pre_rpm_va_log_path, tofile=post_rpm_va_log_path,
+                                 n=0, lineterm=""))
+
+        if modified_rpm_files_diff:
+            self.logger.info(
+                "Comparison of modified rpm files from before and after the conversion:\n%s" % modified_rpm_files_diff)
 
 
 # Code to be executed upon module import
--- a/convert2rhel/unit_tests/systeminfo_test.py
+++ b/convert2rhel/unit_tests/systeminfo_test.py
@@ -35,7 +35,6 @@ from convert2rhel.systeminfo import system_info
 
 
 class TestSysteminfo(unittest.TestCase):
-
     class RunSubprocessMocked(unit_tests.MockFunction):
         def __init__(self, output_tuple=('output', 0)):
             self.output_tuple = output_tuple
@@ -47,12 +46,58 @@ class TestSysteminfo(unittest.TestCase):
             self.used_args.append(args)
             return self.output_tuple
 
+    class PathExistsMocked(unit_tests.MockFunction):
+        def __init__(self, return_value=True):
+            self.return_value = return_value
+
+        def __call__(self, filepath):
+            return self.return_value
+
     class GenerateRpmVaMocked(unit_tests.MockFunction):
         def __init__(self):
             self.called = 0
 
         def __call__(self):
             self.called += 1
+
+    class GetLoggerMocked(unit_tests.MockFunction):
+        def __init__(self):
+            self.task_msgs = []
+            self.info_msgs = []
+            self.warning_msgs = []
+            self.critical_msgs = []
+
+        def __call__(self, msg):
+            return self
+
+        def critical(self, msg):
+            self.critical_msgs.append(msg)
+            raise SystemExit(1)
+
+        def task(self, msg):
+            self.task_msgs.append(msg)
+
+        def info(self, msg):
+            self.info_msgs.append(msg)
+
+        def warn(self, msg, *args):
+            self.warning_msgs.append(msg)
+
+        def warning(self, msg, *args):
+            self.warn(msg, *args)
+
+        def debug(self, msg):
+            pass
+
+    class GetFileContentMocked(unit_tests.MockFunction):
+        def __init__(self, data):
+            self.data = data
+            self.as_list = True
+            self.called = 0
+
+        def __call__(self, filename, as_list):
+            self.called += 1
+            return self.data[self.called -1]
 
     ##########################################################################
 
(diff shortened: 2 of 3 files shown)


convert2rhel/systeminfo.py
```diff
@@ -154,6 +154,27 @@ def _generate_rpm_va(self):
         self.logger.info("The 'rpm -Va' output has been stored in the %s file"
                          % output_file)
 
+    def generate_modified_rpms(self):
+        """Generate a log and stdout message with modified rpms after the convert process """
```
ZhukovGreen, line 158: Should be a dot at the end of the sentence

ZhukovGreen: without space

convert2rhel/systeminfo.py
```diff
@@ -154,6 +154,27 @@ def _generate_rpm_va(self):
...
+        if tool_opts.no_rpm_va:
+            self.logger.info("Skipping execution of 'rpm -Va'.")
+            return
+
+        before_rpm_va = utils.get_file_content(os.path.join(logger.LOG_DIR, "rpm_va.log"), True)
+
+        rpm_va, _ = utils.run_subprocess("rpm -Va", print_output=False)
+        output_file = os.path.join(logger.LOG_DIR, "rpm_va_after_conversion.log")
+        utils.store_content_to_file(output_file, rpm_va)
+        difference = list(set(before_rpm_va).symmetric_difference(set(rpm_va.split())))
```
ZhukovGreen, line 168: cool, never saw this method before

convert2rhel/systeminfo.py
```diff
@@ -154,6 +154,27 @@ def _generate_rpm_va(self):
...
+
+        if difference:
+            output_file = os.path.join(logger.LOG_DIR, "rpm_va_difference.log")
+            utils.store_content_to_file(output_file, difference)
+            self.logger.info("INFO - List of modified rpm after the conversion process:")
+            for i in difference:
```
ZhukovGreen, line 174: better to avoid such variable names like `i`, maybe there is something more meaningful?

convert2rhel/systeminfo.py
SpyTec, line 168: Can be refactored to reduce code duplication with

https://github.com/oamg/convert2rhel/blob/8a31c2c17eac2628a2d29e1fc1b5a717d180a6e7/convert2rhel/systeminfo.py#L140-L154

convert2rhel/systeminfo.py
SpyTec: Not sure I like that the function both generates a log file as well as outputting changes to stdout, should be split

convert2rhel/systeminfo.py
```diff
@@ -154,6 +154,27 @@ def _generate_rpm_va(self):
...
+
+        before_rpm_va = utils.get_file_content(os.path.join(logger.LOG_DIR, "rpm_va.log"), True)
+
+        rpm_va, _ = utils.run_subprocess("rpm -Va", print_output=False)
+        output_file = os.path.join(logger.LOG_DIR, "rpm_va_after_conversion.log")
+        utils.store_content_to_file(output_file, rpm_va)
+        difference = list(set(before_rpm_va).symmetric_difference(set(rpm_va.split())))
+
+        if difference:
+            output_file = os.path.join(logger.LOG_DIR, "rpm_va_difference.log")
```
SpyTec, line 171: Since the logs are stored we don't need to store the difference in a separate file. Instead we can just log the differences

convert2rhel/systeminfo.py
SpyTec: Yes, both `i` and `difference` should be renamed to be more verbose

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -68,9 +78,33 @@ def tearDown(self):
         if os.path.exists(unit_tests.TMP_DIR):
             shutil.rmtree(unit_tests.TMP_DIR)
 
+    @unit_tests.mock(tool_opts, "no_rpm_va", True)
+    def test_generate_modified_rpms_with_no_rpm_va(self):
+        self.assertEqual(system_info.generate_modified_rpms(), None)
```
SpyTec, line 83: This is too imprecise, from this test we can't tell whether it respects `no_rpm_va` or simply returns no differences. Change this to be similar to:

https://github.com/oamg/convert2rhel/pull/90/files#diff-b42c590ebd33c41f1d6d03b3372d8da0R120-R127

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -68,9 +78,33 @@ def tearDown(self):
...
+
+    @unit_tests.mock(tool_opts, "no_rpm_va", False)
+    @unit_tests.mock(logger, "LOG_DIR", unit_tests.TMP_DIR)
+    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked(("rpm1\nrpm2\n", 0)))
+    @unit_tests.mock(utils, "get_file_content", GetFileContentMocked(filename='rpm_va.log', data=['rpm1', 'rpm2']))
+    def test_generate_modified_rpms_without_difference(self):
+        self.assertEqual(system_info.generate_modified_rpms(), None)
```
SpyTec, line 90: Likewise as above, we don't know if this returns `None` because of `no_rpm_va` being implemented incorrectly or it returns the differences

convert2rhel/systeminfo.py
fellipeh: This code was re-factored...

SpyTec: @fellipeh you forgot to push the commit to GitHub

convert2rhel/systeminfo.py
```diff
@@ -138,21 +138,38 @@ def _get_gpg_key_fingerprints(self):
...
+            return
+
+        before_rpm_va = utils.get_file_content(os.path.join(logger.LOG_DIR, "rpm_va.log"), True)
+        rpm_va = utils.get_file_content(os.path.join(logger.LOG_DIR, "rpm_va_after_conversion.log"), True)
+        diff_packages_list = list(set(before_rpm_va).symmetric_difference(set(rpm_va)))
+
+        if diff_packages_list:
+            self.logger.info("INFO - List of modified rpm after the conversion process:")
+            for pkg in diff_packages_list:
+                self.logger.info(" - %s" % pkg)
```
bocekm, line 171:
```suggestion
        if modified_rpm_files_diff:
            self.logger.info("Comparison of modified rpm files from before and after the conversion:\n%s" % modified_rpm_files_diff)
```
The prefix, similar to INFO, is being added by the logger formatter.

convert2rhel/unit_tests/systeminfo_test.py
SpyTec: Still applies after refactor to `log_modified_rpms_diff()`. Too imprecise

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -68,26 +113,43 @@ def tearDown(self):
...
+        self.assertEqual(system_info.log_modified_rpms_diff(), None)
+
+    @unit_tests.mock(tool_opts, "no_rpm_va", False)
     @unit_tests.mock(logger, "LOG_DIR", unit_tests.TMP_DIR)
-    @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked(
-        ("rpmva\n", 0)))
+    @unit_tests.mock(utils, "get_file_content", GetFileContentMocked(data=[['rpm1', 'rpm2'],
+                                                                           ['rpm1', 'rpm2']]))
+    def test_generate_modified_rpms_without_difference(self):
+        self.assertEqual(system_info.log_modified_rpms_diff(), None)
```
SpyTec, line 125: No longer is dependent on `no_rpm_va`, both tests are misleading and imprecise. Test name isn't precise enough to tell me what it tests

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -68,26 +113,43 @@ def tearDown(self):
...
+
+    @unit_tests.mock(os.path, "exists", PathExistsMocked(True))
+    @unit_tests.mock(tool_opts, "no_rpm_va", False)
+    @unit_tests.mock(logger, "LOG_DIR", unit_tests.TMP_DIR)
+    @unit_tests.mock(utils, "get_file_content", GetFileContentMocked(data=[['rpm1'],
+                                                                           ['rpm1', 'rpm2']]))
+    def test_generate_modified_rpms_with_difference(self):
+        self.assertEqual(system_info.log_modified_rpms_diff(), True)
```
SpyTec, line 133: Too imprecise, test name isn't precise enough to tell me what it tests

convert2rhel/systeminfo.py
```diff
@@ -149,10 +151,30 @@ def _generate_rpm_va(self):
                          " minutes. It can be disabled by using the"
                          " --no-rpm-va option.")
         rpm_va, _ = utils.run_subprocess("rpm -Va", print_output=False)
-        output_file = os.path.join(logger.LOG_DIR, "rpm_va.log")
+        output_file = os.path.join(logger.LOG_DIR, log_filename)
```
SpyTec, line 154: Test to check that it defaults to the global variable

convert2rhel/systeminfo.py
```diff
@@ -138,9 +139,10 @@ def _get_gpg_key_fingerprints(self):
     def _get_pkg_blacklist(self):
         return self._get_cfg_opt("pkg_blacklist").split()
 
-    def _generate_rpm_va(self):
-        """Verify the installed files on the system with the rpm metadata
-        for debug and support purposes."""
+    def _generate_rpm_va(self, log_filename=DEFAULT_RPM_VA_LOG_FILENAME):
```
SpyTec, line 142: Another test to check whether `log_filename` param works as expected

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -47,13 +46,59 @@ def __call__(self, *args, **kwargs):
...
+
+    class GetFileContentMocked(unit_tests.MockFunction):
+        def __init__(self, data):
+            self.data = data
+            self.as_list = True
+            self.called = -1
+
+        def __call__(self, filename, as_list):
+            self.called += 1
+            return self.data[self.called]
```
SpyTec, line 100:
```suggestion
            self.called = 0

        def __call__(self, filename, as_list):
            self.called += 1
            return self.data[self.called + 1]
```

fellipeh:
Rebased with updated master

convert2rhel/systeminfo.py
fellipeh: I believe there's no need to test the `log_filename`, because we accept any value, any value you put that will be used, even blank. If you put blank value, function `store_content_to_file` will be raise a exception and will close the file cursor as expected.

convert2rhel/systeminfo.py
fellipeh: I believe there's no need to test the `log_filename`, because we accept any value, any value you put that will be used, even blank. If you put blank value, will used a const, if you pass any value, will be used.. so, no reason to test this argument since we accept any value.

convert2rhel/unit_tests/systeminfo_test.py
fellipeh: This will raise: `IndexError: list index out of range` since list start in 0 and in test will add only one item..

SpyTec: That's why I suggest `self.data[self.called - 1]`. It does not make sense for `called` to be off by one

convert2rhel/systeminfo.py
SpyTec: It should be tested, though I made a mistake. The tests seem to cover both using the parameter and the default value as well.

convert2rhel/systeminfo.py
SpyTec: As I mentioned in the other thread, seems like there are tests and I missed it.

SpyTec (approved):

convert2rhel/systeminfo.py
```diff
@@ -138,21 +138,38 @@ def _get_gpg_key_fingerprints(self):
...
+
+    def generate_modified_rpms(self):
+        """Generate a log and print stdout message with modified rpms after the convert process."""
+        if not self._generate_rpm_va(log_filename='rpm_va_after_conversion.log'):
+            self.logger.info("Unable to run 'rpm -Va' after the conversion.")
+            return
+
+        before_rpm_va = utils.get_file_content(os.path.join(logger.LOG_DIR, "rpm_va.log"), True)
+        rpm_va = utils.get_file_content(os.path.join(logger.LOG_DIR, "rpm_va_after_conversion.log"), True)
+        diff_packages_list = list(set(before_rpm_va).symmetric_difference(set(rpm_va)))
```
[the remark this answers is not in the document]
SpyTec: @bocekm shouldn't we keep all imports at the top?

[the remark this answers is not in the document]
bocekm: Yeah, we should. Thanks for catching it. @fellipeh, please move it to the top.

SpyTec (approved):

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -68,26 +113,42 @@ def tearDown(self):
...
+    def test_generate_modified_rpms_without_rpm_different_after_convert(self):
+        self.assertEqual(system_info.log_modified_rpms_diff(), None)
+
+    @unit_tests.mock(os.path, "exists", PathExistsMocked(True))
+    @unit_tests.mock(tool_opts, "no_rpm_va", False)
+    @unit_tests.mock(logger, "LOG_DIR", unit_tests.TMP_DIR)
+    @unit_tests.mock(utils, "get_file_content", GetFileContentMocked(data=[['rpm1'],
+                                                                           ['rpm1', 'rpm2']]))
+    def test_generate_modified_rpms_with_difference_after_convert(self):
+        self.assertEqual(system_info.log_modified_rpms_diff(), True)
```
bocekm, line 132:
```suggestion
        system_info.log_modified_rpms_diff()
        self.assertEqual('S.5....T.  c /etc/yum.conf' in system_info.logger.warning_msgs)
```

bocekm (changes requested):

fellipeh:
@bocekm I believe this PR is ready to merge, right?

bocekm (approved):
Tested and works correctly. Thanks.

PR #92, opened 2020-07-29
add: Clear yum versionlock

fellipeh:
Users may have locked some of the installed packages to a certain version using `yum-plugin-versionlock`. This would most probably cause the conversion to fail because the RHEL equivalent of the package may easily have a different version.

This PR check if `yum-plugin-versionlock` is installed, if so, ask user if want to continue, if yes clear the versionlock list, if no conversion will stop.

See https://access.redhat.com/solutions/98873 for more information.

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -72,6 +72,9 @@ def main():
         redhatrelease.yum_conf.backup()
         subscription.rhn_reg_file.backup()
 
+        loggerinst.task("Prepare: Clear yum version locks")
+        pkghandler.clear_yum_versionlock()
+
         # begin conversion process
         process_phase = ConversionPhase.PRE_PONR_CHANGES
         pre_ponr_conversion()
@@ -213,6 +216,8 @@ def rollback_changes():
     redhatrelease.system_release_file.restore()
     subscription.rhn_reg_file.restore()
     redhatrelease.yum_conf.restore()
+    pkghandler.versionlock_file.restore()
+
     return
 
 
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -28,6 +28,8 @@ from convert2rhel.toolopts import tool_opts
 # Limit the number of loops over yum command calls for the case there was
 # an error.
 MAX_YUM_CMD_CALLS = 2
+_VERSIONLOCK_FILE_PATH = '/etc/yum/pluginconf.d/versionlock.list'
+versionlock_file = utils.RestorableFile(_VERSIONLOCK_FILE_PATH)  # pylint: disable=C0103
 
 
 class PkgWFingerprint(object):
@@ -524,3 +526,19 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
 def is_rhel_kernel_installed():
     installed_rhel_kernels = get_installed_pkgs_by_fingerprint(system_info.fingerprints_rhel, name="kernel")
     return len(installed_rhel_kernels) > 0
+
+
+def clear_yum_versionlock():
+    loggerinst = logging.getLogger(__name__)
+
+    if os.path.isfile(_VERSIONLOCK_FILE_PATH) and os.path.getsize(_VERSIONLOCK_FILE_PATH) > 0:
+        loggerinst.warn("yum versionlock plugin is in use. It may cause the conversion to fail.")
+        loggerinst.info("Upon continuing, we will clear all package version locks.")
+        utils.ask_to_continue()
+
+        versionlock_file.backup()
+
+        loggerinst.info("Clearing package versions locks...")
+        call_yum_cmd("versionlock clear", print_output=False)
+    else:
+        loggerinst.info("Usage of yum versionlock plugin not detected.")
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -18,6 +18,7 @@
 import glob
 import os
 import re
+import sys
 import yum
 
 from convert2rhel import logger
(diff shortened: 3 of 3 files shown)


convert2rhel/pkghandler.py
```diff
@@ -508,3 +507,21 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
...
+
+def check_yum_versionlock_plugin_enabled():
+    loggerinst = logging.getLogger(__name__)
+    loggerinst.info("Checking if yum versionlock plugin is installed and actived ...")
+    _VERSIONLOCK_FILE = '/etc/yum/pluginconf.d/versionlock.list'
+
+    if os.path.isfile(_VERSIONLOCK_FILE):
+        loggerinst.info("\nYum versionlock plugin detected!")
+        loggerinst.info("\nConvert2RHEL detected you are using versionlock yum plugin. This plugin may cause conversion"
+                        " issues, by continue Convert2RHEL will disable versionlock plugin on yum.")
```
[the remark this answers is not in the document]
bocekm: Still to be changed..

bocekm (changes requested):
A rollback of the version lock is needed.
Backing up and then copying back the `_VERSIONLOCK_FILE` is enough to do that.

fellipeh:
Rebased

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -79,6 +71,58 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
+        def __init__(self, file_size):
+            self.file_size = file_size
+
+        def __call__(self, *args, **kwargs):
+            return self.file_size
+
+
+    @unit_tests.mock(os.path, "isfile", IsFileMocked(is_file=False))
+    def test_clear_yum_versionlock_plugin_not_enabled(self):
+        self.assertFalse(pkghandler.clear_yum_versionlock())
```
SpyTec, line 99: Add an assertion for the logger as well

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -79,6 +71,58 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
+    @unit_tests.mock(utils, "ask_to_continue", DumbCallableObject())
+    @unit_tests.mock(os.path, "isfile", IsFileMocked(is_file=True))
+    @unit_tests.mock(os.path, "getsize", GetSizeMocked(file_size=1))
+    @unit_tests.mock(pkghandler, "call_yum_cmd", CallYumCmdMocked())
+    @unit_tests.mock(utils.RestorableFile, "backup", DumbCallableObject)
+    @unit_tests.mock(utils.RestorableFile, "restore", DumbCallableObject)
+    def test_clear_yum_versionlock_user_says_yes(self):
+        pkghandler.clear_yum_versionlock()
+        self.assertEqual(pkghandler.call_yum_cmd.called, 1)
+
```
SpyTec, line 110: Ideally extend `CallYumCmdMocked` to store the command called, and then assert that we call `yum versionlock clear`

SpyTec (changes requested):
Missing tests for making sure backup works. As well as a function for restoring the backup and relevant tests

convert2rhel/pkghandler.py
```diff
@@ -507,3 +507,25 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
...
     installed_rhel_kernels = get_installed_pkgs_by_fingerprint(system_info.fingerprints_rhel, name="kernel")
     return len(installed_rhel_kernels) > 0
+
+
+def clear_yum_versionlock():
+    loggerinst = logging.getLogger(__name__)
+    _VERSIONLOCK_FILE = '/etc/yum/pluginconf.d/versionlock.list'
+    _vers_lock_file = utils.RestorableFile(_VERSIONLOCK_FILE)
+
+    _vers_lock_file.backup()
```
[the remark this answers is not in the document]
fellipeh: backup file already handle with this, here: https://github.com/oamg/convert2rhel/blob/master/convert2rhel/utils.py#L385

convert2rhel/pkghandler.py
```diff
@@ -507,3 +507,25 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
...
+        loggerinst.info("\nIf you choose to continue, all the package version locks will get cleared"
+                        " to not cause the conversion to fail.")
+        utils.ask_to_continue()
+
+        loggerinst.info("Clearing package versions locks...")
+        call_yum_cmd("versionlock clear", print_output=False)
+    else:
+        loggerinst.info("Usage of yum versionlock plugin not detected.")
+
+    _vers_lock_file.restore()
```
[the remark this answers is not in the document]
fellipeh: Backup this file was requested by @bocekm 

Regarding to restore:   agree here is not better place, but, rollback isn't the better place,because it's called once we cancel the process.. and the goal here is:  backup the original lock file, do all process and restore the original file.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -79,6 +71,58 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
+    @unit_tests.mock(utils, "ask_to_continue", AskDumbCallableObject())
+    @unit_tests.mock(os.path, "isfile", IsFileMocked(is_file=True))
+    @unit_tests.mock(pkghandler, "call_yum_cmd", CallYumCmdMocked())
+    def test_clear_yum_versionlock_user_says_no(self):
+        self.assertRaises(SystemExit, pkghandler.clear_yum_versionlock)
+
```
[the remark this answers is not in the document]
fellipeh: Could you please elaborate more?

convert2rhel/pkghandler.py
```diff
@@ -507,3 +507,25 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
...
+    _VERSIONLOCK_FILE = '/etc/yum/pluginconf.d/versionlock.list'
+    _vers_lock_file = utils.RestorableFile(_VERSIONLOCK_FILE)
+
+    _vers_lock_file.backup()
+
+    if os.path.isfile(_VERSIONLOCK_FILE) and os.path.getsize(_VERSIONLOCK_FILE) > 0:
+        # The file may exist and be empty => no version locks
+        loggerinst.warn("yum versionlock plugin is in use")
+        loggerinst.info("\nIf you choose to continue, all the package version locks will get cleared"
+                        " to not cause the conversion to fail.")
```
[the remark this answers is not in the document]
fellipeh: That's not truth.. version locks will not be cleared, it's will be cleared during the process, and it's supposed to restore to the original file after the process... that's the reason for backup/restore the file.

convert2rhel/pkghandler.py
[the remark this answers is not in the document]
SpyTec: The goal here is to backup the lock file, and if conversion fails we roll it back. Restoring the file post migration will presumably have invalid entries due to the switch of packages to redhat repo, @bocekm do correct me if I'm wrong about the invalid entries.

@fellipeh, what you're doing here is backing up the file, clearing the file, then restoring it - all in the same function. This means the lock file is not changed and the function has no side effects except for touching the file. `yum versionlock clear` only clears the file, it does not do anything else. Yum itself checks this file whenever it handles packages whether a package it's handling is locked or not.

convert2rhel/unit_tests/pkghandler_test.py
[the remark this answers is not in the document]
SpyTec: You need to assert that `call_yum_cmd()` was never called here.

Having multiple assertions avoids most false positives and ensures that the application runs as you expect it to.

For instance, ask yourself: would the tests pass if `call_yum_cmd()` was called before `ask_to_continue()`? If it would, then the tests are incorrect

convert2rhel/pkghandler.py
```diff
@@ -524,3 +526,20 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
...
+
+def clear_yum_versionlock():
+    loggerinst = logging.getLogger(__name__)
+
+    if os.path.isfile(_VERSIONLOCK_FILE_PATH) and os.path.getsize(_VERSIONLOCK_FILE_PATH) > 0:
+        # The file may exist and be empty => no version locks
+        VERSIONLOCK_FILE.backup()
+
+        loggerinst.warn("yum versionlock plugin is in use. It may cause the conversion to fail.")
+        loggerinst.info("\nIf you choose to continue, the package version locks will be cleared.")
```
SpyTec, line 539: I wasn't too keen on this sentence as it was using passive voice when asking the user for confirmation. I was struggling to come up with an active voice that worked, but Zack Hawkins helped and came up with this

```suggestion
		loggerinst.info("Do you want to continue? This will clear package version locks.")
```

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -19,6 +19,7 @@
 import os
 import re
 import yum
+import sys
```
SpyTec, line 22: Alphabetical import order
```suggestion
import sys
import yum
```

SpyTec (changes requested):
A final round of changes, once it's all done I think this is good to merge

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -79,6 +73,93 @@ def __call__(self, cmd, print_cmd=True, print_output=True):
...
+        def __call__(self, *args, **kwargs):
+            return self.is_file
+
+    class DumbCallableObject(unit_tests.MockFunction):
+        def __call__(self, *args, **kwargs):
+            return
+
+    class AskDumbCallableObject(unit_tests.MockFunction):
+        def __call__(self, *args, **kwargs):
+            sys.exit(1)
```
SpyTec, line 89: Can also refactor this to be `SysExitCallableObject`

convert2rhel/pkghandler.py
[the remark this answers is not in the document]
bocekm: Yeah, the file should not be restored after the conversion. Since every package reinstalled during the conversions, there's a huge chance the restored version lock would not be valid anymore.
So, as Eric wrote, restore it only during the rollback.

fellipeh:
@bocekm @SpyTec I believe this PR is ready to merge, right?

convert2rhel/pkghandler.py
```diff
@@ -524,3 +526,19 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
...
+
+
+def clear_yum_versionlock():
+    loggerinst = logging.getLogger(__name__)
+
+    if os.path.isfile(_VERSIONLOCK_FILE_PATH) and os.path.getsize(_VERSIONLOCK_FILE_PATH) > 0:
+        VERSIONLOCK_FILE.backup()
+
+        loggerinst.warn("yum versionlock plugin is in use. It may cause the conversion to fail.")
+        loggerinst.info("Do you want to continue? This will clear package version locks.")
```
bocekm, line 538: I know that this has been changed two times already but the question _Do you want to continue?_ is kind of duplicate because `ask_to_continue()` prints _Continue with the system conversion? [y/n]_.
I actually liked the previous _"If you choose to continue, the package version locks will be cleared."_. But since you don't want to use the passive form, @SpyTec, what about:
```suggestion
        loggerinst.info("Upon continuing, we will clear all package version locks.")
```
I'm not sure if "we" is the best choice, but it's hard to choose something here when not using the passive form.

convert2rhel/pkghandler.py
```diff
@@ -28,6 +28,8 @@
 # Limit the number of loops over yum command calls for the case there was
 # an error.
 MAX_YUM_CMD_CALLS = 2
+_VERSIONLOCK_FILE_PATH = '/etc/yum/pluginconf.d/versionlock.list'
+VERSIONLOCK_FILE = utils.RestorableFile(_VERSIONLOCK_FILE_PATH)
```
bocekm, line 32:
```suggestion
versionlock_file = utils.RestorableFile(_VERSIONLOCK_FILE_PATH)  # pylint: disable=C0103
```
To be consistent in the project, use lowercase as is the case of all the other RestorableFile instances (e.g. `system_release_file` or `yum_conf`).

bocekm (changes requested):

convert2rhel/main.py
```diff
@@ -60,6 +60,9 @@ def main():
         loggerinst.task("Prepare: End user license agreement")
         user_to_accept_eula()
 
+        loggerinst.task("Prepare: Clear yum version locks")
+        pkghandler.clear_yum_versionlock()
```
bocekm, line 64: Please move this below the `subscription.rhn_reg_file.backup()`. Doing it as a first thing, before even checking if the system is possible to convert (gathering system info), is not right.

convert2rhel/pkghandler.py
```diff
@@ -524,3 +526,19 @@ def install_additional_rhel_kernel_pkgs(additional_pkgs):
 def is_rhel_kernel_installed():
     installed_rhel_kernels = get_installed_pkgs_by_fingerprint(system_info.fingerprints_rhel, name="kernel")
     return len(installed_rhel_kernels) > 0
+
+
+def clear_yum_versionlock():
+    loggerinst = logging.getLogger(__name__)
+
+    if os.path.isfile(_VERSIONLOCK_FILE_PATH) and os.path.getsize(_VERSIONLOCK_FILE_PATH) > 0:
+        versionlock_file.backup()
```
bocekm, line 535: Move the backup below the `utils.ask_to_continue()`. It makes no sense to back the file up in case they don't want to continue.
As you can see from the output, the information about the backup just adds a noise around the important message under it:
```
[10/06/2020 14:04:37] TASK - [Prepare: Clear yum version locks] *********************************
Backing up /etc/yum/pluginconf.d/versionlock.list
Copying /etc/yum/pluginconf.d/versionlock.list to /var/lib/convert2rhel/
WARNING - yum versionlock plugin is in use. It may cause the conversion to fail.
Do you want to continue? This will clear package version locks.

Continue with the system conversion? [y/n]: y
```

bocekm (changes requested):

convert2rhel/pkghandler.py
SpyTec: You're right, I thought it was just printing out `[y/n]` so we could be flexible. But yeah, that change SGTM. We should be using active voice as much as possible as it conveys better and sets a more direct tone

bocekm (approved):

bocekm:
I'm merging this despite the one test failing. It seems to be a testing farm issue which I've filed at https://gitlab.com/testing-farm/cruncher/-/issues/11.

bocekm:
I can't merge it before @SpyTec approves it.

SpyTec (approved):

PR #97, opened 2020-09-02
Do not check repoids with disabled RHSM

bocekm:
When using custom repos instead of RHSM, checking for expected repoids and warning when they're not available is confusing to customers as they fear something is wrong.
Related: OAMG-3144


fellipeh (approved):
LGTM

PR #102, opened 2020-09-15
Simplify getting /etc/system-release path

bocekm:
The function was more complicated that necessary. The reason is that when the tool supported conversions to RHEL 5, at that system the system release file had a different name. Support for RHEL 5 has been dropped and on RHEL 6+ (and its derivatives_ the path is always `/etc/system-release`.

Related: https://github.com/oamg/convert2rhel/pull/75/.


fellipeh (approved):
LGTM

SpyTec (approved):

PR #103, opened 2020-09-24
Fix: RHSM repo availibility method not found 

SpyTec:
Tests makes sure call order is what we expect, as a benefit it also catches the error introduced in #99

### DIFF
--- a/convert2rhel/unit_tests/main_test.py
+++ b/convert2rhel/unit_tests/main_test.py
@@ -23,12 +23,18 @@ try:
 except ImportError:
     import unittest
 
+from collections import OrderedDict
 from convert2rhel import main
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
 from convert2rhel import redhatrelease
+from convert2rhel import repo
 from convert2rhel import subscription
 from convert2rhel import utils
+from convert2rhel import pkghandler
+from convert2rhel.toolopts import tool_opts
 
+def mock_calls(class_or_module, method_name, mock_obj):
+    return unit_tests.mock(class_or_module, method_name, mock_obj(method_name))
 
 class TestMain(unittest.TestCase):
 
@@ -56,6 +62,9 @@ class TestMain(unittest.TestCase):
         def __call__(self, msg):
             return self
 
+        def task(self, msg):
+            pass
+
         def critical(self, msg):
             self.critical_msgs.append(msg)
             raise SystemExit(1)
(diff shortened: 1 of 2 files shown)


convert2rhel/unit_tests/main_test.py
```diff
@@ -23,12 +23,18 @@
 except ImportError:
     import unittest
 
+from collections import OrderedDict
```
bocekm, line 26: Unfortunately, `OrderedDict` is not available with python 2.6 (https://docs.python.org/2/library/collections.html#collections.OrderedDict), i.e. it is not expected to work on CentOS/OL 6. I use "not expected to" because there's a small chance someone has backported the feature in RHEL. Please check that.

bocekm (changes requested):
Thanks for catching this.

convert2rhel/unit_tests/main_test.py
SpyTec: You're right. Luckily it seems to be backported in CentOS 6. There seems to be a pip package for backwards compatibility though so it shouldn't be any issues installing that when if we ever run tests on OL.
```py
Python 2.6.6 (r266:84292, Jun 20 2019, 14:14:55) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-23)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import OrderedDict
>>> OrderedDict()
OrderedDict()
```

bocekm: We can't rely on anything available in pip. We are unable to handle such dependency in the rpm spec file. It's either part of the standard library provided by the python rpm packages on CentOS/OL, or we need to find an alternative.

On CentOS 6.10 it's available. I've checked it quickly on RHEL and it's been available since RHEL 6.5.

So, if we go with this change, we basically support conversions of CentOS/OL 6.5+. Which is ok with me, we just need to be aware of that and document it.

SpyTec: I'd disagree, we can rely on pip for packages to help with testing. The OrderedDict is only being used within tests and isn't something the customer will ever encounter. Hence why I'm saying it's fine as it is and if we ever want to support running tests on older versions or Oracle Linux we can simply add a pip package to the container we're testing with

bocekm: Aha, you're right. I haven't noticed it's used in a test only. Sorry.

sturivny (approved):

bocekm (approved):

PR #104, opened 2020-09-30
feat: Echo which platform we're testing in

SpyTec:
Small change to echo the platform we're running tests in - purely to help local development


sturivny (approved):

sturivny:
LGTM

PR #108, opened 2020-10-14
Resolve nose not being available though pip

bocekm:
We can simply install the nose CentOS package.


SpyTec (approved):
Well that certainly makes it easier

PR #109, opened 2020-10-15
Enable repos enabled through RHSM in yum calls

bocekm:
There was a bug causing that when subscription-manager was used for the
conversions, no repository got enabled when calling yum. And because by
default convert2rhel disables all repos in yum calls except those to be
specifically enabled, the conversion could not proceed.

The issue got introduced in https://github.com/oamg/convert2rhel/pull/84.


fellipeh (approved):
LGTM

I believe we can merge it @bocekm , since the packit still facing internal error..

PR #114, opened 2020-11-05
fix(Utils): Avoid NULL byte error by ASCII-encoding subprocess command

SpyTec:
Fixes `TypeError: execve() argument 1 must be encoded string without NULL bytes, not str` error with `subprocess` and `shlex` for python 2.6 with OL6/CentOS 6

Gotten the bugfix to work but be wary in testing as the error it fixes may not happen every time

### DIFF
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -150,9 +150,9 @@ class TestUtils(unittest.TestCase):
         self.assertEqual("rpm -i --replacepkgs pkg1 pkg2 pkg3", utils.run_subprocess.cmd)
 
     def test_run_subprocess(self):
-        output, code = utils.run_subprocess("echo ahoj")
+        output, code = utils.run_subprocess("echo foobar")
 
-        self.assertEqual(output, "ahoj\n")
+        self.assertEqual(output, "foobar\n")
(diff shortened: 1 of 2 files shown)


bocekm:
```
File "/root/git-source/convert2rhel/utils.py", line 129, in run_subprocess
cmd = shlex.split(cmd.encode("ascii"), False)
File "/usr/lib64/python3.6/shlex.py", line 305, in split
return list(lex)
File "/usr/lib64/python3.6/shlex.py", line 295, in __next__
token = self.get_token()
File "/usr/lib64/python3.6/shlex.py", line 105, in get_token
raw = self.read_token()
File "/usr/lib64/python3.6/shlex.py", line 136, in read_token
nextchar = self.instream.read(1)
AttributeError: 'bytes' object has no attribute 'read'
```

SpyTec:
Ack, saw that. Wondering if there's a different solution or if we should simply do a check for py2.6 and encode it then

SpyTec:
@bocekm @fellipeh I believe that's it. I doubt we'll be able to notice whether it's fixed or not that easily since it's an intermittent error. I vote for merging it after an initial test that it doesn't break anything

bocekm (approved):
I've tried it twice on OL 6, with and without RHSM, and no issue.

PR #115, opened 2020-11-05
Remove all things RHN

bocekm:
RHN has been shut down for good. And the RHN-related package rhn-client-tools started causing the conversion to fail on OL7.8.


SpyTec (approved):
LGTM

PR #116, opened 2020-11-09
add: check/warn user if same repo in both enable / disable repo options

fellipeh:
Warn the user when they provide the same repo in both the --disablerereo and --enablerepo option.

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -222,6 +222,7 @@ def get_installed_pkg_objects(name=""):
         return _get_installed_pkg_objects_dnf(name)
 
 
+
 def _get_installed_pkg_objects_yum(name):
     yum_base = pkgmanager.YumBase()
     # Disable plugins (when kept enabled yum outputs useless text every call)
@@ -676,3 +677,15 @@ def clear_versionlock():
         call_yum_cmd("versionlock clear", print_output=False)
     else:
         loggerinst.info("Usage of YUM/DNF versionlock plugin not detected.")
+
+
+def has_duplicate_repos_across_disablerepo_enablerepo_options():
+    loggerinst = logging.getLogger(__name__)
+
+    duplicate_repos = set(tool_opts.disablerepo) & set(tool_opts.enablerepo)
+    if duplicate_repos:
+        message = "Duplicate repositories was found across disablerepo and enablerepo options:"
+        for repo in duplicate_repos:
+            message += "\n%s" % repo
+        message += "\nThis ambiguity may have unintended consequences."
+        loggerinst.warning(message)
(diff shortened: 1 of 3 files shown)


convert2rhel/pkghandler.py
```diff
@@ -590,3 +589,14 @@ def clear_versionlock():
...
+
+def is_disable_and_enable_repos_has_same_repo():
+    loggerinst = logging.getLogger(__name__)
+
+    same_repos = set(tool_opts.disablerepo) & set(tool_opts.enablerepo)
+    if same_repos:
+        message = ("Convert2RHEL found identical repos in disablerepo and enablerepo options: %s\n"
+                   "Please use only in one option and run Convert2RHEL again." % same_repos)
+        print(message)
+        loggerinst.critical(message)
```
SpyTec, line 602: I don't think we need to throw a critical message, we can just warn the user instead. This change also makes it a prettier print
```suggestion
        message = "Duplicate repositories was found across disablerepo and enablerepo options:"
        for repo in same_repos:
            message += "\n%s" % repo
        message += "\nThis ambiguity may have unintended consequences."
        loggerinst.warning(message)
```

convert2rhel/pkghandler.py
```diff
@@ -590,3 +589,14 @@ def clear_versionlock():
         call_yum_cmd("versionlock clear", print_output=False)
     else:
         loggerinst.info("Usage of YUM/DNF versionlock plugin not detected.")
+
+
+def is_disable_and_enable_repos_has_same_repo():
```
SpyTec, line 594:
```suggestion
def has_duplicate_repos_across_disablerepo_enablerepo_options():
```

convert2rhel/pkghandler.py
```diff
@@ -590,3 +589,14 @@ def clear_versionlock():
...
+    loggerinst = logging.getLogger(__name__)
+
+    same_repos = set(tool_opts.disablerepo) & set(tool_opts.enablerepo)
```
SpyTec, line 597: `same_repos` can probably be `duplicate_repos`

SpyTec (changes requested):

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -853,6 +853,19 @@ def test_list_third_party_pkgs(self):
...
+    @unit_tests.mock(logger.CustomLogger, "warning", LogMocked())
+    def test_is_disable_and_enable_repos_has_same_repo(self):
+        pkghandler.has_duplicate_repos_across_disablerepo_enablerepo_options()
+        self.assertTrue("Duplicate repositories was found" in logger.CustomLogger.warning.msg)
+
+
+    @unit_tests.mock(tool_opts, "disablerepo", ['*'])
+    @unit_tests.mock(tool_opts, "enablerepo", ['rhel-7-extras-rpm'])
+    def test_is_disable_and_enable_repos_doesnt_thas_same_repo(self):
+        self.assertEqual(None, pkghandler.has_duplicate_repos_across_disablerepo_enablerepo_options())
```
SpyTec, line 867: This will also pass if there are duplicate repos. Add a mock for CustomLogger and assert that it is not called
```suggestion
    @unit_tests.mock(logger.CustomLogger, "warning", LogMocked())
    def test_is_disable_and_enable_repos_doesnt_thas_same_repo(self):
        pkghandler.has_duplicate_repos_across_disablerepo_enablerepo_options()
        self.assertEqual(logger.CustomLogger.warning.called, 0)
```

SpyTec (approved):

PR #125, opened 2020-12-02
Do not rename repofiles

bocekm:
When using RHSM, we currently rename all repofiles not coming from Red Hat during the conversion so that they are not used for the conversion in any way. That causes a problem when somebody passes --enablerepo together with using RHSM, and this repo is in some other repofile than /etc/yum.repos.d/redhat.repo. Because of the repofile renaming, this repository is then unexpectedly not used during the conversion.

The repofile renaming is not needed if we call yum with the right parameters (--enablerepo and --disablerepo) and that's what we've managed under https://github.com/oamg/convert2rhel/pull/109/.

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -178,9 +178,6 @@ def pre_ponr_conversion():
         subscription.disable_repos()
         loggerinst.task("Convert: Subscription Manager - Enable RHEL repositories")
         subscription.enable_repos(rhel_repoids)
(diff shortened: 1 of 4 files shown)


fellipeh (dismissed):
LGTM

bocekm:
Let's wait for merging the https://github.com/oamg/convert2rhel/pull/126.

bocekm:
Rebased. @fellipeh, please re-review.

fellipeh (approved):

PR #127, opened 2020-12-07
feat: backup repo files

SpyTec:
This backs up repo files and restores them on rollback. This prevents accidental deletion of repos due to package removals.

OAMG-4141

### DIFF
--- a/convert2rhel/repo.py
+++ b/convert2rhel/repo.py
@@ -16,8 +16,11 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import logging
+import os
+import shutil
 
 from convert2rhel.systeminfo import system_info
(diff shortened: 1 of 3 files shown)


convert2rhel/repo.py
```diff
@@ -16,8 +16,11 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import logging
+import os
+import shutil
 
 from convert2rhel.systeminfo import system_info
+from convert2rhel.utils import TMP_DIR
```
bocekm, line 23: We now have BACKUP_DIR in utils.py. Please use that. Otherwise LGTM.

bocekm (changes requested):

bocekm (approved):

PR #129, opened 2020-12-09
[WIP] Packit to build rpms in a custom epel-6-like chroot

bocekm:
Pavel Raiskup from the team behind Copr has suggested us to use the custom-1-x86_64 chroot available in Fedora Copr and set it up to behave the same as the original removed epel-6-x86_64 chroot.


sturivny (approved):
LGTM

bocekm:
epel-6 chroot is back! No need for custom chroot anymore.

PR #112, opened 2020-10-22
Add support for converting OL 8

bocekm:
- Added missing dnf-utils dependency for yumdownloader.
- Replaced function getting package NVRA with getting NEVRA as that's what yum/dnf is printing.
- Improved function for downloading packages that uses yumdownloader to be more robust.
- The get_installed_pkgs_w_different_fingerprint function, when using DNF, filters packages by name only, not by package NEVRA as with YUM.
- Fixed getting the redhat-release package name, it's different on RHEL 8 than on RHEL 6 and 7.

To convert OL 8, run the following under root:
```
dnf install -y http://mirror.centos.org/centos/8/BaseOS/x86_64/os/Packages/python3-syspurpose-1.26.20-1.el8_2.x86_64.rpm # temporary before we fix deps
curl https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/repo/epel-8/group_oamg-convert2rhel-epel-8.repo -o /etc/yum.repos.d/copr_convert2rhel.repo
yum install -y convert2rhel*pr112*

mkdir -p /usr/share/convert2rhel/redhat-release/Server/
curl http://<pulp_mirror_hostname>/content/dist/rhel8/8.2/x86_64/baseos/os/Packages/r/redhat-release-8.2-1.0.el8.x86_64.rpm -o /usr/share/convert2rhel/redhat-release/Server/redhat-release-8.2-1.0.el8.x86_64.rpm 

cat >/etc/yum.repos.d/rhel8.repo <<EOFE
[rhel-8-for-x86_64-baseos-rpms]
name=RHEL 8.2 BaseOS for x86_64
baseurl=http://<pulp_mirror_hostname>/content/dist/rhel8/8.2/x86_64/baseos/os/
enabled=0
gpgcheck=0

[rhel-8-for-x86_64-appstream-rpms]
name=RHEL 8.2 AppStream for x86_64
baseurl=http://<pulp_mirror_hostname>/content/dist/rhel8/8.2/x86_64/appstream/os/
enabled=0
gpgcheck=0
EOFE

convert2rhel --debug -y --no-rpm-va --disable-submgr --enablerepo rhel-8-for-x86_64-baseos-rpms --enablerepo rhel-8-for-x86_64-appstream-rpms
```

Notes:
- It won't be needed to install the python3-syspurpose package manually after merging https://github.com/oamg/convert2rhel/pull/107.
- It won't be needed to download the redhat-release package after merging https://github.com/oamg/convert2rhel/pull/105.
[...]

PR #130, opened 2020-12-09
Fix not replacing same version kernel

bocekm:
Without this fix the original kernel silently remains not replaced with the RHEL one in case the kernel version on the original system is the same as the version available for RHEL.

Related change that caused this bug: https://github.com/oamg/convert2rhel/pull/112/files#diff-ed1ca2e4a8b27f1bee13d608562d4e94eceb2972790eccf4d84f366b750f221dR433


fellipeh (approved):
LGTM

sturivny (approved):

SpyTec (approved):

PR #132, opened 2020-12-10
[WIP] Add CI for making releases

sturivny:
For creating a release using: https://github.com/marketplace/actions/create-a-release
For making an `rpm` using:  https://github.com/naveenrajm7/rpmbuild
Probably it should be updated with `mock`.


bocekm:
I don't think we'll be able to set the RPM builds easily. There are couple of issues:

- There's no official Github action to create the RPMs
- There's this action https://github.com/naveenrajm7/rpmbuild which lacks the EL6 env and allows uploading only SRPMs, not RPMs
- Then there's its fork https://github.com/redwain/rpmbuild/ which has EL6 env but this env does not work because it uses CentOS 6 docker image which uses old not accessible CentOS repos (as tested [here](https://github.com/bocekm/convert2rhel/runs/1557596695?check_suite_focus=true)), and as well as the option above, it does not allow uploading the built RPMs

One option could be forking the latter, but it would require updating shell scripts like this one: https://github.com/redwain/rpmbuild/blob/el6/entrypoint.sh.

At this point the solution looks a bit hacky, prone to errors and unreliable.

bocekm:
Closing in favor of https://github.com/oamg/convert2rhel/pull/135.

PR #133, opened 2020-12-11
fix: Print package vendor or fallback to packager

fellipeh:
With RHEL 8.4, dnf will be fixed in a way that it will be possible to… get pkg vendor through its python API: https://bugzilla.redhat.com/show_bug.cgi?id=1876561

This commit print vendor if is possible or fallback to packager

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -222,7 +222,6 @@ def get_installed_pkg_objects(name=""):
         return _get_installed_pkg_objects_dnf(name)
 
 
-
 def _get_installed_pkg_objects_yum(name):
     yum_base = pkgmanager.YumBase()
     # Disable plugins (when kept enabled yum outputs useless text every call)
@@ -293,12 +292,14 @@ def print_pkg_info(pkgs):
     (https://bugzilla.redhat.com/show_bug.cgi?id=1876561).
     """
     max_nvra_length = max(map(len, [get_pkg_nvra(pkg) for pkg in pkgs]))
-    max_packager_length = max(max(map(len, [get_packager(pkg) for pkg in pkgs])), len("Packager"))
+    max_packager_length = max(
+        max(map(len, [get_vendor(pkg) if hasattr(pkg, "vendor") else get_packager(pkg) for pkg in pkgs])),
+        len("Vendor/Packager"))
 
     header = "%-*s  %-*s  %s" % (max_nvra_length, "Package", max_packager_length,
-                                 "Packager", "Repository") + "\n"
+                                 "Vendor/Packager", "Repository") + "\n"
     header_underline = "%-*s  %-*s  %s" % (max_nvra_length, "-" * len("Package"),
-                                           max_packager_length, "-" * len("Packager"),
+                                           max_packager_length, "-" * len("Vendor/Packager"),
                                            "-" * len("Repository")) + "\n"
 
     pkg_list = ""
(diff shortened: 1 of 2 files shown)


convert2rhel/pkghandler.py
```diff
@@ -316,7 +317,7 @@ def print_pkg_info(pkgs):
             from_repo = pkg._from_repo
 
         pkg_list += "%-*s  %-*s  %s" % (max_nvra_length, get_pkg_nvra(pkg),
-                                        max_packager_length, get_packager(pkg), from_repo) + "\n"
+                                        max_packager_length, get_vendor(pkg), from_repo) + "\n"
```
SpyTec, line 320: Need to check for vendor and have fallback here. <8.3 will otherwise only show N/A. I think it would be better if we have another function that does this check for us and calls `get_packager` and `get_vendor` as needed

SpyTec (changes requested):

convert2rhel/pkghandler.py
```diff
@@ -293,12 +292,14 @@ def print_pkg_info(pkgs):
     (https://bugzilla.redhat.com/show_bug.cgi?id=1876561).
     """
     max_nvra_length = max(map(len, [get_pkg_nvra(pkg) for pkg in pkgs]))
-    max_packager_length = max(max(map(len, [get_packager(pkg) for pkg in pkgs])), len("Packager"))
+    max_packager_length = max(
+        max(map(len, [get_vendor(pkg) if hasattr(pkg, "vendor") else get_packager(pkg) for pkg in pkgs])),
+        len("Vendor/Packager"))
```
ZhukovGreen, line 297: Nit:
It is not that readable, here you could find some concepts to improve the readability:
```python
max(itertools.chain((n*"*" for n in range(5)), ("******",)), key=len)
```

convert2rhel/pkghandler.py
```diff
@@ -366,6 +367,14 @@ def get_packager(pkg_obj):
     return packager.split("<", 1)[0].rstrip()
 
 
+def get_vendor(pkg_obj):
+    # The vendor may not be set for all packages
+    if hasattr(pkg_obj, "vendor") and pkg_obj.vendor:
+        return pkg_obj.vendor
+    else:
+        return "N/A"
```
ZhukovGreen, line 375: Nit:
I think more pythonic will be:
```python
try:
    return pkg_obj.vendor
except AttributeError:
    return "N/A"
```

ZhukovGreen (commented):
Couple of nits

convert2rhel/pkghandler.py
fellipeh: Nice catch!

fellipeh: Actually, I wrote like that, because, for some package vendor can come as None, use try..block will return None, on my code, I check it..

SpyTec (approved):

convert2rhel/pkghandler.py
```diff
@@ -366,6 +369,14 @@ def get_packager(pkg_obj):
     return packager.split("<", 1)[0].rstrip()
 
 
+def get_vendor(pkg_obj):
+    # The vendor may not be set for all packages
+    if hasattr(pkg_obj, "vendor") and pkg_obj.vendor:
```
sturivny, line 374: Looks good. I'd use here `try, except` construction. It will look simpler and will work faster

SpyTec: Artem brought it up in his review, but Fellipe mentioned vendor attribute may be `None`, meaning try-catch would not work as expected.

Anyhow, PR is already merged ;)

PR #134, opened 2020-12-11
add: Highlight warning PONR message and adds colors to INFO/WARNING messages

fellipeh:
Visually highlight the warning message about the Point of No Return and adds colors to INFO and WARNING

Here is a example:

image


SpyTec:
I like the tasks being colored, creates a clear separation. I'm not sure but I think that warning could use with only the "Warning" text being colored. Likewise, Critical messages needs to be colored too, possibly just having its background be red, and should only be for the "Critical" text, not the whole message.

I'd vote for
- tasks: green
- warning: yellow
- critical: red

fellipeh:
+1 @SpyTec  I agree with you, I like task colored, what do you think about this @bocekm ?

bocekm:
Sounds good, go for it.

fellipeh:
Updated the description with a result image. @bocekm  @SpyTec

convert2rhel/logger.py
```diff
@@ -117,6 +116,18 @@ def info(self, msg, *args, **kwargs):
...
+class bcolors:
+    HEADER = '\033[95m'
+    OKBLUE = '\033[94m'
+    OKCYAN = '\033[96m'
+    OKGREEN = '\033[92m'
+    WARNING = '\033[93m'
+    FAIL = '\033[91m'
+    ENDC = '\033[0m'
+    BOLD = '\033[1m'
+    UNDERLINE = '\033[4m'
```
SpyTec, line 128:
```suggestion
    OKGREEN = '\033[92m'
    WARNING = '\033[93m'
    FAIL = '\033[91m'
    ENDC = '\033[0m'
```

convert2rhel/unit_tests/logger_test.py
```diff
@@ -85,7 +85,7 @@ def test_log_format(self):
             logging.DEBUG, tempstr)
         self.check_formatter_result(
             logging.INFO, self.test_msg)
-        tempstr = "WARNING - %s" % self.test_msg
+        tempstr = "\x1b[93mWARNING - %s\x1b[0m" % self.test_msg
```
SpyTec, line 88: We should be consistent with how we defined it in logger
```suggestion
        tempstr = "\033[93mWARNING - %s\033[0m" % self.test_msg
```

SpyTec (changes requested):

convert2rhel/logger.py
```diff
@@ -31,7 +31,6 @@
 import sys
 from convert2rhel.utils import format_msg_with_datetime
 
-
```
ZhukovGreen, line 34: Nit: I guess two lines after imports is OK

convert2rhel/logger.py
```diff
@@ -117,6 +116,13 @@ def info(self, msg, *args, **kwargs):
         super(CustomLogger, self).info(msg, *args, **kwargs)
 
 
+class bcolors:
```
ZhukovGreen, line 119: Nit: I think enum.Enum is exactly for such cases. If compare with your solution I think the difference is that enum is immutable, uses less memory, and has a lot of nice features out of the box (i.e. iterator protocol, nice __str__ and __repr__ ...)

ZhukovGreen (commented):
Just a few nits:)

convert2rhel/logger.py
SpyTec: Unfortunately, `enum.Enum` is from `python >3.4` and we need to support 2.6. Python future doesn't seem to cover it easily either. Too much work :)

ZhukovGreen: :sob:

convert2rhel/unit_tests/logger_test.py
```diff
@@ -85,7 +85,7 @@ def test_log_format(self):
             logging.DEBUG, tempstr)
         self.check_formatter_result(
             logging.INFO, self.test_msg)
-        tempstr = "WARNING - %s" % self.test_msg
+        tempstr = "\033[93mWARNING - %s\033[0m" % self.test_msg
         self.check_formatter_result(
             logging.WARNING, tempstr)
```
SpyTec, line 90: Add a test for when removing/escaping coloring
```suggestion
        tempstr = "\033[93mWARNING - %s\033[0m" % self.test_msg
        self.check_formatter_result(
            logging.WARNING, tempstr)

        custom_formatter.escape_colors(True)
        tempstr = "WARNING - %s" % self.test_msg
        self.check_formatter_result(
            logging.WARNING, tempstr)
```

convert2rhel/logger.py
```diff
@@ -79,6 +78,7 @@ def initialize_logger(log_name):
         os.makedirs(LOG_DIR)
     handler = logging.FileHandler(os.path.join(LOG_DIR, log_name), "a")
     formatter = CustomFormatter("%(message)s")
+    formatter.escape_colors(False)
```
SpyTec, line 81:
```suggestion
    formatter.escape_colors(True)
```

convert2rhel/logger.py
```diff
@@ -117,24 +117,44 @@ def info(self, msg, *args, **kwargs):
...
+
+
 class CustomFormatter(logging.Formatter, object):
     """Custom formatter to handle different logging formats based on logging level
 
     Python 2.6 workaround - logging.Formatter class does not use new-style
         class and causes 'TypeError: super() argument 1 must be type, not
         classobj' so we use multiple inheritance to get around the problem.
     """
+    escape_color = True
```
SpyTec, line 134:
```suggestion
    escape_color = False
```

SpyTec (changes requested):

convert2rhel/logger.py
```diff
@@ -117,24 +117,44 @@ def info(self, msg, *args, **kwargs):
...
+
+    def escape_colors(self, value):
+        self.escape_color = value
 
     def format(self, record):
         if record.levelno == LogLevelTask.level:
             temp = '*' * (90 - len(record.msg) - 25)
-            self._fmt = "\n[%(asctime)s] %(levelname)s - [%(message)s] " + temp
+            fmt_orig = "\n[%(asctime)s] %(levelname)s - [%(message)s] " + temp
+            new_fmt = bcolors.OKGREEN + fmt_orig + bcolors.ENDC if self.escape_color else fmt_orig
```
SpyTec, line 143:
```suggestion
            new_fmt = fmt_orig if self.escape_color else bcolors.OKGREEN + fmt_orig + bcolors.ENDC
```

convert2rhel/logger.py
```diff
@@ -117,24 +117,44 @@ def info(self, msg, *args, **kwargs):
...
+            self._fmt = new_fmt
             self.datefmt = "%m/%d/%Y %H:%M:%S"
         elif record.levelno in [logging.INFO, LogLevelFile.level]:
             self._fmt = "%(message)s"
             self.datefmt = ""
         elif record.levelno in [logging.WARNING]:
-            self._fmt = "%(levelname)s - %(message)s"
+            fmt_orig = "%(levelname)s - %(message)s"
+            new_fmt = bcolors.WARNING + fmt_orig + bcolors.ENDC if self.escape_color else fmt_orig
```
SpyTec, line 151:
```suggestion
            new_fmt = fmt_orig if self.escape_color else bcolors.WARNING + fmt_orig + bcolors.ENDC
```

convert2rhel/logger.py
```diff
@@ -117,24 +117,44 @@ def info(self, msg, *args, **kwargs):
...
+            self._fmt = new_fmt
+            self.datefmt = ""
+        elif record.levelno in [logging.CRITICAL]:
+            fmt_orig = "%(levelname)s - %(message)s"
+            new_fmt = bcolors.FAIL + fmt_orig + bcolors.ENDC if self.escape_color else fmt_orig
```
SpyTec, line 156:
```suggestion
            new_fmt = fmt_orig if self.escape_color else bcolors.FAIL + fmt_orig + bcolors.ENDC
```

SpyTec (changes requested):

convert2rhel/unit_tests/logger_test.py
fellipeh: I believe default = True it's better, since we have colors in all info (info, warning and error)... and we don't have colors only in file.

convert2rhel/logger.py
fellipeh: Same comment as above ^^

convert2rhel/logger.py
fellipeh: Same comment as above ^^

convert2rhel/unit_tests/logger_test.py
SpyTec: It doesn't really matter, but due to the choice of words with `escape_colors` rather than something like `enable_colors` it should be `False` by default. Likewise, since CLI is our main use case, we should go by that rather than by logs - i.e. colors enabled by default.

Also I just realized, but I believe that some orchestration and logging tools will listen to stdout rather than logs too. It might make sense to have a CLI option to disable it as well to not mess up their own logging

SpyTec: Just missing the test for when disabling colors like in the suggestion here

SpyTec (changes requested):
Just one note about tests, otherwise LGTM

SpyTec (approved):

PR #136, opened 2020-12-17
Add back pylint 2.6 check

bocekm:
There was an issue with unavailable CentOS 6 repos. They have moved to Vault.

Reverts the removal from https://github.com/oamg/convert2rhel/pull/126.


fellipeh (approved):
LGTM

PR #138, opened 2020-12-29
add: check single-user mode

fellipeh:
Require the system-to-be-converted to be booted in a single-user mode.
That would prevent other user sessions from doing changes to the system during the conversion.

WARNING:
- Need to update the vagrant test box after merge this PR.
- Need to update the Red Hat Knowledgebase article (https://access.redhat.com/articles/2360841)

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -66,6 +66,11 @@ def main():
         loggerinst.task("Prepare: Determine RHEL variant")
         rhelvariant.determine_rhel_variant()
 
+        # We check for the single-user mode only after the resolve_system_info() call because that function stops
+        # the conversion in case the system vendor/major version is not supported and we don't want users to go
+        # through booting into single-user mode to just find this out.
+        utils.require_single_user_mode()
+        
         # backup system release file before starting conversion process
         loggerinst.task("Prepare: Backup System")
         redhatrelease.system_release_file.backup()
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -532,6 +532,21 @@ class RestorablePackage(object):
             loggerinst.warning("Can't access %s" % TMP_DIR)
 
 
+def require_single_user_mode():
+    """
+    Make sure that the system is running in single-user mode and exit if it is not.
+    """
+    loggerinst = logging.getLogger(__name__)
+
+    output, ret_code = run_cmd_in_pty("runlevel", print_cmd=False)
+    runlevel = output.replace("\r\n", "")[-1]
+
+    if ret_code != 0:
+        loggerinst.critical("Unable to determine if the system runs in single-user mode.")
+    elif runlevel != '1':
+        loggerinst.critical("Convert2RHEL requires the system to run in single-user mode.")
+
+
 changed_pkgs_control = ChangedRPMPackagesController()  # pylint: disable=C0103
 
 
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -212,3 +212,54 @@ class TestUtils(unittest.TestCase):
 
     def test_is_rpm_based_os(self):
         assert is_rpm_based_os() in (True, False)
+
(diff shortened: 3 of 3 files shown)


convert2rhel/main.py
```diff
@@ -56,6 +56,9 @@ def main():
 
         process_phase = ConversionPhase.POST_CLI
 
+        # the tool will not run if the system not executed in single-user mode
+        utils.check_if_single_user_mode()
+
```
SpyTec, line 61: Call this after `rhelvariant.determine_rhel_variant()` so we know whether the system can be converted or not. Otherwise they might boot into single-user mode, and not accept EULA or simply not be on a compatible system

SpyTec (changes requested):

ZhukovGreen (commented):
went through, nothing spoted:)

SpyTec (commented):
Sorry Fellipe for a second round of reviews. This should just be a select and batch-commit situation though.

I suggest some changes so that

- We are consistent with other function names
- Fixed a typo
- Renamed tests
- Reworded logging to be in active voice
- More accurate `runlevel` list fetching (`[-1]` instead of `[2:]`)

SpyTec (approved):
LGTM

convert2rhel/main.py
```diff
@@ -66,6 +66,9 @@ def main():
         loggerinst.task("Prepare: Determine RHEL variant")
         rhelvariant.determine_rhel_variant()
 
+        # prohibit converting if system is not in single-user mode
+        utils.require_single_user_mode()
```
bocekm, line 70: I'd move this up, right below calling the `utils.require_root()`. Both functions seem similar in purpose to me, so I feel they should be closer to each other in the code.
@ZhukovGreen, @SpyTec, WDYT?

bocekm (changes requested):

convert2rhel/utils.py
```diff
@@ -532,4 +532,19 @@ def backup(self):
...
+    """
+    loggerinst = logging.getLogger(__name__)
+
+    output, ret_code = run_cmd_in_pty("runlevel", print_cmd=False)
+    runlevel = output.replace("\r\n", "")[-1]
+
+    if ret_code != 0:
+        loggerinst.critical("Unable to determine if system is in single-user mode. Convert2RHEL requires single-user mode to run.")
+    elif runlevel != '1':
+        loggerinst.critical("Convert2RHEL requires single-user mode to run.")
```
bocekm, line 547:
```suggestion
        loggerinst.critical("Unable to determine if the system runs in single-user mode.")
    elif runlevel != '1':
        loggerinst.critical("Convert2RHEL requires the system to run in single-user mode.")
```

bocekm (changes requested):

convert2rhel/main.py
ZhukovGreen: Sounds logical to me @bocekm . However, error handling (starts on line 104) won't work. I think this is OK.

ZhukovGreen (commented):
just answered Michal request

convert2rhel/main.py
SpyTec: @bocekm What is the reason for that? My reasoning for moving it down was so that the user has gone through EULA and know whether the system can be converted.

If we move it further up they might end up rebooting to single-user only to then not go through with the conversion due to incompatible system or not agreeing to EULA

bocekm: @SpyTec, you're right, that makes sense.
@fellipeh, please update the comment to include a comment like:
_We check for the single-user mode only after the resolve_system_info() call because that function stops the conversion in case the system vendor/major version is not supported and we don't want users to go through booting into single-user mode to just find this out._

fellipeh:
Rebased and update requested changes

SpyTec (approved):

bocekm (approved):

bocekm:
@fellipeh, please revert this PR. Single-user mode comes with no network which would hamper all the conversions except using a mounted RHEL DVD ISO (or with RHEL repos available locally in general).

PR #139, opened 2021-01-04
Remove RHEL variant options

fellipeh:
This PR remove the option for choosing a RHEL variant, using "Server" as default.
And also inform the user in case of use this option.


convert2rhel/toolopts.py
```diff
@@ -71,69 +71,68 @@ def _get_argparser():
     def _register_options(self):
         """Prescribe what command line options the tool accepts."""
         self._parser.add_option("-h", "--help", action="help", help="Show "
-                                " help message and exit.")
+                                                                    " help message and exit.")
```
sturivny, line 74: Why did you add this change here? Did you mean "    <text>" not             "<text>"?

fellipeh: It`s a vscode format issue.. fixed.

convert2rhel/toolopts.py
```diff
@@ -30,7 +31,7 @@ def __init__(self):
         self.enablerepo = []
         self.disablerepo = []
         self.pool = None
-        self.variant = None
+        self.variant = "Server"
```
SpyTec, line 34: We should remove it completely. For all usages of variant we should just replace it with the string `server`
```suggestion
```

convert2rhel/toolopts.py
```diff
@@ -56,84 +57,80 @@ def _get_argparser():
...
-        group.add_option("-v", "--variant", help="The RHEL variant to which"
-                         " the system will be converted. Available variants"
-                         " are: Server, Client, Workstation, ComputeNode. If"
-                         " not used, the user is asked to choose a variant.")
+                                        " the user is asked to choose from the available"
+                                        " subscriptionsincorret rhel variant. A list of the available"
+                                        " subscriptions is possible to obtain by running"
+                                        " 'subscription-manager list --available'.")
+        group.add_option("-v", "--variant", help="The RHEL variant params is not used anymore."
+                                                 " Convert2RHEL will user Server variant as default.")
```
SpyTec, line 129: Can't we simply remove this? Or do we keep this just for the logging?

If we leave it in it will still be displayed in the help page and take up space. If there's a way to remove it but still be able to detect the argument to warn the user it would be best. Should be possible with this: https://docs.python.org/2.6/library/optparse.html#optparse.OptionParser.get_option

Otherwise I think we should just remove it and simply ignore the argument and go about the conversion.
```suggestion
```

SpyTec (changes requested):

convert2rhel/toolopts.py
ZhukovGreen: Nit: This string "Server" is a constant, probably it has the sense to move it to the top of the module. (i.e.
```python
DEFAULT_VARIANT = "Server"
```

ZhukovGreen (commented):
my 2 cents

convert2rhel/toolopts.py
SpyTec: We use `toolopts.variant` in two spots I believe, I don't believe it's worth keeping the variable as we'll always use server

convert2rhel/toolopts.py
fellipeh: If we remove this options, will fail when user try to use... on line 201 we check if user trying to use the params.. if we remove this options will fail there.

so, I believe it's a good approach to have this options, and warning the user as I did.. after some versions we can remove it.

SpyTec: That's why I'm asking if it's possible to use `get_option`. Getting `parsed_opts.variant` will inevitably fail if we remove this. We shouldn't have variant in help if it does nothing

fellipeh: The problem is if we use `get_options` we lost the params options.. if you try to pass  you receive something like that: 
```bash
Usage: 
  convert2rhel [-h]
  convert2rhel [-u username] [-p password | -f pswd_file] [--pool pool_id | -a] [--disablerepo repoid] [--enablerepo repoid] [--serverurl url] [--no-rpm-va] [--debug] [--restart] [-y]
  convert2rhel [--disable-submgr] [--disablerepo repoid] [--enablerepo repoid] [--no-rpm-va] [--debug] [--restart] [-y]
  convert2rhel [-k key] [-o organization] [--pool pool_id | -a] [--disablerepo repoid] [--enablerepo repoid] [--serverurl url] [--no-rpm-va] [--debug] [--restart] [-y]

convert2rhel: error: no such option: --variant
```

so we need to decide if we move forward to show a user a info message, or just remove it..

SpyTec: Found two possible solutions. One involves `argparse`, the successor to `optparse`, as it can suppress specific options to be displayed. Meaning it will still parse but not show in `--help` or man pages.

The other option, which I think we can go with now, is by checking if variant is in `sys.argv`. This means we only add variant to the parser if it's already part of the CLI arguments. Just remember to `import sys`.

```suggestion
        if '--variant' in sys.argv:
            group.add_option("-v", "--variant", help="The RHEL variant params is not used anymore."
                                                     " Convert2RHEL will user Server variant as default.")
```

convert2rhel/toolopts.py
```diff
@@ -56,84 +56,80 @@ def _get_argparser():
...
-                         " the user is asked to choose from the available"
-                         " subscriptions. A list of the available"
-                         " subscriptions is possible to obtain by running"
-                         " 'subscription-manager list --available'.")
-        group.add_option("-v", "--variant", help="The RHEL variant to which"
-                         " the system will be converted. Available variants"
-                         " are: Server, Client, Workstation, ComputeNode. If"
-                         " not used, the user is asked to choose a variant.")
+                                        " the user is asked to choose from the available"
+                                        " subscriptionsincorret rhel variant. A list of the available"
```
SpyTec, line 124:
```suggestion
                                        " subscriptions. A list of the available"
```

convert2rhel/toolopts.py
```diff
@@ -201,19 +199,15 @@ def _process_cli_options(self):
...
+        if '--variant' in sys.argv[1:]:
+            loggerinst.warning("Variant option is not supported anymore."
+                               "Convert2RHEL will always use Server variant from now.")
+            utils.ask_to_continue()
 
         if parsed_opts.serverurl:
             if parsed_opts.disable_submgr:
                 loggerinst.warn("Ignoring the --serverurl option. It has no effect when --disable-submgr is used.")
-            else:
+            else:
```
SpyTec, line 210:
```suggestion
            else:
```

SpyTec (approved):

PR #140, opened 2021-01-05
Add missing CentOS 8 packages - to be removed prior the conversion

bocekm:
In CentOS 8.3 the package that used to be named `centos-release` in CentOS 8.2 was renamed to `centos-linux-release`.
Also, the `centos-repos`/`centos-linux-repos` and `centos-obsolete-packages` were missing in the list of pkgs to remove thus not being removed during the conversion.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1913008

### DIFF
--- a/convert2rhel/data/8/x86_64/configs/centos-8-x86_64.cfg
+++ b/convert2rhel/data/8/x86_64/configs/centos-8-x86_64.cfg
@@ -8,10 +8,10 @@ gpg_fingerprints = 05b555b38483c65d
 # List of packages that have to be removed before the system conversion starts.
 # Delimited by any whitespace(s).
 excluded_pkgs =
-  centos-logos{,-httpd}
+  centos-logos*
   centos-indexhtml
-  centos-{,linux-}release
-  centos-{,linux-}repos
+  centos*-release
+  centos*-repos
   centos-obsolete-packages
   redhat-release*
   rhn*
--- a/convert2rhel/data/8/x86_64/configs/centos-8-x86_64.cfg
+++ b/convert2rhel/data/8/x86_64/configs/centos-8-x86_64.cfg
@@ -8,9 +8,11 @@ gpg_fingerprints = 05b555b38483c65d
 # List of packages that have to be removed before the system conversion starts.
 # Delimited by any whitespace(s).
 excluded_pkgs =
-  centos-logos
+  centos-logos{,-httpd}
   centos-indexhtml
-  centos-release*
+  centos-{,linux-}release
+  centos-{,linux-}repos
+  centos-obsolete-packages
   redhat-release*
   rhn*
   python3-rhn*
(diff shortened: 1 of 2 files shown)


convert2rhel/data/8/x86_64/configs/centos-8-x86_64.cfg
```diff
@@ -10,7 +10,7 @@ gpg_fingerprints = 05b555b38483c65d
 excluded_pkgs =
   centos-logos
   centos-indexhtml
-  centos-release*
+  centos*release*
```
ZhukovGreen, line 13: @bocekm maybe more strict globbing. I am not sure how do you read the config, but what about something like:
```bash
centos-{,linux-}release
```

ZhukovGreen (approved):
Optional comment

convert2rhel/data/8/x86_64/configs/centos-8-x86_64.cfg
bocekm: I'll check if such a globbing is supported by dnf.
We would need to change it also [for Oracle Linux](https://github.com/oamg/convert2rhel/blob/main/convert2rhel/data/8/x86_64/configs/oracle-8-x86_64.cfg#L14) to stay consistent.
I'm not sure if making it strict is necessary since in the end we want to get rid of all the centos packages. I wouldn't mind even `centos-*` :)

bocekm: I used your suggested globbing and added the missing centos-repos. And centos-obsolete-packages.

ZhukovGreen (approved):

fellipeh (dismissed):
LGTM

convert2rhel/data/8/x86_64/configs/centos-8-x86_64.cfg
bocekm: Unfortunately, this kind of globbing does not work with dnf API filtering: 
https://github.com/oamg/convert2rhel/blob/416c45574b5c4560d9a2b2afbae9b8895bfa0920/convert2rhel/pkghandler.py#L243
Even though, per docs, it should: https://dnf.readthedocs.io/en/latest/api_queries.html#dnf.query.Query.filter

```
centos-logos{,-httpd} ........................... 0
centos-indexhtml ................................ 0
centos-{,linux-}release ......................... 0
centos-{,linux-}repos ........................... 0
```

bocekm: Filed a bug for dnf https://bugzilla.redhat.com/show_bug.cgi?id=1913418. In the meantime, we'll need to stick with "*".

bocekm:
@ZhukovGreen, @fellipeh, please re-review changes based on https://github.com/oamg/convert2rhel/pull/140#discussion_r552873476.

ZhukovGreen (approved):

ZhukovGreen:
@bocekm Done

PR #141, opened 2021-01-06
Use Vault in CentOS 6 Docker image

bocekm:
CentOS 6 has been retired and its repos moved to the Vault. We use CentOS 6 Docker image for unit testing, so we need to point the yum repos to the Vault now.

Related: #136


ZhukovGreen (approved):
Tested, works as expected. Thx

PR #142, opened 2021-01-06
Allow running tests on non CentOS systems

ZhukovGreen:
After this MR it will be possible to run tests locally on non-CentOS systems.
Also, python2.6, 2.7, 3 compatible `skipIf` decorator is introduced, that allows to skip tests on condition. API is similar to https://docs.python.org/2/library/unittest.html#unittest.skipIf 

Skipped tests are reported in the testing session log:
```
centos6_1  | ............/data/convert2rhel/unit_tests/__init__.py:182: UserWarning: Test <function test_call_yum_cmd_w_downgrades_continuous_fail at 0x1fba7d0> skipped. Reason: Current test runs only on rpm based systems.
centos6_1  |   warn("Test %r skipped. Reason: %s" % (test_item, reason))
centos6_1  | ...................../data/convert2rhel/unit_tests/__init__.py:182: UserWarning: Test <function test_get_rpm_header at 0x1fc41b8> skipped. Reason: Current test runs only on rpm based systems.
centos6_1  |   warn("Test %r skipped. Reason: %s" % (test_item, reason))
centos6_1  | ..................................................../data/convert2rhel/unit_tests/__init__.py:182: UserWarning: Test <function test_modified_rpm_files_diff_with_differences_after_conversion at 0x2034410> skipped. Reason: Current test runs only on rpm based systems.
centos6_1  |   warn("Test %r skipped. Reason: %s" % (test_item, reason))
centos6_1  | .........................
centos6_1  | ----------------------------------------------------------------------
centos6_1  | Ran 110 tests in 0.284s
centos6_1  | 
centos6_1  | OK
convert2rhel_centos6_1 exited with code 0
centos7_1  | ..............................................................................................................
centos7_1  | ----------------------------------------------------------------------
centos7_1  | Ran 110 tests in 1.970s
centos7_1  | 
centos7_1  | OK
centos8_1  | ..............................................................................................................
centos8_1  | ----------------------------------------------------------------------
centos8_1  | Ran 110 tests in 2.061s
centos8_1  | 
centos8_1  | OK
convert2rhel_centos7_1 exited with code 0
convert2rhel_centos8_1 exited with code 0
```

### DIFF
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -28,6 +28,8 @@ from convert2rhel import utils
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
+from convert2rhel.unit_tests import skipIf
+from convert2rhel.utils import is_rpm_based_os
 
 
 class TestPkgHandler(unit_tests.ExtendedTestCase):
@@ -154,6 +156,7 @@ class TestPkgHandler(unit_tests.ExtendedTestCase):
 
         self.assertEqual(utils.run_subprocess.cmd, "yum install -y")
 
+    @skipIf(not is_rpm_based_os(), reason="Current test runs only on rpm based systems.")
     @unit_tests.mock(pkghandler, "call_yum_cmd", CallYumCmdMocked())
     def test_call_yum_cmd_w_downgrades_continuous_fail(self):
         pkghandler.call_yum_cmd.return_code = 1
@@ -372,6 +375,7 @@ class TestPkgHandler(unit_tests.ExtendedTestCase):
 
     @unit_tests.mock(logger.CustomLogger, "warning", LogMocked())
     @unit_tests.mock(rpm, "TransactionSet", TransactionSetMocked())
+    @skipIf(not is_rpm_based_os(), reason="Current test runs only on rpm based systems.")
     def test_get_rpm_header(self):
         pkg = TestPkgHandler.create_pkg_obj(name="pkg1", version="1", release="2")
         hdr = pkghandler.get_rpm_header(pkg)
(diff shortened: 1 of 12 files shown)


convert2rhel/utils.py
```diff
@@ -533,3 +533,13 @@ def backup(self):
 
 
 changed_pkgs_control = ChangedRPMPackagesController()  # pylint: disable=C0103
+
+
+def get_os_release():
+    """Parse /etc/os-release into dict."""
+    os_release = {}
+    with open("/etc/os-release") as f:
```
bocekm, line 541:
```suggestion
    with open("/etc/system-release") as f:
```

`os-release` is not on CentOS/OL/RHEL 6. `system-release` is on all major versions - 6, 7, 8.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -155,6 +155,10 @@ def test_call_yum_cmd(self):
         self.assertEqual(utils.run_subprocess.cmd, "yum install -y")
 
     @unit_tests.mock(pkghandler, "call_yum_cmd", CallYumCmdMocked())
+    @unittest.skipIf(
+        not get_os_release()["NAME"].startswith("CentOS"),
+        reason="This test assumed to be run on CentOS system only."
```
bocekm, line 160: Actually, the test would run fine on CentOS, Oracle Linux and RHEL. I was brave enough to try now and they run on Fedora as well, all pass.
So I'd create another function like `is_rpm_based_os()` where you'd be comparing the what we call [system id](https://github.com/oamg/convert2rhel/blob/main/convert2rhel/systeminfo.py#L74) in convert2rhel (just a lower case first word of `/etc/system-release`) to oracle/centos/red/fedora, and use this function to skip the test if not on rpm bases system.

bocekm (changes requested):

bocekm (approved):

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -131,6 +128,7 @@ def test_modified_rpm_files_diff_without_differences_after_conversion(self):
         data=[['.M.......  g /etc/pki/ca-trust/extracted/java/cacerts'],
               ['.M.......  g /etc/pki/ca-trust/extracted/java/cacerts',
                'S.5....T.  c /etc/yum.conf']]))
+    @skipIf(not is_rpm_based_os(), reason="Current test runs only on rpm based systems.")
```
bocekm, line 131: I'm getting the following on CentOS 6 when running the tests with `make tests`:
```
test_modified_rpm_files_diff_with_differences_after_conversion (convert2rhel.unit_tests.systeminfo_test.TestSysteminfo) ... /data/convert2rhel/unit_tests/__init__.py:182: UserWarning: Test <function test_modified_rpm_files_diff_with_differences_after_conversion at 0x234c140> skipped. Reason: Current test runs only on rpm based systems.
  warn("Test %r skipped. Reason: %s" % (test_item, reason))
o
```

convert2rhel/utils.py
```diff
@@ -533,3 +533,13 @@ def backup(self):
 
 
 changed_pkgs_control = ChangedRPMPackagesController()  # pylint: disable=C0103
+
+
+def is_rpm_based_os():
```
bocekm, line 538: Since this function is used in tests only, it should be rather in `unit_tests/__init__.py`.

convert2rhel/unit_tests/systeminfo_test.py
ZhukovGreen: This will be soved by using @pytest.mark.skipif after #150 will be merged

convert2rhel/utils.py
ZhukovGreen: solved in #151

PR #148, opened 2021-01-14
Improve logging of single-user mode check

bocekm:
The single-user mode check was kind of hidden:
- the check itself was not announced by the tool (TASK),
- ~~executing of the `runlevel` command was not logged,~~ (fixed in https://github.com/oamg/convert2rhel/pull/149)
- ~~the output of the `runlevel` command was always printed to the standard output - that was very mysterious (suddenly seeing `N 3` on the output)~~ (fixed in https://github.com/oamg/convert2rhel/pull/149)

Before:
```
[01/14/2021 22:52:25] TASK - [Prepare: Gather system information] *******************************
Name:                CentOS Linux
OS major version:    7
[01/14/2021 22:52:25] DEBUG - Calling command 'uname -i'
Architecture:        x86_64
Config filename:     centos-7-x86_64.cfg
Skipping the execution of 'rpm -Va'.
N 3
CRITICAL - Convert2RHEL requires the system to run in single-user mode.
[01/14/2021 22:52:25] DEBUG - Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/convert2rhel/main.py", line 69, in main
    utils.require_single_user_mode()
  File "/usr/lib/python2.7/site-packages/convert2rhel/utils.py", line 548, in require_single_user_mode
    loggerinst.critical("Convert2RHEL requires the system to run in single-user mode.")
  File "/usr/lib/python2.7/site-packages/convert2rhel/logger.py", line 106, in critical
    sys.exit(msg)
SystemExit: Convert2RHEL requires the system to run in single-user mode.
```

After:
```
[01/14/2021 22:53:52] TASK - [Prepare: Gather system information] *******************************
Name:                CentOS Linux
OS major version:    7
[01/14/2021 22:53:52] DEBUG - Calling command 'uname -i'
Architecture:        x86_64
Config filename:     centos-7-x86_64.cfg
Skipping the execution of 'rpm -Va'.

[01/14/2021 22:53:52] TASK - [Prepare: Require single-user mode] ********************************
[01/14/2021 22:53:52] DEBUG - Calling command 'runlevel'
CRITICAL - Convert2RHEL requires the system to run in single-user mode.
[01/14/2021 22:53:52] DEBUG - Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/convert2rhel/main.py", line 70, in main
    utils.require_single_user_mode()
  File "/usr/lib/python2.7/site-packages/convert2rhel/utils.py", line 546, in require_single_user_mode
[...]

PR #149, opened 2021-01-15
fix: SingleUser mode check function

fellipeh:
Change `run_cmd_in_pty` to `run_subprocess` into check single user function.

### DIFF
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -538,8 +538,9 @@ def require_single_user_mode():
     """
     loggerinst = logging.getLogger(__name__)
 
-    output, ret_code = run_cmd_in_pty("runlevel", print_cmd=False)
-    runlevel = output.replace("\r\n", "")[-1]
+    runlevel_path = pexpect.which("runlevel")
+    output, ret_code = run_subprocess(cmd=runlevel_path, print_cmd=True, print_output=False)
+    runlevel = output.split()[-1]
 
     if ret_code != 0:
         loggerinst.critical("Unable to determine if the system runs in single-user mode.")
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -0,0 +1,96 @@
+[tool.black]
```
SpyTec, line 1: File shouldn't be part of this PR

fellipeh: You're right... forgot to put into gitignore

convert2rhel/utils.py
```diff
@@ -538,8 +538,9 @@ def require_single_user_mode():
     """
     loggerinst = logging.getLogger(__name__)
 
-    output, ret_code = run_cmd_in_pty("runlevel", print_cmd=False)
-    runlevel = output.replace("\r\n", "")[-1]
+    runlevel_path = pexpect.which("runlevel")
+    output, ret_code = run_subprocess(cmd=runlevel_path, print_cmd=False)
```
bocekm, line 542:
```suggestion
    output, ret_code = run_subprocess(cmd=runlevel_path, print_cmd=True, print_output=False)
```
See https://github.com/oamg/convert2rhel/pull/148.

convert2rhel/utils.py
```diff
@@ -538,8 +538,9 @@ def require_single_user_mode():
...
+    runlevel = output.replace("\r\n", "").replace("\n", "")[-1]
```
bocekm, line 543:
```suggestion
    runlevel = output.split()[-1]
```
The output is `N <number>`, e.g. `N 1`. You can use `split()` to split it to `['N', '1']`.

bocekm (changes requested):

convert2rhel/utils.py
ZhukovGreen: maybe also int(....) to have a valid int

bocekm (approved):

PR #150, opened 2021-01-15
pytest, split requirements to main and app, unify dockerfiles

ZhukovGreen:
Resolves: https://issues.redhat.com/browse/OAMG-1242
This merge request introduces running pytest as the main testing framework. Plus dependencies are systemized and docker files improved.

The motivation for using pytest as a testing framework (not sorted by the importance):
- Availability of many useful plugins in pytest ecosystem
- Very rich default set of functionality (builtins fixtures, tailoring the cli command options) 
- The pytest ecosystem still backward compatible to run unittest testing suits. It even allows mixing the way of configuring or wirte tests (i.e. using @pytest.mark.skipif for unittest test)
- Developers who are not familiar with the unittest can write test in pytest (and wise versa)
- Parametrization and powerful fixtures creation facility with different scopes
- Richer and more configurable ecosystem

The primary motivation in optimizing the dependencies and Dockerfiles is to bring the single source of the dependencies to all development environments (local, docker, Github). 

Todos:
- [x] Github actions doesn't use the provided images, yet. 
- [x] script run_unit_tests.sh left in place, because I expect it has broader usage then I can see it from the github repo
- [x] adjust plans/unittests.fmf to point to pytest (and remove run_tests script then?)
Solved by moving to github actions
- [x] remove pytest config part related to live logging (as a default option, then devs could adjust it locally based on their preferences)
- [x] Solve with @SpyTec how to better specify dependencies 
- [x] create Makefile install command to install the requirements in the virtualenv
- [x] Use pytest-cov for reporting test coverage
- [x] Run linters similarly to tests githhub actions


ZhukovGreen:
@bocekm @fellipeh @sturivny @SpyTec guys, could you please try it out. Is there some issues with this MR affecting your dev workflow. Thx. Any feedback on this very apprecited

SpyTec:
For us the only care we really care about with benefits is making testing easier, anything else is a bonus. Biggest draw to go for pytest or something else was due to the special treatment for py2.6 tests

The `run_unit_tests.sh` is being used by the testing farm, should be replaced there as well from within `plans/unittests.fmf`

I ran the tests in the PR but Convert2RHEL logging is active in the tests and pollutes the output, making the results unusable.

I'm also getting a CentOS 8 image failing, but seems unrelated to this PR 
```
Failed:
  filesystem-3.8-2.el8.x86_64            filesystem-3.8-3.el8.x86_64           

Error: Transaction failed
Error: error building at STEP "RUN dnf update -y && dnf install -y $APP_MAIN_DEPS && dnf clean all": error while running runtime: exit status 1
```

ZhukovGreen:
@SpyTec thx for trying this out. The live logging is easy to fix (https://github.com/oamg/convert2rhel/pull/150/files#diff-fb6a686182f16eb54af3c628f38593f347f68aba31de903983023c560288d7a1R4-R6) . I can remove these lines from the config, this is a matter of taste thing, in my opinion. The benefit of live logging is more for the situations when you run a subset of tests, so you see the logs together with the test session. 

About having separate requirements for each python we test. @SpyTec Could you evolve please why we shouldn't mix major versions of pytest? Maybe, I don't know something. I think each newer version of pytest brings useful reporting and performance improvements (keeping still the same API for tests). For example, reporting warnings.

Thx for showing the fmf things.

fellipeh:
Works fine for me, following these steps:
```bash
$ make images
$ make tests
```

SpyTec:
> @SpyTec thx for trying this out. The live logging is easy to fix ([#150 (files)](https://github.com/oamg/convert2rhel/pull/150/files#diff-fb6a686182f16eb54af3c628f38593f347f68aba31de903983023c560288d7a1R4-R6)) . I can remove these lines from the config, this is a matter of taste thing, in my opinion. The benefit of live logging is more for the situations when you run a subset of tests, so you see the logs together with the test session. Plus you always have a summary line at the end of testing session

Now we're going into the theory of what test suites should be. We want to be able to quickly process the information the test suite tells us, what tests were run, did the tests pass, where are the failing tests in the output. Logging in large does not help with these questions, instead it obscures and hides the results we want.

You mention that we would be able to see logs when running a subset of tests, but that doesn't tell me why we want that. If we want to see that the logs look correctly then we're looking at the wrong type of tests. Unit tests in particular is only there to test units, small pieces of code. We can make assertions about logging in the tests, but it is not helpful for printing it in the test suite output. You could instead argue that printing this out as part of a long e2e test would be good, which is what we're trying to do with the VMs you and Serhii is working on, as we are interesting in making sure everything works as it should.

If the logs were meant to help with debugging, it's the wrong tool for the job. We don't need a hammer for a screw, instead of looking at logs for unit tests you look at the outgoing effects from the unit and find out why the tests are failing/showing false positives and so on.

> 
> About having separate requirements for each python we test. @SpyTec Could you evolve please why we shouldn't mix major versions of pytest? Maybe, I don't know something. I think each newer version of pytest brings useful reporting and performance improvements (keeping still the same API for tests). For example, reporting warnings.

Major versions mean major and breaking changes. Utilizing two major versions is asking for trouble. We are not going to lose out on anything by using 3 instead of 4 as we are already required to support 3 due to py2.6. Reporting is also not ideal to generate by using higher major versions than 3 since it may change how the test suite works, thus giving different reports if you ran it in v3 vs running it in v4. Performance is the only benefit, but we are not running large test suites that takes 30 minutes, our tests suites are very short and do not take more than a minute, so test suite performance should be very low on our list of priorities

ZhukovGreen:
@SpyTec I resolved your comment about logs inside the testing session. However, 

I think it is better to use different requirements.txt for different python versions. If you want I can use https://www.python.org/dev/peps/pep-0508/#environment-markers to have one requirements.txt, which based on the py version decides which package to install. But, in my opinion, multiple txts seems cleaner to me.

This is very typical for python projects to use either separate requirements.txt for each python version or env markers, so you shouldn't each time to look for the minimal common divider:)

And the **unit tests API** (the way we write tests) doesn't change for all pytest versions we have:
- pytest 3.2.5, 4.6.1, 6.2.4 (we have now)
This is not accidentally. And this is the most important fact for us. Managing dependencies separately provides us access to the most recent framework features, still keeping all our multi python compatibility. And itis just easier to manage the dependencies separately for separate pythons.

SpyTec:
> @SpyTec I resolved your comment about logs inside the testing session. However,

Thanks, though I do think it should be verbose enough to display which tests it runs within each file.

> I think it is better to use different requirements.txt for different python versions. If you want I can use [python.org/dev/peps/pep-0508/#environment-markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) to have one requirements.txt, which based on the py version decides which package to install. But, in my opinion, multiple txts seems cleaner to me.
> 
> This is very typical for python projects to use either separate requirements.txt for each python version or env markers, so you shouldn't each time to look for the minimal common divider:)

PEP-0508 is more for installing dependencies for specific versions to make sure your application works as expected. Not mixing dependency versions for seemingly no reason. If we have dependencies we need for specific python version, then yes using multiple requirements.txt can be helpful if you don't want to write it using PEP-0508.

But it is not typical for applications in any language to use different major dependencies willy-nilly. You lock down the dependency versions to prevent issues and save you the hassle of dealing with different versions for no gain whatsoever.

> 
> And the **unit tests API** (the way we write tests) doesn't change for all pytest versions we have:
> 
>     * pytest 3.2.5, 4.6.1, 6.2.4 (we have now)
>       This is not accidentally. And this is the most important fact for us. Managing dependencies separately provides us access to the most recent framework features, still keeping all our multi python compatibility. And itis just easier to manage the dependencies separately for separate pythons.

Which is to be expected given we do not use pytest in our tests, we use unittest. This PR only tackles using pytest as the test suite runner. Once we start utilizing pytest in our tests we will most likely run into issues due to the major version difference, even having minor version differences can cause problems (one feature existing in a new minor version that isn't in the old ones). 

The only thing we should test using multiple dependencies is python itself, which we already do, as we want to be able to support multiple python versions. The other reason for running tests for other specific dependencies would be for peer dependencies where the user themselves install and provide the dependency rather than us putting it in `requirements.txt`, though from my knowledge this isn't something that is widely used in Python.

sturivny (changes requested):
Please, use strategy `matrix` here:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

requirements/centos7.requirements.txt
```diff
@@ -0,0 +1,4 @@
+pylint==1.9.5
+astroid==1.6.6
+pytest==4.6.11
```
bocekm, line 3:
```
ERROR: pip's legacy dependency resolver does not consider dependency conflicts when selecting packages. This behaviour is the source of the following dependency conflicts.
pytest 4.6.11 requires six>=1.10.0, but you'll have six 1.9.0 which is incompatible.
```
It's because the latest python-six package available in centos 7 (installed per the Dockerfiles/centos7.Dockerfile) is python-six-1.9.0-2.el7.
I suppose you need to add six>=1.10.0 to this requirements file.

Makefile
```diff
@@ -12,25 +39,27 @@ clean:
 
 .PHONY: images
 images:
-	@docker build -f Dockerfiles/centos6 -t $(IMAGE)/centos6 .
-	@docker build -f Dockerfiles/centos7 -t $(IMAGE)/centos7 .
-	@docker build -f Dockerfiles/centos8 -t $(IMAGE)/centos8 .
+	@docker build -f Dockerfiles/centos6.Dockerfile -t $(IMAGE)/centos6 .
+	@docker build -f Dockerfiles/centos7.Dockerfile -t $(IMAGE)/centos7 .
+	@docker build -f Dockerfiles/centos8.Dockerfile -t $(IMAGE)/centos8 .
```
bocekm, line 44: Building the centos 8 image fails on:
```
  Upgrading        : filesystem-3.8-3.el8.x86_64                         13/256 
Error unpacking rpm package filesystem-3.8-3.el8.x86_64
  Installing       : glibc-langpack-en-2.28-127.el8.x86_64               14/256 
error: unpacking of archive failed on file /proc: cpio: chown
error: filesystem-3.8-3.el8.x86_64: install failed
```
It's similar to other issues we've seen with installing filesystem: https://issues.redhat.com/browse/OAMG-4310
Is the `/proc/` read-writeable?

bocekm: The docker image is being updated from CentOS 8.2 to 8.3:
```
STEP 10: RUN dnf update -y && dnf install -y $APP_MAIN_DEPS && dnf clean all
CentOS-8 - AppStream                            952 kB/s | 6.3 MB     00:06    
CentOS-8 - Base                                 926 kB/s | 2.3 MB     00:02    
CentOS-8 - Extras                                31 kB/s | 8.6 kB     00:00    
Dependencies resolved.
================================================================================
 Package                     Arch   Version                     Repo       Size
================================================================================
Installing:
 centos-linux-release        noarch 8.3-1.2011.el8              BaseOS     22 k
     replacing  centos-release.x86_64 8.2-2.2004.0.1.el8
     replacing  centos-repos.x86_64 8.2-2.2004.0.1.el8
```
Is it because my previous image was CentOS 8.2? How can I remove it completely so it does not try to get updated to 8.3? 
Per https://hub.docker.com/_/centos the latest image available should be 8.3.

bocekm (changes requested):

CONTRIBUTING.md
```diff
@@ -4,7 +4,28 @@
...
+```
+this command will create the virtual environment and installs needed dependencies.
+Now you have to activate the environment:
+```bash
+. .venv3/bin/activate
+```
+
+and you're ready to run tests with:
+```bash
+pytest
```
bocekm, line 25: When running pytest in the .venv2.7 and .venv3, I'm getting the following for many tests:
`E   ImportError: No module named pexpect`
and 
`E   ModuleNotFoundError: No module named 'pexpect'`

Shouldn't pexpect be added to requirements/centos{7,8}.requirements.txt? Then you could remove the pexpect installation trhough yum/dnf from the Dockerfiles.

CONTRIBUTING.md
```diff
@@ -4,7 +4,28 @@
 1. All python code must be python 2.6/2.7/3.6 compatible
 1. The code should follow linting from pylint
 
-## Unit tests
+## Developing locally
+
+Dependencies: 
+https://virtualenv.pypa.io/en/latest/installation.html
```
bocekm, line 10:
```suggestion
```
I don't think mentioning this is important.

CONTRIBUTING.md
```diff
@@ -4,7 +4,28 @@
...
+
+```bash
+make install  # for centOS 8 dependencies
+CENTOS_VER=7 make install  # for centOS 7 dependencies
+CENTOS_VER=6 make install  # for centOS 6 dependencies
+```
+this command will create the virtual environment and installs needed dependencies.
```
bocekm, line 17:
```suggestion
The command bellow creates a virtual environment and installs needed dependencies.
```bash
make install  # for CentOS 8-like environment
CENTOS_VER=7 make install  # for CentOS 7-like environment
CENTOS_VER=6 make install  # for CentOS 6-like environment
```
```

File: CONTRIBUTING.md [hunk 6]
```diff
@@ -4,7 +4,28 @@
...
+Now you have to activate the environment:
+```bash
+. .venv3/bin/activate
+```
+
+and you're ready to run tests with:
+```bash
+pytest
+```
+
```
bocekm, line 27:
```suggestion
After that, you can deactivate the environment:
```bash
deactivate
```
```

File: CONTRIBUTING.md [hunk 7]
```diff
@@ -4,7 +4,28 @@
...
+```bash
+make install  # for centOS 8 dependencies
+CENTOS_VER=7 make install  # for centOS 7 dependencies
+CENTOS_VER=6 make install  # for centOS 6 dependencies
+```
+this command will create the virtual environment and installs needed dependencies.
+Now you have to activate the environment:
+```bash
+. .venv3/bin/activate
```
bocekm, line 20:
```suggestion
. .venv3/bin/activate # for CentOS 8-like environment
. .venv2.7/bin/activate # for CentOS 7-like environment
. .venv2.6/bin/activate # for CentOS 6-like environment
```

CONTRIBUTING.md
```diff
@@ -4,7 +4,28 @@
...
+```
+
+and you're ready to run tests with:
+```bash
+pytest
+```
+
+## Unit tests (inside the container)
 Our unit tests are run within containers. To first create the container images, run:
```
bocekm, line 29:
```suggestion
You can run unit tests also within containers. To first create the container images, run:
```

CONTRIBUTING.md
```diff
@@ -4,7 +4,28 @@
...
-## Unit tests
+## Developing locally
+
+Dependencies: 
+https://virtualenv.pypa.io/en/latest/installation.html
+
+```bash
+make install  # for centOS 8 dependencies
+CENTOS_VER=7 make install  # for centOS 7 dependencies
+CENTOS_VER=6 make install  # for centOS 6 dependencies
```
bocekm, line 15: I don't know about other systems but on Fedora 32 and greater this is not possible.
- python26 is available on Fedora 32 but not on 33 anymore
- even with python26 installed on F32, you can't use it with virtualenv (https://developer.fedoraproject.org/tech/languages/python/multiple-pythons.html)
```
Running virtualenv with interpreter /usr/bin/python2.6
ERROR: this script requires Python 2.7 or greater.
```

bocekm (changes requested):

sturivny:
I see that on the `centos-6` checks some tests were skipped
```
 convert2rhel/unit_tests/pkghandler_test.py::TestPkgHandler::test_call_yum_cmd_w_downgrades_continuous_fail
  /data/convert2rhel/unit_tests/__init__.py:194: UserWarning: Test <function test_call_yum_cmd_w_downgrades_continuous_fail at 0x3303578> skipped. Reason: Current test runs only on rpm based systems.
    warn("Test %r skipped. Reason: %s" % (test_item, reason))
```
centos-6 is an rpm based systems

bocekm:
@sturivny, this warning appears already in the main branch and is not caused by this PR. See https://github.com/oamg/convert2rhel/pull/142#discussion_r559968157. It can be fixed in a separate PR.

CONTRIBUTING.md
ZhukovGreen: ok, fixed

CONTRIBUTING.md
ZhukovGreen: if you would really wish to have the development env in py26, then you could get py2.6 from the pyenv.
I tested local development in centos6 and CENTOS_6 make install works as expected

CONTRIBUTING.md
ZhukovGreen: Fixed

CONTRIBUTING.md
ZhukovGreen: fixed

CONTRIBUTING.md
ZhukovGreen: Need to think about how to make it better. From one side I would like to keep app and dev deps in a separate category. But I understand that it is going to fail for the local install case. I think I know how to address this:
extending the existing requirements (you could add there `-r requirementsfilename` and other deps)

CONTRIBUTING.md
ZhukovGreen: fixed

Makefile
ZhukovGreen: I believe it used a cached version of the base image (centos:8) and then tries to update it. I think my proposal of building with --no-cache option should be enough to resolve your issue

requirements/centos7.requirements.txt
ZhukovGreen: This is interesting. I can successfully build the cento7 container (which installs dependencies from the cento7.requirements.txt). I will create a short meeting on Monday to look into this (it seems) docker issues.

ZhukovGreen (commented):
@bocekm thank you for your review. Most of the things resolved. Except for the issues with the docker builds. Let's sync on Monday to solve them

ZhukovGreen:
@sturivny please re review. I added matrices to configs. Thx

sturivny (dismissed):

CONTRIBUTING.md
```diff
@@ -4,8 +4,33 @@
...
+```bash
+# for CentOS8-like system
+make install  # for CentOS 8-like system
+# for CentOS7-like system
+# CENTOS_VER=7 make install
+# for CentOS6-like system
+#. .venv2.7/bin/activate
+# CENTOS_VER=6 make install
+#. .venv2.6/bin/activate
+```
```
bocekm, line 19:
```suggestion
The commands bellow activate a virtual environment with all the necessary dependencies installed:
```bash
# for CentOS8-like system
make install  # for CentOS 8-like system
. .venv3/bin/activate
# or for CentOS7-like system
CENTOS_VER=7 make install
. .venv2.7/bin/activate
# or for CentOS6-like system
CENTOS_VER=6 make install
. .venv2.6/bin/activate
```

requirements/centos7.requirements.txt
bocekm: Still the same issue. Or even worse now:
```
ERROR: pip's legacy dependency resolver does not consider dependency conflicts when selecting packages. This behaviour is the source of the following dependency conflicts.
sphinx-autobuild 0.7.1 requires PyYAML>=3.10, which is not installed.
pytest 4.6.11 requires six>=1.10.0, but you'll have six 1.9.0 which is incompatible.
```
IMHO this isn't much about containers but about virtualenv behaviour.

bocekm (changes requested):

CONTRIBUTING.md
bocekm: I'm getting `E   ImportError: No module named rpm` in `.venv2.7` now (not in `.venv3`).

requirements/centos7.requirements.txt
bocekm: It seems the tests work fine even with this error :) So let it be there.
BTW, I've tried older versions of pytest, down to 3.9.1, and even this one requires six>=1.10.0.

CONTRIBUTING.md
ZhukovGreen: Now this chapter looks simplier:
```md
## Developing locally

The commands below installs a python3 virtual environment
with all the necessary dependencies installed:
```bash
make install
```

and you're ready to run tests with:
```bash
make tests-locally
```
```

File: pytest.ini [hunk 11]
```diff
@@ -0,0 +1,6 @@
+[pytest]
+testpaths = "convert2rhel/unit_tests"
+addopts = --cov-fail-under=83 --tb=native
+;log_cli = false
+;log_cli_level = DEBUG
+;log_cli_format = "| %(asctime)s | %(name)s | %(levelname)s | %(filename)s | %(message)s"
```
bocekm, line 6: Is this commented out intentionally?

bocekm (approved):

pytest.ini
ZhukovGreen: yes, I'd like to give a quick path to enable live logging when running tests

PR #153, opened 2021-01-20
add: pre-commit hook

fellipeh:
This PR will add pre-commit hook on our environment.

### Reason

Today we rely on developers to verify important settings and lines of code. Which is not very reliable, since we can forget some line and end up breaking the master branch.

With this PR, pre-commit will check:
* pylint
* merge-conflict
* large files
* symlinks
* debug statements
* some private keys
* end of file

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,25 @@ repos:
       - id: debug-statements
       - id: detect-private-key
       - id: end-of-file-fixer
+  - repo: "https://github.com/psf/black"
+    rev: "20.8b1"
+    hooks:
+      - id: "black"
+        name: "Format code (black)"
+        language_version: "python3"
+  - repo: "https://github.com/asottile/blacken-docs"
+    rev: "v1.9.2"
+    hooks:
+      - id: "blacken-docs"
+        name: "Format docs (blacken-docs)"
+        language_version: "python3"
+        args: [ "-l", "64" ]
+        additional_dependencies:
+          - "black==20.8b1"
+  - repo: "https://github.com/timothycrosley/isort"
+    rev: 5.7.0
+    hooks:
+      - id: isort
+        language_version: "python3"
+        additional_dependencies: [ toml ]
+
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,18 @@
+exclude: ^(src/blib2to3/|profiling/|unit_tests/data/)
+
+repos:
+  - repo: https://github.com/pycqa/pylint
+    rev: pylint-2.6.0
+    hooks:
+      - id: pylint
+  - repo: git://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: check-added-large-files
+      - id: check-merge-conflict
+      - id: check-symlinks
+      - id: check-yaml
+        args: [--unsafe]
+      - id: debug-statements
+      - id: detect-private-key
+      - id: end-of-file-fixer
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,6 +4,47 @@
 1. All python code must be python 2.6/2.7/3.6 compatible
 1. The code should follow linting from pylint
 
+## Prerequisites
+
+### Install pre-commit
+
+We use `pre-commit` to run code formatting/linting tools on each commit. `pre-commit` operates only on **staged** files, and will only run checks that are applicable to files that have changes staged.
+
+```console
+# Ubuntu/Fedora
+# (installs pre-commit to $HOME/bin - make sure this is in your PATH)
+$ curl https://pre-commit.com/install-local.py | python -
+
+# macOS
+$ brew install pre-commit
+```
+
+If you want to format and verify your code passes on `git commit`, you can install pre-commit hooks using:
+
+```console
+$ pre-commit install
+```
+
+At anytime if you need to disable pre-commit hooks, run:
+
+```console
+$ pre-commit uninstall
+```
+
+> **Note:** You can also run these hooks manually at any time by running the following inside the repository:
+>
+> ```console
+> $ pre-commit run
+> ```
+
+> **Note:** It's usually a good idea to run the hooks against all of the files when adding new hooks (usually `pre-commit` will only run on the changed files during git hooks)
+>
+> ```console
+> $ pre-commit run --all-files
+> ```
+
+
+
 ## Unit tests
 Our unit tests are run within containers. To first create the container images, run:
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,20 @@
+[tool.black]
+line-length=79
+target-version=["py38"]
+
+[tool.isort]
+atomic=true
+line_length = 79
+lines_after_imports=2
+lines_between_types=1
+use_parentheses = true
+balanced_wrapping = true
+include_trailing_comma = true
+multi_line_output = 3
+known_third_party=[
+    "pytest",
+]
+known_first_party=[
+    "convert2rhel",
+]
+


ZhukovGreen:
@fellipeh I like pre-commit hooks very much. It gives you a very fast way to check your commits passes the hooks. Its advantage over just pylint:
1. Run just on commits diffs, which makes it almost instant
2. Proactive features, doing fixes for you (like end of file)
3. Allows to combine multiple tools and manage their dependencies in an easy way

So I think this is good benefit for our team. Now things, that should be done in this MR, from my perspective:
- [ ] Ensure that the current pre-commit config is stricter or equal to the current linting configuration
- [ ] Ensure that pre-commit hooks can be run on all our python versions (https://pre-commit.com/#hooks-language_version)
- [ ] Replace pylint with pre-commit run --all-files (so we'll achieve identical environments for local and github linting environments)

CONTRIBUTING.md
```diff
@@ -4,6 +4,47 @@
...
+```console
+# Ubuntu/Fedora
+# (installs pre-commit to $HOME/bin - make sure this is in your PATH)
+$ curl https://pre-commit.com/install-local.py | python -
+
+# macOS
+$ brew install pre-commit
+```
+
+If you want to format and verify your code passes on `git commit`, you can install pre-commit hooks using:
```
SpyTec, line 22: This makes it sound like it's not necessary to do. If you want pre-commit to work you have to install it. We're still talking about `pre-commit` at this point in the file so should be changed to an instruction on how to enable it

.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+exclude: ^(src/blib2to3/|profiling/|unit_tests/data/)
+
+repos:
+  - repo: https://github.com/pycqa/pylint
+    rev: pylint-2.6.0
+    hooks:
+      - id: pylint
```
SpyTec, line 7: This needs to be the same pylint we run in our GitHub actions since pylint returns non-0 codes on anything.

It also doesn't seem to run in the virtualenv, which means it will just never work correctly for me

.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+exclude: ^(src/blib2to3/|profiling/|unit_tests/data/)
```
SpyTec, line 1: No need to exclude something we don't have

SpyTec (commented):
Just going to preface by saying I'm never going to use pre-commit hooks in any project if I can avoid it. I usually disable them if they're enabled by default

> Replace pylint with pre-commit run --all-files (so we'll achieve identical environments for local and github linting environments)

I think it's better to only run the `pylint` for the files you've changed like it does now. When I make a change in one file I don't care about code that is unchanged in a separate untouched file. But for PRs it does make sense to run for all files at it gives a good indication of how the code at large.

The `CONTRIBUTING.md` file suggest using it for changes to the hooks though, which makes sense since it will affect the whole project.

CONTRIBUTING.md
```diff
@@ -4,6 +4,47 @@
...
+## Prerequisites
+
+### Install pre-commit
+
+We use `pre-commit` to run code formatting/linting tools on each commit. `pre-commit` operates only on **staged** files, and will only run checks that are applicable to files that have changes staged.
+
+```console
+# Ubuntu/Fedora
+# (installs pre-commit to $HOME/bin - make sure this is in your PATH)
+$ curl https://pre-commit.com/install-local.py | python -
```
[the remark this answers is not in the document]
ZhukovGreen: Agree with Eric, with pip approach after #150 will be merged we will be able to pin the version and avoid some surprises therefore

CONTRIBUTING.md
```diff
@@ -4,6 +4,47 @@
 1. All python code must be python 2.6/2.7/3.6 compatible
 1. The code should follow linting from pylint
 
+## Prerequisites
+
+### Install pre-commit
+
+We use `pre-commit` to run code formatting/linting tools on each commit. `pre-commit` operates only on **staged** files, and will only run checks that are applicable to files that have changes staged.
```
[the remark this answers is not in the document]
ZhukovGreen: Agree, that it is better to leave as a recommendation:) Surprisingly, a lot of people don't like pre-commit hooks. So they will be punished for this only when our linters jobs will fail:)

ZhukovGreen:
There is a big value behind identical testing and lining environments. If we introduce pre-commit hooks, it should be used as the only linter. Our pre-commit config now is pretty simple and seems not that important, but it will start growing, and without testing it in the CI, it has no sense to rely on contributors, that they made hooks passed successfully

sturivny:
Having pre-commit hooks is a good idea but we do need to use them in a proper way.

Those pre-commit hooks are CI part and that functionality should be implemented on the `server` side. 
Some parts of that already implemented. 

If we want to add this to our repository it should be optional for the developers. 
I agree that those checks should be implemented but they should be executed before the merge, not before the commit. 

**The best practices for the pre-commit hooks are:** 

- Each commit should have a specific file
- Each commit should have a specific description (commit message). For example, each commit message should have a number of a Jira task inside. 
- Each commit should have a file with a specific syntax. For example, a file for the TMT tests has a specific syntax. Checking that before the commit is great. It shouldn't run tests and `stuck` because a developer made a mistake in the syntax or in the structure of a file.

Pre-commit checks you suggested are a part of an IDE (Pycharm, VSCode, etc.) Developers should be responsible for it before a commit. Having those as a gating is a great idea but not as pre-commit hooks. Pre-commit hooks should be implemented as checks for some specific cases.

ZhukovGreen:
@sturivny sorry, your comment is a bit confusing to me.
> Having pre-commit hooks is a good idea but we do need to use them in a proper way.

clear, but abstract. We have to be in sync what is a proper way

> Those pre-commit hooks are CI part and that functionality should be implemented on the server side.

if you mean running pre-commit hooks as a job in the CI, then we're thinking identically.

> Some parts of that already implemented.

What do you mean? We don't have any pre-commit yet.

> If we want to add this to our repository it should be optional for the developers.
I agree that those checks should be implemented but they should be executed before the merge, not before the commit.

Even though it is beneficial to have each commit pass through the pre-commit. Practically, I was adding commits without the pre-commit check (git commit -n ...) and run pre-commit run --all-files before pushing to the remote.

Also agree do not have a strong requirement for local installation of the pre-commit. it should be in the interest of the developer to install them, so you could faster check if your code compiles before waiting for the server-side ci job report.  

> The best practices for the pre-commit hooks are:

    Each commit should have a specific file
    Each commit should have a specific description (commit message). For example, each commit message should have a number of a Jira task inside.
    Each commit should have a file with a specific syntax. For example, a file for the TMT tests has a specific syntax. Checking that before the commit is great. It shouldn't run tests and stuck because a developer made a mistake in the syntax or in the structure of a file.

I don't think this is related to the pre-commit hooks. It is possible to create custom hooks to check the syntax of configs for various tools (I do believe that TMT config doesn't have pre-commit linter:). 

> Pre-commit checks you suggested are a part of an IDE (Pycharm, VSCode, etc.) Developers should be responsible for it before a commit. Having those as a gating is a great idea but not as pre-commit hooks. Pre-commit hooks should be implemented as checks for some specific cases.

Sorry, I can't understand this part

sturivny:
> clear, but abstract. We have to be in sync what is a proper way
sure, would be great.

> if you mean running pre-commit hooks as a job in the CI, then we're thinking identically.
yes, that exactly what I meant =)

> What do you mean? We don't have any pre-commit yet.
I meant pep8 and linter here. I think checks like: pylint, merge-conflict, large files, symlinks, debug statements should be a part of CI gates.

> It should be in the interest of the developer to install them, so you could faster check if your code compiles before waiting for the server-side ci job report.
+100500

>  (I do believe that TMT config doesn't have pre-commit linter:).
Not yet ;)

> Sorry, I can't understand this part
The idea is, developers should be responsible for the code they push. They could simply make all the checks by using IDE. And as you also mentioned do not have a strong requirement for local installation of the pre-commit.

ZhukovGreen:
> @ZhukovGreen
> 
> > clear, but abstract. We have to be in sync what is a proper way
> 
> sure, would be great.
> 
> > if you mean running pre-commit hooks as a job in the CI, then we're thinking identically.
> 
> yes, that exactly what I meant =)
great!
> 
> > What do you mean? We don't have any pre-commit yet.
> 
> I meant pep8 and linter here. I think checks like: pylint, merge-conflict, large files, symlinks, debug statements should be a part of CI gates.
all these linters are available as the pre-commit hooks. That's why I was advocating to run the conde linter in thee CI in the same way how the developer might do this locally - pre-commit run --all-files. 
> 
> > It should be in the interest of the developer to install them, so you could faster check if your code compiles before waiting for the server-side ci job report.
> 
> +100500
> 
> > (I do believe that TMT config doesn't have pre-commit linter:).
> 
> Not yet ;)
:)
> 
> > Sorry, I can't understand this part
> 
> The idea is, developers should be responsible for the code they push. They could simply make all the checks by using IDE. And as you also mentioned do not have a strong requirement for local installation of the pre-commit.

Ok, now I understand. Yes, our CI will protect us from not careful devs:)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,40 @@
...
+        args: [--unsafe]
+      - id: debug-statements
+      - id: detect-private-key
+      - id: end-of-file-fixer
+  - repo: "https://github.com/psf/black"
+    rev: "20.8b1"
+    hooks:
+      - id: "black"
+        name: "Format code (black)"
+        language_version: "python3"
```
SpyTec, line 24: For the time being I would not include this in the PR, we do not know if we want to use black yet and it is quite low in our backlog. See #67 #68 [OAMG-3343](https://issues.redhat.com/browse/OAMG-3343).

ZhukovGreen: I see black and isort as something like low-hanging fruits which has an excellent positive effect on the developer productivity. I believe the whole  this MR is not driven by the priorities in the backlog, but if we're adding pre-commit, then it will be a mistake to skip black and isort.

fellipeh:
Closing for now, since we have more urgent stuff to deal.. in the future, I can go back and reopen...

PR #155, opened 2021-01-21
Provide skip mechanism for single user mode check

ZhukovGreen:
If environment variable SKIP_SINGLE_USER_MODE is set,
then the check won't be performed.


bocekm:
We'll be removing the single-user mode requirement: https://github.com/oamg/convert2rhel/pull/138#issuecomment-764903015

PR #158, opened 2021-01-27
Use pytest.mark.skipif instead of our handcrafted version

ZhukovGreen:
Use pytest.mark.skipif and remove my DIY version of this


sturivny (approved):
LGTM

PR #159, opened 2021-01-27
Now tests*,lint* targets automatically installs required dependencies

ZhukovGreen:
Adding some Makefile good practices. Now targets have dependencies. For example, running:

- make tests, will automatically build images for you (only first time)
- make tests-locally, will create the venv (also only first time)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,43 +1,53 @@
+.PHONY: \
+	force-rebuild \
+	install \
+	tests-locally \
+	lint-locally \
+	clean \
+	images \
+	tests \
+	lint \
+	lint-errors \
+	tests8 \
+
 # Project constants
 IMAGE ?= convert2rhel
 PYTHON ?= python3
 PIP ?= pip3
 VENV ?= .venv3
 
-.PHONY: all
 all: clean images tests
 
-.PHONY: install
-install:
+install: .install
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,58 +1,65 @@
...
 VENV ?= .venv3
 
-.PHONY: all
 all: clean images tests
 
-.PHONY: install
-install:
+install: .install
+
+.install:
```
SpyTec, line 23: Could you explain what this means? Why would this only run once if you call tests? Due to `touch`?

ZhukovGreen: as far as I know, the target `install` requires as a dependency a not PHONY target (which is the file or another target). Once we touch the target name (touch $@) then a file is created and next call to make tests will find that the file dependency (.images) is presented. Otherwise it will run the images target

ZhukovGreen: more interesting example is here https://www.cmcrossroads.com/article/rebuilding-when-files-checksum-changes

sturivny (approved):
LGTM

Makefile
ZhukovGreen: Resolved @SpyTec ?

fellipeh (approved):
LGTM

PR #167, opened 2021-02-01
Feature/make java opendjdk work

ZhukovGreen:
**Blocked by:** #179  **and contains its commits**

Ref https://issues.redhat.com/browse/OAMG-4245

In accordance with https://access.redhat.com/solutions/3573891
create a /var/lib/rpm-state/ dir when we see java package on host.

This MR also introduces a new module for handling such special cases

TODO:
- [x] fix failing tests due to blind caplog fixture
    - [x] use logger propagate True and fix tests [BLOCKED]

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -20,7 +20,7 @@ import os
 import sys
 
 from convert2rhel import cert
-from convert2rhel import logger
+from convert2rhel import logger, special_cases
 from convert2rhel import pkghandler
 from convert2rhel import redhatrelease
 from convert2rhel import repo
@@ -157,6 +157,11 @@ def pre_ponr_conversion():
     # remove excluded packages
     loggerinst.task("Convert: Remove excluded packages")
     pkghandler.remove_excluded_pkgs()
+
+
+    # handle special cases
+    loggerinst.task("Convert: resolve possible edge cases")
+    special_cases.check_and_resolve()
 
     if not toolopts.tool_opts.disable_submgr:
         loggerinst.task("Convert: Subscription Manager - Download packages")
--- a/convert2rhel/special_cases.py
+++ b/convert2rhel/special_cases.py
@@ -1,0 +1,29 @@
+import logging
+
+from convert2rhel.systeminfo import system_info
+from convert2rhel.utils import mkdir_p
+
+
+OPENJDK_RPM_STATE_DIR = "/var/lib/rpm-state/"
+
+logger = logging.getLogger(__name__)
+
+
+def perform_java_openjdk_workaround():
+    if system_info.is_rpm_installed(name="java-1.7.0-openjdk"):
+        logger.info(
+            "Package java-1.7.0-openjdk found. Applying workaround in"
+            "accordance with https://access.redhat.com/solutions/3573891"
+        )
+        try:
+            mkdir_p(OPENJDK_RPM_STATE_DIR)
+        except OSError:
+            logger.warning(
+                "Can't create %s directory." % OPENJDK_RPM_STATE_DIR
+            )
+        else:
+            logger.info("openjdk workaround applied successfully.")
+
+
+def check_and_resolve():
+    perform_java_openjdk_workaround()
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -14,6 +14,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from convert2rhel.utils import run_subprocess
 
 from collections import namedtuple
 try:
@@ -211,6 +212,13 @@ class SystemInfo(object):
         if modified_rpm_files_diff:
             self.logger.info(
                 "Comparison of modified rpm files from before and after the conversion:\n%s" % modified_rpm_files_diff)
+
+    @staticmethod
+    def is_rpm_installed(name):
+        _, return_code = run_subprocess(
+            "rpm -q '%s'" % name, print_cmd=False, print_output=False
+        )
+        return return_code == 0
 
 
 # Code to be executed upon module import
--- a/convert2rhel/unit_tests/main_test.py
+++ b/convert2rhel/unit_tests/main_test.py
@@ -146,6 +146,7 @@ class TestMain(unittest.TestCase):
     @unit_tests.mock(main.logging, "getLogger", GetLoggerMocked())
     @unit_tests.mock(tool_opts, "disable_submgr", False)
     @unit_tests.mock(cert.SystemCert, "_get_cert_path", unit_tests.MockFunction)
+    @mock_calls(main.special_cases, "check_and_resolve", CallOrderMocked)
     @mock_calls(pkghandler, "remove_excluded_pkgs", CallOrderMocked)
     @mock_calls(subscription, "replace_subscription_manager", CallOrderMocked)
     @mock_calls(pkghandler, "remove_repofile_pkgs", CallOrderMocked)
@@ -165,6 +166,7 @@ class TestMain(unittest.TestCase):
         intended_call_order = OrderedDict()
         intended_call_order["list_third_party_pkgs"] = 1
         intended_call_order["remove_excluded_pkgs"] = 1
+        intended_call_order["check_and_resolve"] = 1
         intended_call_order["download_rhsm_pkgs"] = 1
         intended_call_order["replace_subscription_manager"] = 1
         intended_call_order["install"] = 1
@@ -185,6 +187,7 @@ class TestMain(unittest.TestCase):
     @unit_tests.mock(main.logging, "getLogger", GetLoggerMocked())
     @unit_tests.mock(tool_opts, "disable_submgr", False)
     @unit_tests.mock(cert.SystemCert, "_get_cert_path", unit_tests.MockFunction)
+    @mock_calls(main.special_cases, "check_and_resolve", CallOrderMocked)
     @mock_calls(pkghandler, "remove_excluded_pkgs", CallOrderMocked)
     @mock_calls(subscription, "replace_subscription_manager", CallOrderMocked)
     @mock_calls(pkghandler, "remove_repofile_pkgs", CallOrderMocked)
@@ -205,6 +208,7 @@ class TestMain(unittest.TestCase):
 
         intended_call_order["list_third_party_pkgs"] = 1
         intended_call_order["remove_excluded_pkgs"] = 1
+        intended_call_order["check_and_resolve"] = 1
 
         # Do not expect this one to be called - related to RHSM
         intended_call_order["download_rhsm_pkgs"] = 0
--- a/convert2rhel/unit_tests/special_cases_test.py
+++ b/convert2rhel/unit_tests/special_cases_test.py
@@ -1,0 +1,98 @@
+import sys
+
+import pytest
+
+from convert2rhel import special_cases
+from convert2rhel.special_cases import (
+    OPENJDK_RPM_STATE_DIR,
+    check_and_resolve,
+    perform_java_openjdk_workaround,
+)
+
+
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
+
+
+@pytest.mark.parametrize(
+    (
+        "has_openjdk",
+        "can_successfully_apply_workaround",
+        "mkdir_p_should_raise",
+        "check_message_in_log",
+        "check_message_not_in_log",
+    ),
+    [
+        # All is fine case
+        (
+            True,
+            True,
+            None,
+            "openjdk workaround applied successfully.",
+            "Can't create %s" % OPENJDK_RPM_STATE_DIR,
+        ),
+        # openjdk presented, but OSError when trying to apply workaround
+        (
+            True,
+            False,
+            OSError,
+            "Can't create %s" % OPENJDK_RPM_STATE_DIR,
+            "openjdk workaround applied successfully.",
+        ),
+        # No openjdk
+        (False, False, None, None, None),
+    ],
+)
+def test_perform_java_openjdk_workaround(
+    has_openjdk,
+    can_successfully_apply_workaround,
+    mkdir_p_should_raise,
+    check_message_in_log,
+    check_message_not_in_log,
+    monkeypatch,
+    caplog,
+):
+    mkdir_p_mocked = (
+        mock.Mock(side_effect=mkdir_p_should_raise())
+        if mkdir_p_should_raise
+        else mock.Mock()
+    )
+    has_rpm_mocked = mock.Mock(return_value=has_openjdk)
+
+    monkeypatch.setattr(
+        special_cases,
+        "mkdir_p",
+        value=mkdir_p_mocked,
+    )
+    monkeypatch.setattr(
+        special_cases.system_info,
+        "is_rpm_installed",
+        value=has_rpm_mocked,
+    )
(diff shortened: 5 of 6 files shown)


convert2rhel/systeminfo.py
```diff
@@ -188,6 +189,16 @@ def modified_rpm_files_diff(self):
...
 
+    @staticmethod
+    def has_rpm(name):
+        _, return_code = run_subprocess(
+            "rpm -q '%s'" % name, print_cmd=False, print_output=False
+        )
+        if return_code == 0:
+            return True
+        else:
+            return False
```
ZhukovGreen, line 200: return True if return__code==0 else False:)

convert2rhel/unit_tests/main_test.py
```diff
@@ -123,6 +123,7 @@ def test_rollback_changes(self):
 
 
     @unit_tests.mock(main.logging, "getLogger", GetLoggerMocked())
+    @unit_tests.mock(main.special_cases, "check_and_resolve", lambda: None)
```
ZhukovGreen, line 126: Do not allow to run this

convert2rhel/unit_tests/special_cases_test.py
```diff
@@ -0,0 +1,80 @@
...
+        (
+            True,
+            False,
+            OSError,
+            "Can't create %s" % OPENJDK_RPM_STATE_DIR,
+            "openjdk workaround applied successfully.",
+        ),
+        # No openjdk
+        (False, False, None, None, None),
+    ],
```
ZhukovGreen, line 39: What other technically possible scenarious?

pytest.ini
```diff
@@ -1,6 +1,6 @@
 [pytest]
 testpaths = "convert2rhel/unit_tests"
-addopts = --cov-fail-under=83 --tb=native
+addopts = --cov-fail-under=71 --tb=native
```
ZhukovGreen, line 3: previously coverage setup was incorrect, with added coverage we have to lower our acceptance limit

ZhukovGreen (commented):
Couple of comments to my MR

ZhukovGreen:
@bocekm @SpyTec @fellipeh @sturivny could you guys glance at the general approach.

pytest.ini
SpyTec: Lets put the coverage changes in a separate PR

convert2rhel/systeminfo.py
SpyTec, line 200:
```suggestion
        return return_code == 0:
```

convert2rhel/unit_tests/special_cases_test.py
```diff
@@ -0,0 +1,80 @@
...
+        (
+            True,
+            True,
+            None,
+            "openjdk workaround applied successfully.",
+            "Can't create %s" % OPENJDK_RPM_STATE_DIR,
+        ),
+        # openjdk presented, but OSError when trying to apply workaround
+        (
+            True,
```
SpyTec, line 31: Would this styling be possible? Makes it a bit easier to read when there's a lot of parameters
```suggestion
            has_openjdk=True,
```

pytest.ini
ZhukovGreen: @SpyTec Ok, I'll cherry pick this commit into another MR

convert2rhel/unit_tests/special_cases_test.py
```diff
@@ -0,0 +1,80 @@
...
+            OSError,
+            "Can't create %s" % OPENJDK_RPM_STATE_DIR,
+            "openjdk workaround applied successfully.",
+        ),
+        # No openjdk
+        (False, False, None, None, None),
+    ],
+)
+def test_perform_java_openjdk_workaround(
+    monkeypatch,
```
SpyTec, line 42: Never heard of [`monkeypatch`](https://docs.pytest.org/en/stable/monkeypatch.html). You have it at the start and at the end of the parameters in the two functions, we should stick to having it in one position for consistency's sake

convert2rhel/main.py
```diff
@@ -163,6 +163,11 @@ def pre_ponr_conversion():
     # package analysis
     loggerinst.task("Convert: List third-party packages")
     pkghandler.list_third_party_pkgs()
+
+    # handle special cases
+    loggerinst.task("Convert: check and resolve possible special cases.")
+    special_cases.check_and_resolve()
```
SpyTec, line 169:
```suggestion
    loggerinst.task("Convert: resolve possible edge cases.")
    special_cases.check_and_resolve()
```

convert2rhel/special_cases.py
```diff
@@ -0,0 +1,29 @@
+import logging
```
SpyTec, line 1: Wondering if the file `special_cases.py` should be called something else, or just moved into pkghandler for now. The file name is rather undescriptive without context

convert2rhel/systeminfo.py
```diff
@@ -188,6 +189,16 @@ def modified_rpm_files_diff(self):
             self.logger.info(
                 "Comparison of modified rpm files from before and after the conversion:\n%s" % modified_rpm_files_diff)
 
+    @staticmethod
+    def has_rpm(name):
```
SpyTec, line 193: This feels more like it belongs in `utils` since it's more a utility than systeminfo function

convert2rhel/unit_tests/main_test.py
SpyTec: Should be using `CallOrderMocked` here to check that the function gets called, same on L163

convert2rhel/unit_tests/special_cases_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+    monkeypatch.setattr(
+        special_cases,
+        "mkdir_p",
+        value=mkdir_p_mocked,
+    )
+    monkeypatch.setattr(
+        special_cases.system_info,
+        "has_rpm",
+        has_rpm_mocked,
```
SpyTec, line 65:
```suggestion
        value=has_rpm_mocked,
```

convert2rhel/unit_tests/special_cases_test.py
```diff
@@ -0,0 +1,80 @@
...
+            "openjdk workaround applied successfully.",
+        ),
+        # No openjdk
+        (False, False, None, None, None),
+    ],
+)
+def test_perform_java_openjdk_workaround(
+    monkeypatch,
+    has_openjdk,
+    can_successfully_apply_workaround,
```
SpyTec, line 44: Unused parameter

SpyTec (commented):
Never written how pytest wants so this style of writing tests I'm not used to yet. May have missed something

convert2rhel/special_cases.py
```diff
@@ -0,0 +1,29 @@
...
+
+from convert2rhel.systeminfo import system_info
+from convert2rhel.utils import mkdir_p
+
+
+OPENJDK_RPM_STATE_DIR = "/var/lib/rpm-state/"
+
+logger = logging.getLogger(__name__)
```
SpyTec, line 9: @bocekm there was something about getting the logger in the top-level that had issues. Was it related to importing the file multiple times or something? I don't quite remember

convert2rhel/main.py
ZhukovGreen: @oamg/convert2rhel how do you think we should call this special cases python module. I am ok with the idea of @SpyTec . I'd like to have one name:
- for longer task (Convert: resolve possible {name}.)
- for python module

convert2rhel/special_cases.py
ZhukovGreen: me either. Already raised this question one comment above

convert2rhel/systeminfo.py
ZhukovGreen: my idea was to write something like 
```python
if system.has_rmp:
    blabla
```
and I found system_info, which is close. But I don't any strong arguments for both solutions.
WDYT @oamg/convert2rhel

convert2rhel/systeminfo.py
ZhukovGreen: shame on me:) thx @SpyTec

ZhukovGreen: Will be added in the next commit

convert2rhel/unit_tests/main_test.py
ZhukovGreen: ok, will do

convert2rhel/special_cases.py
bocekm: It shouldn't be a problem: https://issues.redhat.com/browse/OAMG-3365. Artem can leave it like this. When we have more time, we can change it in every file like that.

convert2rhel/unit_tests/special_cases_test.py
ZhukovGreen: I don't know any better option. Technically it is possible to adopt the dict: Dict[str, List[Any]] as a parameterization arg, but I don't think it will improve the readability too much. Another option to make a table like crazy formatting:) But what I am usually doing here is controlling the number of parameterization params. So it should exceed 4-5 params, so I can keep them in memory:)

convert2rhel/unit_tests/special_cases_test.py
ZhukovGreen: I believe you meant fixtures? I put them at the start (monkeypatch) and at the end (caplog). ok, let's agree to put fixtures as last args:
```python
def test_perform_java_openjdk_workaround(
    has_openjdk,
    mkdir_p_should_raise,
    check_message_in_log,
    check_message_not_in_log,
    monkeypatch,
    caplog,
):
```

convert2rhel/unit_tests/special_cases_test.py
ZhukovGreen: yes, thx. Fixed

convert2rhel/unit_tests/special_cases_test.py
ZhukovGreen: fixed

ZhukovGreen (commented):
some answers to @SpyTec review

ZhukovGreen:
@oamg/convert2rhel it is ready for review, but before, #179 should be merged (this branch is on top of #179)

convert2rhel/main.py
bocekm: What about `checks.py`? We'll be adding more checks in close future (e.g. for HA clustering, or for making sure repos passed through `--enablerepo` are accessible).

SpyTec: Not really a check if it makes changes to the system though

convert2rhel/systeminfo.py
bocekm: Having it here seems fine. I'd just rename it to `is_rpm_installed`. Because `if system_info.has_rpm("bogus")` does not convey the same as `system.has_rpm`.

convert2rhel/main.py
bocekm: You're right. Then I suggest `workload.py` or `apps.py`. Under that we could have:
- this java check
- checks requested by stakeholders:
  - pacemaker checks
  - HA cluster check
  - check for used kmods N/A in RHEL
- check for packages N/A in RHEL
- possibly move there the existing check for third party pkgs (`list_third_party_pkgs`)

ZhukovGreen: I think I understand your idea @bocekm. you were right in saying that the heaviest work in the development is naming:)  However, I would split checks and workarounds.

workload - seams not intuitive name to me
apps - assumes some apps registration and discovery mechanism... 

Maybe let's create a separate ticket to refactor checks and special cases, but for this MR leave it more or less as it is now. I am ok to change special_cases to edge_cases or checks.py.

convert2rhel/special_cases.py
ZhukovGreen: ok, we are handling this in the thread above. closing this one

convert2rhel/systeminfo.py
ZhukovGreen: @SpyTec I would go with @bocekm solution.

.coveragerc
```diff
@@ -1,4 +1,6 @@
 [run]
 omit=
     convert2rhel/unit_tests/*
-source=convert2rhel
\ No newline at end of file
+source=convert2rhel
+[report]
+show_missing=true
```
SpyTec, line 6: Out of scope for this PR

convert2rhel/main.py
SpyTec: What about

- `mischandler`,
- `systemhandler`, or just
- `system`

convert2rhel/systeminfo.py
SpyTec: SGTM

.coveragerc
ZhukovGreen: Sent as a separate MR #188

convert2rhel/main.py
ZhukovGreen: I'll create a Jira ticket for that. Let's resolve this comment for now.

ZhukovGreen: The ticket is OAMG-4474

ZhukovGreen:
@oamg/convert2rhel need your review guys

convert2rhel/main.py
SpyTec: I'd still change the task string to "Convert: resolve possible edge cases"

convert2rhel/special_cases.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def perform_java_openjdk_workaround():
+    if system_info.is_rpm_installed(name="java-1.7.0-openjdk"):
+        logger.info(
+            "Package java-1.7.0-openjdk found. Applying workaround "
```
SpyTec, line 15:
```suggestion
            "Package java-1.7.0-openjdk found. Applying workaround in "
```

convert2rhel/unit_tests/main_test.py
```diff
@@ -183,6 +185,7 @@ def test_pre_ponr_conversion_order_with_rhsm(self):
                 self.assertEqual(expected, actual)
 
     @unit_tests.mock(main.logging, "getLogger", GetLoggerMocked())
+    @unit_tests.mock(main.special_cases, "check_and_resolve", lambda: None)
```
SpyTec, line 188: Should test call order here as well

convert2rhel/special_cases.py
```diff
@@ -0,0 +1,29 @@
...
+        except OSError:
+            logger.warning(
+                "Can't create %s directory." % OPENJDK_RPM_STATE_DIR
+            )
+        else:
+            logger.info("openjdk workaround applied successfully.")
+
+
+def check_and_resolve():
+    perform_java_openjdk_workaround()
```
SpyTec, line 29: Was never tested. Removing `perform_java_openjdk_workaround` from here will not trigger any failing tests

SpyTec (changes requested):

convert2rhel/main.py
ZhukovGreen: @SpyTec will do

convert2rhel/special_cases.py
ZhukovGreen: why not:) fixed

convert2rhel/special_cases.py
ZhukovGreen: thx, really 29 line is not covered

convert2rhel/unit_tests/main_test.py
ZhukovGreen: fixed, thx, missed that

SpyTec (approved):

bocekm (approved):

PR #169, opened 2021-02-03
Forbid python to create pycache in containers

ZhukovGreen:
Due to running the tests inside the containers, there is no sense to cache pyc files, forever this creates additional overhead for the user to remove them before next tests.

This is fixed now


sturivny (approved):
LGTM

PR #162, opened 2021-01-28
Update EULA to the latest available

bocekm:
It's a copy paste from the pdf linked in the text file, with no ASCII characters.
Related: https://github.com/oamg/convert2rhel/issues/161

PR #107, opened 2020-10-12
feat: install subscription-manager through yum, not rpm and spec file

SpyTec:
We currently install dependencies for RHSM through spec files so we can install it through RPM. Instead we should utilize yum to resolve the dependencies for us.

See https://github.com/oamg/convert2rhel/pull/82#issuecomment-680937882 

OAMG-3843

TODO:

- [x] Merge https://github.com/oamg/convert2rhel/pull/105 first.

PR #171, opened 2021-02-04
Bump version to 0.16

bocekm:
This version includes the following changes:
- Not requiring users to download the following packages prior the conversion when using RHSM:
  - redhat-release (#105)
  - subscription-manager and its dependencies (#100)
- Removed subscription-manager dependencies from the convert2rhel spec file by installing subscription-manager through yum/dnf during the conversion instead of through rpm (#107)
- The --variant option for choosing RHEL variant was broken. Instead of fixing it we removed all the related code. The system is always converted to RHEL Server variant in case of RHEL 6 and 7. RHEL 8 has no variants. Using the --variant option now prints a warning. (#139)
- Updated Red Hat EULA to the 2019 version (#162)
- Messages of importance are printed in color in the terminal (#134)
- When printing a list of packages, we print the package vendor by default, or the packager when the vendor information is not available (#133)
- Unit tests are possible to run on non-rpm based systems (#142)
- Using pytest instead of nosetests for unit testing, added possibility to run tests locally, introduced coverate threshold (#150)

### DIFF
--- a/packaging/convert2rhel.spec
+++ b/packaging/convert2rhel.spec
@@ -9,7 +9,7 @@
 %endif
 
 Name:           convert2rhel
-Version:        0.15
+Version:        0.16
 Release:        1%{?dist}
 Summary:        Automates the conversion of RHEL derivative distributions to RHEL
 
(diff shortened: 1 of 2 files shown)


fellipeh (approved):
LGTM

sturivny (approved):

sturivny:
LGTM

PR #176, opened 2021-02-05
Provide --version argument

ZhukovGreen:
```
[root@eda63b990163 data]# convert2rhel --version
0.16
```

Also moved the optparser to the very beginning, with the idea that tool should unconditionally have working cli


convert2rhel/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.16"
```
fellipeh, line 1: Not big deal, but we need to figure out how to set this version automatically, other wise we could miss to change this file.

convert2rhel/main.py
```diff
@@ -41,6 +41,9 @@ class ConversionPhase(object):
 def main():
     """Perform all steps for the entire conversion process."""
 
+    # handle command line arguments
+    toolopts.CLI()
+
```
fellipeh, line 46: Maybe pass this line after `utils.require_root()` ?  Since we need to check root permission before running anything...

convert2rhel/__init__.py
ZhukovGreen: it is not differ from the previous way, where we needed to specify the version in two places: setup.py and rpm spec. Now instead of setup.py it is convert2rhel/__init__.py.
However, not ideal. There is https://github.com/fedora-python/pyp2rpm that I wish to try one day

convert2rhel/main.py
ZhukovGreen: My logic is that cli should be available always. But I added https://github.com/oamg/convert2rhel/pull/176/commits/3d76e0f4d1ba2d958216f364f1754f3d3a1a2451 . 
What do you think guys @oamg/convert2rhel ?

bocekm: Agreed with @ZhukovGreen.

convert2rhel/toolopts.py
```diff
@@ -65,15 +66,20 @@ def _get_argparser():
...
                                      usage=usage,
-                                     add_help_option=False)
+                                     add_help_option=False,
+                                     version=__version__)
 
     def _register_options(self):
         """Prescribe what command line options the tool accepts."""
         self._parser.add_option("-h", "--help", action="help", help="Show "
                                 " help message and exit.")
+        self._parser.add_option('--version', action='version', help="Show version of the application.")
```
SpyTec, line 82:
```suggestion
        self._parser.add_option('--version', action='version', help="Show Convert2RHEL version and exit.")
```

setup.py
```diff
@@ -25,10 +27,14 @@
 def read(fname):
     return open(os.path.join(os.path.dirname(os.path.abspath(__file__)), fname)).read()
 
+def get_version():
+    with open("convert2rhel/__init__.py") as f:
+        version = re.findall(r'^__version__ = "([^"]+)"\r?$', f.readline(), re.M)[0]
```
SpyTec, line 32: This can be `re.search`

SpyTec (changes requested):

convert2rhel/toolopts.py
bocekm, line 82:
```suggestion
        self._parser.add_option('--version', action='version', help="Show version.")
```
Using _"application"_ would be inconsistent with the rest of the messages where we use _"tool"_. Even in this PR you're adding _"WARNING: The tool needs to be run under the root user"_ :)

I'd rather see us using _"utility"_ instead of _"tool"_ but let's keep it as tool for now.

bocekm (changes requested):

convert2rhel/__init__.py
bocekm: I'm fine with moving the version out of setup.py. This solution seems to me actually even cleaner.

convert2rhel/toolopts.py
bocekm: In this case I'd maybe use the lower case "convert2rhel" to emphasize we talk about the cli utility and not the project.

bocekm: Haha, @SpyTec was faster.

convert2rhel/__init__.py
bocekm: Just add a new line here please.

convert2rhel/toolopts.py
SpyTec: Fixed

setup.py
ZhukovGreen: yeap, no preferences in either solution. But I am lazy:)

convert2rhel/toolopts.py
ZhukovGreen: ok:)

ZhukovGreen: also fixed

setup.py
ZhukovGreen: actually search returns the whole matching string, but I need a capturing group, so findall is better

bocekm (dismissed):
It's interesting that when I run `convert2rhel --vers` or any substring of "version", I get the version. That's probably a feature of the optparse library.

fellipeh (dismissed):
LGTM

setup.py
```diff
@@ -25,10 +27,29 @@
...
+        except IndexError:
+            raise ValueError(
+                (
+                    "Unable to extract the version from %s file. Make sure the "
+                    "first line has the following form: `__version__ = "
+                    '"some.version.here"`'
+                )
+                % version_source
+            )
+
```
ZhukovGreen, line 48: @SpyTec not sure how you were able to comment on the old commit, but this is how it looks like

SpyTec: Yeah I forgot to delete it and somehow clicked "add review". Thought I deleted the comment right after submitting it

SpyTec, line 48: You do not loop through the file, it will fail if the first line is not `__version__`

ZhukovGreen: well, that was intentional. However, I can read the whole file. It is anyway one time overhead

ZhukovGreen: fixed

fellipeh (approved):
LGTM

bocekm (approved):

SpyTec (approved):

PR #178, opened 2021-02-05
Use RHSM repos for downloading same version kernel

bocekm:
For the case when the all installed kernels have the same version as those available in RHEL repos, we need to download one of the RHEL kernels and force replace the already installed one. However for the download we were using only the RHEL repos passed through the --enablerepo option. That would cause the download to fail if the tool uses RHSM and not custom repos (--enablerepo).

TODO:

- [x] Cover by unit tests

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -612,9 +612,12 @@ def replace_non_rhel_installed_kernel(version):
 
     pkg = "kernel-%s" % version
 
+    # For downloading the RHEL kernel we need to use the RHEL repositories. Those are available either through the
+    # attached subscription (submgr_enabled_repos) or through custom repositories (tool_opts.enablerepo).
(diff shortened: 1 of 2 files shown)


convert2rhel/pkghandler.py
```diff
@@ -612,9 +612,12 @@ def replace_non_rhel_installed_kernel(version):
 
     pkg = "kernel-%s" % version
 
+    # For downloading the RHEL kernel we need to use the RHEL repositories. Those are available either through the
+    # attached subscription (submgr_enabled_repos) or through custom repositories (tool_opts.enablerepo).
+    repos_to_enable = system_info.submgr_enabled_repos if not tool_opts.disable_submgr else tool_opts.enablerepo
```
bocekm, line 617: FYI, this is 1:1 with https://github.com/oamg/convert2rhel/blob/main/convert2rhel/pkghandler.py#L114

ZhukovGreen (approved):
Can't find anything that blocks from merging @bocekm

ZhukovGreen (approved):

PR #179, opened 2021-02-08
Allow caplog see logger entries

ZhukovGreen:
Caplog adds its own handler to the root logger to track the records, however
our logger setup blocks the handler from being propagated to all child loggers.
This makes the caplog blind to the emitted entries.

Also, tests are rewritten in pytest because the old way was tested the logger
indirectly, switching to propagate=True led to inconsistent results across
different Python versions, which makes it hard to test.

The logger module coverage increased from 93 -> 100%

Also added helper fixtures to capture stdout and stderr

### DIFF
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -44,10 +44,11 @@ class LogLevelFile(object):
     label = "FILE"
 
 
-def initialize_logger(log_name):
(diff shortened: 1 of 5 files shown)


sturivny (approved):
LGTM

bocekm (approved):

PR #181, opened 2021-02-08
refactor: Sat6 Activation Key

fellipeh:
This PR is a small refactor when use activation key, now show to user when the user pass the activation key on cli

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -250,6 +250,7 @@ def attach_subscription():
     loggerinst = logging.getLogger(__name__)
 
     if tool_opts.activation_key:
+        loggerinst.info("Using the activation key provided through the command line...")
         return True
 
     if tool_opts.auto_attach:
--- a/convert2rhel/unit_tests/subscription_test.py
+++ b/convert2rhel/unit_tests/subscription_test.py
@@ -26,6 +26,7 @@ from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
 from convert2rhel import logger, pkghandler, subscription, utils
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
+from . import GetLoggerMocked
 
 
 class TestSubscription(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


convert2rhel/unit_tests/subscription_test.py
```diff
@@ -89,38 +90,38 @@ def __init__(self):
...
-
-        def warn(self, msg, *args):
-            self.warning_msgs.append(msg)
-
-        def warning(self, msg, *args):
-            self.warn(msg, *args)
-
-        def debug(self, msg):
-            pass
+    # class GetLoggerMocked(unit_tests.MockFunction):
```
ZhukovGreen, line 93: Nit: it is WIP, right?

ZhukovGreen (commented):
some nits:)

convert2rhel/subscription.py
```diff
@@ -250,6 +250,7 @@ def attach_subscription():
     loggerinst = logging.getLogger(__name__)
 
     if tool_opts.activation_key:
+        loggerinst.info("Using Activation Key informed by the user...")
```
bocekm, line 253: We should be direct in the messages, i.e. you instead of user. But in this case I'd go with:
```suggestion
        loggerinst.info("Using the activation key provided through the command line...")
```

ZhukovGreen:
@thrix Hi Miroslav, I have troubles with investigating the reason for failing of packit/testing-farm-epel-6-x86_64 job. Do you have some documentation on that? Or maybe some upcoming improvements.

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -89,38 +90,38 @@ def __init__(self):
...
+    #         self.info_msgs.append(msg)
+    #
+    #     def warn(self, msg, *args):
+    #         self.warning_msgs.append(msg)
+    #
+    #     def warning(self, msg, *args):
+    #         self.warn(msg, *args)
+    #
+    #     def debug(self, msg):
+    #         pass
```
SpyTec, line 124:
```suggestion

```

SpyTec (changes requested):
Two things to change, remove the comment and merge Michal's suggestion

ZhukovGreen (approved):

SpyTec (approved):

Issue #166, opened 2021-01-31
Fail when convert CentOS 7.9 to RHEL 7.9


jerry-wjl:
Hi Experts,
I installed a centos 7.9 server to test conver2rhel, the server is installed as "server with gui" and has totally 1348 pkgs installed. Below are some original info from the box.

[root@localhost ~]# cat /etc/centos-release
CentOS Linux release 7.9.2009 (Core)

[root@localhost ~]# uname  -r
3.10.0-1160.el7.x86_64

[root@localhost ~]# rpm -qa | wc -l
1348

[root@localhost ~]# yum repolist
Loaded plugins: fastestmirror, langpacks
Loading mirror speeds from cached hostfile
 * base: mirrors.huaweicloud.com
 * extras: mirrors.ustc.edu.cn
 * updates: mirrors.ustc.edu.cn
repo id                                                                    repo name                                                                    status
base/7/x86_64                                                              CentOS-7 - Base                                                              10,072
extras/7/x86_64                                                            CentOS-7 - Extras                                                               448
updates/7/x86_64                                                           CentOS-7 - Updates                                                            1,216
repolist: 11,736

As the instruction, I use rhel 7.9 installation media to make a yum.

[root@localhost yum.repos.d]# cat > local.repo << EOF
> [rhel-7.9]
> name=rhel-7.9
> enabled=1
> gpgcheck=0
> baseurl=file:/media
> EOF

[root@localhost yum.repos.d]# ls
CentOS-Base.repo  CentOS-Debuginfo.repo  CentOS-Media.repo    CentOS-Vault.repo          local.repo
CentOS-CR.repo    CentOS-fasttrack.repo  CentOS-Sources.repo  CentOS-x86_64-kernel.repo

[root@localhost yum.repos.d]# yum repolist
Loaded plugins: fastestmirror, langpacks
Loading mirror speeds from cached hostfile
 * base: mirrors.163.com
 * extras: mirrors.ustc.edu.cn
 * updates: mirrors.ustc.edu.cn
rhel-7.9                                                                                                                               | 2.8 kB  00:00:00
(1/2): rhel-7.9/group                                                                                                                  | 628 kB  00:00:00
(2/2): rhel-7.9/primary                                                                                                                | 2.1 MB  00:00:00
rhel-7.9                                                                                                                                            5230/5230
repo id                                                                    repo name                                                                    status
base/7/x86_64                                                              CentOS-7 - Base                                                              10,072
[...]

PR #182, opened 2021-02-09
Do not check deps when force replacing kernel

bocekm:
Under typical circumstances, kernel is installed using yum and with that the system-release/redhat-release pkg is installed as a dependency. However when the original OR kernel and RHEL kernel have the same version, we install it through rpm which causes the "system-release is needed by kernel" error because system-release/redhat-release is not installed at that point.

Reported in https://github.com/oamg/convert2rhel/issues/166.


fellipeh (approved):
LGTM

bocekm:
Just for the record, here's a more complex solution to the same problem: https://github.com/oamg/convert2rhel/pull/70#discussion_r493718252

PR #105, opened 2020-10-05
feat: bundle pem files instead of installing redhat-release

SpyTec:
Stops installing `redhat-release` and instead copies over pem files

Ready for review!

OAMG-3414

PR #189, opened 2021-02-17
Rework patching yum.conf

hosekadam:
Since https://github.com/oamg/convert2rhel/pull/105 we do no longer install the redhat-release package but instead we copy the RHSM certificates to the system.

Yet currently we are commenting out the distroverpkg variable from the /etc/yum.conf. For example on CentOS 7 it's distroverpkg=centos-release. This variable tells yum at which package to look at to be able to expand the $releasever variable. But in https://github.com/oamg/convert2rhel/pull/105 we've started using the --releasever yum option for every yum call.

The conf patching is not necessary anymore, except one use case - when user touches the yum.conf prior executing the conversion (`rpm -V yum`). Then during the conversion yum is replaced but not this config file, so it keeps the distroverpkg variable. If we don't patch the conf in this case, after the conversion every yum call would fail to expand the releasever var.

The above case can be safely done as one of the last steps of the conversion. We don't need to do that at the beginning as we do that now. Leaving it up to the end has the advantage of not needing to backup and restore the file in case of conversion failure.

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -217,6 +217,13 @@ def post_ponr_conversion():
     pkghandler.replace_non_red_hat_packages()
     loggerinst.task("Convert: List remaining non-Red Hat packages")
     pkghandler.list_non_red_hat_pkgs_left()
+    
+    # if user modified /etc/yum.conf, then comment out the distroverpkg variable in yum.conf
+    output, return_code = utils.run_subprocess("rpm -V yum", False, False)
+    if return_code != 0:
+        loggerinst.task("Convert: Patch yum configuration file")
+        redhatrelease.YumConf().patch()
+
     return
 
 
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -78,7 +78,6 @@ def main():
         # backup system release file before starting conversion process
         loggerinst.task("Prepare: Backup System")
         redhatrelease.system_release_file.backup()
-        redhatrelease.yum_conf.backup()
         repo.backup_yum_repos()
 
         loggerinst.task("Prepare: Clear YUM/DNF version locks")
@@ -197,10 +196,6 @@ def pre_ponr_conversion():
         loggerinst.task("Convert: Subscription Manager - Enable RHEL repositories")
         subscription.enable_repos(rhel_repoids)
 
-    # comment out the distroverpkg variable in yum.conf
-    loggerinst.task("Convert: Patch yum configuration file")
-    redhatrelease.YumConf().patch()
-
     # perform final checks before the conversion
     loggerinst.task("Convert: Final system checks before main conversion")
     checks.perform_pre_ponr_checks()
@@ -217,12 +212,9 @@ def post_ponr_conversion():
     pkghandler.replace_non_red_hat_packages()
     loggerinst.task("Convert: List remaining non-Red Hat packages")
     pkghandler.list_non_red_hat_pkgs_left()
-    
-    # if user modified /etc/yum.conf, then comment out the distroverpkg variable in yum.conf
-    output, return_code = utils.run_subprocess("rpm -V yum", False, False)
-    if return_code != 0:
-        loggerinst.task("Convert: Patch yum configuration file")
-        redhatrelease.YumConf().patch()
+
+    loggerinst.task("Convert: Patch yum configuration file")
+    redhatrelease.YumConf().patch()
 
     return
 
@@ -244,7 +236,6 @@ def rollback_changes():
     utils.changed_pkgs_control.restore_pkgs()
     repo.restore_yum_repos()
     redhatrelease.system_release_file.restore()
-    redhatrelease.yum_conf.restore()
     pkghandler.versionlock_file.restore()
     system_cert = cert.SystemCert()
     system_cert.remove()
--- a/convert2rhel/redhatrelease.py
+++ b/convert2rhel/redhatrelease.py
@@ -20,7 +20,7 @@ import logging
 import os
 import re
 
-from convert2rhel import utils
+from convert2rhel import pkgmanager, utils
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
 
@@ -69,9 +69,15 @@ class YumConf(object):
         release version ($releasever) based on the installed redhat-release
         package.
         """
-        self._comment_out_distroverpkg_tag()
-        self._write_altered_yum_conf()
-        loggerinst.info("%s patched." % self._yum_conf_path)
+        if YumConf.is_modified():
+            # When the user touches the yum.conf before executing the conversion, then during the conversion yum as a
+            # package is replaced but this config file is left unchanged and it keeps the original distroverpkg setting.
+            self._comment_out_distroverpkg_tag()
+            self._write_altered_yum_conf()
+            loggerinst.info("%s patched." % self._yum_conf_path)
+        else:
+            loggerinst.info("Skipping patching, yum configuration file not modified")
+
         return
 
     def _comment_out_distroverpkg_tag(self):
@@ -89,7 +95,16 @@ class YumConf(object):
     def get_yum_conf_filepath():
         return YumConf._yum_conf_path
 
+    @staticmethod
+    def is_modified():
+        """Return true if the YUM/DNF configuration file has been modified by the user."""
+        conf = "/etc/yum.conf" if pkgmanager.TYPE == "yum" else "/etc/dnf/dnf.conf"
+
+        output, _ = utils.run_subprocess("rpm -Vf %s" % conf, print_output=False)
+        # rpm -Vf does not return information about the queried file but about all files owned by the rpm
+        # that owns the queried file. Character '5' on position 3 means that the file was modified.
+        return True if re.search(r"^.{2}5.*? %s$" % conf, output, re.MULTILINE) else False
+
 
 # Code to be executed upon module import
 system_release_file = utils.RestorableFile(get_system_release_filepath())  # pylint: disable=C0103
-yum_conf = utils.RestorableFile(YumConf.get_yum_conf_filepath())  # pylint: disable=C0103
--- a/convert2rhel/unit_tests/redhatrelease_test.py
+++ b/convert2rhel/unit_tests/redhatrelease_test.py
@@ -16,12 +16,22 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import glob
+import sys
 import unittest
+
+import pytest
+
+
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
 
 from collections import namedtuple
 
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
-from convert2rhel import redhatrelease, utils
+from convert2rhel import pkgmanager, redhatrelease, utils
+from convert2rhel.redhatrelease import YumConf
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
 
@@ -86,6 +96,50 @@ class TestRedHatRelease(unittest.TestCase):
 
             self.assertTrue("#distroverpkg=" in yum_conf._yum_conf_content)
             self.assertEqual(yum_conf._yum_conf_content.count("#distroverpkg="), 1)
+
+
+@pytest.mark.parametrize(
+    ("pkg_type, subprocess_ret, expected_result"),
+    (
+        ("dnf", "S.5....T.  c /etc/dnf/dnf.conf", True),
+        ("dnf", ".......T.  c /etc/dnf/dnf.conf", False),
+        ("dnf", ".M.......  g /var/lib/dnf", False),
+        ("yum", "S.5....T.  c /etc/yum.conf", True),
+        ("yum", "", False),
+        ("yum", ".......T.  c /etc/yum.conf", False),
+        ("unknown", "anything", False),
+    ),
+)
+def test_yum_is_modified(monkeypatch, pkg_type, subprocess_ret, expected_result):
+    monkeypatch.setattr(pkgmanager, "TYPE", value=pkg_type)
+
+    run_subprocess = mock.Mock(return_value=(subprocess_ret, "anything"))
+    monkeypatch.setattr(utils, "run_subprocess", value=run_subprocess)
+
+    assert YumConf.is_modified() == expected_result
+
+
+@pytest.mark.parametrize("modified", (True, False))
+def test_yum_patch(monkeypatch, modified, caplog):
+    is_modified = mock.Mock(return_value=modified)
+    monkeypatch.setattr(YumConf, "is_modified", value=is_modified)
+    _comment_out_distroverpkg_tag = mock.Mock()
+    monkeypatch.setattr(
+        YumConf,
+        "_comment_out_distroverpkg_tag",
+        value=_comment_out_distroverpkg_tag,
+    )
+    _write_altered_yum_conf = mock.Mock()
+    monkeypatch.setattr(YumConf, "_write_altered_yum_conf", value=_write_altered_yum_conf)
+
+    YumConf().patch()
+
+    if modified:
+        _comment_out_distroverpkg_tag.assert_called_once()
+        assert "patched" in caplog.text
+    else:
+        _comment_out_distroverpkg_tag.assert_not_called()
+        assert "Skipping patching, yum configuration file not modified" in caplog.text
 
 
 YUM_CONF_WITHOUT_DISTROVERPKG = """[main]
--- a/tests/integration/changed-yum-conf/main.fmf
+++ b/tests/integration/changed-yum-conf/main.fmf
@@ -1,0 +1,5 @@
+summary: changed-yum-conf
+
+tag+:
+    - changed-yum-conf
+test: pytest -svv
--- a/tests/integration/changed-yum-conf/test_patch_yum_conf.py
+++ b/tests/integration/changed-yum-conf/test_patch_yum_conf.py
@@ -1,0 +1,32 @@
+import pytest
+
+
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+
+from envparse import env
+
+
+def test_yum_patch(convert2rhel, shell):
+    """Test the scenario in which the user modifies /etc/yum.conf before the conversion.
+    In that case during the conversion the config file does not get replaced with the config file from the RHEL package
+    (%config(noreplace)) and we need to make sure that we patch the config file to get rid of the distroverpkg config
+    key. Leaving the distroverpkg key there would lead to errors when calling yum after the conversion for not
+    expanding the $releasever variable properly.
+    """
+    shell("echo '#random text' >> /etc/yum.conf")
+
+    with convert2rhel(
+        ("-y --no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug").format(
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        c2r.expect("/etc/yum.conf patched.")
+    assert c2r.exitstatus == 0
+
+    assert shell("yum update -y").returncode == 0
--- a/convert2rhel/unit_tests/main_test.py
+++ b/convert2rhel/unit_tests/main_test.py
@@ -130,7 +130,6 @@ class TestMain(unittest.TestCase):
         unit_tests.CountableMockObject(),
     )
     @unit_tests.mock(repo, "restore_yum_repos", unit_tests.CountableMockObject())
-    @unit_tests.mock(redhatrelease.yum_conf, "restore", unit_tests.CountableMockObject())
     @unit_tests.mock(subscription, "rollback", unit_tests.CountableMockObject())
     @unit_tests.mock(
         pkghandler.versionlock_file,
@@ -144,7 +143,6 @@ class TestMain(unittest.TestCase):
         self.assertEqual(utils.changed_pkgs_control.restore_pkgs.called, 1)
         self.assertEqual(repo.restore_yum_repos.called, 1)
         self.assertEqual(redhatrelease.system_release_file.restore.called, 1)
-        self.assertEqual(redhatrelease.yum_conf.restore.called, 1)
         self.assertEqual(subscription.rollback.called, 1)
         self.assertEqual(pkghandler.versionlock_file.restore.called, 1)
         self.assertEqual(cert.SystemCert.remove.called, 1)
@@ -160,7 +158,6 @@ class TestMain(unittest.TestCase):
     @mock_calls(subscription, "verify_rhsm_installed", CallOrderMocked)
     @mock_calls(pkghandler, "remove_repofile_pkgs", CallOrderMocked)
     @mock_calls(cert.SystemCert, "install", CallOrderMocked)
-    @mock_calls(redhatrelease.YumConf, "patch", CallOrderMocked)
     @mock_calls(pkghandler, "list_third_party_pkgs", CallOrderMocked)
     @mock_calls(subscription, "subscribe_system", CallOrderMocked)
     @mock_calls(repo, "get_rhel_repoids", CallOrderMocked)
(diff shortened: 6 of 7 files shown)


fellipeh (dismissed):
LGTM

bocekm:
Congratulations to your first PR, Adam!

convert2rhel/main.py
```diff
@@ -198,6 +194,13 @@ def post_ponr_conversion():
     pkghandler.replace_non_red_hat_packages()
     loggerinst.task("Convert: List remaining non-Red Hat packages")
     pkghandler.list_non_red_hat_pkgs_left()
+    
+    # if user modified /etc/yum.conf, then comment out the distroverpkg variable in yum.conf
+    output, return_code = utils.run_subprocess("rpm -V yum", False, False)
```
bocekm, line 199: We need to find a better way to detect that the `/etc/yum.conf` file has been modified by the user.

What I've found out, the `rpm -V yum` is not what would give us that information.
I was thinking that maybe `rpm -Vf /etc/yum.conf` would do, but not really.

I'll try to find that out. You can also try to search for the solution. Or maybe @pirat89 would know?

bocekm (changes requested):

convert2rhel/main.py
hosekadam: I think, that this could be way: `rpm -Vf /etc/yum.cong | grep 5` or `rpm -V yum | grep 5.` According to this table in this [article](https://www.g-loaded.eu/2012/03/26/restore-original-configuration-files-from-rpm-packages/) and [documentation](http://ftp.rpm.org/max-rpm/s1-rpm-verify-output.html) the number 5 means, that the file was modified - checksum is different. So when grep finds 5, it returns the line and exit code is 0, otherwise 1. 

I hope, that the grep is standard part of CentOS and Oracle Linux.

What's your opinion?

ZhukovGreen:
@hosekadam thanks for the excellent description to the MR

convert2rhel/main.py
ZhukovGreen: Then we would need another MR to implement pipes in run_subprocess (see https://stackoverflow.com/questions/9393425/python-how-to-execute-shell-commands-with-pipe-but-without-shell-true) . @bocekm do you know  how often we would need the pipe in shell commands? Would it worse the feature?

but I think particularly in grep case, you could solve it just with python filter (https://docs.python.org/3/library/functions.html#filter)

bocekm: We haven't had the need for the pipe yet. There's always been more pythonic approach at hand.

The `/etc/yum.conf` is applicable on systems derived from RHEL 6 and 7. On RHEL 8 we need to query `/etc/dnf/dnf.conf`.

@ZhukovGreen, I'd use regexp in this case since `run_subprocess` returns a multiline string, not a list of lines. So you'd first need to split it and then it starts to look convoluted (`list(filter(lambda line: "/etc/yum.conf" in line, output.split()))`).

@hosekadam, we should strive for keeping the `main.py` without any logic and "low level" code as much as possible. In the `main.py` we should be ideally just calling functions/methods defined in other files. So, I'd call there just the `redhatrelease.YumConf().patch()` and then the conf modification check should happen within the `patch()` method.

# You can add a method similar to this one to the `YumConf` class
```python
@staticmethod
def _is_modified()
    """Return true if the YUM/DNF configuration file has been modified by the user."""
    if pkgmanager.TYPE == 'yum':
        conf = "/etc/yum.conf"
    elif pkgmanager.TYPE == 'dnf':
        conf = "/etc/dnf/dnf.conf"

    output, _ = utils.run_subprocess("rpm -Vf %s" % conf, print_output=False)
    # rpm -Vf does not return information about the queried file but about all files owned by the rpm
    # that owns the queried file
    return True if re.search(r".{2}5.*? %s$" % conf, output, re.MULTILINE) else return False
```

Please add unit test as well. Thanks.

hosekadam:
There are no tests, just edit as Michal said. I tried to push the changes, but maybe I didn't select the best way with force-push...

convert2rhel/main.py
pirat89: @bocekm no better way, I would do something like you did in the last snippet :)

convert2rhel/main.py
```diff
@@ -219,6 +219,11 @@ def post_ponr_conversion():
     pkghandler.replace_non_red_hat_packages()
     loggerinst.task("Convert: List remaining non-Red Hat packages")
     pkghandler.list_non_red_hat_pkgs_left()
+    
+    # if it's necessary, patch the yumconf (modificication the file by user check inside)
```
SpyTec, line 223: Mostly self-explanatory so we can remove this
```suggestion
```

convert2rhel/redhatrelease.py
```diff
@@ -92,6 +97,22 @@ def _write_altered_yum_conf(self):
...
+    @staticmethod
+    def _is_modified():
+        """Return true if the YUM/DNF configuration file has been modified by the user."""
+        if pkgmanager.TYPE == 'yum':
+            conf = "/etc/yum.conf"
+        elif pkgmanager.TYPE == 'dnf':
+            conf = "/etc/dnf/dnf.conf"
+        else:
+            #raise NotImplementedError() return False corresponds better to the description
+            return False
```
SpyTec, line 109: We can assume the system has either yum or dnf, otherwise we'd have issues already
```suggestion
        else:
            conf = "/etc/dnf/dnf.conf"
```

convert2rhel/redhatrelease.py
```diff
@@ -69,9 +70,13 @@ def patch(self):
...
         """
-        self._comment_out_distroverpkg_tag()
-        self._write_altered_yum_conf()
-        loggerinst.debug("%s patched." % self._yum_conf_path)
+        if YumConf._is_modified():
+            self._comment_out_distroverpkg_tag()
+            self._write_altered_yum_conf()
+            loggerinst.debug("%s patched." % self._yum_conf_path)
+        else:
+            loggerinst.debug("Patching wasn't necessary, user didn't modify the config file")
```
SpyTec, line 78:
```suggestion
            loggerinst.debug("Skipping patching, yum configuration file not modified")
```

convert2rhel/unit_tests/redhatrelease_test.py
```diff
@@ -85,6 +96,82 @@ def test_patch_yum_conf_existing_distroverpkg(self):
...
+    monkeypatch.setattr(pkgmanager, "TYPE", value=pkg_type)
+
+    run_subprocess = mock.Mock(return_value=(subprocess_ret, "anything"))
+    monkeypatch.setattr(utils, "run_subprocess", value=run_subprocess)
+
+    assert YumConf._is_modified() == expected
+
+
+@pytest.mark.parametrize("is_modified, patch_call", [(True, True), (False, False)])
+def test_patch(monkeypatch, is_modified, patch_call):
```
SpyTec, line 150: Could `is_modified` and `patch_call` be consolidated to just `is_modified`? Doesn't seem like we need two parameters here

SpyTec (changes requested):
Not had time to test it, but code looks good! Only some minor changes needed

convert2rhel/unit_tests/redhatrelease_test.py
hosekadam: Yes, of course :) Sorry for that, I feel little bit awkwardly when I see the line

SpyTec: Happens to everyone, don't worry about it!

convert2rhel/main.py
```diff
@@ -219,6 +219,10 @@ def post_ponr_conversion():
     pkghandler.replace_non_red_hat_packages()
     loggerinst.task("Convert: List remaining non-Red Hat packages")
     pkghandler.list_non_red_hat_pkgs_left()
+    
+    loggerinst.task("Convert: Patch yum configuration file")
+    redhatrelease.YumConf().patch()
```
bocekm, line 224: Since you moved the patching here, you should remove the patching from the `pre_ponr_conversion` function.

Also, as described in the PR description, the config file does not need to be backed up and restored anymore. That means:

- remove `redhatrelease.yum_conf.backup()` from `main()`
- remove `redhatrelease.yum_conf.restore()` from `rollback_changes()`
- remove `yum_conf = utils.RestorableFile(YumConf.get_yum_conf_filepath())  # pylint: disable=C0103` from `redhatrelease.py `

convert2rhel/redhatrelease.py
```diff
@@ -69,9 +70,13 @@ def patch(self):
         release version ($releasever) based on the installed redhat-release
         package.
         """
-        self._comment_out_distroverpkg_tag()
-        self._write_altered_yum_conf()
-        loggerinst.debug("%s patched." % self._yum_conf_path)
+        if YumConf._is_modified():
```
bocekm, line 73:
```suggestion
        if YumConf._is_modified():
            # When the user touches the yum.conf before executing the conversion, then during the conversion yum as a
            # package is replaced but this config file is left unchanged and it keeps the original distroverpkg setting.
```

convert2rhel/redhatrelease.py
```diff
@@ -92,6 +97,19 @@ def _write_altered_yum_conf(self):
...
+    def _is_modified():
+        """Return true if the YUM/DNF configuration file has been modified by the user."""
+        if pkgmanager.TYPE == 'yum':
+            conf = "/etc/yum.conf"
+        else:
+            conf = "/etc/dnf/dnf.conf"
+
+        output, _ = utils.run_subprocess("rpm -Vf %s" % conf, print_output=False)
+        # rpm -Vf does not return information about the queried file but about all files owned by the rpm
+        # that owns the queried file
```
bocekm, line 110:
```suggestion
        # that owns the queried file. Character '5' on position 3 means that the file was modified.
```

bocekm (changes requested):

hosekadam:
Tested on vagrant, successfully patch the file. But problems with unit tests after edits.

convert2rhel/unit_tests/redhatrelease_test.py
```diff
@@ -85,6 +96,82 @@ def test_patch_yum_conf_existing_distroverpkg(self):
...
+    monkeypatch.setattr(pkgmanager, "TYPE", value=pkg_type)
+
+    run_subprocess = mock.Mock(return_value=(subprocess_ret, "anything"))
+    monkeypatch.setattr(utils, "run_subprocess", value=run_subprocess)
+
+    assert YumConf._is_modified() == expected
+
+
+@pytest.mark.parametrize("is_modified", [True, False])
+def test_patch(monkeypatch, is_modified):
```
ZhukovGreen, line 150: These tests are a bit hard to understand. The reason in that, usually test has there steps inside:
1. Setting the environment of the test function (i.e. mocking)
2. Running the testing function in this environment
3. Check expected outputs (asserting)

I understand your motivation to reuse common steps, but I think we can hit two goals at once

convert2rhel/redhatrelease.py
```diff
@@ -92,7 +99,19 @@ def _write_altered_yum_conf(self):
     def get_yum_conf_filepath():
         return YumConf._yum_conf_path
 
+    @staticmethod
+    def _is_modified():
```
ZhukovGreen, line 103: why are you doing it protected?

convert2rhel/unit_tests/redhatrelease_test.py
ZhukovGreen: try to solve the problem with one test only, something like this:
```python
@pytest.mark.parametrize(
    (
        "pkg_manager",
        "run_subprocess_returns",
        "should_raise",
        "expected_result",
    ),
    (
        (
            "yum",
            "bla bla bla",
            None,
            "bla bla bla",
        ),
        (
            "dnf",
            "bla bla bla",
            None,
            "bla bla bla",
        ),
        (
            "unknown",
            "bla bla bla",
            Exception,
            "bla bla bla",
        ),
    ),
)
def test_yum_conf_is_modified(
    pkg_manager, run_subprocess_returns, should_raise, expected_result, caplog
):
    """TODO for Adam.

    - Use caplog for asserting logs
    - should_raise is optional, and may exists here for checking that exceptions
        were raised

    """
```
And compare with your solution

ZhukovGreen (commented):
@hosekadam some hints regarding testing

convert2rhel/redhatrelease.py
bocekm: I suppose because I suggested it under https://github.com/oamg/convert2rhel/pull/189#discussion_r583114922.
You're right there's no reason to have this method protected. Adam, you can change it to:
```suggestion
    def is_modified():
```

convert2rhel/redhatrelease.py
```diff
@@ -89,7 +95,19 @@ def _write_altered_yum_conf(self):
...
+        """Return true if the YUM/DNF configuration file has been modified by the user."""
+        if pkgmanager.TYPE == "yum":
+            conf = "/etc/yum.conf"
+        else:
+            conf = "/etc/dnf/dnf.conf"
+
+        output, _ = utils.run_subprocess("rpm -Vf %s" % conf, print_output=False)
+        # rpm -Vf does not return information about the queried file but about all files owned by the rpm
+        # that owns the queried file. Character '5' on position 3 means that the file was modified.
+        return True if re.search(r".{2}5.*? %s$" % conf, output, re.MULTILINE) else False
```
ZhukovGreen, line 109: Is there any reason for re.MULTILINE flag?
then you use ".", but in regex this means any character. If you want `.` character specifically, then you should use `\.`
Also, you're only interested in this 5 at 3rd position. So maybe it will be simpler to solve it as:
return True if output[2] == "5" else False?

But if you want regex, you also can simplify it:
return True if re.search(r"\.{2}(5).+", "..5....T.  c /etc/dnf/dnf.conf") else False

convert2rhel/unit_tests/redhatrelease_test.py
```diff
@@ -88,6 +97,50 @@ def test_patch_yum_conf_existing_distroverpkg(self):
...
+        ("dnf", "S.5....T.  c /etc/dnf/dnf.conf", True),
+        ("dnf", ".......T.  c /etc/dnf/dnf.conf", False),
+        ("dnf", ".M.......  g /var/lib/dnf", False),
+        ("yum", "S.5....T.  c /etc/yum.conf", True),
+        ("yum", "", False),
+        ("yum", ".......T.  c /etc/yum.conf", False),
+        ("unknown", "anything", False),
+    ),
+)
+def test_is_modified(monkeypatch, pkg_type, subprocess_ret, expected_result):
```
ZhukovGreen, line 112: Nicely written test!

convert2rhel/unit_tests/redhatrelease_test.py
```diff
@@ -88,6 +97,50 @@ def test_patch_yum_conf_existing_distroverpkg(self):
...
+    monkeypatch.setattr(pkgmanager, "TYPE", value=pkg_type)
+
+    run_subprocess = mock.Mock(return_value=(subprocess_ret, "anything"))
+    monkeypatch.setattr(utils, "run_subprocess", value=run_subprocess)
+
+    assert YumConf.is_modified() == expected_result
+
+
+@pytest.mark.parametrize("modified", [True, False])
```
ZhukovGreen, line 121: Nit: if you just need an iterable, without the need to modify it, then always use tuple (True, False) rather than a mutable list. That is especially important when you provide a default value in the function argument (read this https://web.archive.org/web/20200221224620/http://effbot.org/zone/default-values.htm)

ZhukovGreen (commented):
Nothing serious, generally excellent work!
When you write the test, it is better to use explicit naming, i.e.
instead of def test_patch, use def test_yum_patch etc. It is because when you run pytest, you see only the names of the the test functions. You also have a higher chance to end up with the same test name if you give less explicit naming

ZhukovGreen:
@hosekadam I see your pre-commit fails. Install it locally (see https://github.com/oamg/convert2rhel/blob/main/CONTRIBUTING.md#work-with-pre-commit-hooks)

convert2rhel/redhatrelease.py
hosekadam: I think, that re.MULTILINE is necessary because the output from `rpm -Vf` can have more lines. The dots there are ok, there can be any characters. 

This part gave me @bocekm, so maybe he'll correct me if I'm saying something wrong. It was after a research because I had some idea, but it showed up as a little bit more complicated as we expected :)

ZhukovGreen: I think -Vf gives the output only for the given file. but I might be wrong. I tried to modify those files, but always get only one line. @bocekm should know more about rpm utility.

convert2rhel/redhatrelease.py
```diff
@@ -69,9 +69,15 @@ def patch(self):
...
         """
-        self._comment_out_distroverpkg_tag()
-        self._write_altered_yum_conf()
-        loggerinst.info("%s patched." % self._yum_conf_path)
+        if YumConf.is_modified():
+            # When the user touches the yum.conf before executing the conversion, then during the conversion yum as a
+            # package is replaced but this config file is left unchanged and it keeps the original distroverpkg setting.
+            self._comment_out_distroverpkg_tag()
+            self._write_altered_yum_conf()
+            loggerinst.info("%s patched." % self._yum_conf_path)
```
hosekadam, line 77: Changed to `loggerinst.info` to merge conflicts

convert2rhel/redhatrelease.py
bocekm: Against all common sense, `rpm -Vf <file>` outputs all changed files of the package to which `<file>` belongs. So the multiline is correct.

bocekm:
```
# touch /etc/yum.conf
# touch /etc/yum/version-groups.conf
# rpm -Va yum
.......T.  c /etc/yum.conf
.......T.  c /etc/yum/version-groups.conf
# rpm -Vf /etc/yum.conf
.......T.  c /etc/yum.conf
.......T.  c /etc/yum/version-groups.conf
```

ZhukovGreen: ok, then do not forget a `^` in the regex, but the filename should be there

bocekm:
```suggestion
        return True if re.search(r"^.{2}5.*? %s$" % conf, output, re.MULTILINE) else False
```

bocekm:
Please create an integration test. The test should:
1. Take CentOS Linux 7. I think this one system is enough to verify the change works as expected (i.e. no need for running it on CentOS Linux 8 or Oracle Linux 7 and 8).
2. `touch /etc/yum.conf`
3. convert the system using RHSM (with RHSM we can be sure there's $releasever in the repo baseurls)
1. run yum update and check the exit code. If the yum.conf is not patched (and the original distroverpkg key stays there), the yum call would fail on expanding the $releasever.

bocekm (changes requested):

plans/integration/changed-yum-conf/vm_centos7.fmf
```diff
@@ -0,0 +1,14 @@
...
+
+/good:
+    discover+:
+        filter+:
+            - 'tag: good_test'
+    prepare:
+        how: ansible
+        playbook:
+            - tests/ansible_collections/main.yml
+            - tests/integration/changed-yum-conf/ansible/change_yum_conf.yml
```
ZhukovGreen, line 14:
```suggestion
    prepare+:
        - name: change yum config
          how: ansible
          playbook:
            - tests/integration/changed-yum-conf/ansible/change_yum_conf.yml
```

ZhukovGreen: You need to extend the metadata, rather than overwrite it

tests/integration/changed-yum-conf/ansible/change_yum_conf.yml
```diff
@@ -0,0 +1,5 @@
+- hosts: all
+  tasks:
+
+    - name: Edit /etc/yum.conf
+      shell: echo '#random text' >> /etc/yum.conf
```
ZhukovGreen, line 5: Nit: I think it will be more clear, and simple if you do this inside the pytest with help of shell fixture, like for example here:
https://github.com/oamg/convert2rhel/blob/main/tests/integration/remove-excluded-pkgs/test_remove_excluded_pkgs.py#L15-L20.

tests/integration/changed-yum-conf/main.fmf
```diff
@@ -0,0 +1,8 @@
+summary: changed-yum-conf
+
+/good:
+  tag+:
+    - good_test
```
ZhukovGreen, line 5: this is not needed, right?

tests/integration/changed-yum-conf/test_patch_yum_conf.py
```diff
@@ -0,0 +1,26 @@
...
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+
+from envparse import env
+
+
+@pytest.mark.good_tests
+def test_yum_patch(convert2rhel, shell):
+    with convert2rhel(
```
ZhukovGreen, line 14: Add here:
shell("echo '#random text' >> /etc/yum.conf") and you're done. No need additional ansible playbook

plans/integration/changed-yum-conf/vm_centos7.fmf
bocekm: What about https://github.com/oamg/convert2rhel/pull/189#discussion_r668835784? I.e. merge main.fmf and vm_centos7.fmf into one.

tests/integration/changed-yum-conf/main.fmf
bocekm: I think so as well: https://github.com/oamg/convert2rhel/pull/189#discussion_r668837673

tests/integration/changed-yum-conf/test_patch_yum_conf.py
bocekm: And actually you can run just shell("touch /etc/yum.conf"). Touching it is enough. A touched file (modified date gets updated) is considered changed by the rpm tooling.

bocekm: I'm taking the _touch_ idea back. We are [checking the MD5 checksum](https://github.com/oamg/convert2rhel/pull/189/files#diff-f06a581f26424ede8990e290a85424d0d56edf4a5c5a18c97c293ee9dd72364bR106), and touching does not change that.

tests/integration/changed-yum-conf/main.fmf
```diff
@@ -0,0 +1,8 @@
...
+    - changed-yum-conf
+  test: |
+    pytest -svv --tb=no -m good_tests
```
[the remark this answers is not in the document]
ZhukovGreen: we have set default duration at 30 mins in the root fmf, so no need to have it here

plans/integration/changed-yum-conf/vm_centos7.fmf
ZhukovGreen: yes, sure, we're no longer creating separate fmf for different vms. Agree.
@hosekadam example of merging https://github.com/oamg/convert2rhel/blob/main/plans/integration/releasever-as-mapping/main.fmf

tests/integration/changed-yum-conf/test_patch_yum_conf.py
bocekm: It brings me to the question - should we be checking the MD5 checksum? I have a hunch that whatever change happens to the file, yum won't replace it during reinstallation. If that's the case, we'd need to be checking only whether the file appears in the `rpm -V` output. I need to verify that.

bocekm: I checked the yum spec on RHEL 7 and the /etc/yum.conf is `%config(noreplace)`.
My testing shows, that when I change the mode (chmod) or owner or timestamp, the config gets replaced.
So, no change needed, you're doing it right ;)

tests/integration/changed-yum-conf/test_patch_yum_conf.py
```diff
@@ -0,0 +1,26 @@
...
+
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+
+from envparse import env
+
+
+def test_yum_patch(convert2rhel, shell):
```
bocekm, line 12:
```suggestion
def test_yum_patch(convert2rhel, shell):
    """Test the scenario in which the user modifies /etc/yum.conf before the conversion.

    In that case during the conversion the config file does not get replaced with the config file from the RHEL package
    (%config(noreplace)) and we need to make sure that we patch the config file to get rid of the distroverpkg config
     key. Leaving the distroverpkg key there would lead to errors when calling yum after the conversion for not
    expanding the $releasever variable properly.
    """
```

bocekm (changes requested):
It looks ready to merge. Just one last thing - describe what the integration test does as a best practice - to not require the reader to read through the code to figure out what the test is about.

bocekm (approved):

ZhukovGreen (approved):

SpyTec (approved):

PR #191, opened 2021-02-19
Feature/use logger instead of custom logger

ZhukovGreen:
This merge request introduces a set of fixes and improvements to the app logger facility.

Fixes:

1. Testing session logger was different from the logger during the app run (not an instance of CustomLogger etc.). This makes invisible CustomLogger functionality (i.e.  not expected sys.exit during logger.critical call during the test run). In this MR this issue is solved and the same logger is used for the app and testing session

Improvements:

1. Now there is no difference whether you get logger at the module level or inside the functions, due to the singleton pattern used in getting the logger and  auto propogating the handlers from the convert2rhel logger.
2. The setup of the logger performed before each testing unit and logs are available in the tmpdir (fixture) with automatic teardown at the end of the testing session
3. Provide write access to the app root dir within the docker

TODO
- [x] fix existing places where the logger was called from inside each function

PR #193, opened 2021-02-19
Feature/inhibit if kmods are not supported

ZhukovGreen:
~~**blocked by:** #191 (rebased on top)~~

This MR introduces a check of host kernel modules against available kernel modules on rhel

TODO:
- [x] verify 
- [x] Move 3rd party kmods check right after system facts check
- [x] Use repoqery -l {pkgs} instead of loop for each package

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -101,8 +101,10 @@ def ensure_compatibility_of_kmods():
         #  doesn't work. We have `%s` as output instead. Make it work
         logger.critical(
             (
-                "The following kernel module(s) are not "
-                "supported in RHEL:\n{kmods}"
+                "The following kernel modules are not "
+                "supported in RHEL:\n{kmods}\n"
+                "Uninstall or disable them and run convert2rhel "
+                "again to continue with the conversion."
             ).format(kmods=not_supported_kmods)
         )
     else:
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -16,12 +16,28 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 
+import itertools
 import logging
 import os
+import re
+import subprocess
 
-from convert2rhel.pkghandler import call_yum_cmd
+from convert2rhel.systeminfo import system_info
+from convert2rhel.utils import run_subprocess
+
 
 logger = logging.getLogger(__name__)
+
+KERNEL_REPO_RE = re.compile("^.+:(?P<version>.+).el.+$")
+KERNEL_REPO_VER_SPLIT_RE = re.compile("\W+")
+LINK_KMODS_RH_POLICY = "https://access.redhat.com/third-party-software-support"
+
+
+def perform_pre_checks():
+    """Early checks after system facts should be added here."""
+    check_uefi()
+    check_tainted_kmods()
+
 
 def check_uefi():
     """Inhibit the conversion when UEFI detected."""
@@ -37,10 +53,236 @@ def check_uefi():
     logger.debug("Converting BIOS system")
 
 
-def perform_pre_checks():
-    """Perform all 'registered' checks
+def check_tainted_kmods():
+    """Stop the conversion when a loaded tainted kernel module is detected.
 
-    Every check function in this file should be performed from here.
-    This is the entry-point for this module.
+    Tainted kmods ends with (...) in /proc/modules, for example:
+        multipath 20480 0 - Live 0x0000000000000000
+        linear 20480 0 - Live 0x0000000000000000
+        system76_io 16384 0 - Live 0x0000000000000000 (OE)  <<<<<< Tainted
+        system76_acpi 16384 0 - Live 0x0000000000000000 (OE) <<<<< Tainted
     """
-    check_uefi()
+    unsigned_modules, _ = run_subprocess("grep '(' /proc/modules")
+    module_names = "\n  ".join(
+        [mod.split(" ")[0] for mod in unsigned_modules.splitlines()]
+    )
+    if unsigned_modules:
+        logger.critical(
+            "Tainted kernel module(s) detected. "
+            "Third-party components are not supported per our "
+            "software support policy\n%s\n\n"
+            "Uninstall or disable the following module(s) and run convert2rhel "
+            "again to continue with the conversion:\n  %s",
+            LINK_KMODS_RH_POLICY,
+            module_names,
+        )
+
+
+def perform_pre_ponr_checks():
+    """Late checks before ponr should be added here."""
+    ensure_compatibility_of_kmods()
+
+
+def ensure_compatibility_of_kmods():
+    """Ensure if the host kernel modules are compatible with RHEL."""
+    host_kmods = get_installed_kmods()
+    rhel_supported_kmods = get_rhel_supported_kmods()
+    if is_unsupported_kmod_installed(host_kmods, rhel_supported_kmods):
+        kernel_version = run_subprocess("uname -r")[0].rstrip("\n")
+        not_supported_kmods = "\n".join(
+            map(
+                lambda kmod: "/lib/modules/{kver}/{kmod}".format(
+                    kver=kernel_version, kmod=kmod
+                ),
+                host_kmods - rhel_supported_kmods,
+            )
+        )
+        # TODO logger.critical("message %s, %s", "what should be under s")
+        #  doesn't work. We have `%s` as output instead. Make it work
+        logger.critical(
+            (
+                "The following kernel module(s) are not "
+                "supported in RHEL:\n{kmods}"
+            ).format(kmods=not_supported_kmods)
+        )
+    else:
+        logger.debug("Kernel modules are compatible.")
+
+
+def get_installed_kmods():
+    """Get a set of kernel modules.
+
+    Each module we cut part of the path until the kernel release
+    (i.e. /lib/modules/5.8.0-7642-generic/kernel/lib/a.ko.xz ->
+    kernel/lib/a.ko.xz) in order to be able to compare with RHEL
+    kernel modules in case of different kernel release
+    """
+    try:
+        kernel_version, exit_code = run_subprocess("uname -r")
+        assert exit_code == 0
+        kmod_str, exit_code = run_subprocess(
+            'find /lib/modules/{kver} -name "*.ko*"'.format(
+                kver=kernel_version.rstrip("\n")
+            )
+        )
+        assert exit_code == 0
+        assert kmod_str
+    except (subprocess.CalledProcessError, AssertionError):
+        logger.critical("Can't get list of kernel modules.")
+    else:
+        return set(
+            _get_kmod_comparison_key(path)
+            for path in kmod_str.rstrip("\n").split()
+        )
+
+
+def _get_kmod_comparison_key(path):
+    """Create a comparison key from the kernel module abs path.
+
+    Converts /lib/modules/5.8.0-7642-generic/kernel/lib/a.ko.xz ->
+    kernel/lib/a.ko.xz
+
+    Why:
+        The standard kernel modules are located under
+        /lib/modules/{some kernel release}/.
+        If we want to make sure that the kernel package is present
+        on RHEL, we need to compare the full path, but because kernel release
+        might be different, we compare the relative paths after kernel release.
+    """
+    return "/".join(path.split("/")[4:])
+
+
+def get_rhel_supported_kmods():
+    """Return set of target RHEL supported kernel modules."""
+    repoquery_repoids_args = " ".join(
+        "--repoid " + repoid for repoid in system_info.get_enabled_rhel_repos()
+    )
+    # Without the release package installed, dnf can't determine the modularity
+    #   platform ID.
+    setopt_arg = (
+        "--setopt=module_platform_id=platform:el8"
+        if system_info.version.major == 8
+        else ""
+    )
+    # get output of a command to get all packages which are the source
+    # of kmods
+    kmod_pkgs_str, _ = run_subprocess(
+        (
+            "repoquery "
+            "--releasever={releasever} "
+            "{setopt_arg} "
+            "{repoids_args} "
+            "-f /lib/modules/*.ko*"
+        ).format(
+            releasever=system_info.releasever,
+            setopt_arg=setopt_arg,
+            repoids_args=repoquery_repoids_args,
+        ),
+        print_output=False,
+    )
+    # from these packages we select only the latest one
+    kmod_pkgs = get_most_recent_unique_kernel_pkgs(
+        kmod_pkgs_str.rstrip("\n").split()
+    )
+    # querying obtained packages for files they produces
+    rhel_kmods_str, _ = run_subprocess(
+        (
+            "repoquery "
+            "--releasever={releasever} "
+            "{setopt_arg} "
+            "{repoids_args} "
+            "-l {pkgs}"
+        ).format(
+            releasever=system_info.releasever,
+            setopt_arg=setopt_arg,
+            repoids_args=repoquery_repoids_args,
+            pkgs=" ".join(kmod_pkgs),
+        ),
+        print_output=False,
+    )
+    return get_rhel_kmods_keys(rhel_kmods_str)
+
+
+def get_most_recent_unique_kernel_pkgs(pkgs):
+    """Return the most recent versions of all kernel packages.
+
+    When we scan kernel modules provided by kernel packages
+    it is expensive to check each kernel pkg. Since each new
+    kernel pkg do not deprecate kernel modules we only select
+    the most recent ones.
+
+    All RHEL kmods packages starts with kernel* or kmod*
+
+    For example, we have the following packages list:
+        kernel-core-0:4.18.0-240.10.1.el8_3.x86_64
+        kernel-core-0:4.19.0-240.10.1.el8_3.x86_64
+        kmod-debug-core-0:4.18.0-240.10.1.el8_3.x86_64
+        kmod-debug-core-0:4.18.0-245.10.1.el8_3.x86_64
+    ==> (output of this function will be)
+        kernel-core-0:4.19.0-240.10.1.el8_3.x86_64
+        kmod-debug-core-0:4.18.0-245.10.1.el8_3.x86_64
+
+    _repos_version_key extract the version of a package
+        into the tuple, i.e.
+        kernel-core-0:4.18.0-240.10.1.el8_3.x86_64 ==>
+        (4, 15, 0, 240, 10, 1)
+
+
+    :type pkgs: Iterable[str]
+    :type pkgs_groups:
+        Iterator[
+            Tuple[
+                package_name_without_version,
+                Iterator[package_name, ...],
+                ...,
+            ]
+        ]
+    """
+
+    pkgs_groups = itertools.groupby(
+        pkgs, lambda pkg_name: pkg_name.split(":")[0]
+    )
+    return (
+        max(distinct_kernel_pkgs[1], key=_repos_version_key)
+        for distinct_kernel_pkgs in pkgs_groups
+        if distinct_kernel_pkgs[0].startswith(("kernel", "kmod"))
+    )
+
+
+def _repos_version_key(pkg_name):
+    try:
+        rpm_version = KERNEL_REPO_RE.search(pkg_name).group("version")
+    except AttributeError:
+        logger.critical(
+            "Unexpected package:\n%s\n is a source of kernel modules.",
+            pkg_name,
+        )
+    else:
+        return tuple(
+            map(
+                _convert_to_int_or_zero,
+                KERNEL_REPO_VER_SPLIT_RE.split(rpm_version),
+            )
+        )
+
+
+def _convert_to_int_or_zero(s):
+    try:
+        return int(s)
+    except ValueError:
+        return 0
+
+
+def get_rhel_kmods_keys(rhel_kmods_str):
+    return set(
+        _get_kmod_comparison_key(kmod_path)
+        for kmod_path in filter(
+            lambda path: path.endswith(("ko.xz", "ko")),
+            rhel_kmods_str.rstrip("\n").split(),
+        )
+    )
+
+
+def is_unsupported_kmod_installed(host_kmods, rhel_supported_kmods):
+    """Return True if any of the installed kernel modules is not available in RHEL repositories."""
+    return not host_kmods.issubset(rhel_supported_kmods)
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -16,30 +16,402 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import os
+import subprocess
+import sys
+
+import pytest
+
+from convert2rhel import checks, unit_tests
+from convert2rhel.checks import (
+    check_tainted_kmods,
+    ensure_compatibility_of_kmods,
+    get_installed_kmods,
+    get_most_recent_unique_kernel_pkgs,
+    get_rhel_supported_kmods,
+    perform_pre_checks,
+    perform_pre_ponr_checks,
+)
+from convert2rhel.unit_tests import GetLoggerMocked
+from convert2rhel.utils import run_subprocess
+
+
 try:
     import unittest2 as unittest  # Python 2.6 support
 except ImportError:
     import unittest
 
-from convert2rhel import checks, unit_tests
-from convert2rhel.unit_tests import GetLoggerMocked
+
+try:
+    import unittest2 as unittest  # Python 2.6 support
+except ImportError:
+    import unittest
 
 
-class TestChecks(unittest.TestCase):
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
 
+
+HOST_MODULES_STUB_GOOD = (
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/a.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/b.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/c.ko.xz\n"
+)
+HOST_MODULES_STUB_BAD = (
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/d.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/e.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/f.ko.xz\n"
+)
+REPOQUERY_F_STUB_GOOD = (
+    "kernel-core-0:4.18.0-240.10.1.el8_3.x86_64\n"
+    "kernel-core-0:4.18.0-240.15.1.el8_3.x86_64\n"
+    "kernel-debug-core-0:4.18.0-240.10.1.el8_3.x86_64\n"
+    "kernel-debug-core-0:4.18.0-240.15.1.el8_3.x86_64\n"
+)
+REPOQUERY_F_STUB_BAD = (
+    "kernel-idontexpectyou-core-sdsdsd.ell8_3.x86_64\n"
+    "kernel-core-0:4.18.0-240.15.1.el8_3.x86_64\n"
+    "kernel-debug-core-0:4.18.0-240.10.1.el8_3.x86_64\n"
+    "kernel-debug-core-0:4.18.0-240.15.1.el8_3.x86_64\n"
+)
+REPOQUERY_L_STUB_GOOD = (
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/a.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/a.ko\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/b.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/c.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/c.ko\n"
+)
+REPOQUERY_L_STUB_BAD = (
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/d.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/d.ko\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/e.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/f.ko.xz\n"
+    "/lib/modules/5.8.0-7642-generic/kernel/lib/f.ko\n"
+)
+
+
+def _run_subprocess_side_effect(*stubs):
+    def factory(*args, **kwargs):
+        for kws, result in stubs:
+            if all(kw in args[0] for kw in kws):
+                return result
+        else:
+            return run_subprocess(*args, **kwargs)
+
+    return factory
+
+
+def test_perform_pre_checks(monkeypatch):
+    check_thirdparty_kmods_mock = mock.Mock()
+    check_uefi_mock = mock.Mock()
+    monkeypatch.setattr(
+        checks,
+        "check_uefi",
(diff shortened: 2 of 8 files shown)


SpyTec:
Is this ready for review @ZhukovGreen?

ZhukovGreen:
@SpyTec Not yet Eric

shaded-enmity:
The code should also check currently loaded modules, something like:
```bash
lsmod | cut -d' ' -f1 | tail -n +2 | sort -R | xargs modinfo | grep '^filename' | tr -s ' ' | cut -d' ' -f2
```

The main rationale here is that if I'm using non-standard kmods, more than likely they live in non-standard locations.

ZhukovGreen:
@shaded-enmity but even custom kernel modules should live in /lib/modules/{kernel-version}. This is considered in the implementation

shaded-enmity:
@ZhukovGreen It is a convention that all kmods in EL live in these folders, yes. But there's nothing preventing me from placing my kmods in, say, `/opt/myCompany/myKmods`, kernel doesn't really care from what directory it loads these mods. A lot of 3rd parties do exactly this -- place everything under `/opt/myCompany`.

If we want to separate the implementation into two parts I'm fine with that:
- Kernel module files installed in standard location (this PR)
- Kernel modules actively loaded on the system from *any* location (my suggestion)

This is all a matter of setting the right expectations for QE and support folks.

ZhukovGreen:
@shaded-enmity yep, I would rather split it into two parts. Based on sync with the CEE guys, they are ok if we cover standard location first

convert2rhel/unit_tests/conftest.py
```diff
@@ -68,3 +70,29 @@ def pkg_root(is_py2):
 @pytest.fixture(autouse=True)
 def setup_logger(tmpdir):
     initialize_logger(log_name="convert2rhel", log_dir=tmpdir)
+
+
+@pytest.fixture()
+def replace_data_dir_for_centos8(monkeypatch, pkg_root):
+    """Use our """
```
SpyTec, line 77: Missing doc

convert2rhel/checks.py
```diff
@@ -16,13 +16,190 @@
...
+    # TODO what letter is after "(" should be inhibited
+    #   after (O|E|P|U|D) (exceptional case)
+    unsigned_modules, _ = run_subprocess("grep '(' /proc/modules")
+    if unsigned_modules:
+        logger.critical(
+            "An unsigned kernel module(s) has been detected. "
+            "These are not supported as defined in the third "
+            "party software support policy\n%s.\n\n "
+            "Uninstalling or disabling these modules may allow "
+            "the conversion to proceed.",
```
SpyTec, line 184:
```suggestion
            "Unsigned kernel module(s) has been detected. "
            "Third-party components are not supported per our "
            "software support policy\n%s.\n\n "
            "Uninstall or disable the module(s) and run convert2rhel "
            "again to continue with the conversion.",
```

convert2rhel/checks.py
```diff
@@ -16,13 +16,190 @@
...
+
+
+def ensure_compatibility_of_kmods(_refresh_system_info=False):
+    """Ensure if the host kernel modules are compatible with RHEL."""
+    check_thirdparty_kmods()
+    host_kmods = get_host_kmods()
+    rhel_supported_kmods = get_rhel_supported_kmods(_refresh_system_info)
+    if not host_kmods.issubset(rhel_supported_kmods):
+        logger.critical(
+            "The following kernel modules are not supported "
```
SpyTec, line 196:
```suggestion
            "The following kernel module(s) are not supported "
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -16,30 +16,401 @@
...
+        check_thirdparty_kmods()
+
+
+def test_mo_kmod_check(caplog):
+    tool_opts.no_kmods_check = True
+    perform_final_checks_before_ponr()
+    assert (
+        "No check of kernel modules support will be performed"
+        in caplog.records[-1].message
+    )
```
SpyTec, line 366: Out of scope for next release
```suggestion
```

Also has a typo in the method name

convert2rhel/checks.py
```diff
@@ -16,13 +16,190 @@
...
+        pkgs, lambda pkg_name: pkg_name.split(":")[0]
+    )
+    return (
+        max(distinct_kernel_pkgs[1], key=_repos_version_key)
+        for distinct_kernel_pkgs in pkgs_groups
+        if distinct_kernel_pkgs[0].startswith(("kernel", "kmod"))
+    )
+
+
+def get_rhel_supported_kmods(_refresh_system_info=False):
```
SpyTec, line 127: Could use more inline comments to explain what some pieces of code does

convert2rhel/unit_tests/checks_test.py
```diff
@@ -16,30 +16,401 @@
...
+        "ensure_compatibility_of_kmods",
+        value=ensure_compatibility_of_kmods_mock,
+    )
+    monkeypatch.setattr(
+        checks,
+        "check_uefi",
+        value=check_uefi_mock,
+    )
+    perform_final_checks_before_ponr()
+    perform_pre_checks()
```
SpyTec, line 109: Tests shouldn't be testing two methods at the same time, should split into two tests

convert2rhel/checks.py
```diff
@@ -16,13 +16,190 @@
...
+        )
+
+
+def get_most_recent_unique_kernel_pkgs(pkgs):
+    """Return the most recent versions of all kernel packages.
+
+    When we do scanning of kernel modules provided by kernel packages,
+    it is expensive to check each kernel pkg. Considering the fact,
+    that each new kernel pkg do not deprecate kernel modules we select only
+    the most recent ones.
```
SpyTec, line 112:
```suggestion
    When we scan kernel modules provided by kernel packages
    it is expensive to check each kernel pkg. Since each new
    kernel pkg do not deprecate kernel modules we only select
    the most recent ones.
```

convert2rhel/toolopts.py
```diff
@@ -98,6 +99,10 @@ def _register_options(self):
                                                                          " to show you what rpm files have been affected by the conversion."
                                                                          % (PRE_RPM_VA_LOG_FILENAME,
                                                                             POST_RPM_VA_LOG_FILENAME))
+        self._parser.add_option("--no-kmods-check", action="store_true", help="WARNING: Experimental feature! "
+                                                                              "Skip check of unsupported kernel modules. "
+                                                                              "By default the convertion will be stopped if such "
+                                                                              "modules will be identified.")
```
SpyTec, line 105: Out of scope
```suggestion
```

convert2rhel/checks.py
```diff
@@ -38,9 +215,20 @@ def check_uefi():
...
+        ensure_compatibility_of_kmods()
+    else:
+        logger.warning(
+            (
+                "No check of kernel modules support will be performed. "
+                "Asked explicitly via a command line argument. "
+                "If your system has custom kernel modules you may need "
+                "to reinstall them after the conversion."
+            )
+        )
```
SpyTec, line 234: Out of scope
```suggestion
    ensure_compatibility_of_kmods()
```

SpyTec (changes requested):
Overall implementation seems sound by just looking through the code, albeit a bit complex at parts. There's instances where a new variable could describe better what the returned list is as opposed to doing it inline. 

Allowing the user to overwrite the kmods inhibitor is out of scope for the next release, should be removed before merging. That's the only concern with the implementation - the rest of my comments is with the tests, some notes for docs and logging

I have not tested this PR manually yet

convert2rhel/checks.py
```diff
@@ -16,13 +16,190 @@
...
+
+
+def _repos_version_key(pkg_name):
+    try:
+        rpm_version = KERNEL_REPO_RE.search(pkg_name).group("version")
+    except AttributeError:
+        raise NotImplementedError(
+            "Unexpected package:\n%s\n is a source of kernel modules."
+            % pkg_name
+        )
```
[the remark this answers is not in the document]
ZhukovGreen: Unexpected package:
kmod-ssss:wrong-version.no-el8.no-cpu-arch
is a source of kernel modules.

convert2rhel/checks.py
ZhukovGreen: thx

convert2rhel/checks.py
ZhukovGreen: yeap

convert2rhel/checks.py
```diff
@@ -16,13 +16,185 @@
...
+                    rhel_kmods_bytes.decode().rstrip("\n").split(),
+                )
+            )
+        )
+    return rhel_kmods
+
+
+def check_thirdparty_kmods():
+    # TODO what letter is after "(" should be inhibited
+    #   after (O|E|P|U|D) (exceptional case)
```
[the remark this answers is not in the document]
ZhukovGreen: Just removed this, do not create confusion. I'll create a jira ticket as a follow up, which will contain this info

convert2rhel/checks.py
ZhukovGreen: thx

ZhukovGreen (commented):
@SpyTec  thx for your review, all comments except one is addressed

convert2rhel/unit_tests/checks_test.py
```diff
@@ -16,30 +16,401 @@
...
+        with pytest.raises(SystemExit):
+            ensure_compatibility_of_kmods(_refresh_system_info=True)
+    else:
+        ensure_compatibility_of_kmods(_refresh_system_info=True)
+    if host_kmods == HOST_MODULES_STUB_GOOD:
+        assert "Kernel modules are compatible" in caplog.records[-1].message
+    else:
+        assert (
+            "Kernel modules are compatible" not in caplog.records[-1].message
+        )
```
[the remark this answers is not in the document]
ZhukovGreen: Let me leave my impl. It is important to keep if host_kmods == HOST_MODULES_STUB_GOOD

convert2rhel/unit_tests/conftest.py
ZhukovGreen: just removed it, and made private. it is not assumed to be used outside of the pretend_centos8

convert2rhel/unit_tests/conftest.py
```diff
@@ -68,3 +70,29 @@ def pkg_root(is_py2):
...
+    monkeypatch.setattr(
+        utils,
+        "DATA_DIR",
+        value=str(pkg_root / "convert2rhel/data/8/x86_64/"),
+    )
+
+
+@pytest.fixture()
+def pretend_centos8(monkeypatch, replace_data_dir_for_centos8):
+    # TODO create similar for rest systems and document its usage
```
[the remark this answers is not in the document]
ZhukovGreen: yes

convert2rhel/unit_tests/main_test.py
```diff
@@ -224,6 +230,8 @@ def test_pre_ponr_conversion_order_without_rhsm(self):
 
         intended_call_order["remove_repofile_pkgs"] = 1
         intended_call_order["patch"] = 1
+        intended_call_order["perform_final_checks_before_ponr"] = 1
+        intended_call_order["perform_pre_checks"] = 1
```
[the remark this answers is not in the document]
ZhukovGreen: thx

ZhukovGreen (commented):
@SpyTec added another set of fixes

shaded-enmity:
Test results from CentOS7 with custom kmod (both installed and just present on the FS)
```
# cat main.c
#include <linux/module.h>	/* Needed by all modules */
#include <linux/kernel.h>	/* Needed for KERN_INFO */

int init_module(void)
{
	printk(KERN_INFO "Hello world 1.\n");
	return 0;
}

void cleanup_module(void)
{
	printk(KERN_INFO "Goodbye world 1.\n");
}

# ls -lah
total 224
-rw-r--r--. 1 root root    256 Mar 22 09:30 main.c
-rw-r--r--. 1 root root 100992 Mar 22 09:36 main.ko
-rw-r--r--. 1 root root    907 Mar 22 09:36 main.mod.c
-rw-r--r--. 1 root root  58856 Mar 22 09:36 main.mod.o
-rw-r--r--. 1 root root  45072 Mar 22 09:36 main.o
-rw-r--r--. 1 root root    172 Mar 22 09:36 Makefile
-rw-r--r--. 1 root root     26 Mar 22 09:36 modules.order
-rw-r--r--. 1 root root      0 Mar 22 09:36 Module.symvers

# modinfo main.ko
filename:       /root/kmod/main.ko
retpoline:      Y
rhelversion:    7.9
srcversion:     140276773A3090F6F33891F
depends:        
vermagic:       3.10.0-1160.21.1.el7.x86_64 SMP mod_unload modversions

# lsmod | grep main
main                   12426  0

# cat /proc/modules | grep '('
main 12426 0 - Live 0xffffffffc06c6000 (POE)

# dmesg
<<snip>>
[  797.912337] main: module verification failed: signature and/or required key missing - tainting kernel
[  797.912442] Hello world 1.

# convert2rhel
<<snip>>
[03/22/2021 09:55:11] TASK - [Convert: Final system checks before main conversion] **************
main 12426 0 - Live 0xffffffffc06c6000 (POE)
CRITICAL - Unsigned kernel module(s) has been detected. Third-party components are not supported per our software support policy
https://access.redhat.com/third-party-software-support.

 Uninstall or disable the module(s) and run convert2rhel again to continue with the conversion.
WARNING - Abnormal exit! Performing rollback ...

# rmmod main
# dmesg
<<snip>>
[ 1876.068697] Goodbye world 1.

# cp main.ko /lib/modules/3.10.0-1160.el7.x86_64/kernel/kernel
# convert2rhel
<<snip>>
CRITICAL - The following kernel module(s) are not supported in RHEL:
 kernel/kernel/main.ko
WARNING - Abnormal exit! Performing rollback ...
```

shaded-enmity:
Same test on CentOS 8.3:
```
# cat main.c
#include <linux/module.h>	/* Needed by all modules */
#include <linux/kernel.h>	/* Needed for KERN_INFO */

int init_module(void)
{
	printk(KERN_INFO "Hello world 1.\n");
	return 0;
}

void cleanup_module(void)
{
	printk(KERN_INFO "Goodbye world 1.\n");
}
# ls -lah
-rw-r--r--.  1 root root  256 Mar 22 10:38 main.c
-rw-r--r--.  1 root root 211K Mar 22 10:39 main.ko
-rw-r--r--.  1 root root  826 Mar 22 10:39 main.mod.c
-rw-r--r--.  1 root root 109K Mar 22 10:39 main.mod.o
-rw-r--r--.  1 root root 104K Mar 22 10:39 main.o
-rw-r--r--.  1 root root  154 Mar 22 10:38 Makefile
-rw-r--r--.  1 root root   21 Mar 22 10:39 modules.order
-rw-r--r--.  1 root root    0 Mar 22 10:39 Module.symvers
# modinfo main.ko
filename:       /root/main.ko
rhelversion:    8.3
srcversion:     140276773A3090F6F33891F
depends:        
name:           main
vermagic:       4.18.0-240.15.1.el8_3.x86_64 SMP mod_unload modversions 
# insmod main.ko
# cat /proc/modules | grep "("
main 16384 0 - Live 0xffffffffc03f3000 (POE)
# dmesg
<<snip>>
[  690.281620] main: module verification failed: signature and/or required key missing - tainting kernel
[  690.281702] Hello world 1.
# convert2rhel
[03/22/2021 10:50:28] TASK - [Convert: Final system checks before main conversion] **************
main 16384 0 - Live 0xffffffffc03f3000 (POE)
CRITICAL - Unsigned kernel module(s) has been detected. Third-party components are not supported per our software support policy
https://access.redhat.com/third-party-software-support.

 Uninstall or disable the module(s) and run convert2rhel again to continue with the conversion.
WARNING - Abnormal exit! Performing rollback ...
# rmmod main.ko

# cp main.ko /lib/modules/4.18.0-240.15.1.el8_3.x86_64/kernel/kernel/
# convert2rhel
<<snip>>
CRITICAL - The following kernel module(s) are not supported in RHEL:
 kernel/kernel/main.ko
WARNING - Abnormal exit! Performing rollback ...

```

convert2rhel/checks.py
```diff
@@ -16,13 +16,194 @@
...
+
+def ensure_compatibility_of_kmods(_refresh_system_info=False):
+    """Ensure if the host kernel modules are compatible with RHEL."""
+    check_thirdparty_kmods()
+    host_kmods = get_host_kmods()
+    rhel_supported_kmods = get_rhel_supported_kmods(_refresh_system_info)
+    if not host_kmods.issubset(rhel_supported_kmods):
+        logger.critical(
+            "The following kernel module(s) are not supported "
+            "in RHEL:\n %s" % "\n".join(host_kmods - rhel_supported_kmods)
```
shaded-enmity, line 201: This currently prints path relative to `/lib/modules/$(uname -r)/` which can be confusing, we should print absolute path here.

convert2rhel/checks.py
```diff
@@ -16,13 +16,194 @@
...
+    unsigned_modules, _ = run_subprocess("grep '(' /proc/modules")
+    if unsigned_modules:
+        logger.critical(
+            "Unsigned kernel module(s) has been detected. "
+            "Third-party components are not supported per our "
+            "software support policy\n%s.\n\n "
+            "Uninstall or disable the module(s) and run convert2rhel "
+            "again to continue with the conversion.",
+            LINK_KMODS_RH_POLICY,
+        )
```
shaded-enmity, line 190: 1. The output of grep is visible in stdout
2. It's not clear which kmod is the problem

```suggestion
    unsigned_modules, _ = run_subprocess("grep '(' /proc/modules")
    module_names = "\n  ".join([mod.split(' ')[0] for mod in unsigned_modules.splitlines()])
    if unsigned_modules:
        logger.critical(
            "Unsigned kernel module(s) has been detected. "
            "Third-party components are not supported per our "
            "software support policy\n%s.\n\n"
            "Uninstall or disable the following module(s) and run convert2rhel "
            "again to continue with the conversion:\n  %s",
            LINK_KMODS_RH_POLICY,
            module_names
        )
```

shaded-enmity (changes requested):
Added 2 comments to make the output more clear about locations & names. Other than that my tests are passing and code in general LGTM.

convert2rhel/checks.py
```diff
@@ -16,13 +16,194 @@
...
+
+
+def get_host_kmods():
+    try:
+        # TODO make it work with utils.run_subprocess (now fails)
+        kernel_version = check_output(["uname", "-r"]).decode().rstrip()
+        kmod_str = check_output(
+            ["find", "/lib/modules/" + kernel_version, "-name", "*.ko*"]
+        ).decode()
+        assert kmod_str
```
bocekm, line 79:
```suggestion
        kernel_version, ret_code = utils.run_subprocess("uname -r")
        assert ret_code == 0
        kmod_str, ret_code = utils.run_subprocess("find /lib/modules/" + kernel_version.rstrip() + " -name *.ko*")
        assert kmod_str
        assert ret_code == 0
```

It works for me using the `run_subprocess` on CentOS 6/7/8.

bocekm (changes requested):

convert2rhel/checks.py
SpyTec: Updated your suggestion @shaded-enmity to remove a dot after the URL. Some resolve the URL with the dot in it

So it will now be

```
CRITICAL - Unsigned kernel module(s) has been detected. Third-party components are not supported per our software support policy
https://access.redhat.com/third-party-software-support
```

instead of

```
CRITICAL - Unsigned kernel module(s) has been detected. Third-party components are not supported per our software support policy
https://access.redhat.com/third-party-software-support.
```

convert2rhel/checks.py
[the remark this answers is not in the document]
SpyTec: Right, I just don't know what it means

convert2rhel/unit_tests/checks_test.py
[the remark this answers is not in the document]
SpyTec: In this case we only have good and bad, so one if-else is enough. My suggestion is ultimately the same code, just shortened 

Can also add an explicit `elif host_kmods == HOST_MODULES_STUB_GOOD` in my suggestion too

convert2rhel/checks.py
bocekm: Calling `find` is not so pythonic. I'd prefer using for example `glob`, like [here](https://github.com/oamg/convert2rhel/blob/v0.18/convert2rhel/pkghandler.py#L728). But I'll not block the PR on that.

convert2rhel/checks.py
```diff
@@ -16,13 +16,194 @@
...
+        for distinct_kernel_pkgs in pkgs_groups
+        if distinct_kernel_pkgs[0].startswith(("kernel", "kmod"))
+    )
+
+
+def get_rhel_supported_kmods(_refresh_system_info=False):
+    """Return set of target RHEL supported kernel modules."""
+    # allowing to rerun resolve_system_info (for the needs of tests)
+    if _refresh_system_info:
+        system_info.resolve_system_info()
```
bocekm, line 131: I'm not in favor of code that is present only for the purpose of unit tests. I'm sure there's a way to write the tests to not require this workaround.

convert2rhel/checks.py
ZhukovGreen: added

convert2rhel/checks.py
ZhukovGreen: applied

convert2rhel/checks.py
ZhukovGreen: me either but I was unable to find the way to solve it differently. The reason is such things like https://github.com/oamg/convert2rhel/blob/main/convert2rhel/systeminfo.py#L225. So the class initialized during the modules loading and  I can't affect it, without reinitialization. 

I believe it will be better to try to resolve this in a separate fix.

ZhukovGreen (commented):
@bocekm done with fixing the review

bocekm (changes requested):

convert2rhel/checks.py
```diff
@@ -38,9 +219,10 @@ def check_uefi():
...
-    """Perform all 'registered' checks
-
-    Every check function in this file should be performed from here.
-    This is the entry-point for this module.
-    """
+    """Early checks after system facts should be added here."""
     check_uefi()
+
+
+def perform_pre_ponr_checks():
```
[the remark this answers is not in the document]
ZhukovGreen: There is a reason, that pre_ponr check needs subscription manger installed, because we are using repoquery

convert2rhel/checks.py
```diff
@@ -16,13 +16,194 @@
...
+
+def _get_kmod_comparison_key(path):
+    """Create a comparison key from the kernel module abs path.
+
+    Converts /lib/modules/5.8.0-7642-generic/kernel/lib/a.ko.xz ->
+    kernel/lib/a.ko.xz
+
+    Why:
+        all kernel modules lives in /lib/modules/{some kernel release}
+        if we want to make sure that the kernel package is really presented
```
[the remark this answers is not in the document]
ZhukovGreen: applied

convert2rhel/checks.py
```diff
@@ -16,13 +16,194 @@
...
+        on RHEL, we need to compare the full path, but because kernel release
+        might be different, we compare the relative paths after kernel release.
+    """
+    return "/".join(path.split("/")[4:])
+
+
+def get_host_kmods():
```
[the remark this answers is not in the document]
ZhukovGreen: renamed, added docstrings

ZhukovGreen (commented):
@bocekm your turn:)

convert2rhel/checks.py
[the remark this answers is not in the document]
pirat89: In such a case, I suggest to mention it in the docstring and commit message so it's clear.

convert2rhel/checks.py
pirat89: @ZhukovGreen why it cannot reinitialized before the call of the function? To be honest I don't see the problem with that. what is preventing possibility to set the `system_info` in the test before you call the function?

shaded-enmity: DI or dependency inversion could be used here:
```python
def get_rhel_supported_kmods(system_info):
```

I.e. make it responsibility of the caller to provide `system_info`. Another benefit is that in testing you can supply the object without the need for mocking/monkeypatching.

shaded-enmity (commented):
Generally LGTM, I will test again after 2pm today.

convert2rhel/checks.py
bocekm: This is essentially dead code - code that can never be executed at run-time.
After talking to @shaded-enmity, I agreed on not blocking the PR on things like this that do not directly affect the functionality. Here's a high priority task to remove the dead code and resolve the unit tests in a different way: https://issues.redhat.com/browse/OAMG-4634.

ZhukovGreen: As a DI it will work as well, but will be inconsistent, as this will be the only place where we would do this (at least now). @pirat89 advice works nicely, thx. OK, now we don't have this hack and it is moved to the fixture

convert2rhel/checks.py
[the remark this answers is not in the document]
ZhukovGreen: We have docstrings for that:
```python
def perform_pre_checks():


def perform_pre_ponr_checks():
    """Late checks before ponr should be added here."""
```
I believe it is clear enough, @pirat89 wdyt?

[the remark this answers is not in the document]
bocekm: I'd rename them first one to `perform_early_checks` to make it more descriptive.

convert2rhel/checks.py
```diff
@@ -16,13 +16,207 @@
...
+        max(distinct_kernel_pkgs[1], key=_repos_version_key)
+        for distinct_kernel_pkgs in pkgs_groups
+        if distinct_kernel_pkgs[0].startswith(("kernel", "kmod"))
+    )
+
+
+def get_rhel_supported_kmods():
+    """Return set of target RHEL supported kernel modules."""
+    repoquery_repoids_args = " ".join(
+        "--repoid " + repoid for repoid in system_info.default_rhsm_repoids
```
bocekm, line 127: This would work only in case of using RHSM, not the custom repos.

```suggestion
        "--repoid " + repoid for repoid in get_enabled_rhel_repos()
```

Create the following new function in systeminfo.py:
```python
get_enabled_rhel_repos():
    """Return a list of repoids containing RHEL packages.

    These are either the repos enabled through RHSM or the custom repositories passed though CLI.
    """
    return system_info.submgr_enabled_repos if not tool_opts.disable_submgr else tool_opts.enablerepo
```
Use this new function also on two other places:
- https://github.com/oamg/convert2rhel/blob/v0.18/convert2rhel/pkghandler.py#L122
- https://github.com/oamg/convert2rhel/blob/v0.18/convert2rhel/pkghandler.py#L637

bocekm (changes requested):

bocekm (changes requested):

convert2rhel/checks.py
```diff
@@ -16,13 +16,207 @@
...
+    When we scan kernel modules provided by kernel packages
+    it is expensive to check each kernel pkg. Since each new
+    kernel pkg do not deprecate kernel modules we only select
+    the most recent ones.
+
+    :type pkgs: Iterable[str]
+    """
+
+    pkgs_groups = itertools.groupby(
+        pkgs, lambda pkg_name: pkg_name.split(":")[0]
```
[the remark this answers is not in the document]
ZhukovGreen: Added

convert2rhel/checks.py
```diff
@@ -16,13 +16,207 @@
...
+    )
+    return (
+        max(distinct_kernel_pkgs[1], key=_repos_version_key)
+        for distinct_kernel_pkgs in pkgs_groups
+        if distinct_kernel_pkgs[0].startswith(("kernel", "kmod"))
```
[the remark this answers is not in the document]
ZhukovGreen: All RHEL kmods packages starts with kernel* or kmod*. Added to docstring

convert2rhel/checks.py
```diff
@@ -16,13 +16,207 @@
...
+        _get_kmod_comparison_key(kmod_path)
+        for kmod_path in filter(
+            lambda path: path.endswith(("ko.xz", "ko")),
+            rhel_kmods_str.rstrip("\n").split(),
+        )
+    )
+
+
+def check_thirdparty_kmods():
+    """Identify tainted kmods.
```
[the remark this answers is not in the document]
ZhukovGreen: I renamed it for consistency

convert2rhel/checks.py
```diff
@@ -16,13 +16,207 @@
...
+        system76_acpi 16384 0 - Live 0x0000000000000000 (OE) <<<<< Tainted
+
+    """
+    unsigned_modules, _ = run_subprocess("grep '(' /proc/modules")
+    module_names = "\n  ".join(
+        [mod.split(" ")[0] for mod in unsigned_modules.splitlines()]
+    )
+    if unsigned_modules:
+        logger.critical(
+            "Unsigned kernel module(s) has been detected. "
```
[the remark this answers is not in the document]
ZhukovGreen: thx

convert2rhel/checks.py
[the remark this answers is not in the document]
ZhukovGreen: ok, but this is out of scope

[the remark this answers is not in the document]
ZhukovGreen: will create a ticket for that

ZhukovGreen (commented):
@bocekm done with the fixing review

shaded-enmity (approved):
Tested scenarios:

- CentOS 7 RHSM
- CentOS 7 Custom repositories
- CentOS 8 RHSM
- CentOS 8 Custom repositories

All tests were **green**. Approving.

bocekm (approved):

convert2rhel/checks.py
```diff
@@ -16,13 +16,233 @@
...
+    kernel/lib/a.ko.xz) in order to be able to compare with RHEL
+    kernel modules in case of different kernel release
+    """
+    try:
+        kernel_version, exit_code = run_subprocess("uname -r")
+        assert exit_code == 0
+        kmod_str, exit_code = run_subprocess(
+            'find /lib/modules/{kver} -name "*.ko*"'.format(
+                kver=kernel_version.rstrip("\n")
+            )
```
bocekm, line 73:
```suggestion
            ), print_output=False
```

convert2rhel/checks.py
```diff
@@ -16,13 +16,233 @@
...
+def get_rhel_supported_kmods():
+    """Return set of target RHEL supported kernel modules."""
+    repoquery_repoids_args = " ".join(
+        "--repoid " + repoid for repoid in system_info.get_enabled_rhel_repos()
+    )
+    # get output of a command to get all packages which are the source
+    # of kmods
+    kmod_pkgs_str, _ = run_subprocess(
+        (
+            "repoquery "
```
bocekm, line 160: You need to handle the repoquery commands in a special way on CentOS/OL 8, otherwise you'll be getting:
```
Modular dependency problems:

 Problem 1: conflicting requests
  - nothing provides module(platform) needed by module ant:1.10:820181213135032:5ea3b708-0.x86_64
 Problem 2: conflicting requests
  - nothing provides module(platform:el8) needed by module container-tools:rhel8:8030020201124131330:830d479e-0.x86_64
```

For yum calls, we use the `--setopt=module_platform_id=platform:el8` option: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/pkghandler.py#L113.

convert2rhel/checks.py
```diff
@@ -16,13 +16,233 @@
...
+            "--releasever={releasever} "
+            "{repoids_args} "
+            "-f /lib/modules/*.ko*"
+        ).format(
+            releasever=system_info.releasever,
+            repoids_args=repoquery_repoids_args,
+        )
```
bocekm, line 167:
```suggestion
        ), print_output=False
```

bocekm (changes requested):

ZhukovGreen (commented):
@bocekm done

convert2rhel/checks.py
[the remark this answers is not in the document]
bocekm: https://issues.redhat.com/browse/OAMG-4668

bocekm (approved):

shaded-enmity (approved):
LGTM!

Issue #123, opened 2020-11-19
Error during unpacking rpm package filesystem-2.4.30-3.el6.x86_64 when converting from CENTOS6 to RHEL6


ahilden:
I am getting the following errors when running convert2rhel on all CENTOS6 systems.  Although it does look the package is installed.  This causes the convert2rhel script to fail. After it fails what are the possible recovery paths to complete the conversion?

Resolving Dependencies
--> Running transaction check
---> Package filesystem.x86_64 0:2.4.30-3.el6 will be reinstalled
--> Finished Dependency Resolution
Dependencies Resolved
================================================================================
 Package         Arch        Version              Repository               Size
================================================================================
Reinstalling:
 filesystem      x86_64      2.4.30-3.el6         rhel-6-server-rpms      1.0 M
Transaction Summary
================================================================================
Reinstall     1 Package(s)
Total download size: 1.0 M
Installed size: 0  
Downloading Packages:
Running rpm_check_debug
Running Transaction Test
Transaction Test Succeeded
Running Transaction
  Installing : filesystem-2.4.30-3.el6.x86_64                               1/1Error unpacking rpm package filesystem-2.4.30-3.el6.x86_64
error: unpacking of archive failed on file /sys: cpio: chown
  Verifying  : filesystem-2.4.30-3.el6.x86_64                               1/1 
Failed:
  filesystem.x86_64 0:2.4.30-3.el6                                              
Complete!
Received return code: 1
Resolving dependency errors ... 
Found protected packages: []
Found deps packages: []
Found multilib packages: []
Found req packages: []
No other package to try to downgrade in order to resolve yum dependency errors.
[11/19/2020 16:50:51] CRITICAL - Could not resolve yum errors.

SpyTec:
Hi!

Are you able to get the machine working by rebooting and running `yum update -y` after the error has occurred? Does this happen on version 0.13 or an older version? `yum list convert2rhel`

bocekm:
Hi, thanks for the report.

Please make sure that your `/` and `/sys/` folders are not mounted as read-only (https://unix.stackexchange.com/questions/173896/sys-cpio-chown-failed-read-only-file-system-building-passenger-rpm-using-d).

Regarding the possible recovery - the failure happened after the point where conver2rhel is able to do a rollback.
But you can try going to `/var/lib/convert2rhel/` and run `rpm -i --replacepkgs *rpm`. That should at least get you to a point from where you can run convert2rhel again.

[...]

PR #194, opened 2021-02-22
add: Check if /mnt /sys is read/write

fellipeh:
This commit will check if `/mnt` and `/sys` are read/write, if not, will abort the conversion.

Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1887513

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -149,6 +149,10 @@ def pre_ponr_conversion():
     # check if user pass some repo to both disablerepo and enablerepo options
     pkghandler.has_duplicate_repos_across_disablerepo_enablerepo_options()
 
+    # checking if /mnt and /sys are read-write
+    loggerinst.task("Convert: Checking /mnt and /sys are read-write")
+    utils.check_readonly_mounts()
+
     # package analysis
     loggerinst.task("Convert: List third-party packages")
     pkghandler.list_third_party_pkgs()
--- a/convert2rhel/unit_tests/main_test.py
+++ b/convert2rhel/unit_tests/main_test.py
@@ -19,7 +19,6 @@
 import os
 import unittest
 
-
 try:
     from collections import OrderedDict
 except ImportError:
@@ -35,7 +34,6 @@ from convert2rhel import (
     subscription,
     utils,
 )
-from convert2rhel.systeminfo import SystemInfo
 from convert2rhel.toolopts import tool_opts
 
 
@@ -54,6 +52,10 @@ class TestMain(unittest.TestCase):
     @unit_tests.mock(utils, "DATA_DIR", eula_dir)
     def test_show_eula(self):
         main.show_eula()
+
+    class GetFakeFunctionMocked(unit_tests.MockFunction):
+        def __call__(self, filename):
+            pass
 
     class GetFileContentMocked(unit_tests.MockFunction):
         def __call__(self, filename):
@@ -100,8 +102,6 @@ class TestMain(unittest.TestCase):
         @classmethod
         def reset(cls):
             cls.calls = OrderedDict()
-
-
 
     class CallYumCmdMocked(unit_tests.MockFunction):
         def __init__(self):
@@ -159,6 +159,7 @@ class TestMain(unittest.TestCase):
     @mock_calls(subscription, "disable_repos", CallOrderMocked)
     @mock_calls(subscription, "enable_repos", CallOrderMocked)
     @mock_calls(subscription, "download_rhsm_pkgs", CallOrderMocked)
+    @unit_tests.mock(utils, "check_readonly_mounts", GetFakeFunctionMocked)
     def test_pre_ponr_conversion_order_with_rhsm(self):
         self.CallOrderMocked.reset()
         main.pre_ponr_conversion()
@@ -200,6 +201,7 @@ class TestMain(unittest.TestCase):
     @mock_calls(subscription, "disable_repos", CallOrderMocked)
     @mock_calls(subscription, "enable_repos", CallOrderMocked)
     @mock_calls(subscription, "download_rhsm_pkgs", CallOrderMocked)
+    @unit_tests.mock(utils, "check_readonly_mounts", GetFakeFunctionMocked)
     def test_pre_ponr_conversion_order_without_rhsm(self):
         self.CallOrderMocked.reset()
         main.pre_ponr_conversion()
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -31,6 +31,8 @@ import traceback
 from six import moves
 
 
+loggerinst = logging.getLogger(__name__)
+
 class Color(object):
     PURPLE = '\033[95m'
     CYAN = '\033[96m'
@@ -110,7 +112,6 @@ def store_content_to_file(filename, content):
 
 
 def restart_system():
-    loggerinst = logging.getLogger(__name__)
     from convert2rhel.toolopts import tool_opts
     if tool_opts.restart:
         run_subprocess("reboot")
@@ -124,7 +125,6 @@ def run_subprocess(cmd="", print_cmd=True, print_output=True):
     output (print_output=True). Switching off printing the command can be useful in case it contains
     a password in plain text.
     """
-    loggerinst = logging.getLogger(__name__)
     if print_cmd:
         loggerinst.debug("Calling command '%s'" % cmd)
 
@@ -170,7 +170,6 @@ def run_cmd_in_pty(cmd="", print_cmd=True, print_output=True, columns=120):
     :return: The output (combined stdout and stderr) and the return code of the executed command
     :rtype: tuple
     """
-    loggerinst = logging.getLogger(__name__)
     if print_cmd:
         loggerinst.debug("Calling command '%s'" % cmd)
 
@@ -178,7 +177,7 @@ def run_cmd_in_pty(cmd="", print_cmd=True, print_output=True, columns=120):
         # https://github.com/pexpect/pexpect/issues/134
         def setwinsize(self, rows, cols):
             super(PexpectSizedWindowSpawn, self).setwinsize(0, columns)
-    
+
     process = PexpectSizedWindowSpawn(cmd, env={'LC_ALL': 'C'}, timeout=None)
 
     # The setting of window size is super unreliable
@@ -198,7 +197,6 @@ def run_cmd_in_pty(cmd="", print_cmd=True, print_output=True, columns=120):
 
 def let_user_choose_item(num_of_options, item_to_choose):
     """Ask user to enter a number corresponding to the item they choose."""
-    loggerinst = logging.getLogger(__name__)
     while True:  # Loop until user enters a valid number
         opt_num = prompt_user("Enter number of the chosen %s: "
                               % item_to_choose)
@@ -232,7 +230,6 @@ def ask_to_continue():
     """Ask user whether to continue with the system conversion. If no,
     execution of the tool is stopped.
     """
-    loggerinst = logging.getLogger(__name__)
     from convert2rhel.toolopts import tool_opts
     if tool_opts.autoaccept:
         return
@@ -245,7 +242,6 @@ def ask_to_continue():
 
 
 def prompt_user(question, password=False):
-    loggerinst = logging.getLogger(__name__)
     color_question = Color.BOLD + question + Color.END
     if password:
         response = getpass.getpass(color_question)
@@ -259,7 +255,6 @@ def log_traceback(debug):
     """Log a traceback either to both a file and stdout, or just file, based
     on the debug parameter.
     """
-    loggerinst = logging.getLogger(__name__)
     traceback_str = get_traceback_str()
     if debug:
         # Print the traceback to the user when debug option used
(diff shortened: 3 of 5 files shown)


convert2rhel/main.py
```diff
@@ -151,6 +151,10 @@ def pre_ponr_conversion():
     # check if user pass some repo to both disablerepo and enablerepo options
     pkghandler.has_duplicate_repos_across_disablerepo_enablerepo_options()
 
+    # checking if /mnt and /sys are read-write
+    loggerinst.task("Convert: Checking /mnt and /sys are read-write")
+    utils.check_if_mount_is_readonly()
```
ZhukovGreen, line 156: I created a special python module checks.py (see #193 ) for doing this kind of checks. Could you please grab it into your MR, so we'll have your check also in checks.py? WDYT?

convert2rhel/utils.py
```diff
@@ -560,4 +560,17 @@ def backup(self):
             loggerinst.warning("Can't access %s" % TMP_DIR)
 
 
+def check_if_mount_is_readonly():
+    loggerinst = logging.getLogger(__name__)
```
ZhukovGreen, line 564: let's merge it after #191 and use logger on the module level? I fixed this in other places

convert2rhel/utils.py
```diff
@@ -560,4 +560,17 @@ def backup(self):
...
+
+    mounts = get_file_content('/proc/mounts', as_list=True)
+    for line in mounts:
+        device, mount_point, filesystem, flags, __, __ = line.split()
+        flags = flags.split(',')
+        if mount_point == '/mnt' or mount_point == '/sys':
```
ZhukovGreen, line 570: Nit:
```suggestion
        if mount_point in ('/mnt',  '/sys'):
```

ZhukovGreen (commented):
@fellipeh let's merge it after #191 and use checks.py from #193

convert2rhel/main.py
SpyTec: They're not related, so whichever get merged first should adhere to the other - not create dependencies. Either PR can be pushed back due to reviews, so keep stacked PRs to a minimum.

convert2rhel/utils.py
SpyTec: Same as my previous comment, #191 can make this change if this PR is merged first. Or this PR can make the change if #191 is merged first

convert2rhel/main.py
ZhukovGreen: @SpyTec It is not stacking MRs, but synchronizing on the solution to implement different kinds of checks. I am reviewing this MR and I see that Fellipe implement check. I know that I also do this. My intention is to synchronize with the team on doing checks.

I would call this as weak dependency.

bocekm: No need to merge #193. We can merge this one and then, upon rebasing #193, @ZhukovGreen you can just move this check to the new checks.py.

convert2rhel/utils.py
bocekm: #191 merged. @fellipeh, please rebase and use logger defined at the module level.

convert2rhel/main.py
ZhukovGreen: @bocekm Definitely no need to wait for #193

convert2rhel/utils.py
```diff
@@ -560,4 +560,17 @@ def backup(self):
             loggerinst.warning("Can't access %s" % TMP_DIR)
 
 
+def check_if_mount_is_readonly():
```
bocekm, line 563:
```suggestion
def check_readonly_mounts():
    """Add a docstring that clearly describes what we're trying to achieve."""
```

The docstring can mention:

- Mounting directly to /mnt/ is not in line with Unix FS (https://en.wikipedia.org/wiki/Unix_filesystem).
- Having /mnt/ and /sys/ read-only causes the installation of the filesystem package to fail (https://bugzilla.redhat.com/show_bug.cgi?id=1887513, https://github.com/oamg/convert2rhel/issues/123).

convert2rhel/utils.py
```diff
@@ -560,4 +560,17 @@ def backup(self):
...
+    loggerinst = logging.getLogger(__name__)
+
+    mounts = get_file_content('/proc/mounts', as_list=True)
+    for line in mounts:
+        device, mount_point, filesystem, flags, __, __ = line.split()
```
bocekm, line 568:
```suggestion
        _, mount_point, _, flags, _, _ = line.split()
```

convert2rhel/utils.py
```diff
@@ -560,4 +560,17 @@ def backup(self):
...
+        flags = flags.split(',')
+        if mount_point == '/mnt' or mount_point == '/sys':
+            (loggerinst.critical(
+                "%s are not read-write. Conversion can not continue." % mount_point) if 'ro' in flags else loggerinst.debug(
+                "/mnt and /sys mount are readwrite."))
```
bocekm: As you have it now, the _"/mnt and /sys mount are readwrite."_ would be printed twice.

```suggestion
        if 'ro' not in flags and mount_point not in ('/mnt',  '/sys'):
            continue
        if mount_point == '/mnt':
            loggerinst.critical("/mnt/ is read-only - that would cause the conversion to fail.\n"
                                "Do not mount directly to /mnt,  mount to its subdirectories instead.")
        else:  # /sys
            loggerinst.critical("/sys/ is read-only - that would cause the conversion to fail.\n"
                                "Make sure it is mounted in read-write mode before executing convert2rhel.")
        loggerinst.debug("%s mounted in read-write mode." % mount_point)
```

bocekm (changes requested):

convert2rhel/utils.py
```diff
@@ -560,4 +545,27 @@ def backup(self):
...
+            continue
+        else:
+            if 'ro' in flags:
+                if mount_point == '/mnt':
+                    loggerinst.critical("/mn/ is read-only - that would cause the conversion to fail.\n"
+                                        "Do not mount directly to /mnt,  mount to its subdirectories instead.")
+                else:  # /sys
+                    loggerinst.critical("/sys is read-only - that would cause the conversion to fail.\n"
+                                        "Make sure it is mounted in read-write mode before executing convert2rhel.")
+            loggerinst.debug("%s is not read-only." % mount_point)
```
SpyTec, line 568: * Else not necessary for, if-statement continues to next iteration. Keeping the else for the sys critical logger call for readability reasons.
* Changed the text for both log calls

```suggestion
        if 'ro' in flags:
            if mount_point == '/mnt':
                loggerinst.critical("Detected read-only mount to /mnt directory. This will cause the conversion to fail.\n"
                                    "Mount to a subdirectory of /mnt to have read-write.")
            else:  # /sys
                loggerinst.critical("Detected read-only mount to /sys directory. This will cause the conversion to fail.\n"
                                    "Mount in read-write mode before executing convert2rhel.")
        loggerinst.debug("%s mount is not read-only." % mount_point)
```

SpyTec (changes requested):

convert2rhel/unit_tests/main_test.py
```diff
@@ -19,6 +19,7 @@
 import os
 import unittest
 
+from convert2rhel.unit_tests import GetFileContentMocked
```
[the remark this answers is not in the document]
fellipeh: It's a completely different kind of mock. From now we are using the `unit_test.py` GetFileContentMocked. but, if I remove the mock from line 64, I need to refactor all unit test using it, which is out of this PT scope. So, I believe we can move forward with this, and make a new PR with unit test refactored, what you think about this approach?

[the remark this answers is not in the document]
SpyTec: Yes, then this line needs to be deleted if it's not used in the test file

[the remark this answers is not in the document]
fellipeh: I believe you doesn't understand me.. :)  on my test I used the "global"  GetFileContentMocked, (if you look at that you can see the difference between the both mocks and the reason I need to use the global one)  that's the reason we have that line..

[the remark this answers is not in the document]
SpyTec: Two things, you didn't write a test in this file. You only updated some existing tests with another mock for `GetFakeFunctionMocked`, but nothing was changed with `GetFileContentMocked`.

This import has the same name as the mock on line 64, which means you have two declarations of `GetFileContentMocked` in the file

https://github.com/fellipeh/convert2rhel/blob/f409f404ea11688e63094c418a9ffa9b07211c63/convert2rhel/unit_tests/main_test.py#L64-L66

If you didn't mean to update the `GetFileContentMocked` in this file the import should be deleted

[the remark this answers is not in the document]
fellipeh: You're right.. my mistake.. I was thinking about `utils_test` not main_test.. my mistake

TomasTomecek:
sadly testing farm does not seem to support RHEL/CentOS 6 :/

SpyTec (dismissed):

convert2rhel/utils.py
```diff
@@ -560,4 +545,26 @@ def backup(self):
...
+        if mount_point not in ('/mnt', '/sys'):
+            continue
+        if 'ro' in flags:
+            if mount_point == '/mnt':
+                loggerinst.critical("Detected read-only mount to /mnt directory. This will cause the conversion to fail.\n"
+                                    "Mount to a subdirectory of /mnt to have read-write.")
+            else:  # /sys
+                loggerinst.critical("Detected read-only mount to /sys directory. This will cause the conversion to fail.\n"
+                                    "Mount in read-write mode before executing convert2rhel.")
+        loggerinst.debug("%s mount is not read-only." % mount_point)
```
bocekm, line 567: In the `mount` man page, they use "mount filesystem/device [on/at] a directory". So they are not consistent, but they do not use "mount filesystem/device to a directory". _Mount_ is a verb, and _mount point_ is the directory on/at which you mount a filesystem/device.

```suggestion
                loggerinst.critical("Detected /mnt as a read-only mount point. That would cause the conversion to fail.\n"
                                    "Use any subdirectory of /mnt as a mount point instead to have /mnt writeable.")
            else:  # /sys
                loggerinst.critical("Detected /mnt as a read-only mount point. That would cause the conversion to fail.\n"
                                    "Mount the directory in read-write mode before executing convert2rhel.")
        loggerinst.debug("%s mount point is not read-only." % mount_point)
```

bocekm (changes requested):

convert2rhel/utils.py
SpyTec: > That would cause the conversion to fail

Why past tense rather than present? IMO if we describe it in past tense here it indicates we're being nice to the user rather than simply stating we can't continue. That we're doing them a favor of stopping the conversion early

> Use any subdirectory of /mnt as a mount point instead to have /mnt writeable.

I find this rather complex rather, this would be better: "Mount to a subdirectory of /mnt to have a [read-write|writable] mount point."

> Mount the directory in read-write mode before executing convert2rhel

We're mounting to a directory in /mnt, not mounting a specific directory

bocekm: > Why past tense rather than present? IMO if we describe it in past tense here it indicates we're being nice to the user rather than simply stating we can't continue. That we're doing them a favor of stopping the conversion early

The "would" does not IMHO imply past tense here. It implies a possibility in the future. I.e. should we not stop you here to make amendments the conversion would fail.

> I find this rather complex rather, this would be better: "Mount to a subdirectory of /mnt to have a [read-write|writable] mount point."

That's indeed better, thanks.

> We're mounting to a directory in /mnt, not mounting a specific directory

Agreed. What about "Make sure this mount point is writeable before executing convert2rhel" to workaround it?

SpyTec: > The "would" does not IMHO imply past tense here. It implies a possibility in the future. I.e. should we not stop you here to make amendments the conversion would fail.

Hm. I think we can skip it altogether actually. We don't need to tell the user it will fail, as we just stopped it - so it won't ever "fail".

What about doing:

> Stopping conversion due to read-only mount to /mnt directory
> Mount to a subdirectory of /mnt to have read-write.

---

> Agreed. What about "Make sure this mount point is writeable before executing convert2rhel" to workaround it?

I would go with "ensure" rather than "make sure":

> Ensure mount point is writable before executing convert2rhel

bocekm: > I would go with "ensure" rather than "make sure":

Sure.

The _"Mount to a subdirectory of /mnt to have read-write."_ sounds strange. Especially the "having read-write". What is read-write and why should I have it? :)
Also, as I wrote, the man page does not use _mount to_, but _mount on_ or _mount at_. So what about:

```
Stopping the conversion due to read-only mount at /mnt directory.
Mount at a subdirectory of /mnt instead to have /mnt writeable.
```

SpyTec: Almost agree, I'd drop `instead` as it's an unneeded filler :)

bocekm: Fine with me.

SpyTec: Something like this then

```suggestion
                loggerinst.critical("Stopping conversion due to read-only mount to /mnt directory.\n"
                                    "Mount at a subdirectory of /mnt to have /mnt writeable.")
            else:  # /sys
                loggerinst.critical("Stopping conversion due to read-only mount to /sys directory.\n"
                                    "Ensure mount point is writable before executing convert2rhel.")
        loggerinst.debug("%s mount point is not read-only." % mount_point)
```

fellipeh:
Fixed unit_test. Please re-review again and move to approve @bocekm @SpyTec

SpyTec (dismissed):

SpyTec (approved):

bocekm (approved):

PR #195, opened 2021-03-02
tests: tmt setup

ZhukovGreen:
Introducing a new way of running tests for our project, including integration tests.
This MR is my attempt to make it work with https://github.com/psss/tmt


plans/centos8/integration.fmf
```diff
@@ -0,0 +1,7 @@
+discover:
+    how: fmf
+#    filter: 'tag: integration & tag: centos8'
```
ZhukovGreen, line 3: @psss @thrix could you guys help me to realize what I am doing wrong with this filter. My expected behavior is that the integration plan will pick tests only with these tags

plans/centos8/unit.fmf
```diff
@@ -0,0 +1,3 @@
+discover:
+  how: fmf
+#  filter: 'tag: unit & tag: centos8'
```
ZhukovGreen, line 3: the same here

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
```diff
@@ -0,0 +1,9 @@
+summary: kmods tests
+# TODO Add prepare with running ansible playbooks
+# TODO why prepare step is ignored?
+prepare:
+  how: shell
+  script: |
+    echo "some preparation script here"
```
ZhukovGreen, line 7: @psss this step is ignored (when do tmt run). Should I create a separate fmf for running test? i.e. this one remains + add run_tests.fmf with test: pytest?

ZhukovGreen (commented):
Could you guys take a look at my attempts to integrate TMT @psss @thrix

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
```diff
@@ -0,0 +1,9 @@
+summary: kmods tests
+# TODO Add prepare with running ansible playbooks
```
ZhukovGreen, line 2: I read thorugh https://tmt.readthedocs.io/en/stable/spec/steps.html#ansible and unable to find how to specify the path to the playbooks

plans/centos8/integration.fmf
psss: Sure, checking what's wrong:

```
> tmt
Found 2 tests: /tests/integration/inhibit-if-kmods-is-not-supported and /tests/unit.
Found 2 plans: /plans/centos8/integration and /plans/centos8/unit.
Found 0 stories.
> tmt test ls
/tests/integration/inhibit-if-kmods-is-not-supported
/tests/unit
> tmt test ls --filter tag:integration

> tmt test show
/tests/integration/inhibit-if-kmods-is-not-supported
     summary kmods tests
        test pytest
        path /tests/integration/inhibit-if-kmods-is-not-supported
      manual no
    duration 5m
     enabled yes
      result respect

/tests/unit
     summary kmods tests
        test pytest .
        path /tests/unit
      manual no
    duration 5m
     enabled yes
      result respect
```

So the reason is that none of your tests has the `tag` defined.
Let's see why:

```
> tmt test show -v integration
/tests/integration/inhibit-if-kmods-is-not-supported
     summary kmods tests
        test pytest
        path /tests/integration/inhibit-if-kmods-is-not-supported
      manual no
    duration 5m
     enabled yes
      result respect
     sources /home/psss/git/github/convert2rhel/tests/integration/main.fmf
             /home/psss/git/github/convert2rhel/tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
      fmf-id name: /tests/integration/inhibit-if-kmods-is-not-supported
             url: https://github.com/oamg/convert2rhel.git
             ref: pr195

> cat /home/psss/git/github/convert2rhel/tests/integration/main.fmf
summary: Integration tests
tags: [integration, centos8]
prepare:
  how: shell
  # TODO install rpm script
  script: |
    echo "Installing rpm here"
```

Ah, I see. The correct name for the attribute is [tag](https://tmt.readthedocs.io/en/latest/spec/tests.html#tag). All test attributes are kept consistent in singular.

psss: Also note, that if a particular test needs a package installed on the guest, you can use the [require](https://tmt.readthedocs.io/en/latest/spec/tests.html#require) attribute for this. The [prepare](https://tmt.readthedocs.io/en/latest/spec/steps.html#prepare) step can be used in plans only.

plans/centos8/unit.fmf
psss: Using `tag` key should fix this.

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
thrix: I believe putting there directly the relative path from the project root should be fine:

```
prepare:
    how: ansible
    playbook:
        - some-path/common.yml
```

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
psss: The `prepare` step can be configured under a plan. To better understand the difference between plan (L2) and test (L1) have a look at the [spec overview](https://tmt.readthedocs.io/en/latest/spec.html).

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
psss: Hm, this should be better documented: https://github.com/psss/tmt/pull/599

plans/centos8/integration.fmf
ZhukovGreen: thx @psss, do you think it is possible to create a json schema for TMT configs? if yes, than it would be possible to upload it https://github.com/schemastore/schemastore/ and a lot of editors will start to validate fmf  files. Added https://github.com/psss/tmt/issues/602

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
ZhukovGreen: OK, thx @psss. I created a ticket to track this https://github.com/psss/tmt/issues/603

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
ZhukovGreen: ok, clear now. Thx

ZhukovGreen:
@psss @thrix thank you a lot guys

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
ZhukovGreen: OK, my plan was to have a possibility to add playbooks for each integration test. So in this case potentially each test should have own plan, right @psss ?

thrix:
rerun tests as there were some changes to the xunit viewer

plans/inhibit-if-kmods-is-not-supported/centos8/main.fmf
```diff
@@ -0,0 +1,14 @@
+discover:
+    how: fmf
+    # TODO filter inheritance doesn't work
```
psss, line 3: In order to [extend](https://fmf.readthedocs.io/en/stable/features.html#merging) values inherited from parent use the `+` suffix. Here's an example plan:

```yaml
discover:
    how: fmf
    filter:
      - tag:common-filter
execute:
    how: tmt

/centos8:
    discover+:
        filter+:
          - tag:centos8
/centos9:
    discover+:
        filter+:
          - tag:centos9
```

Which gives this:

```
> tmt plan show
/plans/example/centos8
    discover 
         how fmf
      filter tag:common-filter and tag:centos8

/plans/example/centos9
    discover 
         how fmf
      filter tag:common-filter and tag:centos9
```

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
```diff
@@ -1,5 +1,5 @@
 summary: kmods tests
 # TODO Add prepare with running ansible playbooks
-tag+: [centos6, centos7, centos8]
+tag: [centos6, centos7, centos8, inhibit-if-kmods-is-not-supported]
```
psss, line 3: Hint: Instead of using `tag` for selecting relevant tests you might want to use the context [adjust](https://tmt.readthedocs.io/en/latest/spec/core.html#adjust) feature which gives a bit more flexibility:

```yaml
adjust:
    enabled: false
    when: distro == centos-8
```
Providing context on the command line works like this:

```
tmt --context distro=centos-8 run ...
```

plans/inhibit-if-kmods-is-not-supported/centos8/main.fmf
ZhukovGreen: Thanks

plans/inhibit-if-kmods-is-not-supported/centos8/main.fmf
```diff
@@ -0,0 +1,14 @@
...
+        - 'tag: centos8'
+        - 'tag: inhibit-if-kmods-is-not-supported'
+provision:
+    how: container
+    image: convert2rhel/centos8
+    pull: false
+prepare:
+    how: ansible
+    playbook:
+        - tests/integration/inhibit-if-kmods-is-not-supported/ansible/setup.yml
```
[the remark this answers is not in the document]
psss: Hm... Seems that the `unshare` command is available in `podman` but not in `docker`? @thrix any idea?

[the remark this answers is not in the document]
ZhukovGreen: I would expect running something like `docker exec convert2rhel/centos8 ansible-playbook ....`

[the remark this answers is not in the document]
thrix: prepare has different paths @ZhukovGreen as I showed in the jira ticket ...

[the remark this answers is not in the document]
thrix: the feature to have the paths as one would expect .. there is an issue for that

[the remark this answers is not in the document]
thrix: yeah, the docs we updated is misleading again, due to this problem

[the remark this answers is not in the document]
thrix: docker is not supporter, only podman :)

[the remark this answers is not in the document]
thrix: we never tested with docker ... let's stick to podman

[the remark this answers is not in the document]
thrix: Isn't the problem as I described OAMG-4515  ?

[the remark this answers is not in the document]
thrix: > I would expect running something like docker exec convert2rhel/centos8 ansible-playbook ....

We do not want to run ansible inside docker. That would require installing Ansible in the container, would not work for older RHELs, etc. We run ansible on the "localhost" using podman Ansible connection plugin to connect to the container.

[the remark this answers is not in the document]
ZhukovGreen: @thrix thx for sharing this. You were right, this is due to docker issue. After installation of podman, it seems almost working. As I understood tmt is now only capable to run particular playbooks, rather than roles?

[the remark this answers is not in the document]
ZhukovGreen: I would vote for running the ansible playbook inside the container. It seems wrong to me to rely on the host. For example I have Debian and docker. Now I am in trouble to continue. Is it the same behavior for VMs?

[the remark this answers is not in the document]
thrix: > I would vote for running the ansible playbook inside the container.

There are use cases where the control host must be outside (like restart of the system). So there is a reason why we do it like this. Of course, that does not apply to the case of containers. Still I am not sure we could easily run ansible for RHEL6 for example, but from outside it should work ...

If you think running ansible from inside is better, we can have it, but it should be optional and we can discuss what the default should be if that is done

> It seems wrong to me to rely on the host.
Well you already rely on host with tmt right? :)

> I have Debian and docker. 

But you know, there is also an ansible docker plugin :) so it just needs few patches to make the "backend" 
use docker instead ... `https://docs.ansible.com/ansible/latest/collections/community/docker/docker_connection.html#ansible-collections-community-docker-docker-connection`

> Now I am in trouble to continue.

Try to replicate the podman provision plugin to docker one, it should not be hard.

>  Is it the same behavior for VMs?

Yes, it is.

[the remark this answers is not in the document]
ZhukovGreen: @thrix Thx Miroslav, I understand your point, I'll try to workaround it somehow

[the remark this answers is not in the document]
thrix: thanks for the discussion, it is good feedback to have!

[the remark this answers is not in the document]
ZhukovGreen: OK, I was able to achieve the possibility to run playbooks inside the container. But without TMT yet. It seems podman plugin doesn't work as expected and runs the playbook against the host rather than against the container. So preparation in this sense relates to the host and not to the container. Submitted a ticket https://github.com/psss/tmt/issues/621

ZhukovGreen:
No longer valid. Closing

PR #196, opened 2021-03-03
add: check function to validate custom repos

fellipeh:
When using custom repos but the repoid passed through --enablerepo is not valid (typo in the repoid, the baseurl not accessible, etc.), convert2rhel continues without a problem beyond the point of no return (PONR) and then it fails, leaving the user with a message that they need to fix their system manually.

This PR check the validity os repos before anything.

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -22,8 +22,9 @@ import os
 import re
 import subprocess
 
-from convert2rhel.pkghandler import get_installed_pkg_objects, get_pkg_fingerprint
+from convert2rhel.pkghandler import call_yum_cmd, get_installed_pkg_objects, get_pkg_fingerprint
 from convert2rhel.systeminfo import system_info
+from convert2rhel.toolopts import tool_opts
 from convert2rhel.utils import get_file_content, run_subprocess
 
 
@@ -48,6 +49,7 @@ def perform_pre_checks():
     check_tainted_kmods()
     check_readonly_mounts()
     check_rhel_compatible_kernel_is_used()
+    check_custom_repos_are_valid()
 
 
 def check_uefi():
@@ -118,6 +120,34 @@ def check_readonly_mounts():
 def perform_pre_ponr_checks():
     """Late checks before ponr should be added here."""
     ensure_compatibility_of_kmods()
+
+
+def check_custom_repos_are_valid():
+    """To prevent failures past the PONR, make sure that the enabled custom repositories are valid.
+
+    What is meant by valid:
+    - YUM/DNF is able to find the repoids (to rule out a typo)
+    - the repository "baseurl" is accessible and contains repository metadata
+    """
+    logger.task("Prepare: Checking if --enablerepo repositories are accessible")
+
+    if not tool_opts.disable_submgr:
+        logger.info("Skipping the check of repositories due to the use of RHSM for the conversion.")
+        return
+
+    # Without clearing the metadata cache, the `yum makecache` command may return 0 (everything's ok) even when
+    # the baseurl of a repository is not accessible. That would happen when the repository baseurl is changed but yum
+    # still uses the previous baseurl stored in its cache.
+    call_yum_cmd(command="clean", args="metadata")
+
+    output, ret_code = call_yum_cmd(command="makecache", args="-v --setopt=*.skip_if_unavailable=False")
+    if ret_code != 0:
+        logger.critical(
+            "Unable to access the repositories passed through the --enablerepo option. "
+            "See the above YUM/DNF output for more details."
+        )
+
+    logger.info("The repositories passed through the --enablerepo option are all accessible.")
 
 
 def ensure_compatibility_of_kmods():
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -39,6 +39,8 @@ from convert2rhel.checks import (
     perform_pre_ponr_checks,
 )
 from convert2rhel.pkghandler import get_pkg_fingerprint
+from convert2rhel.systeminfo import system_info
+from convert2rhel.toolopts import tool_opts
 from convert2rhel.unit_tests import GetFileContentMocked, GetLoggerMocked
 from convert2rhel.utils import run_subprocess
 
@@ -114,6 +116,7 @@ def test_perform_pre_checks(monkeypatch):
     check_thirdparty_kmods_mock = mock.Mock()
     check_uefi_mock = mock.Mock()
     check_readonly_mounts_mock = mock.Mock()
+    check_custom_repos_are_valid_mock = mock.Mock()
     check_rhel_compatible_kernel_is_used_mock = mock.Mock()
     monkeypatch.setattr(
         checks,
@@ -134,6 +137,11 @@ def test_perform_pre_checks(monkeypatch):
         checks,
         "check_rhel_compatible_kernel_is_used",
         value=check_rhel_compatible_kernel_is_used_mock,
+    )
+    monkeypatch.setattr(
+        checks,
+        "check_custom_repos_are_valid",
+        value=check_custom_repos_are_valid_mock,
     )
 
     perform_pre_checks()
@@ -652,3 +660,42 @@ class TestReadOnlyMountsChecks(unittest.TestCase):
         self.assertTrue(
             "Stopping conversion due to read-only mount to /sys directory" not in checks.logger.critical_msgs[0]
         )
+        self.assertTrue("/sys mount point is not read-only." in checks.logger.debug_msgs[0])
+
+    class CallYumCmdMocked(unit_tests.MockFunction):
+        def __init__(self, ret_code, ret_string):
+            self.called = 0
+            self.return_code = ret_code
+            self.return_string = ret_string
+            self.fail_once = False
+            self.command = None
+
+        def __call__(self, command, *args, **kwargs):
+            if self.fail_once and self.called == 0:
+                self.return_code = 1
+            if self.fail_once and self.called > 0:
+                self.return_code = 0
+            self.called += 1
+            self.command = command
+            return self.return_string, self.return_code
+
+    @unit_tests.mock(system_info, "version", namedtuple("Version", ["major", "minor"])(7, 0))
+    @unit_tests.mock(checks, "call_yum_cmd", CallYumCmdMocked(ret_code=0, ret_string="Abcdef"))
+    @unit_tests.mock(checks, "logger", GetLoggerMocked())
+    @unit_tests.mock(tool_opts, "disable_submgr", True)
+    def test_custom_repos_are_valid(self):
+        checks.check_custom_repos_are_valid()
+        self.assertEqual(len(checks.logger.info_msgs), 1)
+        self.assertTrue(
+            "The repositories passed through the --enablerepo option are all accessible." in checks.logger.info_msgs
+        )
+
+    @unit_tests.mock(system_info, "version", namedtuple("Version", ["major", "minor"])(7, 0))
+    @unit_tests.mock(checks, "call_yum_cmd", CallYumCmdMocked(ret_code=1, ret_string="Abcdef"))
+    @unit_tests.mock(checks, "logger", GetLoggerMocked())
+    @unit_tests.mock(tool_opts, "disable_submgr", True)
+    def test_custom_repos_are_invalid(self):
+        self.assertRaises(SystemExit, checks.check_custom_repos_are_valid)
+        self.assertEqual(len(checks.logger.critical_msgs), 1)
+        self.assertEqual(len(checks.logger.info_msgs), 0)
+        self.assertTrue("Unable to access the repositories passed through " in checks.logger.critical_msgs[0])
--- a/tests/integration/check-custom-repo/test_custom-repos.py
+++ b/tests/integration/check-custom-repo/test_custom-repos.py
@@ -47,8 +47,6 @@ def test_good_convertion_without_rhsm(shell, convert2rhel):
         # send Ctrl-C
         c2r.send(chr(3))
 
-    assert c2r.exitstatus == 0
-
 
 @pytest.mark.bad_tests
 def test_bad_convertion_without_rhsm(shell, convert2rhel):
@@ -59,7 +57,7 @@ def test_bad_convertion_without_rhsm(shell, convert2rhel):
     with convert2rhel("-y --no-rpm-va --disable-submgr --enablerepo fake-rhel-8-for-x86_64-baseos-rpms --debug") as c2r:
         c2r.expect(
             "CRITICAL - Unable to access the repositories passed through the --enablerepo option. See the above "
-            "YUM/DNF output for more details. "
+            "YUM/DNF output for more details."
         )
 
     assert c2r.exitstatus == 1
--- a/tests/integration/check-custom-repo/test_custom-repos.py
+++ b/tests/integration/check-custom-repo/test_custom-repos.py
@@ -1,0 +1,65 @@
+import re
+
+from collections import namedtuple
+
+import pytest
+
+
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+
+
+def get_system_version(system_release_content=None):
+    """Return a namedtuple with major and minor elements, both of an int type.
+
+    Examples:
+    Oracle Linux Server release 6.10
+    Oracle Linux Server release 7.8
+    CentOS release 6.10 (Final)
+    CentOS Linux release 7.6.1810 (Core)
+    CentOS Linux release 8.1.1911 (Core)
+    """
+    match = re.search(r".+?(\d+)\.(\d+)\D?", system_release_content)
+    if not match:
+        return "not match"
+    version = namedtuple("Version", ["major", "minor"])(int(match.group(1)), int(match.group(2)))
+
+    return version
+
+
+@pytest.mark.good_tests
+def test_good_convertion_without_rhsm(shell, convert2rhel):
+    """
+    Test if --enablerepo are not skipped when  subscription-manager are disabled and test if repo passed are valid.
+    """
+    with open("/etc/system-release", "r") as file:
+        system_release = file.read()
+        system_version = get_system_version(system_release_content=system_release)
+        if system_version.major == 7:
+            enable_repo_opt = "--enablerepo rhel-7-server-rpms"
+        elif system_version.major == 8:
+            enable_repo_opt = "--enablerepo rhel-8-for-x86_64-baseos-rpms --enablerepo rhel-8-for-x86_64-appstream-rpms"
+
+    with convert2rhel("-y --no-rpm-va --disable-submgr {} --debug".format(enable_repo_opt)) as c2r:
+        c2r.expect("The repositories passed through the --enablerepo option are all accessible.")
+        # send Ctrl-C
+        c2r.send(chr(3))
+
+    assert c2r.exitstatus == 0
+
+
+@pytest.mark.bad_tests
+def test_bad_convertion_without_rhsm(shell, convert2rhel):
+    """
+    Test if --enablerepo are not skipped when  subscription-manager are disabled and test the convertion will stop
+    with non-valid repo
+    """
+    with convert2rhel("-y --no-rpm-va --disable-submgr --enablerepo fake-rhel-8-for-x86_64-baseos-rpms --debug") as c2r:
+        c2r.expect(
+            "CRITICAL - Unable to access the repositories passed through the --enablerepo option. See the above "
+            "YUM/DNF output for more details. "
+        )
+
+    assert c2r.exitstatus == 1
--- a/plans/integration/check-custom-repo/main.fmf
+++ b/plans/integration/check-custom-repo/main.fmf
@@ -1,0 +1,7 @@
+discover+:
+    filter:
+        - 'tag: check-custom-repo'
+
+provision:
+    how: libvirt
+    develop: true
--- a/plans/integration/check-custom-repo/vm_centos7.fmf
+++ b/plans/integration/check-custom-repo/vm_centos7.fmf
@@ -1,0 +1,14 @@
+discover+:
+    filter+:
+        - 'tag: centos7'
+provision+:
+    origin_vm_name: c2r_centos7_template
+
+/good:
+    discover+:
+        filter+:
+            - 'tag: good_test'
+/bad:
+    discover+:
+        filter+:
+            - 'tag: bad_test'
--- a/plans/integration/check-custom-repo/vm_centos8.fmf
+++ b/plans/integration/check-custom-repo/vm_centos8.fmf
@@ -1,0 +1,14 @@
+discover+:
+    filter+:
+        - 'tag: centos8'
+provision+:
+    origin_vm_name: c2r_centos8_template
+
+/good:
+    discover+:
+        filter+:
+            - 'tag: good_test'
+/bad:
+    discover+:
+        filter+:
+            - 'tag: bad_test'
--- a/plans/integration/check-custom-repo/vm_oracle7.fmf
+++ b/plans/integration/check-custom-repo/vm_oracle7.fmf
@@ -1,0 +1,14 @@
+discover+:
+    filter+:
+        - 'tag: oracle7'
+provision+:
+    origin_vm_name: c2r_oracle7_template
+
+/good:
+    discover+:
+        filter+:
+            - 'tag: good_test'
+/bad:
+    discover+:
+        filter+:
+            - 'tag: bad_test'
--- a/plans/integration/check-custom-repo/vm_oracle8.fmf
+++ b/plans/integration/check-custom-repo/vm_oracle8.fmf
@@ -1,0 +1,14 @@
+discover+:
+    filter+:
+        - 'tag: oracle8'
+provision+:
+    origin_vm_name: c2r_oracle8_template
+
+/good:
+    discover+:
+        filter+:
+            - 'tag: good_test'
+/bad:
+    discover+:
+        filter+:
+            - 'tag: bad_test'
--- a/tests/integration/check-custom-repo/main.fmf
+++ b/tests/integration/check-custom-repo/main.fmf
@@ -10,10 +10,10 @@ tag+:
   tag+:
     - good_test
   test: |
-    pytest -svv --tb=no -m good_tests
+    pytest -svv -m good_tests
 
 /bad:
   tag+:
     - bad_test
   test: |
-    pytest -svv --tb=no -m bad_tests
+    pytest -svv -m bad_tests
--- a/tests/integration/check-custom-repo/main.fmf
+++ b/tests/integration/check-custom-repo/main.fmf
@@ -1,0 +1,19 @@
+summary: check-custom-repo
+tag+:
+  - centos7
+  - centos8
+  - oracle7
+  - oracle8
+  - check-custom-repo
+
+/good:
+  tag+:
+    - good_test
+  test: |
+    pytest -svv --tb=no -m good_tests
+
+/bad:
+  tag+:
+    - bad_test
+  test: |
+    pytest -svv --tb=no -m bad_tests


convert2rhel/checks.py
```diff
@@ -0,0 +1,35 @@
...
+from convert2rhel.pkghandler import call_yum_cmd
+
+logger = logging.getLogger(__name__)
+
+
+def check_repos_are_valid():
+    output, ret_code = call_yum_cmd(command="repolist", args="-v")
+    if "failure" in output or ret_code != 0:
+        logger.critical("Couldn't validate custom repositories. Please check custom repo config files.")
+    logger.debug("All custom repo appears to be valid.")
```
SpyTec, line 30: We need to validate that the repos the user has passed with `--enablerepo` is a valid repo, not that all the repos are valid.

Adding to that, this is the output I get with misconfigured URL when running `repolist`, the check wouldn't catch that
```
[root@c2r-20210305121855 ~]# yum repolist -v
Failed to set locale, defaulting to C
Loading "product-id" plugin
Loading "search-disabled-repos" plugin
Loading "subscription-manager" plugin
Updating Subscription Management repositories.
Config time: 3.201
Yum Version: 3.2.29
Repository rhel-6-server-rpms is listed more than once in the configuration
Repository rhel-6-server-optional-rpms is listed more than once in the configuration
Repository rhel-6-server-extras-rpms is listed more than once in the configuration
https://download.caopr.fedorainfracloud.org/results/%40oamg/convert2rhel/epel-6-x86_64/repodata/repomd.xml: [Errno 14] PYCURL ERROR 6 - "Couldn't resolve host 'badurl.copr.fedorainfracloud.org'"
Trying other mirror.
Setting up Package Sacks
pkgsack time: 0.040
Repo-id      : copr:copr.fedorainfracloud.org:group_oamg:convert2rhel
Repo-name    : Copr repo for convert2rhel owned by @oamg
Repo-revision: 1614936293
Repo-updated : Fri Mar  5 10:24:53 2021
Repo-pkgs    : 54
Repo-size    : 5.0 M
Repo-baseurl : https://badurl.copr.fedorainfracloud.org/results/@oamg/convert2rhel/epel-6-x86_64/
Repo-expire  : 21600 second(s) (last: Fri Mar  5 11:21:47 2021)
Repo-excluded: 54

Repo-id      : rhel-6-server-rpms
Repo-name    : Red Hat Enterprise Linux 6 Server (RPMs)
Repo-revision: 1609129994
Repo-updated : Mon Dec 28 05:33:13 2020
Repo-pkgs    : 21210
Repo-size    : 48 G
Repo-baseurl : https://cdn.redhat.com/content/dist/rhel/server/6/6Server/x86_64/os
Repo-expire  : 86400 second(s) (last: Fri Mar  5 11:26:57 2021)

repolist: 21264

[root@c2r-20210305121855 ~]# echo $?
0

```

convert2rhel/checks.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def perform_pre_checks():
+    logger.task("Prepare: Checking if custom repos are valid")
```
pirat89, line 34: @fellipeh  the log message should be part of the specific check. this function should just call each *check function* in the file

convert2rhel/checks.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright(C) 2016 Red Hat, Inc.
```
ZhukovGreen, line 3: Nit: we need to update the copyright year. Not related to the scope of this MR, just for reference

convert2rhel/checks.py
ZhukovGreen: Agree with @pirat89, now I see that the only exception is when we're doing rollbacks. Otherwise all the logger.task are resides in main.py

ZhukovGreen: https://github.com/oamg/convert2rhel/search?q=loggerinst.task

convert2rhel/main.py
```diff
@@ -30,9 +30,9 @@
     systeminfo,
     toolopts,
     utils,
+    checks
```
ZhukovGreen, line 33: Nit: isort!

convert2rhel/checks.py
ZhukovGreen: ref #208

convert2rhel/checks.py
fellipeh: Hi @j-mracek Did we have some way to check a specific repo using yum ?

j-mracek: Please could you provide additional information. Did you mean yum in RHEL7 or RHEL8? It is you but two different component.
What do you suggest as a check? Repo is well defined, available on server, data are downloadable or something else?

fellipeh: @j-mracek
I need to validate if individual repo are valid, in other word: if I can use this repo for download and install packages, for example: we had 6 repos listed on config file, and I need to test if one of that are valid, and I can use to download/install packages.

fellipeh: Using `makecache --setopt=reposdir=/etc/yum.repos.d -v --setopt=*.skip_if_unavailable=False` as @j-mracek suggestion

fellipeh:
@bocekm @SpyTec @ZhukovGreen rebase and ready for review!

convert2rhel/checks.py
```diff
@@ -293,7 +295,17 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
...
+            - set(system_info.kmods_to_ignore)
     )
+
+
+def check_repos_are_valid():
+    """Check if all used repositories are valid."""
+    output, ret_code = call_yum_cmd(command="makecache",
+                                    args="--setopt=reposdir=%s -v --setopt=*.skip_if_unavailable=False" % YUM_REPO_DIR)
+    errors_strs = ['error', 'err', 'failure']
+    if ret_code != 0 or any(err in output.lower() for err in errors_strs):
```
bocekm, line 309: @fellipeh, have you discussed with @j-mracek the meaning of return codes of dnf makecache? I.e. is it safe to assume that return code 0 means that the repo is accessible and any other return code means there's a problem?

Also, is there a particular reason for looking for strings ['error', 'err', 'failure'] in the makecache command output when it returns 0?

convert2rhel/checks.py
```diff
@@ -293,7 +295,17 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
...
-        - set(system_info.kmods_to_ignore)
+            host_kmods
+            - rhel_supported_kmods
+            - set(system_info.kmods_to_ignore)
     )
+
+
+def check_repos_are_valid():
+    """Check if all used repositories are valid."""
+    output, ret_code = call_yum_cmd(command="makecache",
```
bocekm, line 306: Have you checked that the same command works identically on CentOS 6/7/8? I.e. that the command works not only with dnf but with yum as well.

bocekm (changes requested):

convert2rhel/checks.py
fellipeh: I just cheking the list of errors string to be more care, if we have any error but escape 0 as Eric show us above.

convert2rhel/checks.py
fellipeh: Tested with yum without issue.

convert2rhel/checks.py
```diff
@@ -293,7 +295,17 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
...
+                                    args="--setopt=reposdir=%s -v --setopt=*.skip_if_unavailable=False" % YUM_REPO_DIR)
```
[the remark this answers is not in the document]
fellipeh: You have point.. change the code here, removing this `reposdir` and same result, so.. I'll send a new commit with this change,.

convert2rhel/checks.py
```diff
@@ -22,21 +22,28 @@
...
 def perform_pre_checks():
     """Early checks after system facts should be added here."""
     check_uefi()
     check_tainted_kmods()
+    check_repos_are_valid()
+
+
+def perform_pre_ponr_checks():
+    """Late checks before ponr should be added here."""
+    ensure_compatibility_of_kmods()
```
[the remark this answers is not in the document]
fellipeh: I just move it to the top. seams to be better to see all _performs_ functions near it self.

convert2rhel/checks.py
bocekm: That was a different command that was printing out errors even when returning 0. The output of `makecache` may be different.

In my testing, when making the baseurl incorrect, I'm getting this on CentOS 7:
```
[root@c2r-20210408201227 ~]# yum makecache -v --setopt=*.skip_if_unavailable=False --disablerepo="*" --enablerepo=base
Loading "fastestmirror" plugin
Config time: 0.006
Yum version: 3.4.3
Loading mirror speeds from cached hostfile
Could not retrieve mirrorlist http://badurl/ error was
14: curl#6 - "Could not resolve host: badurl; Unknown error"
 * base: mirror.karneval.cz
base                                                                                                                                                                                                                   | 3.6 kB  00:00:00     
(1/2): base/other_db                                                                                                                                                                                                   | 2.6 MB  00:00:02     
(2/2): base/filelists_db                                                                                                                                                                                               | 7.2 MB  00:00:03     
Metadata Cache Created
[root@c2r-20210408201227 ~]# echo $?
0
```
The reason for that is that it uses cached metadata. So, rather than relying on a certain keywords we just assume might appear, it seems to me better to clean the yum cache first. Then every inaccessible url or non-existing repoid ends up with non-0 return code.
```
[root@c2r-20210408201227 ~]# yum clean metadata
Loaded plugins: fastestmirror
Cleaning repos: base copr:copr.fedorainfracloud.org:group_oamg:convert2rhel extras updates
12 metadata files removed
12 sqlite files removed
0 metadata files removed
[root@c2r-20210408201227 ~]# yum makecache -v --setopt=*.skip_if_unavailable=False --disablerepo="*" --enablerepo=base
Loading "fastestmirror" plugin
Config time: 0.005
Yum version: 3.4.3
Loading mirror speeds from cached hostfile
Could not retrieve mirrorlist http://badurl/ error was
14: curl#6 - "Could not resolve host: badurl; Unknown error"


 One of the configured repositories failed (Unknown),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:

     1. Contact the upstream for the repository and get them to fix the problem.

     2. Reconfigure the baseurl/etc. for the repository, to point to a working
        upstream. This is most often useful if you are using a newer
        distribution release than is supported by the repository (and the
        packages for the previous distribution release still work).

     3. Run the command with the repository temporarily disabled
            yum --disablerepo=<repoid> ...

     4. Disable the repository permanently, so yum won't use it by default. Yum
        will then just ignore the repository until you permanently enable it
        again or use --enablerepo for temporary usage:

            yum-config-manager --disable <repoid>
        or
            subscription-manager repos --disable=<repoid>

     5. Configure the failing repository to be skipped, if it is unavailable.
        Note that yum will try to contact the repo. when it runs most commands,
        so will have to try and fail each time (and thus. yum will be be much
        slower). If it is a very temporary problem though, this is often a nice
        compromise:

            yum-config-manager --save --setopt=<repoid>.skip_if_unavailable=true

Cannot find a valid baseurl for repo: base
```

convert2rhel/checks.py
```diff
@@ -293,7 +294,15 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
...
+            - set(system_info.kmods_to_ignore)
     )
+
+
+def check_repos_are_valid():
+    """Check if all used repositories are valid."""
+    output, ret_code = call_yum_cmd(command="makecache", args="-v --setopt=*.skip_if_unavailable=False")
+    if ret_code != 0:
+        logger.critical("Couldn't validate custom repositories. Please check custom repo config files.")
+    logger.debug("All custom repo appears to be valid.")
```
bocekm, line 308:
```suggestion
    logger.debug("The repositories passed through the --enablerepo option are all accessible.")
```

convert2rhel/checks.py
```diff
@@ -293,7 +294,15 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
...
-        - set(system_info.kmods_to_ignore)
+            host_kmods
+            - rhel_supported_kmods
+            - set(system_info.kmods_to_ignore)
     )
+
+
+def check_repos_are_valid():
+    """Check if all used repositories are valid."""
+    output, ret_code = call_yum_cmd(command="makecache", args="-v --setopt=*.skip_if_unavailable=False")
```
bocekm, line 305:
```suggestion
    logger.task("Prepare: Checking if custom repositories are accessible")
    output, ret_code = call_yum_cmd(command="makecache", args="-v --setopt=*.skip_if_unavailable=False")
```

convert2rhel/checks.py
```diff
@@ -293,7 +294,15 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
...
-        host_kmods
-        - rhel_supported_kmods
-        - set(system_info.kmods_to_ignore)
+            host_kmods
+            - rhel_supported_kmods
+            - set(system_info.kmods_to_ignore)
     )
+
+
+def check_repos_are_valid():
```
bocekm, line 303: This function runs under all conditions. That is even when `--disable-submgr` is not used. Didn't you want to check only custom repos?

bocekm (changes requested):

convert2rhel/unit_tests/checks_test.py
```diff
@@ -137,34 +140,34 @@ def test_pre_ponr_checks(monkeypatch):
...
-            "Kernel modules are compatible",
-            None,
-        ),
-        (
-            HOST_MODULES_STUB_BAD,
-            SystemExit,
-            None,
-            "Kernel modules are compatible",
-        ),
+            (
```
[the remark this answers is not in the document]
fellipeh: That's weird, my IDE formatter made this...

convert2rhel/checks.py
fellipeh: I already remove the str checks and check the return code.

convert2rhel/checks.py
```diff
@@ -293,7 +310,7 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
     Ignore certain kmods mentioned in the system configs. These kernel modules moved to kernel core, meaning that the
     functionality is retained and we would be incorrectly saying that the modules are not supported in RHEL."""
     return (
-        host_kmods
-        - rhel_supported_kmods
-        - set(system_info.kmods_to_ignore)
+            host_kmods
+            - rhel_supported_kmods
+            - set(system_info.kmods_to_ignore)
```
SpyTec, line 315: Seems like IDE indented this too much

convert2rhel/checks.py
```diff
@@ -83,6 +84,22 @@ def perform_pre_ponr_checks():
...
 
+def check_custom_repos_are_valid():
+    """To prevent failures past the PONR, make sure that the enabled custom repositories are valid.
+
+    What is meant by valid:
+    - YUM/DNF is able to find the repoids (to rule out a typo)
+    - the repository "baseurl" is accessible and contains repository metadata
+    """
+    logger.task("Prepare: Checking if custom repositories are accessible")
+    output, ret_code = call_yum_cmd(command="makecache", args="-v --setopt=*.skip_if_unavailable=False")
```
SpyTec, line 95: Clear metadata first as Michal mentioned

`yum clean metadata`

convert2rhel/checks.py
```diff
@@ -83,6 +84,22 @@ def perform_pre_ponr_checks():
...
 
 
+def check_custom_repos_are_valid():
+    """To prevent failures past the PONR, make sure that the enabled custom repositories are valid.
+
+    What is meant by valid:
+    - YUM/DNF is able to find the repoids (to rule out a typo)
+    - the repository "baseurl" is accessible and contains repository metadata
+    """
+    logger.task("Prepare: Checking if custom repositories are accessible")
```
SpyTec, line 94: Better to say we're checking `--enablerepo` explicitly than "custom repos", as it can mean pretty much all repos including those we don't verify

```suggestion
    logger.task("Prepare: Checking if --enablerepo repositories are accessible")
```

SpyTec (changes requested):

convert2rhel/checks.py
fellipeh: Yeah.. PyCharm goes crazy, I'll investigate it..

convert2rhel/checks.py
bocekm: Yeah, this is better.
I initially thought it's not enough accurate because one can use --enablerepo also with RHSM to override the default RHSM repos we enable based on our config settings, and we do not validate these overriding RHSM repos, but we don't need to explain 100% everything in the `task` level log message :)

convert2rhel/checks.py
```diff
@@ -83,6 +84,23 @@ def perform_pre_ponr_checks():
...
+    - the repository "baseurl" is accessible and contains repository metadata
+    """
+    logger.task("Prepare: Checking if --enablerepo repositories are accessible")
+    call_yum_cmd(command="clean", args="metadata")
+    output, ret_code = call_yum_cmd(command="makecache", args="-v --setopt=*.skip_if_unavailable=False")
+    if ret_code != 0:
+        logger.critical(
+            "Unable to access the repositories passed through the --enablerepo option. "
+            "See the above YUM/DNF output for more details.")
+    logger.debug("The repositories passed through the --enablerepo option are all accessible.")
```
bocekm, line 101:
```suggestion
    logger.info("The repositories passed through the --enablerepo option are all accessible.")
```

convert2rhel/checks.py
bocekm: Not resolved yet.
Please add a condition right after the line `logger.task("Prepare: Checking if --enablerepo repositories are accessible")` that would print an `info` message and return from the function.

Something like:
```
if not tool_opts.disable_submgr:
    logger.info("Skipping the check of repositories due to the use of RHSM for the conversion.")
    return
```

ZhukovGreen (commented):
Looks good to me, but I was not tested it. Let's write verification tests together.

It seems your formatted went mad:))))

convert2rhel/checks.py
```diff
@@ -83,6 +84,23 @@ def perform_pre_ponr_checks():
...
 
+def check_custom_repos_are_valid():
+    """To prevent failures past the PONR, make sure that the enabled custom repositories are valid.
+
+    What is meant by valid:
+    - YUM/DNF is able to find the repoids (to rule out a typo)
+    - the repository "baseurl" is accessible and contains repository metadata
+    """
+    logger.task("Prepare: Checking if --enablerepo repositories are accessible")
+    call_yum_cmd(command="clean", args="metadata")
```
SpyTec, line 95: Modified Michal's log suggestion a bit:
```suggestion
    logger.task("Prepare: Checking if --enablerepo repositories are accessible")

    if not tool_opts.disable_submgr:
        logger.info("Skipping the check of repositories due to RHSM being used for the conversion.")
        return

    call_yum_cmd(command="clean", args="metadata")
```

SpyTec (changes requested):

convert2rhel/checks.py
```diff
@@ -83,6 +85,30 @@ def perform_pre_ponr_checks():
...
+    - YUM/DNF is able to find the repoids (to rule out a typo)
+    - the repository "baseurl" is accessible and contains repository metadata
+    """
+    logger.task("Prepare: Checking if --enablerepo repositories are accessible")
+
+    if not tool_opts.disable_submgr:
+        logger.info("Skipping the check of repositories due to the use of RHSM for the conversion.")
+        return
+
+    call_yum_cmd(command="clean", args="metadata")
```
bocekm, line 101:
```suggestion
    # Without clearing the metadata cache, the `yum makecache` command may return 0 (everything's ok) even when
    # the baseurl of a repository is not accessible. That would happen when the repository baseurl is changed but yum
    # still uses the previous baseurl stored in its cache.
    call_yum_cmd(command="clean", args="metadata")
```

bocekm (changes requested):

bocekm (dismissed):
Code looks ok. It needs integration test before merging.

bocekm:
@fellipeh, please rebase to resolve conflicts.

SpyTec (dismissed):

tests/ansible_collections/basic_setup.yml
```diff
@@ -3,6 +3,8 @@
     # to be overridden by the tmt
     repo_root:
   tasks:
+    # including rhel internal repos for tests purposes
+    - include: copy_internal_rhel_repos.yml
```
ZhukovGreen, line 7: this is very cool. I had some problems in the past to include other playbooks. This means we can use Ansible roles with vars

tests/ansible_collections/copy_internal_rhel_repos.yml
```diff
@@ -0,0 +1,20 @@
+- name: Copy repofile with internal RHEL 6 repositories
+  copy:
+    src: 'rhel-6-internal.repo'
+    dest: '/etc/yum.repos.d'
+  when: (ansible_facts['distribution'] == "CentOS" and ansible_facts['distribution_major_version'] == "6") or
+        (ansible_facts['distribution'] == "OracleLinux" and ansible_facts['distribution_major_version'] == "6")
+
```
ZhukovGreen, line 7: No need to have 6 centos oracle.

tests/integration/check-custom-repo/test_custom-repos.py
```diff
@@ -0,0 +1,53 @@
...
+                "--debug"
+            ).format(
+                env.str("RHSM_SERVER_URL"),
+                env.str("RHSM_USERNAME"),
+                env.str("RHSM_PASSWORD"),
+                env.str("RHSM_POOL"),
+            )
+        ]
+    )
+    assert convertion.returncode == 1
```
ZhukovGreen, line 52: Let's also capture the log message, so we'll be sure that the conversion fails exactly due to fake-rhel-server-rpms

ZhukovGreen (commented):
Looks great to me, apart from a couple of nits. Excellent work, Fellipe!

ZhukovGreen:
let me try to run integration tests

tests/integration/check-custom-repo/test_custom-repos.py
```diff
@@ -0,0 +1,53 @@
...
+                env.str("RHSM_PASSWORD"),
+                env.str("RHSM_POOL"),
+            )
+        ]
+    )
+    assert convertion.returncode == 0
+
+
+@pytest.mark.bad_tests
+def test_with_fake_repo_convertion(shell):
```
ZhukovGreen, line 35: I got:
```
            out: [05/04/2021 16:14:19] TASK - [Prepare: Checking if --enablerepo repositories are accessible] ****
            out: 
            out: Skipping the check of repositories due to the use of RHSM for the conversion.
            out: 
```
This is not expected, right?

ZhukovGreen: And it is inhibited, but not at the check level:
```
[05/04/2021 16:16:56] DEBUG - Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 86, in main
    pre_ponr_conversion()
  File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 189, in pre_ponr_conversion
    subscription.enable_repos(rhel_repoids)
  File "/usr/lib/python3.6/site-packages/convert2rhel/subscription.py", line 394, in enable_repos
    " subscription-manager:\n%s" % output)
  File "/usr/lib/python3.6/site-packages/convert2rhel/logger.py", line 103, in _critical
    sys.exit(msg)
SystemExit: Repos were not possible to enable through subscription-manager:
Error: 'fake-rhel-server-rpms' does not match a valid repository ID. Use "subscription-manager repos --list" to see valid repositories.
```

tests/integration/check-custom-repo/test_custom-repos.py
```diff
@@ -0,0 +1,53 @@
...
+
+from envparse import env
+
+
+@pytest.mark.good_tests
+def test_good_convertion(shell):
+    convertion = shell(
+        command=[
+            (
+                "convert2rhel -y "
```
ZhukovGreen, line 17: You probably should use another command:
```
convert2rhel --disable-submgr --disablerepo "*" --enablerepo rhel-{ version here}-server-rpms --debug
```

ZhukovGreen: Yes, in your implementation you have:
```
  if not tool_opts.disable_submgr:
        logger.info("Skipping the check of repositories due to the use of RHSM for the conversion.")
        return
 ```
So you're skipping the check if rhsm not disabled

**Review by ZhukovGreen — COMMENTED** [review]
Captured significant things during the running the tmt tests

File: tests/integration/check-custom-repo/test_custom-repos.py [hunk 24]
```diff
@@ -0,0 +1,53 @@
...
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+
+from envparse import env
+
+
+@pytest.mark.good_tests
+def test_good_convertion(shell):
```
ZhukovGreen, line 13: When running the test fails with:
```
[05/04/2021 16:21:54] DEBUG - Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 86, in main
    pre_ponr_conversion()
  File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 189, in pre_ponr_conversion
    subscription.enable_repos(rhel_repoids)
  File "/usr/lib/python3.6/site-packages/convert2rhel/subscription.py", line 394, in enable_repos
    " subscription-manager:\n%s" % output)
  File "/usr/lib/python3.6/site-packages/convert2rhel/logger.py", line 103, in _critical
    sys.exit(msg)
SystemExit: Repos were not possible to enable through subscription-manager:
Error: 'rhel-server-rpms' does not match a valid repository ID. Use "subscription-manager repos --list" to see valid repositories.
```

ZhukovGreen (commented):
added output of the good test

tests/integration/check-custom-repo/test_custom-repos.py
fellipeh: Yes, it is expected since we are running with subscription manager.

ZhukovGreen:
@fellipeh please remove repofiles and rebase

fellipeh:
Integration tests passed:
```
centos7   passed
centos8   passed
ol7       passed
ol8       passed
```

tests/integration/check-custom-repo/test_custom-repos.py
```diff
@@ -0,0 +1,84 @@
...
+    match = re.search(r".+?(\d+)\.(\d+)\D?", system_release_content)
+    if not match:
+        return "not match"
+    version = namedtuple("Version", ["major", "minor"])(int(match.group(1)), int(match.group(2)))
+
+    return version
+
+
+@pytest.mark.good_tests
+def test_good_convertion_with_rhsm(shell, convert2rhel):
```
ZhukovGreen, line 37: Is it ok to test your feature good scenario with the rhsm?

bocekm: I'd say it can be removed. The check of the repo validity is skipped with the use of RHSM.

tests/integration/check-custom-repo/test_custom-repos.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Test if --enablerepo are not skipped when  subscription-manager are disabled and test if repo passed are valid.
+    """
+    with open("/etc/system-release", "r") as file:
+        system_release = file.read()
+        system_version = get_system_version(system_release_content=system_release)
+        if system_version.major == 7:
+            repo = "rhel-7-server-rpms"
+        elif system_version.major == 8:
+            repo = "rhel-8-for-x86_64-baseos-rpms"
```
bocekm, line 63: How does one get the repo to the testing VM?

fellipeh: checking the `/etc/system-release` file:
```python
    with open("/etc/system-release", "r") as file:
        system_release = file.read()
        system_version = get_system_version(system_release_content=system_release)
        if system_version.major == 7:
            repo = "rhel-7-server-rpms"
        elif system_version.major == 8:
            repo = "rhel-8-for-x86_64-baseos-rpms"
```

bocekm: But the repofile should be specified somewhere. As a secret in .env file? How should the definition look like?

fellipeh: After a quick call with @ZhukovGreen, we realize the tests are using our vagrantbox, and we have the repofiles there.

bocekm: I see. Unfortunately the standard vagrant setup described in the [README](https://github.com/oamg/convert2rhel/blob/main/tests/integration/README.md) does not include the repofile definition. One option is making it part of the `.env` file https://github.com/oamg/convert2rhel/blob/main/.env.example. @ZhukovGreen, WDYT?

ZhukovGreen: Actually, it does. We are doing vagrant up and our vagrant ansible setup includes role, which adds custom repos. So technically we don't need this step in this MR. However, I just realized that it is going to work just for our local setup. On TFT it won't.

So let's add custom repos, but as a separate MR and into https://github.com/oamg/convert2rhel/tree/main/tests/ansible_collections/roles

tests/integration/check-custom-repo/test_custom-repos.py
```diff
@@ -0,0 +1,67 @@
...
+
+    with convert2rhel("-y --no-rpm-va --disable-submgr --enablerepo {} --debug".format(repo)) as c2r:
```
bocekm, line 49:
```suggestion
            enable_repo_opt = "--enablerepo rhel-7-server-rpms"
        elif system_version.major == 8:
            enable_repo_opt = "--enablerepo rhel-8-for-x86_64-baseos-rpms --enablerepo rhel-8-for-x86_64-appstream-rpms"

    with convert2rhel("-y --no-rpm-va --disable-submgr {} --debug".format(enable_repo_opt)) as c2r:
```
I'm surprised that the conversion passed with only BaseOS enabled. It used to be necessary to enable both BaseOS and AppStream.

tests/integration/check-custom-repo/test_custom-repos.py
```diff
@@ -0,0 +1,67 @@
...
+        c2r.expect("Conversion successful!")
```
bocekm, line 50:
```suggestion
        c2r.expect(""The repositories passed through the --enablerepo option are all accessible.")
        # send Ctrl-C
        c2r.send(chr(3))
```
You can make the tests faster this way.

bocekm (changes requested):

tests/integration/check-custom-repo/test_custom-repos.py
bocekm: But if you're going to make the test faster by sending Ctrl+C, it does not matter and you can leave this as is.

bocekm:
Rebase to include #290.

bocekm (changes requested):
Rebase to include #290.
And update the `convert2rhel` pytest fixture to run the `shell("subscription-manager unregister")` only when subscription-manager is installed. Right now the tests fail on:
```
            out: Executing a command:
            out: subscription-manager unregister
            out: 
            out: /bin/sh: subscription-manager: command not found
            out: FAILED
```

bocekm (approved):

PR #197, opened 2021-03-03
feat: reinstall subscription-manager

SpyTec:
In Oracle Linux we do not have the correct fingerprint after migration.
To address this we simply reinstall subscription manager post migration to get the correct package fingerprint. Subscription remains

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -45,15 +45,14 @@ class PkgWFingerprint(object):
         self.fingerprint = fingerprint
 
 
-def call_yum_cmd_w_downgrades(cmd, fingerprints):
+def call_yum_cmd_w_downgrades(cmd, pkgs):
     """Calling yum command is prone to end up with an error due to unresolved
     dependencies, especially when it tries to downgrade pkgs. This function
     tries to resolve the dependency errors where yum is not able to.
     """
 
     for _ in range(MAX_YUM_CMD_CALLS):
-        output, ret_code = call_yum_cmd(cmd, "%s" % (" ".join(
-            get_installed_pkgs_by_fingerprint(fingerprints))))
+        output, ret_code = call_yum_cmd(cmd, "%s" % (" ".join(pkgs)))
         loggerinst.info("Received return code: %s\n" % str(ret_code))
         # handle success condition #1
         if ret_code == 0:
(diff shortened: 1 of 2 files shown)


fellipeh (dismissed):
LGTM

convert2rhel/pkghandler.py
```diff
@@ -481,6 +481,9 @@ def replace_non_red_hat_packages():
         "Performing reinstallation of the %s packages ..." % system_info.name)
     call_yum_cmd_w_downgrades("reinstall", system_info.fingerprints_orig_os)
 
+    loggerinst.info("Updating subscription manager to a RHEL fingerprint ...")
+    call_yum_cmd("update", "subscription-manager")
```
bocekm, line 485: This is not going to fly in case the subscription-manager in RHEL is of the same or lower version.

bocekm (changes requested):
The issue applies to OL 6 and 7 only. I suggest to limit the issue to OL 6 only by using the UBI 7 repo as a source for subscription-manager, instead of CentOS 7.

convert2rhel/pkghandler.py
SpyTec: Right, changed to a reinstall. My initial testing seems fine for that as well

bocekm: Actually, yum reinstall will reinstall the package only if there's the package with the **same** version available. That's why convert2rhel replaces all the system packages in three phases:

1. yum update ... to get the original system packages to a newer version that is available in RHEL
2. yum reinstall ... replace the packages that have no newer version available in RHEL but there's the same version available
3. yum distro-sync ... that performs a downgrade of the rest of the packages

In theory, you could choose the distro-sync as a do-it-all (update/reinstall/downgrade as needed). But you'd need to test all the options out. Because in practice, using distro-sync for converting all the packages always failed for me on dependencies.

bocekm: Also, this command would replace only the subcription-manager package. But we install more packages from the CentOS repo. On Oracle Linux 6 it's subscription-manager, subscription-manager-rhsm-certificates and subscription-manager-rhsm.

After merging https://github.com/oamg/convert2rhel/pull/198, this issue will be affecting only Oracle Linux 6.

The asterisk should work:
```suggestion
    if system_info.id == "oracle" and system_info.version.major == 6:
        # The subscription-manager packages on Oracle Linux 6 are installed from CentOS Linux 6 repositories. They are
        # not replaced during the system conversion with the RHEL ones because convert2rhel replaces only packages
        # signed by the original system vendor (Oracle).
        call_yum_cmd("distro-sync", "subscription-manager*")
```

SpyTec: I don't really understand what `distro-sync` is supposed to be doing but I'm not able to get it to work like you suggested. Should we simply call update+reinstall?

```
[root@c2r-20210305121855 ~]# yum list subscription-manager*
Failed to set locale, defaulting to C
Loaded plugins: product-id, search-disabled-repos, subscription-manager
Repository rhel-6-server-rpms is listed more than once in the configuration
Repository rhel-6-server-optional-rpms is listed more than once in the configuration
Repository rhel-6-server-extras-rpms is listed more than once in the configuration
Installed Packages
subscription-manager.x86_64                   1.20.10-8.el6     @rhel-6-server-rpms/6.10                          
subscription-manager-rhsm.x86_64              1.20.10-8.el6     @/subscription-manager-rhsm-1.20.10-8.el6.x86_64  
subscription-manager-rhsm-certificates.x86_64 1.20.10-8.el6     @/subscription-manager-rhsm-certificates-1.20.10-8.el6.x86_64
Available Packages
subscription-manager-firstboot.x86_64         1.20.10-8.el6     rhel-6-server-rpms                                
subscription-manager-gnome.x86_64             0.99.19.4-1.el6_3 rhel-6-server-rpms                                
subscription-manager-gui.x86_64               1.20.10-8.el6     rhel-6-server-rpms                                
subscription-manager-migration.x86_64         1.20.10-8.el6     rhel-6-server-rpms                                
subscription-manager-migration-data.noarch    2.0.43-1.el6      rhel-6-server-rpms    

[root@c2r-20210305121855 ~]# yum distro-sync --releasever=6.10 subscription-manager*
Failed to set locale, defaulting to C
Loaded plugins: product-id, search-disabled-repos, subscription-manager
Setting up Distribution Synchronization Process
Repository rhel-6-server-rpms is listed more than once in the configuration
Repository rhel-6-server-optional-rpms is listed more than once in the configuration
Repository rhel-6-server-extras-rpms is listed more than once in the configuration
No Packages marked for Distribution Synchronization
```

bocekm: From yum man page:

> yum distro-sync ... Synchronizes the installed package set with the latest packages available, this is done by either obsoleting, upgrading or downgrading as appropriate. This will "normally" do the same thing as the upgrade command however if you have the package FOO installed at version 4, and the latest available is only version 3, then this command will downgrade FOO to version 3. 

From that it seems it does not do the reinstallation.

> Should we simply call update+reinstall?

No, you wouldn't have the downgrade case covered. This is how would I approach it:
https://github.com/bocekm/convert2rhel/pull/2/files

bocekm:
The code looks good, thanks. I'm going to test it out.

bocekm (approved):

PR #198, opened 2021-03-04
refactor: use ubi7 instead of centos7

SpyTec:


bocekm:
Reasons for this change:
- simplify the codebase
- resolve the issue of installing subscription-manager signed by CentOS on Oracle Linux systems - the package is considered as from third-party (signed by neither CentOS nor Oracle) and is thus ignored during the conversion and not replaced with the RHEL one

bocekm (changes requested):
You can remove the whole `_get_rhsm_cert_on_centos_7` function from `subscription.py`.

bocekm (approved):

PR #203, opened 2021-03-12
Copying of RHSM certs not rolled back

hosekadam:
Instead of requiring the pre-download and installation of redhat-release, we now copy the RHSM .pem certificates to /etc/pki/product-default/ at the beginning of the conversion process. However in case of any failure, we do not rollback this action, leaving the certificate where we copied it. That does not follow our official statement that "until the point of no return, convert2rhel reverts all its actions in case of a failure".

### DIFF
--- a/convert2rhel/cert.py
+++ b/convert2rhel/cert.py
@@ -26,13 +26,15 @@ loggerinst = logging.getLogger(__name__)
 
 
 class SystemCert(object):
-    _system_cert_dir = "/etc/pki/product-default/"
-
     def __init__(self):
-        self._cert_path = self._get_cert_path()
+        self._target_cert_dir = "/etc/pki/product-default/"
+        self._cert_filename, self._source_cert_dir = self._get_cert()
+        self._source_cert_path = self._get_source_cert_path()
+        self._target_cert_path = self._get_target_cert_path()
 
     @staticmethod
-    def _get_cert_path():
+    def _get_cert():
+        """Return name of certificate and his directory."""
         cert_dir = os.path.join(utils.DATA_DIR, "rhel-certs")
         if not os.access(cert_dir, os.R_OK | os.W_OK):
             loggerinst.critical("Error: Could not access %s." % cert_dir)
@@ -43,7 +45,13 @@ class SystemCert(object):
                 break
         if not pem_filename:
             loggerinst.critical("Error: System certificate (.pem) not found in %s." % cert_dir)
-        return os.path.join(cert_dir, pem_filename)
+        return pem_filename, cert_dir
+
+    def _get_source_cert_path(self):
+        return os.path.join(self._source_cert_dir, self._cert_filename)
+
+    def _get_target_cert_path(self):
+        return os.path.join(self._target_cert_dir, self._cert_filename)
 
     def install(self):
         """RHEL certificate (.pem) is used by subscription-manager to
@@ -52,9 +60,19 @@ class SystemCert(object):
         loggerinst.info("Installing RHEL certificate to the system.")
 
         try:
-            utils.mkdir_p(self._system_cert_dir)
-            shutil.copy(self._cert_path, self._system_cert_dir)
+            utils.mkdir_p(self._target_cert_dir)
+            shutil.copy(self._source_cert_path, self._target_cert_dir)
         except OSError as err:
             loggerinst.critical("OSError({0}): {1}".format(err.errno, err.strerror))
 
-        loggerinst.debug("Certificate copied to %s." % self._system_cert_dir)
+        loggerinst.debug("Certificate copied to %s." % self._target_cert_dir)
+
+    def remove(self):
+        """Remove certificate (.pem), which was copied to system's cert dir."""
+        loggerinst.task("Rollback: Removing installed RHSM certificate")
+
+        try:
+            os.remove(self._target_cert_path)
+            loggerinst.info("Certificate %s removed" % self._target_cert_path)
+        except OSError as err:
+            loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -25,6 +25,15 @@ except ImportError:
 
 import os
 import shutil
+import sys
+
+import pytest
+
+
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
 
 from convert2rhel import cert, utils
 from convert2rhel.systeminfo import system_info
@@ -50,7 +59,8 @@ class TestCert(unittest.TestCase):
         for arch, rhel_versions in self.certs.items():
             for rhel_version, pem in rhel_versions.items():
                 utils.DATA_DIR = os.path.join(self.base_data_dir, rhel_version, arch)
-                cert_path = cert.SystemCert._get_cert_path()
+                system_cert = cert.SystemCert()
+                cert_path = system_cert._source_cert_path
                 self.assertEqual(cert_path, "{0}/rhel-certs/{1}".format(utils.DATA_DIR, pem))
 
     @unit_tests.mock(cert, "loggerinst", unit_tests.GetLoggerMocked())
@@ -61,7 +71,7 @@ class TestCert(unittest.TestCase):
         cert_dir = os.path.join(utils.DATA_DIR, "rhel-certs", system_info.arch)
         utils.mkdir_p(cert_dir)
         # Check response for the non-existing certificate in the temporary dir
-        self.assertRaises(SystemExit, cert.SystemCert._get_cert_path)
+        self.assertRaises(SystemExit, cert.SystemCert._get_cert)
         self.assertEqual(len(cert.loggerinst.critical_msgs), 1)
         # Remove the temporary directory tree
         shutil.rmtree(os.path.join(utils.DATA_DIR, "rhel-certs"))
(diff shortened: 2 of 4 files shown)


convert2rhel/cert.py
```diff
@@ -58,3 +60,33 @@ def install(self):
...
+            os.remove(os.path.join(self._system_cert_dir, self._get_cert_filename()))
+        except OSError as err:
+            loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
+
+        loggerinst.debug("Certificate removed")
+
+    @staticmethod
+    def _get_cert_filename():
+        """ Same as _get_cert_path, but returns just filename, not whole path """
+        loggerinst = logging.getLogger(__name__)
```
SpyTec, line 78: Not needed as we initialize the file's logger at the beginning of the file
```suggestion
```

convert2rhel/cert.py
```diff
@@ -58,3 +60,33 @@ def install(self):
...
+        for filename in os.listdir(cert_dir):
+            if filename.endswith(".pem"):
+                pem_filename = filename
+                break
+        if not pem_filename:
+            loggerinst.critical(
+                "Error: System certificate (.pem) not found in %s." % cert_dir
+            )
+
+        return pem_filename
```
SpyTec, line 92: Since this is the exact same implementation as `_get_cert_path` as you mentioned, I'd recommend to consolidate these two into one function

Something like `_get_cert()` that returns a tuple of name and path `return pem_filename, os.path.join(cert_dir, pem_filename)`. Then set these two as class variables in the `__init__`

convert2rhel/unit_tests/main_test.py
```diff
@@ -134,6 +134,8 @@ def test_show_eula_nonexisting_file(self):
     @unit_tests.mock(redhatrelease.yum_conf, "restore", unit_tests.CountableMockObject())
     @unit_tests.mock(subscription, "rollback", unit_tests.CountableMockObject())
     @unit_tests.mock(pkghandler.versionlock_file, "restore", unit_tests.CountableMockObject())
+    @unit_tests.mock(cert.SystemCert, "_get_cert_path", unit_tests.MockFunction)
+    @unit_tests.mock(cert.SystemCert, "remove", unit_tests.CountableMockObject())
```
SpyTec, line 138: `remove` was never asserted if it was called

convert2rhel/cert.py
```diff
@@ -58,3 +60,33 @@ def install(self):
             loggerinst.critical("OSError({0}): {1}".format(err.errno, err.strerror))
 
         loggerinst.debug("Certificate copied to %s." % self._system_cert_dir)
+
+    def remove(self):
+        """Removes certificate (.pem), which was copied in install method"""
```
SpyTec, line 65:
```suggestion
        """Remove certificate (.pem) which was copied to system's cert dir"""
```

SpyTec (changes requested):

convert2rhel/cert.py
bocekm: Just FYI, the `Removes` -> `Remove` change comes from https://www.python.org/dev/peps/pep-0257/:
> [The docstring] prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

convert2rhel/cert.py
```diff
@@ -58,3 +60,33 @@ def install(self):
...
+        loggerinst.info("Removing copy of certificate from %s." % self._system_cert_dir)
+
+        try:
+            os.remove(os.path.join(self._system_cert_dir, self._get_cert_filename()))
```
bocekm, line 69:
```suggestion
            os.remove(self._target_cert_path)
```

There's too many cert paths and dirs, so it's better to distinguish them to easily tell them apart.

```python
def __init__(self):
    self._source_cert_dir = os.path.join(utils.DATA_DIR, "rhel-certs")
    self._target_cert_dir = "/etc/pki/product-default/"
    self._cert_filename = self._get_cert_filename()
    self._source_cert_path = self._get_source_cert_path()
    self._target_cert_path = self._get_target_cert_path()

def _get_cert_filename(self):
    pem_filename = None
    for filename in os.listdir(self._source_cert_dir):
        if filename.endswith(".pem"):
            pem_filename = filename
            break
    if not pem_filename:
        loggerinst.critical(
            "Error: System certificate (.pem) not found in %s." % self._source_cert_dir
        )
    return pem_filename

def _get_source_cert_path(self):
    os.path.join(self._source_cert_dir, self._cert_filename)

def _get_target_cert_path(self):
    os.path.join(self._target_cert_dir, self._cert_filename)
```

bocekm (changes requested):

hosekadam:
Update based on combination of @SpyTec and @bocekm comments.

SpyTec (changes requested):
One minor change, otherwise LGTM

convert2rhel/cert.py
```diff
@@ -52,9 +60,20 @@ def install(self):
...
+    def remove(self):
+        """Remove certificate (.pem), which was copied to system's cert dir."""
+        loggerinst.info("Removing the certificate from %s." % self._target_cert_path)
+
+        try:
+            os.remove(self._target_cert_path)
+        except OSError as err:
+            loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
+
+        loggerinst.debug("Certificate removed")
```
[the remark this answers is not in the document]
hosekadam: Oh, I see. Fixed :)

convert2rhel/unit_tests/cert_test.py
```diff
@@ -70,20 +77,32 @@ def test_get_cert_path_missing_cert(self):
...
+    os_remove = mock.Mock()
+    monkeypatch.setattr(os, "remove", value=os_remove)
+
+    _get_cert = mock.Mock(return_value=("anything", "anything"))
+    monkeypatch.setattr(cert.SystemCert, "_get_cert", value=_get_cert)
+
+    sys_cert = cert.SystemCert()
+    sys_cert.remove()
+
+    os_remove.assert_called_once()
```
ZhukovGreen, line 108: This test is not that useful. You mocking the main piece of the functionality - removing the file. That would be fine if you would like to test if logger messages were emitted. Otherwise, it is better to mock as minimum things as possible.
The goal of your function to remove the target cert path. So instead of mocking the remove function, try to play with the value of the cert path. Check if it is None, not existing file, and existing file

convert2rhel/unit_tests/main_test.py
```diff
@@ -113,6 +113,10 @@ def __call__(self, command, args):
             self.args = args
             return self.return_string, self.return_code
 
+    class GetCertMocked(unit_tests.MockFunction):
+        def __call__(self):
+            return "anything", "anything"
+
```
ZhukovGreen, line 119: Nit: you can use just Mock(return_value=("anything", "anything"), so you'll get what you need + bonus functionality

ZhukovGreen (commented):
@hosekadam mostly nits (nit picks - which optionally to be resolved) + 1 more important comment

convert2rhel/unit_tests/cert_test.py
hosekadam: @ZhukovGreen Fixed, now it tests real remove of file if it exists, otherwise it raise exception (and check, if logger.error was called). I think, that variant with None would crash earlier, than it would come to remove method. But I put it there :)

convert2rhel/unit_tests/main_test.py
hosekadam: @ZhukovGreen My motivation to do it as I did was to not combine two methods of testing in one test case. Or it doesn't matter?

convert2rhel/unit_tests/cert_test.py
```diff
@@ -70,20 +78,53 @@ def test_get_cert_path_missing_cert(self):
...
+        installed_cert_path = os.path.join(system_cert._target_cert_dir, system_cert._cert_filename)
         self.assertTrue(os.path.exists(installed_cert_path))
         shutil.rmtree(unit_tests.TMP_DIR)
+
+
+@pytest.mark.parametrize(
+    ("directory", "filename"),
+    ((unit_tests.TMP_DIR, "existing.file"), (unit_tests.TMP_DIR, unit_tests.NONEXISTING_FILE), (None, None)),
+)
+def test_remove(monkeypatch, directory, filename):
```
ZhukovGreen, line 103: nit: pytest has fixtures `tmpdir` and `caplog`, try to adopt it, the solution should be more readable and simplier

convert2rhel/unit_tests/cert_test.py
```diff
@@ -70,20 +78,53 @@ def test_get_cert_path_missing_cert(self):
...
+    if filename == "existing.file":
+        try:
+            utils.mkdir_p(directory)
+            os.mknod(os.path.join(directory, filename))
+        except FileExistsError:
+            pass
+
+    monkeypatch.setattr(cert.SystemCert, "_get_cert", value=mock.Mock(return_value=("anything", "anything")))
+    if directory is not None or filename is not None:
```
ZhukovGreen, line 112: it is enough to write just:
```suggestion
    if directory or filename:
```

convert2rhel/unit_tests/main_test.py
ZhukovGreen: if yuo mean using mock.Mock in unittest, that's totally fine, moreover there is no mixing, as mock.Mock I snot related to pytest method (if you thought so)

convert2rhel/unit_tests/cert_test.py
bocekm: Lately we've been using caplog, it's easier. E.g. https://github.com/oamg/convert2rhel/blob/main/convert2rhel/unit_tests/subscription_test.py#L432.

convert2rhel/unit_tests/main_test.py
```diff
@@ -137,6 +141,8 @@ def test_show_eula_nonexisting_file(self):
         "restore",
         unit_tests.CountableMockObject(),
     )
+    @unit_tests.mock(cert.SystemCert, "_get_cert", GetCertMocked())
```
bocekm, line 144:
```suggestion
    @unit_tests.mock(cert.SystemCert, "_get_cert", lambda: ("anything", "anything"))
```

And you can remove `class GetCertMocked` ;)

bocekm (changes requested):

convert2rhel/unit_tests/main_test.py
bocekm: Other option: https://github.com/oamg/convert2rhel/pull/203#discussion_r657142772

convert2rhel/unit_tests/cert_test.py
hosekadam: @ZhukovGreen  I've got a problem with 'tmpdir' on older version of python (while running 'make tests' the first one is python2.7.5). I think it's due this code you've shared with me [conftest.py#L22](https://github.com/oamg/convert2rhel/blob/5630fa7b070b0527e0a6426148ce6823108f7610/convert2rhel/unit_tests/conftest.py#L22) because it returns string on older versions, but str hasn't got methods for [creating](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture) file :smiley: 

The test is working correctly while running it on newer compatible version with 'pytest -k <name_of_test>'.

I think, that I could use the 'tmpdir' as the 'unit_tests.TMP_DIR' which I'm using now and it should solve my problem. But with this will go away the advantage of 'tmpdir' fixture.

ZhukovGreen: This fixture was created to solve a problem of py26 compatibility. It is no longer the care and we are not running unit tests on this system anymore. Try to remove this fixture

hosekadam: @ZhukovGreen even worse. Too many type errors `TypeError: coercing to Unicode: need string or buffer, LocalPath found`

hosekadam:
Problem with [tmpdir](https://github.com/oamg/convert2rhel/pull/203#discussion_r656228703) in unit tests.

convert2rhel/unit_tests/cert_test.py
ZhukovGreen: Try to troubleshoot this. If dead end, I'll help on Thu-Fri

hosekadam:
In my opinion, the variant of unit test with `None` isn't necessary. Because the method `_get_cert` which return the filename can't return `None`. It would fail there on the [last condition](https://github.com/oamg/convert2rhel/commit/fbc5e4fd79bb5bc4c6654b382967364fabf6f194#diff-f88fcc3a7b096cd8ad73a72a5b87da4a9b4ab4014ac5adaa9b07110e86fc2f00L44).

I reworked the [unit test](https://github.com/oamg/convert2rhel/commit/fbc5e4fd79bb5bc4c6654b382967364fabf6f194#diff-63e8e1691f03bae8891cc9dc99187aeae3e1ac9b0b6f6b88b6d5ded0cf62dc0aR105) that it works on the older versions. I can't use the `tmpdir` fixture, but when it was a [nit](https://github.com/oamg/convert2rhel/pull/203#discussion_r656228703), I believe, it would be ok.

convert2rhel/cert.py
```diff
@@ -52,9 +60,19 @@ def install(self):
...
+            shutil.copy(self._source_cert_path, self._target_cert_dir)
         except OSError as err:
             loggerinst.critical("OSError({0}): {1}".format(err.errno, err.strerror))
 
-        loggerinst.debug("Certificate copied to %s." % self._system_cert_dir)
+        loggerinst.debug("Certificate copied to %s." % self._target_cert_dir)
+
+    def remove(self):
+        """Remove certificate (.pem), which was copied to system's cert dir."""
+        loggerinst.info("Removing the certificate %s." % self._target_cert_path)
```
bocekm, line 72:
```suggestion
        loggerinst.task("Rollback: Removing installed RHSM certificate")
```

Without this, the output is mixed up with the previous task:
```
[07/13/2021 10:37:09] TASK - [Rollback: Restoring /etc/yum/pluginconf.d/versionlock.list from backup] 
WARNING - /etc/yum/pluginconf.d/versionlock.list hasn't been backed up
Removing the certificate /etc/pki/product-default/69.pem.
[07/13/2021 10:37:09] DEBUG - Certificate removed
```

convert2rhel/cert.py
```diff
@@ -52,9 +60,19 @@ def install(self):
...
+
+        try:
+            os.remove(self._target_cert_path)
+            loggerinst.debug("Certificate removed")
```
bocekm, line 76:
```suggestion
            loggerinst.info("Certificate %s removed" % self._target_cert_path)
```

bocekm (changes requested):

bocekm (approved):

ZhukovGreen (approved):

ZhukovGreen:
@SpyTec @hosekadam @bocekm I am fine if we don't cover this feature by integration tests

SpyTec (approved):

PR #204, opened 2021-03-12
Rename CentOS -> CentOS Linux

hosekadam:
All occurrences of CentOS X, or just CentOS are replaced with CentOS Linux X, or CentOS Linux

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -682,7 +682,7 @@ def remove_non_rhel_kernels():
 
 def fix_default_kernel():
     """
-    Systems converted from Oracle Linux or CentOS Plus may have leftover kernel-uek or kernel-plus in
+    Systems converted from Oracle Linux or CentOS Linux may have leftover kernel-uek or kernel-plus in
     /etc/sysconfig/kernel as DEFAULTKERNEL.
     This function fixes that by replacing the DEFAULTKERNEL setting from kernel-uek or kernel-plus to kernel for RHEL 6,7 and kernel-core for RHEL 8
     """
--- a/packaging/convert2rhel.spec
+++ b/packaging/convert2rhel.spec
@@ -51,8 +51,8 @@ Requires:       pexpect
 The purpose of the convert2rhel tool is to provide an automated way of
 converting the installed other-than-RHEL OS distribution to Red Hat Enterprise
 Linux (RHEL). The tool replaces all the original OS-signed packages with the
-RHEL ones. Available are conversions of CentOS 6/7/8 and Oracle Linux 6/7/8 to
-the respective major version of RHEL.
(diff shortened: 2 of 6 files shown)


convert2rhel/pkghandler.py
```diff
@@ -661,7 +661,7 @@ def remove_non_rhel_kernels():
 
 def fix_default_kernel():
     """
-    Systems converted from Oracle Linux or CentOS Plus may have leftover kernel-uek or kernel-plus in
+    Systems converted from Oracle Linux or CentOS Linux Plus may have leftover kernel-uek or kernel-plus in
```
bocekm, line 664:
```suggestion
    Systems converted from Oracle Linux or CentOS Linux may have leftover kernel-uek or kernel-plus in
```
As it was before, it suggested the there's such a thing as an OS named CentOS Linux Plus. But in reality there's just CentOS Linux on which you can enable the Plus repository and install the kernel-plus package from there.

packaging/convert2rhel.spec
```diff
@@ -51,7 +51,7 @@ Requires:       pexpect
 The purpose of the convert2rhel tool is to provide an automated way of
 converting the installed other-than-RHEL OS distribution to Red Hat Enterprise
 Linux (RHEL). The tool replaces all the original OS-signed packages with the
-RHEL ones. Available are conversions of CentOS 6/7/8 and Oracle Linux 6/7/8 to
+RHEL ones. Available are conversions of CentOS Linux 6/7/8 and Oracle Linux 6/7/8 to
```
bocekm, line 54: Wrap the line to 80 chars. It's I believe required by the spec file format and we'd be getting warnings when building the rpm.

bocekm (changes requested):
Change it also here:
- https://github.com/oamg/convert2rhel/blob/v0.17/convert2rhel/subscription.py#L36

packaging/convert2rhel.spec
```diff
@@ -127,7 +127,7 @@ install -p man/%{name}.8 %{buildroot}%{_mandir}/man8/
   the packager when the vendor information is not available.
 
 * Fri Jan 8 2021 Michal Bocek <mbocek@redhat.com> 0.15-1
-- add missing CentOS 8 packages to be removed prior the conversion
+- add missing CentOS Linux 8 packages to be removed prior the conversion
```
bocekm, line 130: Please do not change the changelog entries. Let's consider it as immutable - not to be changed in the future.

bocekm (changes requested):

packaging/convert2rhel.spec
bocekm: @hosekadam, please revert also the changes in the other changelog entries. Thanks.

bocekm (approved):

PR #209, opened 2021-03-19
Feature/build rpms locally

ZhukovGreen:
Builds rpms for all 3 centos versions locally.
Command make rpms


Makefile
```diff
@@ -63,3 +64,6 @@ lint-errors: images
 
 tests8: images
 	@docker run --rm -v $(shell pwd):/data:Z $(IMAGE)/centos8 pytest
+
+rpms:
```
pirat89, line 68: @ZhukovGreen unless it was expected to have the target `rpms` directly, I sugges to call it `build` instead to be consistent with bunch of other tools and projects.

ZhukovGreen: So you mean
Make builds? @pirat89

ZhukovGreen: It is all three rpms

pirat89:
and curerntly the PR cannot be merged until other two PRs (#193 and #210) are not merged..

Makefile
pirat89: @ZhukovGreen hmm. maybe even `local_build` as it is just for local stuff - it doesn't matter so much whether it's for all systems. when you do the copr build you expect that all expected builds will be created (for every build root unless specified otherwise). I would not do so much difference in this, unless you plan in future provide a target for one particular build. In such a case, maybe `local_build_all` could be what you want. as `build` | `builds` is easy to exchange.

pirat89: But to be honest, I would go most just wiath `local_build` and in case someone would like to specify that only one specific build should be created, I would add possibility to specify an envar for that.

bocekm: I think `make rpms` works fine as well.

bocekm (changes requested):
@ZhukovGreen, rebase to get rid of unrelated commits.

ZhukovGreen:
closing in favour of #211

PR #213, opened 2021-03-19
Feature/integration tests of read only mnt sys

ZhukovGreen:
Tests @fellipeh feature of inhibition of the conversion when /mnt or /sys are read only only #194
Currently tests checks the messages and inhibition on centos7 and 8 for /sys:ro  only


ZhukovGreen:
closing as it is too outdated. Should be solved via libvirt porvisioner and using our state of the art fixtures:

Hint:
```
 mount -o remount,ro /dev/sda6 /srv/html
```
For remounting as read only

PR #218, opened 2021-03-23
Disable centos6 jobs until resolving the issue with ssl

ZhukovGreen:
Currently, container can't be build due to inability to install pip. We need to disable centos6 pipelines until finishing the investigation, in order do not block other MRs

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -11,7 +11,7 @@ jobs:
     name: linting-centos${{ matrix.centos_ver }}
     strategy:
       matrix:
-        centos_ver: [6, 7, 8]
+        centos_ver: [7, 8]
 
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


SpyTec:
We can merge with failing checks, so we don't need to remove it from the CI

ZhukovGreen:
Sorry Eric, but do not close my merge requests please. The CI is important to keep, and it is one of the conditions to merge MRs￼. So I need this MR merged

SpyTec:
Failing checks for PRs are not a blocker for getting PRs merged, we can merge with failing checks

SpyTec:
@ZhukovGreen I see that other tests top if one of them fails, we can instead disable fail-fast to that the other tests for 7 and 8 can still run

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

ZhukovGreen:
@SpyTec this is actually exactly what I did

shaded-enmity (approved):
Approving, @ZhukovGreen can you please create a tracking ticket to get CentOS6 back on track?

PR #224, opened 2021-03-25
Remove cockpit-subscriptions not available in RHEL

bocekm:
Leaving this package on the system leads to a transaction failure. It conflicts with files of the RHEL package cockpit-system that is being installed during the conversion.

### DIFF
--- a/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
+++ b/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
@@ -19,6 +19,7 @@ excluded_pkgs =
   yum-rhn-plugin
   mod_ldap
   mod_proxy_html
+  cockpit-subscriptions
 
 # List of packages that either contain repofiles or affect variables in the repofiles (e.g. $releasever).
 # Delimited by any whitespace(s).


AloisMahdal:
FWIW, the cockpit-subscriptions package is a sub-package of CentOS-7's `cockpit.spec`.

Looking at the package description in CentOS-7:

    [root@localhost jatlab]# yum info cockpit-subscriptions
    Loaded plugins: fastestmirror
    Loading mirror speeds from cached hostfile
     * base: repos-va.psychz.net
     * extras: mirror.centos.iad1.serverforge.org
     * updates: mirror.vcu.edu
    Available Packages
    Name        : cockpit-subscriptions
    Arch        : noarch
    Version     : 160
    Release     : 1.el7.centos
    Size        : 81 k
    Repo        : extras/7/x86_64
    Summary     : Cockpit subscription user interface package
    URL         : http://cockpit-project.org/
    License     : LGPLv2+
    Description : This package contains the Cockpit user interface integration with local
                : subscription management.

looks like removing should be relatively safe.

This package is not available in CentOS-8, it's CentOS-7 only.

AloisMahdal:
I will now re-run the same `max_sst` test with this PR, that should show if the conversion can finish.

Atop of that, this manual test could be worth doing to cover our backs:

 1. On centos-7, install `cockpit` and `cockpit-subscription`
 2. Start the cockpit
     *  IIRC `systemctl start cockpit` should be enough.
 3. Check if the web UI works.
 4. Run the conversion
     *  this should auto-remove the `cockpit-subscription`
 5. Reboot
 6. Check again if the web UI works.
     *  at least you should be able to get system shell.

ZhukovGreen (approved):

bocekm:
I and Alois have not been able to verify this so far (the conversion halts on an unrelated kmod check). Waiting with merging.

AloisMahdal (approved):

PR #226, opened 2021-03-25
Ensure no symlinks will be considered as kmod

ZhukovGreen:


SpyTec (approved):

bocekm:
@ZhukovGreen, @shaded-enmity, please describe the reason for the change for future reference.

PR #228, opened 2021-03-26
Ignore kmods moved to kernel core

bocekm:
Some kernel modules move from kernel modules to kernel core. Instead of
inhibiting the conversion saying that such a module is not available in
RHEL and thus is unsupported, we ignore it.

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -287,7 +287,11 @@ def get_rhel_kmods_keys(rhel_kmods_str):
 
 def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
     """Return a set of those installed kernel modules that are not available in RHEL repositories.
-    
+
     Ignore certain kmods mentioned in the system configs. These kernel modules moved to kernel core, meaning that the
-    functionality is retained and we would be incorrectly saying that the modules are not supported in RHEL. """
-    return host_kmods - rhel_supported_kmods - set(system_info.kmods_to_ignore)
+    functionality is retained and we would be incorrectly saying that the modules are not supported in RHEL."""
+    return (
+        host_kmods
+        - rhel_supported_kmods
+        - set(system_info.kmods_to_ignore)
+    )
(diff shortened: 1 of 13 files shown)


convert2rhel/checks.py
```diff
@@ -87,14 +87,15 @@ def ensure_compatibility_of_kmods():
     """Ensure if the host kernel modules are compatible with RHEL."""
     host_kmods = get_installed_kmods()
     rhel_supported_kmods = get_rhel_supported_kmods()
-    if is_unsupported_kmod_installed(host_kmods, rhel_supported_kmods):
+    unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods):
```
SpyTec, line 90: Syntax error
```suggestion
    unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods)
```

bocekm:
It might be failing on the key in the config missing in other configs. To be fixed soon.

shaded-enmity (approved):
Apart from the syntax error pointed out by @SpyTec LGTM

ZhukovGreen (approved):
Excellent work @bocekm , especially tests:)

bocekm:
Verified and works. On the machine where it was manifesting before.
```
[03/26/2021 18:55:26] DEBUG - Calling command 'uname -r'
3.10.0-1160.15.2.el7.x86_64
[03/26/2021 18:55:26] DEBUG - Calling command 'find /lib/modules/3.10.0-1160.15.2.el7.x86_64 -name "*.ko*" -type f'
[03/26/2021 18:55:26] DEBUG - Calling command 'repoquery --releasever=7Server  --repoid rhel-7-server-rpms -f /lib/modules/*.ko*'
[03/26/2021 18:57:11] DEBUG - Calling command 'repoquery --releasever=7Server  --repoid rhel-7-server-rpms -l kernel-0:3.10.0-1160.21.1.el7.x86_64 ...
```

PR #229, opened 2021-03-26
Do not inhibit conversion on unsupported kmods

bocekm:
There might be false positives so we don't want to inhibit in those
cases.


shaded-enmity (approved):
LGTM

PR #230, opened 2021-03-26
fix: remove dep resolve correctly

SpyTec:
Currently whenever we get dependencies to remove we don't use the correct output. We should grab the output that comes from already resolving the dependency issues.


bocekm (dismissed):
Looks good, thanks.

bocekm (dismissed):
Re-approving after rebase.

bocekm (approved):

bocekm:
@AloisMahdal has verified that the scenario with which we were getting the error before is passing now.

PR #235, opened 2021-03-29
Resolve yum errors with downgrades, not distro-sync

bocekm:
Yum has a problematic resolution of dependencies in general. It
oftentimes ends up with a transaction error even though there's a way to
resolve the transaction, mainly when there are downgrades necessary.
Using yum downgrade instead of distro-sync should be more reliable in
this regards.


bocekm:
It is not more reliable.
It does not handle dependency errors related to python2-hawkey on CentOS 7.
We have actually a regression of the "abrt" issue that prompted https://github.com/oamg/convert2rhel/pull/205 and https://github.com/oamg/convert2rhel/pull/230.

PR #236, opened 2021-03-29
fix: resolve_dep_errors() return last bad output

SpyTec:
Fix of #230

If we get problematic packages and it's not resolved, we returned output incorrectly by returning the yum command instead of the output with the errors. It now uses the correct output in the return


bocekm:
This solution would not work as expected in this scenario:
1. we're in the call_yum_cmd_w_downgrades and there's a dependency error so resolve_dep_errors is called 
2. within that function we call yum distro-sync which ends up successful
3. with the code change you did, the output with the dependency error is returned back to the call_yum_cmd_w_downgrades
4. the packages that were causing the error (but distro-sync resolved that) are being incorrectly removed

Closing in favor of https://github.com/oamg/convert2rhel/pull/237.

PR #240, opened 2021-03-30
Remove non-interactive command suggestion

bocekm:
Currently the feature of suggesting a convert2rhel command to run on other systems to convert in a non-interactive fashion is not working properly. For example, it is missing couple of switches (e.g. --no-rpm-va and --debug) and does not handle well the situation with enabled repoids when RHSM is used.

To reduce the codebase to maintain and for the lack of feedback on this feature, we decided to remove it.

An example of the output being removed:
```
[03/29/2021 19:55:12] TASK - [Final: Non-interactive mode] **************************************
For the non-interactive use of the tool, run the following command:
convert2rhel --disable-submgr  --disablerepo=*  --enablerepo=rhel-7-server-rpms  -y
```

Related: https://issues.redhat.com/browse/OAMG-3141

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -101,10 +101,6 @@ def main():
         loggerinst.task("Final: rpm files modified by the conversion")
         systeminfo.system_info.modified_rpm_files_diff()
 
-        # recommend non-interactive command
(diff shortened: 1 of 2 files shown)


shaded-enmity (approved):
lgtm

PR #242, opened 2021-03-30
fix: Remove variant message from help usage

fellipeh:
This PR remove [-v variant] on help usage message


bocekm:
@fellipeh, this solution is not dealing with the other part of the task:
> The warning message _"Variant option is not supported anymore"_ is extended to the "-v" option as well (currently it's displayed only when "--variant" is used)

Please create another PR for that.

PR #251, opened 2021-04-12
use more universal reboot command

### LINKED ISSUE #241: CentOS 6.10 with Python 2.6.6 fail at reboot (overwriting env vars)
When I run the script it ends in an error near the end when it tries to restart the server.

Here is the error :

> [03/30/2021 10:34:25] TASK - [Final: Non-interactive mode] **************************************
> For the non-interactive use of the tool, run the following command:
> convert2rhel --disable-submgr  --disablerepo=*  --enablerepo=redhat  -y
> 
> 
> ('[%s] %s - %s', '03/30/2021 10:34:25', 'DEBUG', "Calling command 'reboot'")
>   File "/usr/lib/python2.6/site-packages/convert2rhel/main.py", line 109, in main
>     utils.restart_system()
>   File "/usr/lib/python2.6/site-packages/convert2rhel/utils.py", line 117, in restart_system
>     run_subprocess("reboot")
>   File "/usr/lib/python2.6/site-packages/convert2rhel/utils.py", line 140, in run_subprocess
>     env={'LC_ALL': 'C'})
>   File "/usr/lib64/python2.6/subprocess.py", line 642, in __init__
>     errread, errwrite)
>   File "/usr/lib64/python2.6/subprocess.py", line 1238, in _execute_child
>     raise child_exception
> OSError: [Errno 2] No such file or directory
> 
> WARNING - Conversion process interrupted and manual user intervention will be necessary.
> 

I tried with versions 0.20 and 0.19 of the script. I checked that the reboot command existed, but after some testing from a python console I noticed that it worked by removing the "env={'LC_ALL': 'C'}" parameter from the subprocess.popen command.

ZhukovGreen:
fixes #241

reboot command might be missing on CentOS 6. Using a more universal alternative

### DIFF
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -114,7 +114,7 @@ def store_content_to_file(filename, content):
 def restart_system():
     from convert2rhel.toolopts import tool_opts
     if tool_opts.restart:
-        run_subprocess("reboot")
+        run_subprocess("shutdown -r now")
     else:
         loggerinst.warning("In order to boot the RHEL kernel,"
                            " restart of the system is needed.")


ZhukovGreen:
@bocekm do you think we need an integration tests for this case?

convert2rhel/utils.py
```diff
@@ -114,7 +114,7 @@ def store_content_to_file(filename, content):
 def restart_system():
     from convert2rhel.toolopts import tool_opts
     if tool_opts.restart:
-        run_subprocess("reboot")
+        run_subprocess("shutdown -r now")
```
bocekm, line 117: I haven't tested that but I'd say this change is going to have the same outcome. Both `reboot` and `shutdown` are located in `/sbin/` on RHEL 6-based systems, and both are installed through the `upstart` package. My wild guess is that `/sbin/` is not in paths where the python subprocess library looks for executables.

I needed to workaround probably the same type of issue at https://github.com/oamg/convert2rhel/blob/v0.20/convert2rhel/pkghandler.py#L736.

I quickly checked CentOS Linux 6/7/8 now and `/sbin/reboot` exists on all.

```suggestion
        run_subprocess("/sbin/reboot")
```

ZhukovGreen: To follow up, it could be something different, because I have one machine where only shutdown command exists. Also there are vast number of discussions on stack overflow where shutdown -r is recommended over the reboot due to the exactly the reason of possibility of missing the reboot command.

ZhukovGreen: Let me check first why shutdown was present and reboot not, perhaps fall back to your proposal

ZhukovGreen: ok, no success on my side. I asked George to share the output https://github.com/oamg/convert2rhel/issues/241#issuecomment-818515520

bocekm: Ok, feel free to use shutdown. I did not expect some machine to not have reboot. In that case, we might need to update the documentation as well https://access.redhat.com/articles/2360841.

Out of curiosity, can you check what package the shutdown comes from on the machine where reboot is not available (`rpm -qf /sbin/shutdown`)?

ZhukovGreen: [root@d84744d80e20 /]# rpm -qf /sbin/shutdown
error: file /sbin/shutdown: No such file or directory
[root@d84744d80e20 /]# rpm -qf /sbin/reboot  
error: file /sbin/reboot: No such file or directory

But shutdown is available:)

ZhukovGreen: anyway, let's wait for George. I have no idea how env={'LC_ALL': 'C'} affects the existence of reboot

bocekm: In that case you can run `which shutdown` to get the path of the executable and then `rpm -qf <path>`.
BTW, what's the machine on which you're experiencing this?

pirat89: > anyway, let's wait for George. I have no idea how env={'LC_ALL': 'C'} affects the existence of reboot

@ZhukovGreen  becuase it drops all existing evironment variables. It's the same case as I already reported somewhere. corerctly it should get  all existing variables and then set `LC_ALL` in the dict so the system setup is propagated. When the `$PATH` envar is dropped, the function search just in /bin /usr/bin directories. It ignores /sbin and /usr/sbin - which are usually set via $PATH explicitly on the system.

pirat89: > [root@d84744d80e20 /]# rpm -qf /sbin/shutdown
> error: file /sbin/shutdown: No such file or directory
> [root@d84744d80e20 /]# rpm -qf /sbin/reboot
> error: file /sbin/reboot: No such file or directory
> 
> But shutdown is available:)

You should search /usr/sbin/shutdown in such a case

bocekm: @pirat89, but ` /usr/sbin/shutdown` in N/A on CentOS/OL Linux 6. In my testing, ` /sbin/shutdown` is available in all CentOS/OL Linux 6/7/8. And in CentOS/OL Linux 7+ it's a symlink to the ` /usr/sbin/shutdown`. But I wondered what package owns the @ZhukovGreen's shutdown executable. I'd like to find out why some package creates the ` /sbin/shutdown` symlink (on my testing machines) and some not (on Artem's testing machines).

pirat89: @bocekm you are right, on EL/OL 6 it's `/sbin/shutdown/ as there is different FHS.

ZhukovGreen:
@pirat89 @bocekm Then I would vote for the solution do not override the env, but append to it (https://stackoverflow.com/questions/2231227/python-subprocess-popen-with-a-modified-environment/4453495)

bocekm:
@ZhukovGreen, yes, that sounds like a good course of action.
Even though it would be better to find out how's that possible that your machine did not have `reboot` available at all. For that reason, replacing reboot with shutdown can still make sense.

ZhukovGreen:
@bocekm can't find this machine, perhaps removed it during the tests with tmt-pytest. Sorry, about that.  Anyway it seems reliable to implement the feature of extending the process environment

bocekm:
@ZhukovGreen, yes, we'll implement that under https://issues.redhat.com/browse/OAMG-4810.

PR #252, opened 2021-04-14
fix: move CLI init below logger init

SpyTec:
Within CLI init we make use of logger to output messages about the
parameters the user has inputted. We also have a prompt for the
user to continue the conversion, but without our logger being
initialized you cannot say no to the prompt as it loops infinitely


ZhukovGreen (approved):
@SpyTec pls fix linting and tests in centos 6

SpyTec:
> @SpyTec pls fix linting and tests in centos 6

That's the same error we have on all PRs currently, relates to SSL issue we've had previously. Outside the scope of this PR

kokesak:
Tested on version: convert2rhel-0.20-1.20210414143940481771.pr252.14.gc224190.el7.noarch and convert2rhel-0.20-1.20210414143940481771.pr252.14.gc224190.el8.noarch. Both versions are passing!

bocekm (approved):

PR #253, opened 2021-04-14
fix: Moving check_readonly_mounts function from utils.py to checks.py

fellipeh:
Related ticket: OAMG-4697

Moving check_readonly_mounts function from `utils.py` to `checks.py`, since we defined one place for checks in checks.py. This will allow us easier to manage those checks.

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -23,8 +23,7 @@ import re
 import subprocess
 
 from convert2rhel.systeminfo import system_info
-from convert2rhel.utils import run_subprocess
-
+from convert2rhel.utils import run_subprocess, get_file_content
 
 logger = logging.getLogger(__name__)
 
@@ -37,6 +36,7 @@ def perform_pre_checks():
     """Early checks after system facts should be added here."""
     check_uefi()
     check_tainted_kmods()
+    check_readonly_mounts()
 
 
 def check_uefi():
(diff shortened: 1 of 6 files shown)


convert2rhel/checks.py
```diff
@@ -78,6 +77,32 @@ def check_tainted_kmods():
         )
 
 
+def check_readonly_mounts():
```
ZhukovGreen, line 80: This function should be called in `perform_pre_checks` (not in main). So we'll have all pre-checks in one place. Does it make sense?

convert2rhel/unit_tests/checks_test.py
```diff
@@ -489,15 +479,60 @@ def test_check_uefi_efi_detected(self):
...
+    def test_mounted_are_readonly(self):
+        self.assertRaises(SystemExit, checks.check_readonly_mounts)
+        self.assertEqual(len(checks.logger.critical_msgs), 1)
+        self.assertTrue(
+            "Stopping conversion due to read-only mount to /mnt directory"
+            in checks.logger.critical_msgs[0]
+        )
         self.assertTrue(
-            "Converting BIOS system" in checks.logger.debug_msgs[0]
+            "Stopping conversion due to read-only mount to /sys directory"
```
ZhukovGreen, line 536: Nit: In pytest you would be able to do all these tests as one and parameterized it:)

ZhukovGreen (commented):
@fellipeh one important note there

convert2rhel/checks.py
fellipeh: Totally! Fixed in a new commit!

ZhukovGreen (approved):
Looks good to me

bocekm (approved):

PR #254, opened 2021-04-14
Improve messaging around the obsoleted --variant option

bocekm:
- The option was basically now showing in the help message.
- The warning message was not shown when the -v or --variant was used without a value.
- The help message was missing the point that RHEL 8 comes with no variants.

Resolves: https://issues.redhat.com/browse/OAMG-4558

### DIFF
--- a/convert2rhel/unit_tests/toolopts_test.py
+++ b/convert2rhel/unit_tests/toolopts_test.py
@@ -20,33 +20,43 @@
 
 import sys
 import unittest
+import pytest
 
 import convert2rhel.toolopts
+import convert2rhel.utils
 
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
 from convert2rhel.toolopts import tool_opts
 
 
-class TestToolopts(unittest.TestCase):
-    def _params(params):  # pylint: disable=E0213
-        return sys.argv[0:1] + params
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
 
+
+def mock_cli_arguments(args):
+    """Return a list of cli arguments where the first one is always the name of the executable, followed by 'args'."""
+    return sys.argv[0:1] + args
+
+
+class TestToolopts(unittest.TestCase):
     def setUp(self):
         tool_opts.__init__()
 
-    @unit_tests.mock(sys, "argv", _params(["--username", "uname"]))
+    @unit_tests.mock(sys, "argv", mock_cli_arguments(["--username", "uname"]))
     def test_cmdline_interactive_username_without_passwd(self):
         convert2rhel.toolopts.CLI()
         self.assertEqual(tool_opts.username, "uname")
         self.assertFalse(tool_opts.credentials_thru_cli)
 
-    @unit_tests.mock(sys, "argv", _params(["--password", "passwd"]))
+    @unit_tests.mock(sys, "argv", mock_cli_arguments(["--password", "passwd"]))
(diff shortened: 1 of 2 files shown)


convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -68,3 +71,17 @@ def test_cmdline_disablerepo_defaults_to_asterisk(self):
...
+@pytest.mark.parametrize(
+    ("argv"), [
+        _params(["-v", "Server"]),
+        _params(["--variant", "Client"]),
+        _params(["--variant"])
+    ]
+)
+def test_cmdline_obsolete_variant_option(argv, monkeypatch, caplog):
+    monkeypatch.setattr(sys, "argv", argv)
+    monkeypatch.setattr(convert2rhel.utils, "ask_to_continue", lambda: None)
```
ZhukovGreen, line 85: lambda: None is fine, but better to create this mock as mock.Mock() object. This will give you a wider set of things you can check. So you can assert if this object was called, check with which arguments it was called, how many times it was called, or that it wasn't called.

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -68,3 +71,17 @@ def test_cmdline_disablerepo_defaults_to_asterisk(self):
...
     def test_cmdline_exits_on_empty_enablerepo_with_disable_submgr(self):
         self.assertRaises(SystemExit, convert2rhel.toolopts.CLI)
+
+
+@pytest.mark.parametrize(
+    ("argv"), [
+        _params(["-v", "Server"]),
+        _params(["--variant", "Client"]),
+        _params(["--variant"])
+    ]
```
ZhukovGreen, line 81: You're testing only when if the condition is True, but also it should be tested when it False. In this case you can check that ask_to_continue wasn't called, or log records doesn't contains something ...

ZhukovGreen (commented):
@bocekm left some minor things, which might improve the quality of testing

convert2rhel/unit_tests/toolopts_test.py
bocekm: I've seen that in your other tests. But the question to continue does not seem that important from my POV so I intentionally skipped checking it. But since you're mentioning it, I'll add it :)

ZhukovGreen (approved):
@bocekm nice!

fellipeh (dismissed):
LGTM

convert2rhel/unit_tests/toolopts_test.py
ZhukovGreen: yes, reviewing helps to see the same things from the different perspective:)

bocekm:
Rebased to include PR #252.

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -20,17 +20,26 @@
...
 
-class TestToolopts(unittest.TestCase):
-    def _params(params):  # pylint: disable=E0213
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
+
+
+def _params(params):  # pylint: disable=E0213
```
SpyTec, line 38: pylint rule disabling can be removed, and naming could be improved. Suggestion could probably be described better too, append isn't entirely correct

```suggestion
def append_sys_params(params):
```

kokesak:
Test passed using convert2rhel-0.20-1.20210415193341589969.pr254.16.g2005f99.el7.noarch and convert2rhel-0.20-1.20210415193341589969.pr254.16.g2005f99.el8.noarch.

fellipeh (approved):

PR #262, opened 2021-04-21
Delete temp folder

hosekadam:
The utility copies bunch of files to /var/lib/convert2rhel/ (mainly backup for the rollback) and they are not being removed upon uninstalling the rpm. Then it's just a junk data.

I've added the remove of data as the absolutely last step of conversion.

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -103,6 +103,9 @@ def main():
         loggerinst.task("Final: rpm files modified by the conversion")
         systeminfo.system_info.modified_rpm_files_diff()
 
+        loggerinst.task("Final: Remove temporary folder %s" % utils.TMP_DIR)
+        utils.remove_tmp_dir()
+
         loggerinst.info("\nConversion successful!\n")
 
         # restart system if required
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -19,6 +19,14 @@ import re
 import sys
 import unittest
 
+import pytest
+
+
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
+
 from collections import namedtuple
 
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
@@ -280,3 +288,22 @@ def test_get_rpm_header(monkeypatch):
     rpm = get_rpm_mocked()
     monkeypatch.setattr(utils, "rpm", rpm)
     assert utils.get_rpm_header("/path/to.rpm", _open=mock.mock_open())[rpm.RPMTAG_NAME] == "pkg1"
+
+
+@pytest.mark.parametrize("dir_name", ("/existing", "/nonexisting", None))
+# TODO change to tmpdir fixture
+def test_remove_tmp_dir(monkeypatch, dir_name, caplog, tmpdir):
+    if dir_name == "/existing":
+        path = str(tmpdir.mkdir(dir_name))
+    else:
+        path = dir_name
+    monkeypatch.setattr(utils, "TMP_DIR", value=path)
+
+    utils.remove_tmp_dir()
+
+    if dir_name == "/existing":
+        assert "Temporary folder " + str(path) + " removed" in caplog.text
+    elif dir_name == "/nonexisting":
+        assert "Failed removing temporary folder " + dir_name in caplog.text
+    else:
+        assert "TypeError error while removing temporary folder " in caplog.text
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -274,6 +274,17 @@ def get_traceback_str():
     return "".join(traceback.format_exception(exc_type, exc_value, exc_traceback))
 
 
+def remove_tmp_dir():
+    """Remove temporary folder (TMP_DIR), not needed post-conversion."""
+    try:
+        shutil.rmtree(TMP_DIR)
+        loggerinst.info("Temporary folder %s removed" % TMP_DIR)
+    except OSError as err:
+        loggerinst.warn("Failed removing temporary folder %s\nError (%s): %s" % (TMP_DIR, err.errno, err.strerror))
+    except TypeError:
+        loggerinst.warn("TypeError error while removing temporary folder %s" % TMP_DIR)
+
+
 class DictWListValues(dict):
     """Python 2.4 replacement for Python 2.5+ collections.defaultdict(list)."""
 
--- a/plans/integration/conversion/delete-temp-folder/main.fmf
+++ b/plans/integration/conversion/delete-temp-folder/main.fmf
@@ -1,8 +1,13 @@
 discover+:
     filter+:
         - 'tag: delete-temp-folder'
-        - 'tag: centos7'
 provision:
     how: libvirt
     develop: true
-    origin_vm_name: c2r_centos7_template
+
+/centos7:
+    discover+:
+        filter+:
+            - 'tag: centos7'
+    provision+:
+        origin_vm_name: c2r_centos7_template
--- a/plans/integration/conversion/delete-temp-folder/main.fmf
+++ b/plans/integration/conversion/delete-temp-folder/main.fmf
@@ -1,0 +1,8 @@
+discover+:
+    filter+:
+        - 'tag: delete-temp-folder'
+        - 'tag: centos7'
+provision:
+    how: libvirt
+    develop: true
+    origin_vm_name: c2r_centos7_template
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -6,4 +6,5 @@ prepare:
     - name: main preparation step
       how: ansible
       playbook: tests/ansible_collections/main.yml
-environment_from: .env
+environment-file:
+  - .env
--- a/tests/integration/conversion/delete-temp-folder/main.fmf
+++ b/tests/integration/conversion/delete-temp-folder/main.fmf
@@ -1,0 +1,10 @@
+summary: delete-temporary-folder
+tag+:
+  - centos7
+  - delete-temp-folder
+adjust:
+  enabled: false
+  when: >
+    distro != centos-7
+test: |
+  pytest -svv
--- a/tests/integration/conversion/delete-temp-folder/test_delete_temp_folder.py
+++ b/tests/integration/conversion/delete-temp-folder/test_delete_temp_folder.py
@@ -1,0 +1,7 @@
+import os
+
+
+def test_temp_folder():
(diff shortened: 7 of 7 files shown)


convert2rhel/main.py
```diff
@@ -100,6 +100,9 @@ def main():
         loggerinst.task("Final: rpm files modified by the conversion")
         systeminfo.system_info.modified_rpm_files_diff()
 
+        # Remove temporary folder with backup - last step of conversion
+        utils.remove_tmp_dir()
```
bocekm, line 104:
```suggestion
        loggerinst.task("Final: Remove temporary folder %s" % TMP_DIR)
        utils.remove_tmp_dir()
```
Without this, the info log message printed by `remove_tmp_dir` would look like they relate to the previous task ("Final: rpm files modified by the conversion").

bocekm (changes requested):

convert2rhel/utils.py
```diff
@@ -268,6 +268,16 @@ def get_traceback_str():
     """Get a traceback of an exception as a string."""
     exc_type, exc_value, exc_traceback = sys.exc_info()
     return "".join(traceback.format_exception(exc_type, exc_value, exc_traceback))
+                                              
+
+def remove_tmp_dir():
+    """Removes whole temporary folder /var/lib/convert2rhel/"""
```
SpyTec, line 274:
```suggestion
    """Remove temporary folder (TMP_DIR), not needed post-conversion"""
```

convert2rhel/utils.py
```diff
@@ -268,6 +268,16 @@ def get_traceback_str():
...
+    try:
+        shutil.rmtree(TMP_DIR)
+        loggerinst.info("Temporary folder /var/lib/convert2rhel/ removed")
+    except OSError as err:
+        loggerinst.info("Failed removing backup folder /var/lib/convert2rhel/")
```
SpyTec, line 279: If we change TMP_DIR the logger would be inaccurate, and if we did remember to change this we'd have more work for ourselves. Better to use the existing variable
```suggestion
        loggerinst.info("Temporary folder %s removed" % TMP_DIR)
    except OSError as err:
        loggerinst.info("Failed removing backup folder %s" % TMP_DIR)
```

convert2rhel/main.py
```diff
@@ -100,6 +100,10 @@ def main():
         loggerinst.task("Final: rpm files modified by the conversion")
         systeminfo.system_info.modified_rpm_files_diff()
 
+        # Remove temporary folder with backup - last step of conversion
```
SpyTec, line 103: Method and task is self-explanatory, the comment here is not needed. As a rule of thumb for comments, if something is not evidently clear by the way it's written and it cannot be refactored it might make sense to explain what it does. 

Likewise, comments explaining why we're doing something is almost always a good thing. This is most useful within docstrings to explain methods, but can be inline if the method is a bit larger and has a lot of side-effects.

Oh and comments can also indicate start and ends of certain things, like we do here in `main.py` with `# begin conversion process` etc.
```suggestion
```

SpyTec (changes requested):
Missing tests for `remove_tmp_dir`, otherwise minor issues!

convert2rhel/unit_tests/utils_test.py
```diff
@@ -229,3 +240,21 @@ def test_get_rpm_path_from_yumdownloader_output(self):
 
     def test_is_rpm_based_os(self):
         assert is_rpm_based_os() in (True, False)
+
+@pytest.mark.parametrize("raise_exception", [True, False])
+def test_remove_tmp_dir(monkeypatch, raise_exception):
```
ZhukovGreen, line 245: This test can be done better. For example, you force the test to raise OSError, but here could be the bug. Let's say some other exception will be raised. I think you tests scenarios might be as 
1. You mock TMP_DIR and assign it to tmpdir fixture(https://github.com/oamg/convert2rhel/blob/main/convert2rhel/unit_tests/conftest.py#L22), then test your function and log messages (good test)
2. You mock the TMP_DIR to return some not existing path and ensure that you capture exception correctly and correct logs were emited (bad test)

convert2rhel/utils.py
```diff
@@ -268,6 +268,16 @@ def get_traceback_str():
     """Get a traceback of an exception as a string."""
     exc_type, exc_value, exc_traceback = sys.exc_info()
     return "".join(traceback.format_exception(exc_type, exc_value, exc_traceback))
+                                              
+
+def remove_tmp_dir():
+    """Remove temporary folder (TMP_DIR), not needed post-conversion"""
```
ZhukovGreen, line 274: Nit: There is https://www.python.org/dev/peps/pep-0257/ which defines Python docstrings convention. It requires for one-line docstring to end with the dot.

Nit: Also, it might be a little better to rephrase it into something like:
"""Remove temporary directory (TMP_DIR), after successful conversion."""

ZhukovGreen (commented):
@hosekadam nice work, Adam. I think you can improve the test, otherwise, your code is very good.

convert2rhel/utils.py
SpyTec: Just adding the dot here is fine. Message is good as it is since it indicates the reason why we remove the tmp dir
```suggestion
    """Remove temporary folder (TMP_DIR), not needed post-conversion."""
```

convert2rhel/unit_tests/utils_test.py
```diff
@@ -229,3 +240,21 @@ def test_get_rpm_path_from_yumdownloader_output(self):
...
+    shutil_rmtree = mock.Mock()
+    monkeypatch.setattr(shutil, "rmtree", value=shutil_rmtree)
+    loggerinst_info = mock.Mock()
```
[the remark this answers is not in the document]
SpyTec: I'm presuming you'd write it as such, right @ZhukovGreen?

```py
def test_remove_tmp_dir(monkeypatch, caplog, raise_exception):
    # ...
    tmp_dir_msg = "Temporary folder %s removed" % utils.TMP_DIR
    if raise_exception:
        assert tmp_dir_msg not in caplog.text
    else:
        assert tmp_dir_msg in caplog.text
```

convert2rhel/unit_tests/utils_test.py
SpyTec: The only external factor (checking TMP_DIR on the system) is mocked out and we only check our branches. `shutil.rmtree` can raise TypeError if `TMP_DIR` is undefined/None and `FileNotFoundError` if TMP_DIR doesn't exist. The latter is not necessary to test. But the former could be tested like so:

```py
    assert len(utils.TMP_DIR) > 0
    utils.remove_tmp_dir()
    shutil_rmtree.assert_called_once_with(utils.TMP_DIR)
```

This asserts that `utils.TMP_DIR` is defined and that we call `shutils.rmtree` with it

ZhukovGreen: @SpyTec I take as you're evolving my narrative here. So when you're mocking the TMP_DIR, with tmpdir - it is under your control and now you can add files there, make it empty, or make it not exist. And ensure that you properly handling exceptions. The key here is not guessing what will be raised, but model some real world scenarios. As I remember for example that in Centos 6 there is no such exception like FileNotFound

convert2rhel/unit_tests/utils_test.py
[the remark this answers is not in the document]
ZhukovGreen: for example

convert2rhel/utils.py
ZhukovGreen: This is Nit - so up to you guys

ZhukovGreen (commented):
Some comments

convert2rhel/unit_tests/utils_test.py
hosekadam: @SpyTec @ZhukovGreen I've updated the unit test. Please check, if I met with your ideas :)

convert2rhel/unit_tests/utils_test.py
```diff
@@ -280,3 +288,22 @@ def test_get_rpm_header(monkeypatch):
...
+
+
+@pytest.mark.parametrize("path", (unit_tests.TMP_DIR, unit_tests.NONEXISTING_DIR, None))
+def test_remove_tmp_dir(monkeypatch, path, caplog):
+    monkeypatch.setattr(utils, "TMP_DIR", value=path)
+    if path == unit_tests.TMP_DIR:
+        try:
+            utils.mkdir_p(path)
+        except OSError:
+            pass
```
ZhukovGreen, line 300: it is not finished yet, right? You're going to use tmtdir

convert2rhel/utils.py
```diff
@@ -274,6 +274,18 @@ def get_traceback_str():
...
 
 
+def remove_tmp_dir():
+    """Remove temporary folder (TMP_DIR), not needed post-conversion."""
+    try:
+        shutil.rmtree(TMP_DIR)
+        loggerinst.info("Temporary folder %s removed" % TMP_DIR)
+    except OSError as err:
+        loggerinst.info("Failed removing temporary folder %s" % TMP_DIR)
+        loggerinst.info("Error (%s): %s" % (err.errno, err.strerror))
```
bocekm, line 284:
```suggestion
        loggerinst.warn("Failed removing temporary folder %s" % TMP_DIR)
        loggerinst.warn("Error (%s): %s" % (err.errno, err.strerror))
```

convert2rhel/utils.py
```diff
@@ -274,6 +274,18 @@ def get_traceback_str():
...
+    except Exception:
+        loggerinst.info("Other error while removing temporary folder %s" % TMP_DIR)
```
bocekm, line 286:
```suggestion
```
I don't expect other than OSError tree exceptions. And this message would not tell much anyway.

bocekm (changes requested):

convert2rhel/utils.py
ZhukovGreen: Totally agree, moreover bare exception is antipattern in python. The reason is that you need to be explicit about the program behaviour, otherwise you can miss some important exception from your radar

hosekadam: @ZhukovGreen @bocekm Sorry guys, this isn't actual version. As I had problems with the `tmpdir` I didn't update the branch online. 

For the exceptions: now I'm catching `TypeError` due `None` value of `TMP_DIR`. I'll update the branch when I finish the integration test

hosekadam:
Updated integration test. But isn't tested due some problems on my machine.

ZhukovGreen (dismissed):
Nice

ZhukovGreen (approved):
Now test PASSED

ZhukovGreen:
depend on #319

bocekm (approved):

SpyTec (approved):

PR #264, opened 2021-04-29
Feature/pre commit hooks black isort

ZhukovGreen:
https://issues.redhat.com/browse/OAMG-4765

TODO:
Acceptance criteria:
- [x] pre-commit config present in the project tree, and contains two hooks : black and isort
- [x] configurations of black and isort placed in pyproject.toml
- [x] Add a workflow of running pre-commit (pre-commit run --all-files) into the githab actions for cetnos8 image only
- [x] Document the work with the pre-commit hooks
- [x] Test with basic convertability tests

Notes for the reviewer:
Review by commits. 
https://github.com/oamg/convert2rhel/pull/264/commits/5a2aff98ffde5c6048bbdeaf295ba22af9bf4e49 introduces configs
https://github.com/oamg/convert2rhel/pull/264/commits/e52858e87906bd1a20f295e8a55bb88bf40d01a3 - introduces changes by the pre-commit


ZhukovGreen:
@oamg/convert2rhel please review

ZhukovGreen:
ref #153

pyproject.toml
```diff
@@ -0,0 +1,97 @@
+[tool.black]
+line-length=119
```
ZhukovGreen, line 2: I would vote for 79 line length. 119 - is not readable and too long for me.
@oamg/convert2rhel wdyt?

ZhukovGreen:
Tested on CentOS 6,7,8 and OL 6, 7,8

pyproject.toml
fellipeh: @ZhukovGreen  I just think 79 is too small... in some cases (eg: conditionals) will be weird to read.. I vote between this two: 88 line length as `black` default.

fellipeh (commented):
@ZhukovGreen  we'll some command to run isort into our files before the commit ?

ZhukovGreen:
Yes, similar to black

.pre-commit-config.yaml
```diff
@@ -0,0 +1,25 @@
+minimum_pre_commit_version: "2.9.0"
```
bocekm, line 1: Why is this needed?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,25 @@
...
+repos:
+  - repo: "https://github.com/psf/black"
+    rev: "21.4b2"
```
bocekm, line 4: Can't we just omit the `rev` to always get the latest version?

CONTRIBUTING.md
```diff
@@ -6,8 +6,11 @@
 
 ## Developing locally
 
-The commands below installs a python3 virtual environment
-with all the necessary dependencies installed:
+The commands below creates a python3 virtual environment,
+with all the necessary dependencies installed,
+builds needed images,
+creates dotenv file, and
```
bocekm, line 12:
```suggestion
The commands below create a python3 virtual environment with all the necessary dependencies installed,
build needed images,
create .env file, and
```

pyproject.toml
bocekm: It was 80 when I started. Then we decided for 120 about a year ago. So I vote for not getting back to the discussion and stick to the length we have now. I can find the PR where we discussed it if you want.

```suggestion
line-length=120
```

pyproject.toml
```diff
@@ -0,0 +1,97 @@
...
+[tool.coverage.run]
+branch = false
+source = ["convert2rhel"]
+omit = ["convert2rhel/unit_tests"]
+
+[tool.coverage.paths]
+source = ["convert2rhel/"]
+
+[tool.coverage.report]
+fail_under = 87
```
bocekm, line 28: Last week we moved it up to 80. Are you changing the coverage check so that it reports 87% now?

pyproject.toml
```diff
@@ -0,0 +1,97 @@
...
+    "^get.*$",
+    "^factory$",
+    "^_.+$",
+]
+verbose = 2
+quiet = false
+whitelist-regex = []
+color = true
+
+[tool.poetry]
```
bocekm, line 64: What is poetry? How can I verify all its setup is correct?

pyproject.toml
```diff
@@ -0,0 +1,97 @@
+[tool.black]
+line-length=119
+target-version=["py38"]
```
bocekm, line 3: Are you confident that the black formatting for Python 3.8 won't use formatting not compatible with Python 2.6?

bocekm (changes requested):

pyproject.toml
ZhukovGreen: Ok, no need to block on this. Applied

.pre-commit-config.yaml
ZhukovGreen: This is similar to the need of pinning versions in requirements.txt. If we allow having the latest version, then we can't guarantee reproducibility. In case of pre-commit it is even better - you just run:
pre-commit auto-update and these versions are updated automatically

.pre-commit-config.yaml
ZhukovGreen: the same as before

CONTRIBUTING.md
ZhukovGreen: applied

pyproject.toml
ZhukovGreen: that was added accidentally - now no poetry here

pyproject.toml
ZhukovGreen: Tested with py26 unit tests and centos6 vagrant box

pyproject.toml
ZhukovGreen: you're right, made a little mistake in migrating the config. Thx

bocekm (approved):

fellipeh (approved):

bocekm (approved):

pyproject.toml
```diff
@@ -0,0 +1,30 @@
+[tool.black]
+line-length=120
+target-version=["py38"]
+
+[tool.isort]
```
SpyTec, line 5: @ZhukovGreen I noticed this in the doc

https://pycqa.github.io/isort/docs/configuration/black_compatibility/

https://pycqa.github.io/isort/docs/configuration/profiles/

From that we're missing this, should we add that or do the black/isort not collide with each other with current config?
```
force_grid_wrap: 0
ensure_newline_before_comments: true
```

ZhukovGreen: They are not collide in this particular setup. Time will show, however in my projects I never seen the need to solve the conflict. Time will show.

bocekm: I like Eric's suggestion. It would make the config a bit more concise as well:
```
atomic=true
line_length = 120
lines_after_imports=2
lines_between_types=1
use_parentheses = true
balanced_wrapping = true
include_trailing_comma = true
multi_line_output = 3
```
down to:
```
profile = "black"
atomic = true
line_length = 120
lines_after_imports = 2
lines_between_types = 1
balanced_wrapping = true
```

ZhukovGreen: OK, should be addressed by #279

PR #275, opened 2021-05-13
Improve the preparation of the testing environment

ZhukovGreen:
- Instead of separate ansible playbooks the configurable ansible roles are used
- By adjusting ansible group_vars/all.yml it is possible to specify the rpm provider and build mechanism

The tmt-pytest framework documentation was improved and simplified


ZhukovGreen:
@oamg/convert2rhel for a review

bocekm (changes requested):
What's the purpose of the empty `defaults/main.yml` files?

tests/ansible_collections/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+- hosts: all
+  roles:
+
+    - role: oracle-linux-specific
+      when: ansible_facts['distribution'] == "OracleLinux"
+      tags:
+        - oracle
```
bocekm, line 8: What's the purpose of the tags?

bocekm (commented):
I've tried both the local and URL options and both work. Good job.

tests/ansible_collections/main.yml
ZhukovGreen: Similar to defaults in roles. Agree let's remove it unless we'll feel a need in the possibility to reuse roles for maunall runs. The tags allow running certain scenarios, i.e.
ansible-playbook --tags packaging will run only roles in sequence with tag packaging.

ZhukovGreen (commented):
Fixed all comments @bocekm

tests/ansible_collections/roles/oracle-linux-specific/tasks/boot_standard_kernel.yml
```diff
@@ -1,3 +1,4 @@
+---
 - name: Install standard kernel
   yum:
     name: kernel
```
ZhukovGreen, line 4: Inna found an issue, with this task when her machine had kernel-transition package installed, this make this tasks skip to install the RHCK kernel. Should be changed to something like kernel.*

bocekm (approved):

tests/ansible_collections/roles/oracle-linux-specific/tasks/boot_standard_kernel.yml
bocekm: We agreed on resolving this in a separate PR.

PR #279, opened 2021-05-18
Fix/improve pre commit config and docs

ZhukovGreen:
Added black profile + more docs about pre-commit

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,10 @@
 
 ## Developing locally
 
+Required dependencies:
+- virtualenv
+- pre-commit
+
 The commands below create a python3 virtual environment with all the necessary dependencies installed,
 build needed images,
 create .env file, and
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -35,3 +39,22 @@ Or locally
...
+## Work with pre-commit hooks
+
+```bash
+# installs hooks into the repo (included into make install)
+pre-commit install --install-hooks
+
+# run hooks for staged files
+pre-commit run
+
+# run hooks for all files in repo
```
SpyTec, line 52: Running the tool here and not precommit hook
```suggestion
# run pre-commit for staged files
pre-commit run

# run pre-commit for all files in repo
```

convert2rhel/unit_tests/main_test.py
```diff
@@ -25,17 +25,8 @@
...
-    main,
-    pkghandler,
-    redhatrelease,
-    repo,
-    subscription,
-    unit_tests,
-    utils,
-)
+from convert2rhel import main  # Imports unit_tests/__init__.py
+from convert2rhel import cert, checks, pkghandler, redhatrelease, repo, subscription, unit_tests, utils
```
SpyTec, line 29: Oh I think this comment got broken with the initial black refactor. Comment pointed to `unit_tests` initially

Did `main` import here only split from the rest due to the comment? If you remove the comment does the formatting change?

ZhukovGreen: yes, that was due to comment

SpyTec (approved):

PR #281, opened 2021-05-18
Feature/ensure test system is up to date

ZhukovGreen:
ref OAMG-4926

This feature takes care of automatic update and reboot (if needed) the testing environment for each run of the integration tests

### DIFF
--- a/tests/ansible_collections/roles/update-system/tasks/update-system.yml
+++ b/tests/ansible_collections/roles/update-system/tasks/update-system.yml
@@ -7,5 +7,7 @@
 
 - name: Reboot
   reboot:
+  # we need reboot only CentOS, because OL will be
+  #   rebooted in later steps (booting into standard kernel)
   when: (ansible_facts['distribution'] == "CentOS") and
-        (update.msg != "Nothing to do")
+        (update.changed)
--- a/tests/ansible_collections/roles/update-system/tasks/update-system.yml
+++ b/tests/ansible_collections/roles/update-system/tasks/update-system.yml
@@ -1,0 +1,11 @@
+---
+- name: Update system
+  yum:
+    name: "*"
+    state: latest
+  register: update
+
+- name: Reboot
+  reboot:
+  when: (ansible_facts['distribution'] == "CentOS") and
+        (update.msg != "Nothing to do")
--- a/tests/ansible_collections/main.yml
+++ b/tests/ansible_collections/main.yml
@@ -2,6 +2,8 @@
 - hosts: all
   roles:
 
+    - role: update-system
+
     - role: oracle-linux-specific
       when: ansible_facts['distribution'] == "OracleLinux"
 
--- a/tests/ansible_collections/roles/packaging/tasks/build_rpm.yml
+++ b/tests/ansible_collections/roles/packaging/tasks/build_rpm.yml
@@ -2,5 +2,6 @@
 - name: Building rpms
   delegate_to: localhost
   shell: make rpms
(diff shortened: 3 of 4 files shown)


tests/ansible_collections/roles/update-system/tasks/update-system.yml
```diff
@@ -0,0 +1,11 @@
...
+- name: Update system
+  yum:
+    name: "*"
+    state: latest
+  register: update
+
+- name: Reboot
+  reboot:
+  when: (ansible_facts['distribution'] == "CentOS") and
+        (update.msg != "Nothing to do")
```
bocekm, line 11: Shouldn't it be:
```suggestion
        ("Nothing to do" not in update.msg)
```

tests/ansible_collections/roles/update-system/tasks/update-system.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Update system
+  yum:
+    name: "*"
+    state: latest
+  register: update
+
+- name: Reboot
+  reboot:
+  when: (ansible_facts['distribution'] == "CentOS") and
```
bocekm, line 10: Why only on CentOS Linux?

bocekm (changes requested):

tests/ansible_collections/roles/update-system/tasks/update-system.yml
ZhukovGreen: I checked the message is exactly `Nothing to do`, thus I added this type of check

tests/ansible_collections/roles/update-system/tasks/update-system.yml
ZhukovGreen: Because in oracle Linux we are doing the reboot after replacing the kernel

tests/ansible_collections/roles/update-system/tasks/update-system.yml
bocekm: What I'm getting with yum update is:
```
# yum update
Last metadata expiration check: 2:58:02 ago on Wed 19 May 2021 01:10:38 PM UTC.
Dependencies resolved.
Nothing to do.
Complete!
```
But the other messages maybe appear just in terminal, I don't know.

tests/ansible_collections/roles/update-system/tasks/update-system.yml
bocekm: Ok, please add a comment here so it's clear to others who look at it.

tests/ansible_collections/roles/update-system/tasks/update-system.yml
ZhukovGreen: I relied on centos8 tests, where:
```
            out: 
            out: TASK [update-system : Update system] *******************************************
            out: ok: [c2r_centos8_template-clone] => {"changed": false, "msg": "Nothing to do", "rc": 0, "results": []}
            out: 
```

ZhukovGreen: But you're right - now I am testing it on centos 7 and there is a different output

ZhukovGreen: Found a cleaner way to do this:
```
  when: (ansible_facts['distribution'] == "CentOS") and
        (update.changed)
```

SpyTec (approved):

bocekm (approved):

PR #282, opened 2021-05-18
Feature/add possibility to skip function execution

ZhukovGreen:
closes OAMG-4533, OAMG-4777

Provides the mechanism to skip the execution of any function and replace the returned value of this function with the help of environment variable:
```
CONVERT2RHEL_UNSUPPORTED=1 \
CONVERT2RHEL_DEVEL_SKIP=tainted,kmods_check \
convert2rhel ...
```
will skip the execution of the tainted kernel modules and compatibility of kernel modules checks.

The usage:
```python
@available_for_skipping(aliases=("kmods_check", "supported_kmods"))
def ensure_compatibility_of_kmods():
    ...
```
TODO:
- [ ] integration tests

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -24,7 +24,7 @@ import subprocess
 
 from convert2rhel.pkghandler import get_installed_pkg_objects, get_pkg_fingerprint
 from convert2rhel.systeminfo import system_info
-from convert2rhel.utils import get_file_content, run_subprocess
+from convert2rhel.utils import ask_to_continue, get_file_content, run_subprocess
 
 
 logger = logging.getLogger(__name__)
@@ -44,6 +44,7 @@ COMPATIBLE_KERNELS_VERS = {
 
 def perform_pre_checks():
     """Early checks after system facts should be added here."""
+    check_environment()
     check_uefi()
     check_tainted_kmods()
     check_readonly_mounts()
@@ -393,3 +394,9 @@ def _bad_kernel_substring(kernel_release):
         )
         return True
     return False
+
+
+def check_environment():
+    if "CONVERT2RHEL_NOT_SUPPORTED" in os.environ:
+        logger.warning("Variable CONVERT2RHEL_UNSUPPORTED has been detected. Proceed at your own risk.")
+        ask_to_continue()
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -24,7 +24,7 @@ import subprocess
 
 from convert2rhel.pkghandler import get_installed_pkg_objects, get_pkg_fingerprint
 from convert2rhel.systeminfo import system_info
-from convert2rhel.utils import ask_to_continue, get_file_content, run_subprocess
+from convert2rhel.utils import ask_to_continue, available_for_skipping, get_file_content, run_subprocess
 
 
 logger = logging.getLogger(__name__)
@@ -65,6 +65,7 @@ def check_uefi():
     logger.debug("Converting BIOS system")
 
 
+@available_for_skipping(aliases=("tainted", "tainted_kernel", "tainted_kmods"))
 def check_tainted_kmods():
     """Stop the conversion when a loaded tainted kernel module is detected.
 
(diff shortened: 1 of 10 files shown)


convert2rhel/checks.py
```diff
@@ -389,3 +393,9 @@ def _bad_kernel_substring(kernel_release):
         )
         return True
     return False
+
+
+def check_environment():
+    if "CONVERT2RHEL_NOT_SUPPORTED" in os.environ:
```
SpyTec, line 399: We should use `CONVERT2RHEL_UNSUPPORTED`
```suggestion
    if "CONVERT2RHEL_UNSUPPORTED" in os.environ:
```

SpyTec (changes requested):

convert2rhel/checks.py
ZhukovGreen: because?

bocekm: I'm fine with unsupported. It's shorter and has the same meaning :)

ZhukovGreen: The prefix of our vars doesn't say we are seeking for shorter vars:) what about C2R_UNSUPPORTED?

bocekm: I expect the variable to be used not only by developers but also by customers. So far we haven't used the abbreviated form C2R or c2r anywhere publicly and I'd refrain from doing so for professional reasons.

SpyTec: Part of the issue it was agreed to use `CONVERT2RHEL_UNSUPPORTED`. It's also similar to leapp's `LEAPP_UNSUPPORTED`. Makes it more consistent 

Currently you have logging that say `Variable CONVERT2RHEL_UNSUPPORTED has been detected` as well

ZhukovGreen:
Two tests passed at https://github.com/oamg/convert2rhel/commit/80ecdbb85efb689ccc6129a7be6da4e933c90e3b

ZhukovGreen:
Two tests passed at https://github.com/oamg/convert2rhel/commit/0d32e4fcdfc9832096ce9b7e9a829c95d7edc3a6

fellipeh (dismissed):
LGTM

ZhukovGreen:
@bocekm @SpyTec I am merging it?

convert2rhel/checks.py
```diff
@@ -64,6 +65,7 @@ def check_uefi():
     logger.debug("Converting BIOS system")
 
 
+@available_for_skipping(aliases=("tainted", "tainted_kernel", "tainted_kmods"))
```
SpyTec, line 68: This approach wouldn't solve OAMG-4533 in a way that still warns the user. We'd be completely blind to any kernel modules that is unsupported or tainted. The only warning we get is that the method is skipped. It would be to our advantage to still print kmods but just changing it to a warning as opposed to critical

Would it be better if we have a decorator that instead just catches the SysExit? Or get another variable like OAMG-4533 discusses with `CONVERT2RHEL_DEVEL_KMOD_CHECK_WARN_ONLY`

SpyTec (changes requested):

bocekm (changes requested):

convert2rhel/checks.py
ZhukovGreen: In terms of `Allow overriding the kmod inhibitor` it is solved, however it is not verobose, you're right.
I suggest introducing some special option into `available_for_skipping` decorator as a separate MR. The current concept is very flexible for extending.

bocekm: You've a got a good point. I'd also prefer the switch to not block the conversion but still run the check. Even for the purpose of integration testing it's better to run all the code and not skip the checks all together.

But mere catching SystemExit would not solve it since there's a couple of logger.critical() calls in checks.py which are supposed to stop the conversion when something unexpected happens.

bocekm: In other words, it seems to me that the "skip" feature is just an alternative solution to "warn only", both having the same goal.
And the only disadvantage of the "warn only" is that the conversion would take a bit longer. Otherwise it has more advantages - the user gets more information and more of the c2r code is being tested.

My suggestion would be:

In the check functions, instead of calling logger.critical(msg) for inhibiting the conversion, call a new function `inhibit_conversion(msg, warn_only_keyword)`, that would either:

- stop the conversion (the default), or
- print the message just as a warning when `CONVERT2RHEL_DEVEL_WARN_ONLY=override_keyword`

ZhukovGreen:
@oamg/convert2rhel-developers 
Guys came up with a better solution. This MR becomes irrelevant and a new one will be introduced

PR #285, opened 2021-05-19
Improve messaging around the unsupported kmods error

bocekm:
The message around unsupported kernel modules was not asking to make sure the system runs the latest kernel which is an important prerequisite of the kernel module check.

Also, it was stating that disabling the modules will help. But it does not. The module needs to be removed in order to not be picked up by the check.

And printing plainly the `uname -r` call result was confusing.


ZhukovGreen (approved):
@bocekm thank you for this MR. I reran tests without an issue.

ZhukovGreen:
Tests I ran for a different branch:) so they passed. But here was a bug - #286 fixes it

PR #288, opened 2021-05-20
Fix Conversion fails on OL 7 with python-requests

fellipeh:
This PR fix conversion fails on OL7 regarding `python-requests` package.

Jira Ticket: OAMG-4936

PR #290, opened 2021-05-27
Unregister upon finishing a test

bocekm:
We need the system to be unregistered instead of unsubscribed. This way the command fails at the end of a test and the test then fails because of that as well.

Blocks: https://github.com/oamg/convert2rhel/pull/288


ZhukovGreen (approved):
Thanks for this

ZhukovGreen:
@bocekm We can merge it

PR #293, opened 2021-05-27
Rename --disable-submgr option to --no-rhsm

bocekm:
Both are still possible to use and both have the same effect.

The reason is that:
1. we are not disabling the subscription-manager (bad choice of wording) and
2. the same option is used in LEAPP for RHEL system upgrades.


ZhukovGreen (approved):

SpyTec (changes requested):
Needs one unit test to make sure `disable_submgr` still works. Otherwise LGTM

SpyTec (approved):

PR #294, opened 2021-05-27
Allow conversions of offline systems

bocekm:
New option --keep-rhsm is used. The user needs to install the subscription-manager packages before performing the conversion.

To do:

- [x] Verify the subscription-manager is installed when the option is used
- [x] Add unit tests
- [ ] ~~Add integration tests~~ TBD under #310

### DIFF
--- a/convert2rhel/toolopts.py
+++ b/convert2rhel/toolopts.py
@@ -44,6 +44,7 @@ class ToolOpts(object):
         self.arch = None
         self.no_rpm_va = False
         self.disable_colors = False
+        self.keep_rhsm = False
 
         # set True when credentials (username & password) are given through CLI
         self.credentials_thru_cli = False
@@ -61,16 +62,13 @@ class CLI(object):
             "\n"
             "  convert2rhel [-h]\n"
             "  convert2rhel [--version]\n"
-            "  convert2rhel [-u username] [-p password | -f pswd_file]"
-            " [--pool pool_id | -a] [--disablerepo repoid] [--enablerepo"
-            " repoid] [--serverurl url] [--no-rpm-va]"
-            " [--debug] [--restart] [-y]\n"
+            "  convert2rhel [-u username] [-p password | -f pswd_file] [--pool pool_id | -a] [--disablerepo repoid]"
+            " [--enablerepo repoid] [--serverurl url] [--keep-rhsm] [--no-rpm-va] [--debug] [--restart]"
+            " [--disable-colors] [-y]\n"
             "  convert2rhel [--no-rhsm] [--disablerepo repoid]"
-            " [--enablerepo repoid] [--no-rpm-va] [--debug] [--restart] [-y]\n"
-            "  convert2rhel [-k key] [-o organization] [--pool pool_id |"
-            " -a] [--disablerepo repoid] [--enablerepo repoid]"
-            " [--serverurl url] [--no-rpm-va] [--debug]"
-            " [--restart] [--disable-colors] [-y]"
+            " [--enablerepo repoid] [--no-rpm-va] [--debug] [--restart] [--disable-colors] [-y]\n"
+            "  convert2rhel [-k key] [-o organization] [--pool pool_id | -a] [--disablerepo repoid] [--enablerepo"
+            " repoid] [--serverurl url] [--keep-rhsm] [--no-rpm-va] [--debug] [--restart] [--disable-colors] [-y]"
             "\n\n"
             "WARNING: The tool needs to be run under the root user"
         )
@@ -210,6 +208,15 @@ class CLI(object):
             " The default is the Customer Portal Subscription Management service. It is not to be used to specify a"
             " Satellite server. For that, read the product documentation at https://access.redhat.com/.",
         )
+        group.add_option(
+            "--keep-rhsm",
+            action="store_true",
+            help="Keep the already installed Red Hat Subscription Management-related packages. By default,"
+            " during the conversion, these packages are removed, downloaded from verified sources and re-installed."
+            " This option is suitable for environments with no connection to the Internet, or for systems managed by"
+            " Red Hat Satellite. Warning: The system is being re-registered during the conversion and when the"
+            " re-registration fails, there's no automated rollback to the original registration.",
+        )
         self._parser.add_option_group(group)
 
         group = optparse.OptionGroup(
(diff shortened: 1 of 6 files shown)


ZhukovGreen (commented):
I think we need to separate keep-rhsm and keep-subscribed. Also consider if this options are for develop purpose only (then use https://github.com/oamg/convert2rhel/pull/282) otherwise use cli options. Ensure the implementation covers all cases (https://docs.google.com/spreadsheets/d/1hax9e-5oGMolSax915dEB7KqtrqxwlNNlQ0gCcn5Mg0/edit?usp=sharing)

ares:
This does not seem to work. I'm running it like this

```
convert2rhel -y --keep-rhsm --activationkey "convert2rhel_rhe7" --org "Demo" > /root/convert2rhel.log
```

but the output I'm getting is 

```
[06/02/2021 11:54:38] TASK - [Convert: Resolve possible edge cases] *****************************
Checking if java-1.7.0-openjdk is installed.
java-1.7.0-openjdk not installed.

[06/02/2021 11:54:38] TASK - [Convert: Subscription Manager - Download packages] ****************
Skipping due to the use of --keep-rhsm.

[06/02/2021 11:54:38] TASK - [Convert: Subscription Manager - Replace] **************************
Skipping due to the use of --keep-rhsm.

[06/02/2021 11:54:38] TASK - [Convert: Install RHEL certificates for RHSM] **********************
Installing RHEL certificate to the system.
[06/02/2021 11:54:38] DEBUG - Certificate copied to /etc/pki/product-default/.

[06/02/2021 11:54:38] TASK - [Convert: Subscription Manager - Subscribe system] *****************
Building subscription-manager command ... 
    ... activation key not found, username and password required
```

Looking at https://github.com/oamg/convert2rhel/blob/6c56aaeeb48bb9b0016d2f546b3d1352af2ceec6/convert2rhel/subscription.py#L130-L141 it seems that when some steps are skipped due to the `--keep-rhsm`, the `tool_opts.activation_key` is null, even though I specified that on the command line. Without this option all works fine. 

When I added `--keep-rhsm yes` it seems to command it seems to fix the issue. So either the option needs changing or the help should be adjusted.

I was hoping to confirm it works with offline content only, however convert2rhel also requires pexpect package which seems to be only the centos main repo. So first I'll have to sync that. ubi repo seems to only have python36-pexpect.

bocekm:
Thanks, @ares, there was a bug when parsing the keep-rhsm. It should be fine now. 

Regarding the dependencies of convert2rhel, there's more than just pexpect. See the `Requires:` in https://github.com/oamg/convert2rhel/blob/main/packaging/convert2rhel.spec#L34/.

plans/integration/remove-all-submgr-pkgs/main.fmf
```diff
@@ -47,4 +47,4 @@ provision:
     - name: install rhsm
       how: ansible
       playbook:
-        - tests/integration/remove_all_submgr_pkgs/ansible/install_submgr_8.yml
+        - tests/ansible_collections/roles/install-submgr/install_submgr_centos_8.yml
```
ZhukovGreen, line 50: I saw tests/ansible_collections/roles as the roles for the initial preparation of the testing environment. However, I was adding other playbooks in the root of tests/ansible_collections (to be consumed by some specific tests only).

How do you see this tests/ansible_collections/roles/install-submgr/install_submgr_centos_8.yml will be used? Maybe we can create a different directory tree?

bocekm: I considered the ansible_collections as a folder for also roles used by multiple tests, to avoid duplication. At this moment I don't have a problem with the current dir structure. Let me know if you have an idea on how to restructure the tree.

ZhukovGreen: I saw it the same. But I'd do not mix the roles for initial preparation and roles for tests. This will allow us to easy share it then with the leapp team for example

convert2rhel/toolopts.py
```diff
@@ -211,6 +209,15 @@ def _register_options(self):
             " not provided, the subscription-manager defaults will be"
             " used.",
         )
+        group.add_option(
+            "--keep-rhsm",
+            action="store_true",
+            help="Do not remove the already installed Red Hat Subscription Management-related packages. By default,"
```
SpyTec, line 215: Code LGTM, I would change this though as `--keep-rhsm` has a positive connotation while help message starts with a negative one. Comparatively, `--disable-submgr` has negative connotation on both the command and help message.

```suggestion
            help="Keep the already installed Red Hat Subscription Management-related packages. By default,"
```

SpyTec (dismissed):

bocekm:
Manually verified on Oracle and CentOS Linux.
```
[06/30/2021 08:59:14] TASK - [Convert: Subscription Manager - Download packages] ****************
Skipping due to the use of --keep-rhsm.

[06/30/2021 08:59:14] TASK - [Convert: Subscription Manager - Replace] **************************
Skipping due to the use of --keep-rhsm.
```
Steps:
```
yum -y install subscription-manager
curl https://<sat_server>/pub/katello-ca-consumer-latest.noarch.rpm -o /tmp/katello-ca-consumer-latest.noarch.rpm -k
yum install /tmp/katello-ca-consumer-latest.noarch.rpm -y
subscription-manager register --org=<org> --activationkey=<key1>
convert2rhel --keep-rhsm --debug -y --no-rpm-va --org=<org> --activationkey=<key2>
```

convert2rhel/toolopts.py
```diff
@@ -300,6 +307,12 @@ def _process_cli_options(self):
             else:
                 tool_opts.serverurl = parsed_opts.serverurl
 
+        if parsed_opts.keep_rhsm:
+            if parsed_opts.disable_submgr:
```
SpyTec, line 311:
```suggestion
            if parsed_opts.no_rhsm:
```

SpyTec (changes requested):

convert2rhel/toolopts.py
bocekm: It should have been one of these two options:
- `if parsed_opts.no_rhsm or parsed_opts.disable_submgr:`
- `if tool_opts.no_rhsm:`

I've chosen the latter.

SpyTec (approved):

PR #296, opened 2021-05-31
fix: check if subscription manager installed before call unregister

fellipeh:
This commit adds a check if subscription manager is installed before call the unregister process


bocekm:
Not needed in the end. We thought that the following error we were seeing is causing the test failure. But the test was failing because different issue in the test itself (https://github.com/oamg/convert2rhel/pull/196/commits/ac184cc6dc0ddc0a4f87a316bd4db0f77753b83b).
```
Executing a command:
subscription-manager unregister
/bin/sh: subscription-manager: command not found
```

PR #299, opened 2021-06-04
Feature/better handling of releasever

ZhukovGreen:
Use predefined dict for a release version, which  maps CentOS versions to RHEL versions. The config parameter is set to nothing (`releasever=`) by default, however, if set, will take the precedence over the data in predefined dict.

Resolves: https://issues.redhat.com/browse/OAMG-5007
Related: https://bugzilla.redhat.com/show_bug.cgi?id=1962193

TODO:
- [x] integration tests

### DIFF
--- a/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
+++ b/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
@@ -31,6 +31,9 @@ repofile_pkgs =
 # Delimited by any whitespace(s).
 default_rhsm_repoids = rhel-7-server-rpms
 
+# If defined, it overrides the default releasever defined by RELEASE_VER_MAPPING.
+# The value is passed to the yum calls through the --releasever option when accessing RHEL repos. Its purpose is to
+# substitute the $releasever variable in a repository baseurl.
 releasever=
 
 # Some kernel modules move from kernel modules to kernel core. Instead of inhibiting the conversion with a message
--- a/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
+++ b/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
@@ -31,7 +31,7 @@ repofile_pkgs =
 # Delimited by any whitespace(s).
 default_rhsm_repoids = rhel-7-server-rpms
 
-releasever=7Server
+releasever=
 
 # Some kernel modules move from kernel modules to kernel core. Instead of inhibiting the conversion with a message
 # that such a module is not available in RHEL and thus is unsupported, we ignore it.
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -217,7 +217,7 @@ class SystemInfo(object):
             self.logger.critical(
                 "%s of version %d.%d is not allowed for conversion.\n"
                 "Allowed versions are: %s"
-                % (self.name, self.version.major, self.version.minor, RELEASE_VER_MAPPING.keys())
+                % (self.name, self.version.major, self.version.minor, list(RELEASE_VER_MAPPING.keys()))
             )
 
     def _get_kmods_to_ignore(self):
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -14,6 +14,8 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
+import pprint
+
 from collections import namedtuple
 
 from convert2rhel.utils import run_subprocess
@@ -33,10 +35,25 @@ from convert2rhel import logger, utils
 from convert2rhel.toolopts import tool_opts
 
 
+# Supported conversion path to RHEL
+RELEASE_VER_MAPPING = {
+    "CentOS Linux": {
+        "8.4": "8.4",
+        "8.3": "8.3",
+        "7.9": "7.9Server",
+    },
+    "Oracle Linux Server": {
+        "8.4": "8.4",
+        "8.3": "8.3",
+        "7.9": "7.9Server",
+    },
+}
 # For a list of modified rpm files before the conversion starts
 PRE_RPM_VA_LOG_FILENAME = "rpm_va.log"
 # For a list of modified rpm files after the conversion finishes for comparison purposes
 POST_RPM_VA_LOG_FILENAME = "rpm_va_after_conversion.log"
+
+Version = namedtuple("Version", ["major", "minor"])
 
 
 class SystemInfo(object):
@@ -124,7 +141,7 @@ class SystemInfo(object):
             from convert2rhel import redhatrelease
 
             self.logger.critical("Couldn't get system version from %s" % redhatrelease.get_system_release_filepath())
-        version = namedtuple("Version", ["major", "minor"])(int(match.group(1)), int(match.group(2)))
+        version = Version(int(match.group(1)), int(match.group(2)))
 
         self.logger.info("%-20s %d.%d" % ("OS version:", version.major, version.minor))
         return version
@@ -190,7 +207,15 @@ class SystemInfo(object):
         return self._get_cfg_opt("repofile_pkgs").split()
 
     def _get_releasever(self):
-        return self._get_cfg_opt("releasever")
+        might_be_config = self._get_cfg_opt("releasever")
+        try:
+            # return config value or corresponding releasever from the RELEASE_VER_MAPPING
+            return might_be_config or RELEASE_VER_MAPPING[self.name][".".join(map(str, self.version))]
+        except KeyError:
+            self.logger.critical(
+                "Unsupported system name/version met: %s - %s.\n"
+                "Supported paths are:\n%s" % (self.name, self.version, pprint.pformat(RELEASE_VER_MAPPING, width=54))
+            )
 
     def _get_kmods_to_ignore(self):
         return self._get_cfg_opt("kmods_to_ignore").split()
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -35,17 +35,20 @@ from convert2rhel import logger, utils
 from convert2rhel.toolopts import tool_opts
 
 
-# Supported conversion path to RHEL
+# Allowed conversion paths to RHEL. We want to prevent a conversion and minor
+#   version update at the same time.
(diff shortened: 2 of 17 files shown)


convert2rhel/unit_tests/conftest.py
```diff
@@ -73,26 +73,16 @@ def setup_logger(tmpdir):
...
-
-_replace_data_dir_for_centos8 = pytest.fixture(_replace_data_dir_base("8"))
-_replace_data_dir_for_centos7 = pytest.fixture(_replace_data_dir_base("7"))
-
-
-def _pretend_fixture_base(monkeypatch, system_release_version):
-    """Apply common logic in pretend_{os_name} fixtures."""
+@pytest.fixture()
+def pretend_os(request, pkg_root, monkeypatch):
+    system_version, system_name = request.param
```
ZhukovGreen, line 78: Need to document it

ZhukovGreen:
@oamg/convert2rhel-developers pls review

convert2rhel/systeminfo.py
```diff
@@ -33,11 +35,26 @@
 from convert2rhel.toolopts import tool_opts
 
 
+# Supported conversion path to RHEL
+RELEASE_VER_MAPPING = {
```
ZhukovGreen, line 39: Should we no longer require to have the recent version, but any version available in this dict?
Should we update the docs to refer to this dict instead of duplicating this info?

convert2rhel/systeminfo.py
```diff
@@ -33,11 +35,26 @@
...
+    "CentOS Linux": {
+        "8.4": "8.4",
+        "8.3": "8.3",
+        "7.9": "7.9Server",
```
ZhukovGreen, line 43: Should be just 7.9

convert2rhel/systeminfo.py
```diff
@@ -33,11 +35,26 @@
...
+    },
+    "Oracle Linux Server": {
+        "8.4": "8.4",
+        "8.3": "8.3",
+        "7.9": "7.9Server",
```
ZhukovGreen, line 48: the same

bocekm: `7.9` releasever repo contains RHEL 7.9 GA packages, whereas `7Server` contains the latest RHEL 7.9 package updates. That's what we need.

It's not the case for `6.10` and `8.x` releasevers. Since RHEL 6.10 is the latest minor version, we can safely use `6Server` and leave it like that forever. For `8.x`, we can and need to specify the specific minor version (i.e. `8.4` releasever contains the latest RHEL 8.4 package updates, not just GA versions).

bocekm (changes requested):

convert2rhel/systeminfo.py
```diff
@@ -190,7 +207,15 @@ def _get_repofile_pkgs(self):
...
     def _get_releasever(self):
-        return self._get_cfg_opt("releasever")
+        might_be_config = self._get_cfg_opt("releasever")
+        try:
+            # return config value or corresponding releasever from the RELEASE_VER_MAPPING
+            return might_be_config or RELEASE_VER_MAPPING[self.name][".".join(map(str, self.version))]
+        except KeyError:
+            self.logger.critical(
+                "Unsupported system name/version met: %s - %s.\n"
+                "Supported paths are:\n%s" % (self.name, self.version, pprint.pformat(RELEASE_VER_MAPPING, width=54))
```
bocekm, line 217:
```suggestion
                "%s of version %d.%d is not allowed for conversion.\nAllowed versions are: %s"
                % (self.name, self.version.major, self.version.minor, RELEASE_VER_MAPPING.keys())
```

bocekm (changes requested):

convert2rhel/systeminfo.py
```diff
@@ -33,11 +35,26 @@
 from convert2rhel.toolopts import tool_opts
 
 
+# Supported conversion path to RHEL
```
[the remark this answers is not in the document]
SpyTec: Updated your suggestion by removing an extra "a":
```diff
- # Allowed conversion paths to RHEL. We want to prevent a conversion and a minor version update at the same time.
+ # Allowed conversion paths to RHEL. We want to prevent a conversion and minor version update at the same time.
```

convert2rhel/systeminfo.py
SpyTec: As in to say, any releases outside the mapping will not be converted by default. Should there be a new 8.5 RHEL and CentOS release and we haven't updated this mapping, the conversion will also fail (by default)

Not saying that will happen if we update it, but just extending a) a bit

bocekm: Agreed. For that reason I suggested to update the mapping to include 8.5 (https://github.com/oamg/convert2rhel/pull/299#discussion_r648204572).

convert2rhel/systeminfo.py
ZhukovGreen: As @bocekm observe the 7.9 repo is outdated, switched to 7Server

convert2rhel/systeminfo.py
```diff
@@ -33,11 +35,26 @@
...
+RELEASE_VER_MAPPING = {
+    "CentOS Linux": {
```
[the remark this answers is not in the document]
ZhukovGreen: I think it is better to have them separated - as it is another OS. Changes to OL comes more frequently. Even though now we have identical lines

ZhukovGreen (commented):
@bocekm @SpyTec answered your code reviews. Thank you

ZhukovGreen:
@bocekm @SpyTec I rebased this MR on top of #300. Now will be writing integration tests for this feature

convert2rhel/systeminfo.py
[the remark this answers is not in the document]
SpyTec: Until CentOS or Oracle starts to differ there isn't going to be any differences that would cause problems. If we add a mapping for 8.5 now, it will start working for Oracle when their minor version is released and it would introduce no issues for CentOS. There would only be a problem if Oracle added a new minor version that is not based off of RHEL. Does that make sense?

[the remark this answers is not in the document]
ZhukovGreen: ok, just to double confirm. My position is based on just a design sense. No strong arguments

ZhukovGreen:
All 4 tests PASSED, review comments adressed

convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
```diff
@@ -31,7 +31,7 @@ repofile_pkgs =
 # Delimited by any whitespace(s).
 default_rhsm_repoids = rhel-7-server-rpms
 
-releasever=7Server
+releasever=
```
bocekm, line 34:
```suggestion
# If defined, it overrides the default releasever defined by RELEASE_VER_MAPPING.
# The value is passed to the yum calls through the --releasever option when accessing RHEL repos. Its purpose is to
# substitute the $releasever variable in a repository baseurl.
releasever=
```
Please add this to every config.

tests/ansible_collections/roles/install-testing-deps/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
 - name: Install pytest framework dependencies
   pip:
-    name: ["pytest", "pytest-cov", "envparse", "click", "pexpect"]
+    name: ["pytest", "pytest-cov", "envparse", "click", "pexpect", "pdbpp", "dataclasses"]
```
bocekm, line 4:
```suggestion
    name: ["pytest", "pytest-cov", "envparse", "click", "pexpect", "dataclasses"]
```

bocekm (changes requested):
It looks good. Just a few little things.

ZhukovGreen:
Just FYI I also successfully reran the following tests:
➞  tmt run plans --name /plans/integration/inhibit-if-kmods-is-not-supported/.+/good -vvvddd

as kind of basic conversion tests

ZhukovGreen:
Just rebased

SpyTec (changes requested):

convert2rhel/unit_tests/conftest.py
```diff
@@ -101,34 +146,39 @@ def _pretend_fixture_base(monkeypatch, system_release_version):
...
-            "Available are: %s" % (system_release_version, " ".join(release2data_dir_relations)),
-        )
-
-
-pretend_centos8 = pytest.fixture(_pretend_centos_base("8.3.2011"))
-pretend_centos7 = pytest.fixture(_pretend_centos_base("7.9.2009"))
+all_systems = pytest.mark.parametrize(
+    "pretend_os",
+    (
+        ("7.9.1111", "CentOS Linux"),
```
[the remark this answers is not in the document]
ZhukovGreen: Not sure if need to bother with that, as we are not trying to replace the integration tests, but just giving more flexibility in unit testing. Let's leave it like it is until the need to have specific hotfix version (actually in this case it will be also possible just to write raw decorator instead of using aliases

tests/ansible_collections/roles/install-testing-deps/tasks/main.yml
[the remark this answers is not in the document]
ZhukovGreen: this package is backporting dataclasses into py3.6. In py3.7 we wouldn't need this package, but dataclasses would be available.

tests/integration/conftest.py
```diff
@@ -97,3 +104,104 @@ def factory(
...
+                        res[param.lower()] = value.strip('"')
+        return cls(**res)
+
+
+@pytest.fixture()
+def os_release():
+    return OsRelease.create_from_file(Path("/etc/os-release"))
+
+
+class Config:
```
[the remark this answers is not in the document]
ZhukovGreen: Yes, it is supposed to handle whatever text file. I created it for aggregating some useful features with configs (or any other files). So far I see replace line:) Let me add more docstrings there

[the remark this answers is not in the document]
ZhukovGreen: Also, I created c2r_config fixture as it should be used frequently and having this as a fixture allows me to inject os_release fixture and get right config path, without requiring this from the test writer

SpyTec (approved):

SpyTec (approved):

bocekm:
The tests are now failing for me at the line https://github.com/oamg/convert2rhel/pull/299/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56R26.

PR #301, opened 2021-06-09
Add quotes around the activation key

bocekm:
When registering the system through subscription-manager, the key can contain a space character.

Related: https://access.redhat.com/discussions/5676681#comment-2117741


ZhukovGreen (approved):
:)

SpyTec (approved):

PR #305, opened 2021-06-18
Feature/inhibit only if loaded kmods missing

### LINKED ISSUE #241: CentOS 6.10 with Python 2.6.6 fail at reboot (overwriting env vars)

ZhukovGreen:
Use only loaded kmods for a check of supported kernel modules.
Also, an environment vars are fixed for subprocess calls 

closes #241, OAMG-4810, OAMG-5076
replaces #251

# TODO
- [x] Integration tests


ZhukovGreen:
@oamg/convert2rhel-developers ready for a review. 4 integrations tests passed

convert2rhel/utils.py
```diff
@@ -582,4 +581,12 @@ def get_rpm_header(rpm_path, _open=open):
     return rpmhdr
 
 
+def prepare_environment():
+    """Load LCC_ALL=C to the main app process.
+
+    Use default language for output
+    """
+    os.environ.update({"LCC_ALL": "C"})
```
bocekm, line 589: No need for this when we have https://github.com/oamg/convert2rhel/blob/main/convert2rhel/utils.py#L145.

bocekm (changes requested):

convert2rhel/utils.py
ZhukovGreen: This MR removes this line, also resolves the issue, when the main process environment was unavailable in the subprocess

bocekm: > resolves the issue, when the main process environment was unavailable in the subprocess

I believe that the issue lied in the fact that in `run_subprocess` we've been replacing all environment variables with just the LC_ALL one ([OAMG-4810](https://issues.redhat.com/browse/OAMG-4810)).

So, to solve [OAMG-4810](https://issues.redhat.com/browse/OAMG-4810) and https://github.com/oamg/convert2rhel/issues/241, this one-line change should be sufficient:
```
    process = subprocess.Popen(
        cmd,
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        bufsize=1,
        env=dict(os.environ.copy(), **{"LC_ALL": "C"}),
    )
```

ZhukovGreen: You're right. I just solved this in a bit different way. Loading this environment into the main process at the very beginning (so all children processes are inherited this environment by default). 

This os.environ update is very cool, as it is not just updating the dict, but also do os.putenv for each new env var. 

The new solution (with os.environ.update) seems cleaner to me, as we prepare the environment at the beginning of the app runtime.

What do you think?

bocekm: You variant is fine as well.

Because you just set locale, please rename the function to `set_locale()`. And the env var is not LCC_ALL, but LC_ALL.

```python
def set_locale():
    """Set the POSIX default locale for the main process as well as the child processes.

    The reason is to get predictable output from the executables we call, not influenced by non-default locale.
    """
    os.environ.update({"LC_ALL": "C"})
```

bocekm (changes requested):

convert2rhel/checks.py
```diff
@@ -180,24 +180,19 @@ def ensure_compatibility_of_kmods():
...
-            'find /lib/modules/{kver} -name "*.ko*" -type f'.format(kver=system_info.booted_kernel),
-            print_output=False,
-        )
-        assert exit_code == 0
-        assert kmod_str
-    except (subprocess.CalledProcessError, AssertionError):
-        logger.critical("Can't get list of kernel modules.")
-    else:
-        return set(_get_kmod_comparison_key(path) for path in kmod_str.rstrip("\n").split())
+    lsmod_output, _ = run_subprocess("lsmod", print_output=False)
```
bocekm, line 190:
```suggestion
    logger.debug("Getting a list of loaded kernel modules.")
    lsmod_output, _ = run_subprocess("lsmod", print_output=False)
```

bocekm (changes requested):
One additional change needed:
https://github.com/oamg/convert2rhel/blob/v0.22/convert2rhel/checks.py#L246
`Comparing the installed kernel modules` -> `Comparing the loaded kernel modules`

tests/integration/inhibit-if-kmods-is-not-supported/main.fmf
```diff
@@ -5,15 +5,6 @@ tag+:
...
-    - good_test
-  test: |
-    pytest -svv -m good_tests
-
-/bad:
-  tag+:
-    - bad_test
-  test: |
-    pytest -svv -m bad_tests
+duration: 10m
```
bocekm, line 8: Can be removed when we have the default of 30m?

bocekm (approved):

PR #297, opened 2021-06-01
Add a script to submit tests to TFT locally

zhukovgreen:
Steps:
- [x] add script which submits the tmt plans to the testing farm
The quality of this script:
0. cli utility in the scripts directory
1. Automatically detect all plans and environments and translate this into the right API calls
2. Track the status of submitted jobs and show the debug links
3. Secrets are managed with the help of dotenv
- [x] unify the solution of managing secrets to just dot env as a single source of it. Ansible will use lookup plugin and .env defined vars should be loaded first (for local run)
- [x] automate the process of getting the right ref of the repo (param in API call)

This MR introduces new cli for interaction with the TFT API from the local development environment:
```bash
➞  python scripts/tft --help                                                          [git:feature/tft-integration] ✖  
Usage: tft [OPTIONS]

  Script to interact with testing farm service.

  Ensure the content of .env.

  Example commands: ``` # Submit only given plan to tft and use existing
  copr rpm builds python scripts/tft -v -p /plans/integration/inhibit-if-
  kmods-is-not-supported/centos8 --copr-build-id 2353320 --copr-build-id
  2353321

  # Run all plans for some-atypical-remote-name git remote with debug mode
  (-v) python scripts/tft -v -r some-atypical-remote-name ```

Options:
  -p, --plans TEXT          Plan names. i.e. -p /plan/name. Could be multiple
                            [default: /plans/]

  -r, --remote-name TEXT    Git remote name from which the content of the repo
                            will be cloned at current commit. Warning: changes
                            should be pushed to the remote before running this
                            script.  [default: origin]

  -b, --copr-build-id TEXT  Use specified builds ids instead of creating a new
                            copr builds.

  -v, --verbose
  --help                    Show this message and exit.
```

To submit the following tmt plans `/plans/integration/inhibit-if-kmods-is-not-supported/centos8` you need to do run:
```bash
$ rm .install
$ make install
$ source .venv3/bin/activate
$ python scripts/tft -v -p /plans/integration/inhibit-if-kmods-is-not-supported/centos8
```
note: first two commands needed to be run just once (to update the local deps)

PR #317, opened 2021-07-22
Add copr build target in Makefile

ZhukovGreen:
New command `make copr-build`. It will be used in #297

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,14 @@ rpms: images
 	docker cp $$(docker create $(IMAGE)/centos7rpmbuild):/data/.rpms .
 	docker rm $$(docker ps -aq) -f
 
+copr-build: rpms
+	mkdir -p .srpms
+	rm -frv .srpms/*
+	docker cp $$(docker create $(IMAGE)/centos8rpmbuild):/data/.srpms .
+	docker cp $$(docker create $(IMAGE)/centos7rpmbuild):/data/.srpms .
+	docker rm $$(docker ps -aq) -f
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -80,6 +80,14 @@ rpms: images
 	docker cp $$(docker create $(IMAGE)/centos7rpmbuild):/data/.rpms .
 	docker rm $$(docker ps -aq) -f
 
+copr-build: rpms
+	mkdir -p .srpms
+	rm -frv .srpms/*
+	docker cp $$(docker create $(IMAGE)/centos8rpmbuild):/data/.srpms .
+	docker cp $$(docker create $(IMAGE)/centos7rpmbuild):/data/.srpms .
+	docker rm $$(docker ps -aq) -f
+	copr-cli --config .copr.conf build --nowait zhukovgreen/convert2rhel .srpms/*
```
ZhukovGreen, line 89: @bocekm can we build right into `@oamg/convert2rhel`? What do we need for that? Adding my user into the oamg org under copr?

bocekm: It's ok to use `@oamg/convert2rhel`. The only restriction is to not build packages with a bumped [release](https://github.com/oamg/convert2rhel/blob/v0.22/packaging/convert2rhel.spec#L13).  And yes, each user who want's to build in that copr repo needs to ask for permission under https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/permissions/.

bocekm (approved):
It works, thanks.

PR #318, opened 2021-07-26
make tmpdir fixture be LocalPath type for both python3 and 2

ZhukovGreen:
In the previous implementation, the type of the fixture was str in py2. This is fixed now and the type is identical (LocalPath) for both python versions.

ref #262


hosekadam (approved):
LGTM

PR #320, opened 2021-07-29
Use proper comparison of the env var

ZhukovGreen:
Little improvement of the env vars check in ansible-playbook.

the previous solution is deprecated


hosekadam (approved):
LGTM

PR #321, opened 2021-07-30
Adjust environment-file on commit trigger to point to our remote secr…

zhukovgreen:
Blocked by: https://github.com/psss/tmt/issues/852

Solution in this MR depends on the solution in the specified tmt ticket

For running our tests via packit-ci we need other way to send secrets to the testing environment. This MR solves this

Note:
for testing this MR, cert_path in main fmf should be adjusted to the crt in this branch (currently it is refer to the future location in the main branch)


bocekm:
The requested tmt feature related to certificates is not needed: https://github.com/psss/tmt/issues/852#issuecomment-966555647
And regarding loading the environment file from a remote file (related: https://github.com/psss/tmt/pull/795, https://github.com/psss/tmt/pull/881), we're already using that: https://github.com/oamg/convert2rhel/pull/343/files#diff-5ae9adacc9ce4ac4f6d80f8c06e2f4a4967a272f41bc5f2643ed946c4077f685R17.

PR #323, opened 2021-08-03
Convert STI to TMT: Add test using activation key 

kokesak:
Test uses activation key to do the c2r conversion. 

I created a new folder that differentiate the normal conversion and the one using activation key.


hosekadam (approved):
LGTM, just before merging I think there would be needed rebase and put newly created conversion test (flag system as converted) to one of the variant (rhsm or key).

kokesak:
Rebased to the main branch. I moved the 'flag system as converted' test to the rhsm directory.

hosekadam (approved):
LGTM. Also need to merge changes on GitLab which contain updated .env file with secrets.

mkluson (approved):

PR #324, opened 2021-08-05
Flagging system as converted

hosekadam:
Convert2RHEL shall record that the system has been converted. The information shall be stored in the /etc/migration-results file, which will be collected by Insights and sos to help the business to learn more about the number/details of migrations, and to help support engineers to identify what happened to a system.


convert2rhel/breadcrumbs.py
```diff
@@ -0,0 +1,191 @@
...
+import json
+import os
+import sys
+
+from time import gmtime, strftime
+
+from convert2rhel import pkghandler, systeminfo
+
+
+# is importing systeminfo good idea? It creates new instance of SystemInfo after import
```
bocekm, line 27: > It creates new instance of SystemInfo after import

The SystemInfo instance is created just once, during the first import of systeminfo that happens in main.py. So when importing it here, the module is already loaded and the system_info variable already holds the SystemInfo instance with all data populated.

convert2rhel/breadcrumbs.py
```diff
@@ -0,0 +1,191 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright(C) 2016 Red Hat, Inc.
```
bocekm, line 3:
```suggestion
# Copyright(C) 2020 Red Hat, Inc.
```

convert2rhel/breadcrumbs.py
```diff
@@ -0,0 +1,191 @@
...
+
+# is importing systeminfo good idea? It creates new instance of SystemInfo after import
+# but we aren't setting the data, so might be ok
+
+FILE = "/etc/migration-results"
+
+
+class Breadcrumbs(object):
+    def __init__(self):
+        self.executed = ""
```
bocekm, line 35:
```suggestion
        # Version of the JSON schema of the breadcrumbs file. To be changed when the content changes.
        self.version = 1
        # convert2rhel command as executed by the user including all the options
        self.executed = ""
```

convert2rhel/breadcrumbs.py
```diff
@@ -0,0 +1,191 @@
...
+        self.nevra = None
+        self.signature = None
+        self.exit_code = None
+        self.timestamp = None
+        self.source_os = None
+        self.target_os = None
+        self.env = None
+        self.run_id = None
```
bocekm, line 43:
```suggestion
        # Run ID is to be populated by Leapp only. The value should be null in the json generated by convert2rhel.
        self.run_id = None
```

convert2rhel/breadcrumbs.py
```diff
@@ -0,0 +1,191 @@
...
+FILE = "/etc/migration-results"
+
+
+class Breadcrumbs(object):
+    def __init__(self):
+        self.executed = ""
+        self.nevra = None
+        self.signature = None
+        self.exit_code = None
+        self.timestamp = None
```
bocekm, line 39:
```suggestion
        self.activity_started = None
        self.activity_ended = None
```
@shaded-enmity can help with implementing/getting the `activity_started`.

bocekm (changes requested):

convert2rhel/breadcrumbs.py
hosekadam: @bocekm 2021? :smile:

bocekm: Yes :)

hosekadam:
Generated file corresponds with schema. Added intergration test.

bocekm (approved):

PR #325, opened 2021-08-05
fix: allow underscore in package nevr regex

SpyTec:
The regex used to get problematic packages did not catch all package
name variants. Packages like `mod_ldap` has an underscore in it instead of
a dash. This fix makes it possible to fetch packages like it.


bocekm (changes requested):
Please extend unit tests.

bocekm (approved):

PR #326, opened 2021-08-05
feat: handle mismatched packages

SpyTec:
This adds another set to the problematic dict that shows packages that
cause a mismatch of two packages. This mismatch will happen if a package
depends on something that cannot be installed.

See OAMG-4944

---

Currently just an idea and will need polishing

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -45,36 +45,38 @@ class PkgWFingerprint(object):
         self.fingerprint = fingerprint
 
 
-def call_yum_cmd_w_downgrades(cmd, pkgs):
+def call_yum_cmd_w_downgrades(cmd, pkgs, retries=MAX_YUM_CMD_CALLS):
     """Calling yum command is prone to end up with an error due to unresolved
     dependencies, especially when it tries to downgrade pkgs. This function
     tries to resolve the dependency errors where yum is not able to.
     """
 
-    for _ in range(MAX_YUM_CMD_CALLS):
-        output, ret_code = call_yum_cmd(cmd, "%s" % (" ".join(pkgs)))
-        loggerinst.info("Received return code: %s\n" % str(ret_code))
-        # handle success condition #1
-        if ret_code == 0:
-            break
-        # handle success condition #2
-        # false positive: yum returns non-zero code when there is nothing to do
-        nothing_to_do_error_exists = output.endswith("Error: Nothing to do\n")
-        if ret_code == 1 and nothing_to_do_error_exists:
-            break
-        # handle error condition
-        loggerinst.info("Resolving dependency errors ... ")
-        output = resolve_dep_errors(output)
-        # if we have problematic packages, remove them
-        problematic_pkgs = get_problematic_pkgs(output)
-        if problematic_pkgs["errors"]:
-            loggerinst.warning(
-                "Removing problematic packages to continue with conversion:\n%s" % "\n".join(problematic_pkgs["errors"])
-            )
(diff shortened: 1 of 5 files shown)


convert2rhel/pkghandler.py
```diff
@@ -45,36 +45,53 @@ def __init__(self, pkg_obj, fingerprint):
...
+        # no more retries, delete troublesome packages
         problematic_pkgs = get_problematic_pkgs(output)
-        if problematic_pkgs["errors"]:
-            loggerinst.warning(
-                "Removing problematic packages to continue with conversion:\n%s" % "\n".join(problematic_pkgs["errors"])
-            )
-            utils.remove_pkgs(problematic_pkgs["errors"], backup=False, critical=False)
-    else:
+        if not problematic_pkgs["mismatches"]:
+            loggerinst.critical("Could not resolve yum errors.")
```
bocekm, line 71: Why wouldn't you remove `problematic_pkgs["errors"]` when `retries == 1`?
I think we should be removing `errors` and `mismatches` no matter the `retries` variable.

What I would do is just to remove the whole `if retries == 1:` condition block. And leave the package removals to the code starting at line 84 (`# handle error condition`).

convert2rhel/pkghandler.py
```diff
@@ -45,36 +45,53 @@ def __init__(self, pkg_obj, fingerprint):
...
+        loggerinst.warning("Unresolved yum errors present.")
+        loggerinst.warning(
+            "The following packages prevent convert2rhel from continuing. If you choose to not continue manual intervention is required:\n%s"
```
bocekm, line 74: You don't say what's going to happen if they choose to continue, so it's difficult to decide :)
I would actually not ask whether they want to continue because:
- they don't know what is meant by manual intervention (we haven't explored that in detail or documented either). It might not be even possible to get to the pre-conversion state without full system restore.
- they don't know what steps to do to finish the conversion.

So the only real choice for them here should be leaving the transaction error resolution to us. What would be the point in asking them?

convert2rhel/pkghandler.py
```diff
@@ -45,36 +45,53 @@ def __init__(self, pkg_obj, fingerprint):
...
+    # handle error condition
+    loggerinst.info("Resolving dependency errors ... ")
+    output = resolve_dep_errors(output)
+    # if we have problematic packages, remove them
+    problematic_pkgs = get_problematic_pkgs(output)
+    if problematic_pkgs["errors"]:
+        loggerinst.warning(
+            "Removing problematic packages to continue with conversion:\n%s" % "\n".join(problematic_pkgs["errors"])
+        )
+        utils.remove_pkgs(problematic_pkgs["errors"], backup=False, critical=False)
```
bocekm, line 93:
```suggestion
    to_remove = problematic_pkgs["errors"] | problematic_pkgs["mismatches"]
    if to_remove:
        loggerinst.warning(
            "Removing problematic packages to continue with conversion:\n%s" % "\n".join(to_remove)
        )
        utils.remove_pkgs(to_remove, backup=False, critical=False)
```

bocekm (changes requested):

convert2rhel/pkghandler.py
SpyTec: Yeah, if we're not going to ask the user we might as well. Whole implementation was just to handle mismatched packages at the end so we could ask

I'll refactor and simplify without confirmation

SpyTec:
@bocekm everything seems fine except for integration tests. I can't figure out why it doesn't work

Confirmed it to work with Vagrant box

bocekm (approved):

PR #332, opened 2021-08-06
Convert STI to TMT: Add test using satellite for conversion

kokesak:
This PR add test that uses satellite for c2r. PR follows the 
Original STI test: https://gitlab.cee.redhat.com/sturivny/convert_tests/-/blob/master/tests/tests_satellite_scenario.yml

This PR follows up the PR323 (https://github.com/oamg/convert2rhel/pull/323)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,8 @@ RHSM_USERNAME=
 RHSM_SERVER_URL=
 RHSM_KEY=
 RHSM_ORG=
+SATELLITE_KEY=
+SATELLITE_ORG=
 
 # Ansible related secrets
     # options are True|False. if True, then the rpm will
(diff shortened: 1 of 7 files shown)


tests/integration/conftest.py
```diff
@@ -84,7 +84,7 @@ def convert2rhel(shell):
     @contextmanager
     def factory(
         options: str,
-        timeout: int = 30 * 60,
+        timeout: int = 60 * 60,
```
kokesak, line 87: When I was running the tests locally it always timeouts after 30minutes. The conversion is much slower than using activation key or regular RHSM. Maybe we should investigate more?

(Increasing timeout ends in successful conversion)

mkluson (approved):

PR #333, opened 2021-08-06
Convert STI to TMT: Add test using custom repos for conversion

kokesak:
This PR add test that uses custom repos for c2r. PR follows the
Original STI test: https://gitlab.cee.redhat.com/sturivny/convert_tests/-/blob/master/tests/tests_enable_repo.yml

This PR follows up the PR323 (#323)

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -84,7 +84,7 @@ def convert2rhel(shell):
     @contextmanager
     def factory(
         options: str,
-        timeout: int = 30 * 60,
+        timeout: int = 60 * 60,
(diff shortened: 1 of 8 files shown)


tests/integration/conftest.py
```diff
@@ -84,7 +84,7 @@ def convert2rhel(shell):
     @contextmanager
     def factory(
         options: str,
-        timeout: int = 30 * 60,
+        timeout: int = 60 * 60,
```
kokesak, line 87: The same timeout issue as in #332  is happening also here.

mkluson (approved):

PR #334, opened 2021-08-06
Fix/pretend os not isolated from host arch

ZhukovGreen:
Little fix which I observed running tests on mac 😎 

Also fixed @hosekadam unit test, as I ensured the tmpdir is working, but the test wasn't rewritten to consider this

### DIFF
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -99,27 +99,29 @@ class TestCert(unittest.TestCase):
 
 
 @pytest.mark.parametrize(
-    ("filename"),
-    ("existing.file", "nonexisting.file"),
+    (
+        "create_file",
+        "expected_text_in_logs",
+    ),
+    (
+        (True, "/filename removed"),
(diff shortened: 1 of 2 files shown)


hosekadam (approved):
Thanks for the unit test rework. I've forgot to do that after the tmpdir fixture fix.

@ZhukovGreen And just for info: what the `_get_architecture` return on mac? You have the new M1 processor?

ZhukovGreen:
@hosekadam 
> And just for info: what the _get_architecture return on mac? You have the new M1 processor?

it relies on uname -i, what is not working on mac (at least on mine):
```bash
➞  uname -i
uname: illegal option -- i
usage: uname [-amnprsv]

➞  uname -m
x86_64

➞  docker run -it centos:8
[root@a87d1dab2942 /]# uname -m
x86_64
[root@a87d1dab2942 /]# uname -i
x86_64
```

But anyway, the goal of the fixture is to pretend to be another system, so no calls to uname should be done at all

ZhukovGreen:
@bocekm ^^

PR #337, opened 2021-08-17
Bump version to 0.23

bocekm:
Changes in this version include:

- Create /etc/migration-results with a basic information about convert2rhel being executed
  - #324
- Handle yum conflicts where pkg A depends on pkg B, B is being replaced but replacement for A is not available
  - #326
- Inhibit the conversion only if loaded (not installed) kernel modules are not available in RHEL
  - #305
- Remove temporary folder /var/lib/convert2rhel/ after successful conversion
  - #262
- Allow converting Scientific Linux 7
  - #308
- Roll back the installation of an RHSM certificate
  - #203
- Add LC=ALL to existing environment variables when calling shell commands instead of overriding them
  - #305
- Improve handling yum transaction errors by expecting underscore in a pkg name
  - #325
- Improve patching yum.conf so that it's not necessary to back up
  - #189
- Introduce a new make `copr-build` command to create builds in Fedora Copr
  - #317
- Integration tests: Do not capture and print all stdout/err when a test fails
  - #327
- Integration tests: Abstract away from the processor architecture of the system on which tests are executed
  - #334
- Integration tests: Increase the time limit for a reboot to 600 seconds due to slower test systems in the cloud
  - #329
- Integration tests: Use new TMT 'environment-file' option
  - #319
- Integration tests: Limit the tests to run on relevant test systems only
  - #315
- Integration tests: Set the timeout for tests from the default 5 min to 30 min
  - #309


SpyTec (approved):
LGTM!

PR #338, opened 2021-08-23
Install pip as part of install-testing-deps role

fernflower:
The main prepare-all-before-conversion playbook was wrongly
implying that pip was present on the vms which is not always
the case (take centos7 for instance).


fernflower:
@oamg/convert2rhel-developers please review

fernflower:
Ok, with these 2 commits it should be possible to run tmt tests for convert2rhel out of the box with tmt 1.6 (no custom changes needed) as easy as:
```
tmt run -a provision -h connect -g 10.0.138.76  plans --name 'plans/integration/remove_all_submgr_pkgs/centos7
```

bocekm (approved):

PR #339, opened 2021-08-23
Fix cli option sanitization for py2.6

bocekm:
This piece of code was not Python 2.6-compatible so it was failing on CentOS/Oracle Linux 6 .

### DIFF
--- a/convert2rhel/breadcrumbs.py
+++ b/convert2rhel/breadcrumbs.py
@@ -89,7 +89,7 @@ class Breadcrumbs(object):
 
     def _set_executed(self):
         """Set how was Convert2RHEL executed"""
-        cli_options_to_sanitize = {"--password", "-p", "--activationkey", "-k"}
+        cli_options_to_sanitize = frozenset(("--password", "-p", "--activationkey", "-k"))
         self.executed = sanitize_cli_options(sys.argv, cli_options_to_sanitize)
 
     def _set_nevra(self):


convert2rhel/breadcrumbs.py
```diff
@@ -89,7 +89,7 @@ def _set_pkg_object(self):
 
     def _set_executed(self):
         """Set how was Convert2RHEL executed"""
-        cli_options_to_sanitize = {"--password", "-p", "--activationkey", "-k"}
+        cli_options_to_sanitize = ["--password", "-p", "--activationkey", "-k"]
```
abadger, line 92: This would be a more literal translation of the code:

```suggestion
        cli_options_to_sanitize = frozenset(("--password", "-p", "--activationkey", "-k"))
```

Advantages: 
* frozenset is immutable so there's no chance that something could add to cli_options_to_sanitize by mistake.  This could also be enforced by using a tuple: `        cli_options_to_sanitize = ("--password", "-p", "--activationkey", "-k")`
* This data structure is being used in containment tests so a set will have better performance characteristics everytime you loop through the check inside of `sanitize_cli_options()`

Disadvantage:
* Since there's no literal form for a set in python-2.6, you have to use the `frozenset()` function and pass it a tuple which means there's higher startup cost than constructing the list via `[]` or a  tuple via `()` here.

Performance is probably not important in this code but making it immutable will likely be better for future maintainabiity.

bocekm: Hi @abadger, thanks for your valuable insight. I appreciate your focus on robustness and performance. I agree the frozenset is much better option.

Related to performance, you reminded me of one part of the conversion that takes couple of minutes during the conversion - [calling the `yum makecache` command](https://github.com/oamg/convert2rhel/blob/v0.23/convert2rhel/checks.py#L145). I've created an internal ticket for us to improve it: https://issues.redhat.com/browse/OAMG-5570

PR #342, opened 2021-09-03
Add ansible role to set up custom repo for int tests

kokesak:
We have custom repos on machines provided via libvirt. However in the future we want to run tests on machines provided by Artemis or anything else. Those clean vms do not have custom repos. This PR solve this issue.

Solves [OAMG-5485](https://issues.redhat.com/browse/OAMG-5485).

### DIFF
--- a/plans/integration/check-custom-repo/main.fmf
+++ b/plans/integration/check-custom-repo/main.fmf
@@ -2,6 +2,11 @@ discover+:
     filter:
         - 'tag: check-custom-repo'
 
+prepare+:
+    - name: add custom repos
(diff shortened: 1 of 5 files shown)


mkluson:
Could we merge `tests/ansible_collections/roles/add-custom-repos/rhel7-repos.yml` and `tests/ansible_collections/roles/add-custom-repos/rhel8-repos.yml` to one playbook which decides which repos to used based on eg. `ansible_facts['distribution_major_version']` so we don't have to care about it in the test plans?

mkluson (approved):

PR #343, opened 2021-09-10
Arange tests in tiers

mkluson:
Rearanging the test plans to follow tiers structure like this:

- tier0 = tests running on source system
- tier1 = test running on target system
- tier2 = test of other components in convert workflow


plans/tier0.fmf
```diff
@@ -0,0 +1,43 @@
...
+  discover+:
+    test: check-variant-message
+
+/inhibit_if_kmods_is_not_supported:
+  discover+:
+    test: inhibit-if-kmods-is-not-supported
+
+/inhibit_if_oracle_system_uses_not_standard_kernel:
+  discover+:
+    test: inhibit-if-oracle-system-uses-not-standard-kernel'
```
fernflower, line 22: Why the lone '?

fernflower (commented):
I have a couple of questions\remarks:
1. How do I run single tests in your approach? Like I want to run just a NetworkManager test, how do I do it?
2. With tests from tierN running on 1 guest this causes the need for us to track what is actually happening in an SST test, if there are any changes from the last version and if they are destructive (meaning the test can't be run in a tierN anymore). Is the speedup really worth the management complexity it brings? I really loved the "I don't care what's inside" approach to be honest :(
3. How do we mark destructive tests (the one that should be run 1 test per vm)? By not adding them to tiers?

plans/tier0.fmf
```diff
@@ -0,0 +1,43 @@
+discover+:
+    tier: 0
+
+/changed_yum_conf:
+  discover+:
+    test: changed-yum-conf
+
+/check_custom_repo:
```
kokesak, line 8: We need to add the preparation role that adds custom repos on the target machine

plans/tier0.fmf
```diff
@@ -0,0 +1,43 @@
...
+
+/check_variant_message:
+  discover+:
+    test: check-variant-message
+
+/inhibit_if_kmods_is_not_supported:
+  discover+:
+    test: inhibit-if-kmods-is-not-supported
+
+/inhibit_if_oracle_system_uses_not_standard_kernel:
```
kokesak, line 20: Now this plans is only for Centos8, we should not run this one as it's oracle specific

plans/tier1.fmf
```diff
@@ -0,0 +1,42 @@
+discover+:
+    tier: 1
+prepare+:
+  - name: reboot machine
+    how: ansible
+    playbook: tests/ansible_collections/reboot.yml
+
```
kokesak, line 7: Bear in mind that these tests:

- "resolve-dependency" from "rhsm" conversion folder
- "one-kernel-scenarion" from "custom-repos" conversion folder

They need some extra preparation steps. Maybe we could move them to the tier0?

kokesak (commented):
Just few things:

- those "env_from" files are there for some reason?
- I don't like the specification of those tiers, I would swap tier1 and tier0 as those these tests in tier1 are more "basic" and should be run more often. Also some tests in tier0 already do the full conversion (eg. "changed_yum_conf"). However that's just my opinion.

Otherwise I like splitting tests into tiers))

bocekm:
The tests are failing on pytest conftest.py:
```
Warning: Permanently added '10.29.163.132' (ECDSA) to the list of known hosts.

ImportError while loading conftest '/var/ARTIFACTS/work-changed_yum_confqgGEXm/plans/tier0/changed_yum_conf/discover/default/tests/tests/integration/conftest.py'.
../conftest.py:26: in <module>
    logging.basicConfig(level="DEBUG" if env.str("DEBUG") else "INFO", stream=sys.stderr)
/usr/local/lib/python3.6/site-packages/envparse.py:35: in method
    return self.__call__(var, cast=cast, **kwargs)
/usr/local/lib/python3.6/site-packages/envparse.py:99: in __call__
    raise ConfigurationError(error_msg)
E   envparse.ConfigurationError: Environment variable 'DEBUG' not set.
```
It seems the `env.str("DEBUG")` raises an Exception when the requested env var is not set, which is not what we need.
We apparently use this library https://github.com/rconradharris/envparse which is not documented (at least I can't find any documentation) so I'm going to use something more standard.

kokesak:
Broken rollback test is disabled for centos-7 until the [OAMG-5455](https://issues.redhat.com/browse/OAMG-5455) is done.

mkluson:
oh, I can not merge it :-( I am technically still an author. @bocekm could you please approve and merge it?

bocekm (approved):

PR #344, opened 2021-09-14
Add test for tainted kernel inhibitor

kokesak:
This PR adds an integration test verifying that the conversion stops when there's a loaded kernel module marked as tainted. Test builds own kernel module on the testing machine and use it. Kernel is then marked with appropriate flags which makes the kernel 'tainted'. 
This kernel module causes following flags:

- P - proprietary module was loaded
- O - externally-built (“out-of-tree”) module was loaded
- E  - unsigned module was loaded

Conversion of tainted kernel is then inhibited.

Ref. ticket [OAMG-4927](https://issues.redhat.com/browse/OAMG-4927).

### DIFF
--- a/tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
+++ b/tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
@@ -1,3 +1,5 @@
+import platform
+
 from pathlib import Path
 
 import pytest
@@ -20,6 +22,8 @@ def insert_custom_kmod(shell):
 
 
 def test_inhibit_if_custom_module_loaded(insert_custom_kmod, convert2rhel):
+    # This test checks that rpmquery works correctly.
+    # If custom module is loaded the conversion has to be inhibited.
     insert_custom_kmod()
     with convert2rhel(
         ("-y --no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug").format(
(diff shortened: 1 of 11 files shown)


tests/integration/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -34,15 +36,46 @@ def test_inhibit_if_custom_module_loaded(insert_custom_kmod, convert2rhel):
...
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("y")
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("y")
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("y")
         c2r.expect("Kernel modules are compatible.")
-        c2r.send(chr(3))
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("n")
```
SpyTec, line 57: Shouldn't `-y` work if we just expect "Kernel modules are compatible."?

bocekm:
Looks good. Please rebase.

kokesak:
The failing tests are due to the #412. After this gets merged the tests should pass.

danmyway (dismissed):

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -35,14 +39,47 @@ def test_inhibit_if_custom_module_loaded(insert_custom_kmod, convert2rhel):
...
             env.str("RHSM_SERVER_URL"),
             env.str("RHSM_USERNAME"),
             env.str("RHSM_PASSWORD"),
             env.str("RHSM_POOL"),
         )
     ) as c2r:
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("y")
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("y")
```
mkluson, line 54: Do we really need it to have these two lines here twice?

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -35,14 +39,47 @@ def test_inhibit_if_custom_module_loaded(insert_custom_kmod, convert2rhel):
...
+        # On OracleLinux there is one question less than on Centos
+        if platform.platform().find("oracle-8") < 0:
```
mkluson, line 56: `if "oracle-8" in platform.platform()` seems more readable for me

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
kokesak: I'm afraid so. If we cancel the conversion by sending CTRL+C the system will be left in a partially broken state. It's because the rollback function has problem with handling this SIGINT on OL7/C7 systems.  
More on the issue is mentioned here:
https://issues.redhat.com/browse/OAMG-5756
https://github.com/oamg/convert2rhel/pull/411

mkluson: ok, It seemed to me like some copy&paste error

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
kokesak: Thanks, this is a good point. I changed it.

I also changed the same occurrences in other tests so it's consistent and more readable also there. I made a separate commit for this change (https://github.com/oamg/convert2rhel/pull/344/commits/98c8da60beba63308e75b0f7230c34a73dbfbdee).

bocekm (dismissed):

kokesak:
Feel free to merge once the approval is done.

danmyway (approved):

PR #346, opened 2021-09-15
Fix yum distro-sync problem

hosekadam:
On RHEL 8 based systems the `yum distro-sync` command returns 1 when no package passed to it is available in the enabled RHEL repositories. Convert2RHEL did not account for this situation because there's essentially "nothing to do" so it expected return code 0 (the same as when for example trying to update a package).

We have these 2 case when running distro-sync on RHEL 8 based systems:
- If all of the packages passed to distro-sync are unavailable in the enabled repos, yum returns 1 with a message `Error: No packages marked for distribution synchronization.`
- If at least one package from those passed to distro-sync is available in the enabled repos, yum returns 0. No problem occurs in this case.

On top of that, yum distro-sync behaves differently on RHEL 7 and 8. On RHEL 7 yum returns 0 and no error in both above cases (just `No packages marked for distribution synchronization`).

Related: https://issues.redhat.com/browse/OAMG-4600

### DIFF
--- a/plans/tier1.fmf
+++ b/plans/tier1.fmf
@@ -161,3 +161,14 @@ discover+:
   - name: reboot machine
     how: ansible
     playbook: tests/ansible_collections/reboot.yml
+
+/yum_distro_sync:
+  discover+:
+    test: yum-distro-sync
+  prepare+:
+    - name: install problematic package
+      how: shell
+      script: pytest -svv tests/integration/tier1/yum-distro-sync/install_problematic_package.py
+    - name: reboot machine
+      how: ansible
+      playbook: tests/ansible_collections/reboot.yml
--- a/plans/tier1.fmf
+++ b/plans/tier1.fmf
@@ -166,6 +166,9 @@ discover+:
   discover+:
     test: checks-after-conversion
   prepare+:
+    - name: enable extras repo for Oracle Linux
+      how: ansible
+      playbook: tests/integration/tier1/yum-distro-sync/add-extras-repo/main.yml
     - name: install problematic package
       how: shell
       script: pytest -svv tests/integration/tier1/yum-distro-sync/install_problematic_package.py
--- a/plans/tier1.fmf
+++ b/plans/tier1.fmf
@@ -164,11 +164,14 @@ discover+:
 
 /yum_distro_sync:
   discover+:
-    test: yum-distro-sync
+    test: checks-after-conversion
   prepare+:
     - name: install problematic package
       how: shell
       script: pytest -svv tests/integration/tier1/yum-distro-sync/install_problematic_package.py
+    - name: main conversion preparation
+      how: shell
+      script: pytest -svv tests/integration/tier1/yum-distro-sync/test_yum_distro_sync.py
     - name: reboot machine
       how: ansible
       playbook: tests/ansible_collections/reboot.yml
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -79,12 +79,15 @@ def convert2rhel(shell):
     >>>     c2r.expect("Kernel is compatible with RHEL")
     >>> assert c2r.exitstatus == 0
 
+    Unregister means that system wouldn't be unregistered
+    from subscription-manager after conversion.
     """
 
     @contextmanager
     def factory(
         options: str,
         timeout: int = 60 * 60,
+        unregister: bool = True,
     ) -> ContextManager[pexpect.spawn]:
         c2r_runtime = pexpect.spawn(
             f"convert2rhel {options}",
@@ -101,7 +104,8 @@ def convert2rhel(shell):
             c2r_runtime.expect(pexpect.EOF)
             c2r_runtime.close()
         finally:
-            shell("subscription-manager unregister")
+            if unregister:
+                shell("subscription-manager unregister")
 
     return factory
 
--- a/tests/integration/tier1/yum-distro-sync/test_yum_distro_sync.py
+++ b/tests/integration/tier1/yum-distro-sync/test_yum_distro_sync.py
@@ -1,0 +1,66 @@
+from envparse import env
+
+
+def test_yum_distro_sync(convert2rhel, shell):
+    """Test yum distro-sync command edge-cases when given packages aren't in enabled repositories.
+
+    Following is done:
+        1) Install a problematic package (cpaste) - done in preparation step
+        2) Run the conversion
+        3) Run the distro-sync in two variants
+            - just with the problematic package
+            - with some another package, which is in RHEL repositories
+        4) Do the same checks and conditions as is in convert2rhel/pkghandler.py in call_yum_cmd_w_downgrades.
+
+    Another problem is, that yum behaves differently on Centos 7 and Centos 8
+        - on Centos 7 gives 0 and any error in both cases
+        - on Centos 8  gives 0 and any error if in list of packages for distro-sync is at least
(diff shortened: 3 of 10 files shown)


tests/integration/conftest.py
```diff
@@ -79,9 +79,12 @@ def convert2rhel(shell):
     >>>     c2r.expect("Kernel is compatible with RHEL")
     >>> assert c2r.exitstatus == 0
 
+    Unregister set to false means, that system after conversion woudn't be unregistered
+    from subscription-manager. Then that is needed to be done manually.
     """
 
     @contextmanager
+    # def factory(options: str, timeout: int = 60 * 60, unregister: bool = True) -> ContextManager[pexpect.spawn]:
```
hosekadam, line 87: There I have a problem. I need to skip `subscription-manager unregister` which is ran directly after conversion. My plan was to create optional argument `unregister` by default set to True. In this case I would set it to False and unregister the system manually after the checks.

But I don't know how to write that. You can see some attempts in comments, but always I got an error.

tests/integration/conftest.py
```diff
@@ -101,7 +104,10 @@ def factory(
             c2r_runtime.expect(pexpect.EOF)
             c2r_runtime.close()
         finally:
-            shell("subscription-manager unregister")
+            if unregister:
+                shell("subscription-manager unregister")
+            else:
+                pass
```
r0x0d, line 110: This else statement is not actually necessary. If the `unregister` variable is set to `False`, the function flow will continue to the `return factory` either way.

```suggestion
```

tests/integration/yum-distro-sync/install_problematic_package.py
```diff
@@ -0,0 +1,8 @@
+def test_install_problematic_package(shell):
+    """Need to install some package, which is in Centos repos (system, which we are converting)
+    and isn't in RHEL repositories, now is selected package (cpaste) from Centos Extras repo which is
+    enabled in Centos by default.
+
+    When the test starts failing maybe will be needed change of the selected package."""
```
r0x0d, line 6:
```suggestion
    """Need to install a package, which is in CentOS repos (which we are converting) and is not present in RHEL 
    repositories. The package in question (cpaste) is from the CentOS Extras repo which is enabled by default on 
    CentOS systems. 

    When the test starts failing we will need to change the selected package."""
```

r0x0d (commented):
Just a reminder @hosekadam: 

The integration tests on this PR should match the newly created test tiers made by @mkluson and @kokesak. 

PR in questions which was introduced the tests in tiers: https://github.com/oamg/convert2rhel/pull/343

Probably it would be better to rebase this branch with `main` to follow up the fix for integration tests.

tests/integration/conftest.py
hosekadam: I see, not sure why I left that there :D

hosekadam:
@kokesak Hi, could you please take a look to the integration test if it corresponds with new testing way?

plans/tier1.fmf
```diff
@@ -133,3 +133,14 @@ discover+:
   - name: reboot machine
     how: ansible
     playbook: tests/ansible_collections/reboot.yml
+
+/yum_distro_sync:
+  discover+:
+    test: yum-distro-sync
```
kokesak, line 139: In the new test format for tier1 we use `checks-after-conversion` as a test itself , which is a collection of all checks that are run on the destination system (after conversion). Please add `yum-distro-sync` to the preparation phase and make sure the system is rebooted after this phase (after running your `yum-distro-sync`). I see that you put reboot after installing problematic packages. Unless it's necessary to reboot the machine after this part you can add the `yum-distro-sync` before the reboot.

tests/integration/tier1/yum-distro-sync/test_yum_distro_sync.py
```diff
@@ -0,0 +1,66 @@
...
+    Following is done:
+        1) Install a problematic package (cpaste) - done in preparation step
+        2) Run the conversion
+        3) Run the distro-sync in two variants
+            - just with the problematic package
+            - with some another package, which is in RHEL repositories
+        4) Do the same checks and conditions as is in convert2rhel/pkghandler.py in call_yum_cmd_w_downgrades.
+
+    Another problem is, that yum behaves differently on Centos 7 and Centos 8
+        - on Centos 7 gives 0 and any error in both cases
```
kokesak, line 16: Maybe?
```suggestion
        - on Centos 7 returns 0 and any error in both cases
```

tests/integration/tier1/yum-distro-sync/test_yum_distro_sync.py
```diff
@@ -0,0 +1,66 @@
...
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        ),
+        unregister=False,
+    ) as c2r:
+        c2r.expect("Conversion successful!")
+    assert c2r.exitstatus == 0
+
+    out = shell("yum distro-sync cpaste zip")  # any error on Centos 7 and Centos 8
```
kokesak, line 36: I would move the comment above so it's consistent in the code.
```suggestion
    # any error on Centos 7 and Centos 8
    out = shell("yum distro-sync cpaste zip")  
```

tests/integration/tier1/yum-distro-sync/test_yum_distro_sync.py
```diff
@@ -0,0 +1,66 @@
...
+        1) Install a problematic package (cpaste) - done in preparation step
+        2) Run the conversion
+        3) Run the distro-sync in two variants
+            - just with the problematic package
+            - with some another package, which is in RHEL repositories
+        4) Do the same checks and conditions as is in convert2rhel/pkghandler.py in call_yum_cmd_w_downgrades.
+
+    Another problem is, that yum behaves differently on Centos 7 and Centos 8
+        - on Centos 7 gives 0 and any error in both cases
+        - on Centos 8  gives 0 and any error if in list of packages for distro-sync is at least
```
kokesak, line 17: The same as above
```suggestion
        - on Centos 8  returns 0 and any error if in list of packages for distro-sync is at least
```

tests/integration/tier1/yum-distro-sync/test_yum_distro_sync.py
```diff
@@ -0,0 +1,66 @@
...
+        ),
+        unregister=False,
+    ) as c2r:
+        c2r.expect("Conversion successful!")
+    assert c2r.exitstatus == 0
+
+    out = shell("yum distro-sync cpaste zip")  # any error on Centos 7 and Centos 8
+    assert condition_test(out.output, out.returncode)
+
+    out = shell("yum distro-sync cpaste")  # an error on Centos 8, which should be skipped
```
kokesak, line 39: The same as above
```suggestion
    # an error on Centos 8, which should be skipped
    out = shell("yum distro-sync cpaste")  
```

kokesak (commented):
I reviewed the integration test section

kokesak:
Tests for CentoOS are passing except one timeout.
Tests runs for OL (PASSED):
OL7: http://artifacts.osci.redhat.com/testing-farm/cd61be12-c92d-4592-a41e-38f648382020
OL8:http://artifacts.osci.redhat.com/testing-farm/a0beb534-7d30-4031-9ebf-a15380333055

Fail for the released version:
C8: http://artifacts.osci.redhat.com/testing-farm/b761bbf3-1083-49de-99e0-5e0d94c30103/

bocekm (approved):

PR #351, opened 2021-10-14
Fix one-kernel int test

kokesak:
On OL7 there is a problem with python3 package, which is missing in the GA repo. However python3 is mandatory for `pytest` to run tests. Simple solution to this is to remove the python3 package just before the conversion and install it just after the conversion is successful. 
To not interrupt basic conversions using custom-repos I moved this test scenario to it's own directory.


plans/integration/one-kernel-scenario/main.fmf
```diff
@@ -1,27 +1,41 @@
...
     - name: reboot machine
       how: ansible
       playbook: tests/ansible_collections/reboot.yml
-      order: 61
+      order: 62
     - name: remove other kernels
       how: shell
-      script: pytest -svv tests/integration/conversion/custom-repos/one-kernel-scenario/remove_other_kernels.py
-      order: 62
+      script: pytest -svv tests/integration/one-kernel-scenario/remove_other_kernels.py
```
r0x0d, line 24: Just a friendly reminder to change those integration paths accordingly with this PR https://github.com/oamg/convert2rhel/pull/343 when it gets merged

kokesak: Thanks! :D

kokesak:
This PR is no longer relevant as the issue was fixed with the https://github.com/oamg/convert2rhel/pull/343

PR #356, opened 2021-10-20
Add live kernel version checks and package status

r0x0d:
This PR aims to add a couple of checks before the conversion runs. The checks will prevent the conversion from falling before the process is finished. 

The checks are the following: 

* Check if all the system packages are updated (Through yum/dnf API calls) in enabled repos
* Check if the booted kernel is at the latest version (Through a comparison between `rpm` and `uname`)

If both of the conditions are met, then the conversion will proceed normally in the workflow.

Ticket reference: https://issues.redhat.com/browse/OAMG-4659
Linked ticket: https://issues.redhat.com/browse/OAMG-4709

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -380,10 +380,8 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
 def check_rhel_compatible_kernel_is_used():
     """Ensure the booted kernel is signed, is standard (not UEK, realtime, ...), and has the same version as in RHEL.
 
-        By requesting that, we can be confiden    base.conf.substitutions["ocidomain"] = "oracle.com"
-        base.conf.substitutions["ociregion"] = ""
-    t that the RHEL kernel will provide the same capabilities as on the
-        original system.
+    By requesting that, we can be confident that the RHEL kernel will provide the same capabilities as on the
+    original system.
     """
     logger.task("Prepare: Check kernel compatibility with RHEL")
     if any(
@@ -510,12 +508,13 @@ def is_loaded_kernel_latest():
         # Sort out for the most recent kernel with reverse order
         # In case `repoquery` returns more than one kernel in the output
         # We display the latest one to the user.
-        _, most_recent_kernel = sorted(packages, reverse=True)[0]
+        _, latest_kernel = sorted(packages, reverse=True)[0]
+        latest_kernel = latest_kernel.replace('"', "")
 
         # The loaded kernel version
         uname_output, _ = run_subprocess(["uname", "-r"], print_output=False)
         loaded_kernel = uname_output.rsplit(".", 1)[0]
-        match = compare_package_versions(most_recent_kernel, str(loaded_kernel))
+        match = compare_package_versions(latest_kernel, str(loaded_kernel))
 
         if match == 0:
             logger.info("Kernel currently loaded is at the latest version.")
@@ -526,7 +525,7 @@ def is_loaded_kernel_latest():
                 " Current loaded kernel: %s\n\n"
                 "To proceed with the conversion, update the kernel version by executing the following step:\n\n"
                 "1. yum install %s-%s -y\n"
-                "2. reboot" % (most_recent_kernel, loaded_kernel, package_to_check, most_recent_kernel)
+                "2. reboot" % (latest_kernel, loaded_kernel, package_to_check, latest_kernel)
             )
     else:
         # Repoquery failed to detected any kernel or kernel-core packages in it's repositories
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -22,7 +22,13 @@ import os
 import re
 
 from convert2rhel import grub
-from convert2rhel.pkghandler import call_yum_cmd, get_installed_pkg_objects, get_pkg_fingerprint
+from convert2rhel.pkghandler import (
+    call_yum_cmd,
+    compare_package_versions,
+    get_installed_pkg_objects,
+    get_pkg_fingerprint,
+    get_total_packages_to_update,
+)
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
 from convert2rhel.utils import get_file_content, run_subprocess
@@ -45,11 +51,14 @@ COMPATIBLE_KERNELS_VERS = {
 
 def perform_pre_checks():
     """Early checks after system facts should be added here."""
+
     check_efi()
     check_tainted_kmods()
     check_readonly_mounts()
     check_rhel_compatible_kernel_is_used()
     check_custom_repos_are_valid()
+    check_package_updates()
+    is_loaded_kernel_latest()
 
 
 def perform_pre_ponr_checks():
@@ -371,8 +380,10 @@ def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
 def check_rhel_compatible_kernel_is_used():
     """Ensure the booted kernel is signed, is standard (not UEK, realtime, ...), and has the same version as in RHEL.
 
-    By requesting that, we can be confident that the RHEL kernel will provide the same capabilities as on the
-    original system.
+        By requesting that, we can be confiden    base.conf.substitutions["ocidomain"] = "oracle.com"
+        base.conf.substitutions["ociregion"] = ""
+    t that the RHEL kernel will provide the same capabilities as on the
+        original system.
     """
     logger.task("Prepare: Check kernel compatibility with RHEL")
     if any(
@@ -451,3 +462,87 @@ def _bad_kernel_substring(kernel_release):
         )
         return True
     return False
+
+
+def check_package_updates():
+    """Ensure that the system packages installed are up-to-date."""
+    logger.task("Prepare: Checking if the packages are up-to-date")
+
+    packages_to_update = get_total_packages_to_update()
+
+    if len(packages_to_update) > 0:
+        logger.warning(
+            "The system has %s packages not updated.\n"
+            "List of packages to update: %s.\n\n"
+            "If you wish to update then before continuing with the conversion, execute the following step:\n"
+            "Run: yum update -y\n" % (len(packages_to_update), ", ".join(packages_to_update))
+        )
+    else:
+        logger.info("System is up-to-date.")
+
+
+def is_loaded_kernel_latest():
+    """Check if the loaded kernel is behind or of the same version as in yum repos."""
+    logger.task("Prepare: Checking if the loaded kernel version is the most recent")
+
+    # For Oracle/CentOS Linux 8 the `kernel` is just a meta package, instead, we check for `kernel-core`.
+    # But for 6 and 7 releases, the correct way to check is using `kernel`.
+    package_to_check = "kernel-core" if system_info.version.major >= 8 else "kernel"
+
+    # The latest kernel version on repo
+    repoquery_output, _ = run_subprocess(
+        ["repoquery", "--quiet", "--qf", '"%{BUILDTIME}\\t%{VERSION}-%{RELEASE}"', package_to_check], print_output=False
+    )
+
+    # Repoquery doesn't return any text at all when it can't find any matches for the query (when used with --quiet)
+    if len(repoquery_output) > 0:
+        # Convert to an tuple split with `buildtime` and `kernel` version.
+        # We are also detecting if the sentence "listed more than once in the configuration"
+        # appears in the repoquery output. If it does, we ignore it.
+        # This later check is supposed to avoid duplicate repofiles being defined in the system,
+        # this is a super corner case and should not happen everytime, but if it does, we are aware now.
+        packages = [
+            tuple(str(line).split("\t"))
+            for line in repoquery_output.split("\n")
+            if (line.strip() and "listed more than once in the configuration" not in line.lower())
+        ]
+
+        # Sort out for the most recent kernel with reverse order
+        # In case `repoquery` returns more than one kernel in the output
+        # We display the latest one to the user.
+        _, most_recent_kernel = sorted(packages, reverse=True)[0]
+
+        # The loaded kernel version
+        uname_output, _ = run_subprocess(["uname", "-r"], print_output=False)
+        loaded_kernel = uname_output.rsplit(".", 1)[0]
+        match = compare_package_versions(most_recent_kernel, str(loaded_kernel))
+
+        if match == 0:
+            logger.info("Kernel currently loaded is at the latest version.")
+        else:
+            logger.critical(
+                "The current kernel version loaded is different from the latest version in your repos.\n"
+                " Latest kernel version: %s\n"
+                " Current loaded kernel: %s\n\n"
+                "To proceed with the conversion, update the kernel version by executing the following step:\n\n"
+                "1. yum install %s-%s -y\n"
+                "2. reboot" % (most_recent_kernel, loaded_kernel, package_to_check, most_recent_kernel)
+            )
+    else:
+        # Repoquery failed to detected any kernel or kernel-core packages in it's repositories
+        # we allow the user to provide a environment variable to override the functionality and proceed
+        # with the conversion, otherwise, we just throw an critical logging to them.
+        unsupported_skip = os.environ.get("CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK", "0")
+        if unsupported_skip == "1":
+            logger.warning(
+                "Detected 'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' environment variable, we will skip "
+                "the %s comparison.\n"
+                "Beware, this could leave your system in a broken state. " % package_to_check
+            )
+        else:
+            logger.critical(
+                "Could not find any %s from repositories to compare against the loaded kernel.\n"
+                "Please, check if you have any vendor repositories enabled to proceed with the conversion.\n"
+                "If you wish to ignore this message, set the environment variable "
+                "'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' to 1." % package_to_check
+            )
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -19,6 +19,7 @@ import glob
 import logging
 import os
 import re
+import sys
 
 import rpm
 
@@ -968,3 +969,102 @@ def get_pkg_names_from_rpm_paths(rpm_paths):
     for rpm_path in rpm_paths:
         pkg_names.append(utils.get_package_name_from_rpm(rpm_path))
     return pkg_names
+
+
+def compare_package_versions(version1, version2):
+    """Compare two EVR versions against each other.
+
+
+    :param version1: The version to be compared.
+    :type version1: str
+    :param version2: The version to compare against.
+    :type version2: str
+
+    :example:
+
+        >>> match = compare_package_versions("5.14.10-300.fc35", "5.14.15-300.fc35")
+        >>> match # -1
+
+    .. note::
+
+        Since the return type is a int, this could be difficult to understand
+        the meaning of each number, so here is a quick list of the meaning from
+        every possible number:
+            * -1 if the evr1 is less then evr2 version
+            * 0 if the evr1 is equal evr2 version
+            * 1 if the evr1 is greater than evr2 version
+
+    :return: Return a number indicating if the versions match, are less or greater then.
+    :rtype: int
+    """
+
+    evr1 = utils.string_to_version(version1)
+    evr2 = utils.string_to_version(version2)
+
+    return rpm.labelCompare(evr1, evr2)
+
+
+def get_total_packages_to_update():
+    """Return the total number of packages to update in the system
+
+    It uses both yum/dnf depending on weather they are installed on the system,
+    if it's Oracle/CentOS Linux 7 or 6, it will use `yum`, otherwise it should use `dnf`.
+
+    :return: The packages that need to be updated.
+    :rtype: List[str]
+    """
+    packages = []
+
+    if pkgmanager.TYPE == "yum":
+        packages = _get_packages_to_update_yum()
+    elif pkgmanager.TYPE == "dnf":
+        packages = _get_packages_to_update_dnf()
+
+    return packages
+
+
+def _get_packages_to_update_yum():
+    """Query all the packages with yum that has an update pending on the
+    system."""
+    # Check first if this path is not in the `sys.path`
+    if "/usr/share/yum-cli" not in sys.path:
+        # Yum's installation path for the CLI code
+        # https://github.com/rpm-software-management/yum/blob/4ed25525ee4781907bd204018c27f44948ed83fe/bin/yum#L26
+        sys.path.insert(0, "/usr/share/yum-cli")
+
+    from cli import YumBaseCli  # pylint: disable=E0401
+
+    base = YumBaseCli()
+    packages = base.returnPkgLists(["updates"], None)
+    all_packages = []
+    for package in packages.updates:
+        all_packages.append(package.name)
+
+    return all_packages
+
+
+def _get_packages_to_update_dnf():
+    """Query all the packages with dnf that has an update pending on the
+    system."""
+    from dnf import Base  # pylint: disable=E0401
+
+    packages = []
+    base = Base()
+    # Fix the case when we are trying to query packages in Oracle Linux 8
+    # when the DNF API gets called, those variables are not populated by default.
+    if system_info.id == "oracle":
+        base.conf.substitutions["ocidomain"] = "oracle.com"
+        base.conf.substitutions["ociregion"] = ""
+    base.read_all_repos()
+    base.fill_sack()
+
+    # Get a list of all packages to upgrade in the system
+    base.upgrade_all()
+    base.resolve()
+
+    # Iterate over each and every one of them and append to
+    # the packages list
+    for package in base.transaction:
+        packages.append(package.name)
+
+    return packages
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -1059,7 +1059,7 @@ def _get_packages_to_update_dnf():
     # Same thing for CentOS, but this time, the URL for the vault has changed
     # and instead of just ussing $releasever (8.5, 8.4...), we need either
     # using the correct $releasever (8.5.2111) or setting up the $contentdir
-    if system_info.id == "centos":
+    elif system_info.id == "centos":
         base.conf.substitutions["contentdir"] = system_info.id
 
     base.read_all_repos()
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -1055,6 +1055,13 @@ def _get_packages_to_update_dnf():
     if system_info.id == "oracle":
         base.conf.substitutions["ocidomain"] = "oracle.com"
         base.conf.substitutions["ociregion"] = ""
+
+    # Same thing for CentOS, but this time, the URL for the vault has changed
+    # and instead of just ussing $releasever (8.5, 8.4...), we need either
+    # using the correct $releasever (8.5.2111) or setting up the $contentdir
+    if system_info.id == "centos":
+        base.conf.substitutions["contentdir"] = system_info.id
+
     base.read_all_repos()
     base.fill_sack()
 
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -22,13 +22,15 @@ from collections import namedtuple
 
 import pytest
 
-from convert2rhel import checks, grub, unit_tests
+from convert2rhel import checks, grub, systeminfo, unit_tests
 from convert2rhel.checks import (
     _bad_kernel_package_signature,
     _bad_kernel_substring,
     _bad_kernel_version,
+    check_package_updates,
     check_rhel_compatible_kernel_is_used,
     get_loaded_kmods,
+    is_loaded_kernel_latest,
 )
 from convert2rhel.pkghandler import get_pkg_fingerprint
 from convert2rhel.systeminfo import system_info
@@ -43,12 +45,10 @@ try:
 except ImportError:
     import unittest
 
-
 if sys.version_info[:2] <= (2, 7):
     import mock  # pylint: disable=import-error
 else:
     from unittest import mock  # pylint: disable=no-name-in-module
-
 
 MODINFO_STUB = (
     "/lib/modules/5.8.0-7642-generic/kernel/lib/a.ko.xz\n"
@@ -101,6 +101,9 @@ def test_perform_pre_checks(monkeypatch):
     check_readonly_mounts_mock = mock.Mock()
     check_custom_repos_are_valid_mock = mock.Mock()
     check_rhel_compatible_kernel_is_used_mock = mock.Mock()
+    check_package_updates_mock = mock.Mock()
+    is_loaded_kernel_latest_mock = mock.Mock()
+
     monkeypatch.setattr(
         checks,
         "check_efi",
@@ -126,6 +129,13 @@ def test_perform_pre_checks(monkeypatch):
         "check_custom_repos_are_valid",
         value=check_custom_repos_are_valid_mock,
     )
+    monkeypatch.setattr(
+        checks,
+        "check_custom_repos_are_valid",
+        value=check_custom_repos_are_valid_mock,
+    )
+    monkeypatch.setattr(checks, "check_package_updates", value=check_package_updates_mock)
+    monkeypatch.setattr(checks, "is_loaded_kernel_latest", value=is_loaded_kernel_latest_mock)
 
     checks.perform_pre_checks()
 
@@ -133,6 +143,8 @@ def test_perform_pre_checks(monkeypatch):
     check_efi_mock.assert_called_once()
     check_readonly_mounts_mock.assert_called_once()
     check_rhel_compatible_kernel_is_used_mock.assert_called_once()
+    check_package_updates_mock.assert_called_once()
+    is_loaded_kernel_latest_mock.assert_called_once()
 
 
 def test_pre_ponr_checks(monkeypatch):
@@ -816,3 +828,135 @@ class TestReadOnlyMountsChecks(unittest.TestCase):
         self.assertEqual(len(checks.logger.critical_msgs), 1)
         self.assertEqual(len(checks.logger.info_msgs), 0)
         self.assertTrue("Unable to access the repositories passed through " in checks.logger.critical_msgs[0])
+
+
+@pytest.mark.parametrize(
+    ("packages", "exception", "expected"),
+    (
+        (["package-1", "package-2"], True, "The system has {} packages not updated."),
+        ([], False, "System is up-to-date."),
+    ),
+)
+def test_check_package_updates(packages, exception, expected, monkeypatch, caplog):
+    monkeypatch.setattr(checks, "get_total_packages_to_update", value=lambda: packages)
+
+    check_package_updates()
+    if exception:
+        expected = expected.format(len(packages))
+
+    assert expected in caplog.records[-1].message
+
+
+@pytest.mark.parametrize(
+    ("repoquery_version", "uname_version", "return_code", "major_ver", "package_name", "raise_system_exit"),
+    (
+        ("1634146676\t3.10.0-1160.45.1.el7", "3.10.0-1160.42.2.el7.x86_64", 0, 8, "kernel-core", True),
+        ("1634146676\t3.10.0-1160.45.1.el7", "3.10.0-1160.45.1.el7.x86_64", 0, 7, "kernel", False),
+        ("1634146676\t3.10.0-1160.45.1.el7", "3.10.0-1160.45.1.el7.x86_64", 0, 6, "kernel", False),
+    ),
+)
+def test_is_loaded_kernel_latest(
+    repoquery_version, uname_version, return_code, major_ver, package_name, raise_system_exit, monkeypatch, caplog
+):
+    Version = namedtuple("Version", ("major", "minor"))
+    monkeypatch.setattr(
+        checks.system_info,
+        "version",
+        value=Version(major=major_ver, minor=0),
+    )
+
+    run_subprocess_mocked = mock.Mock(
+        spec=run_subprocess,
+        side_effect=run_subprocess_side_effect(
+            (
+                (
+                    "repoquery",
+                    "--quiet",
+                    "--qf",
+                    '"%{BUILDTIME}\\t%{VERSION}-%{RELEASE}"',
+                    package_name,
+                ),
+                (
+                    repoquery_version,
+                    return_code,
+                ),
+            ),
+            (("uname", "-r"), (uname_version, return_code)),
+        ),
+    )
+    monkeypatch.setattr(
+        checks,
+        "run_subprocess",
+        value=run_subprocess_mocked,
+    )
+
+    if raise_system_exit:
+        with pytest.raises(SystemExit):
+            is_loaded_kernel_latest()
+
+        repoquery_kernel_version = repoquery_version.split("\t", 1)[1]
+        uname_kernel_version = uname_version.rsplit(".", 1)[0]
+        assert "Latest kernel version: %s\n" % repoquery_kernel_version in caplog.records[-1].message
+        assert "Current loaded kernel: %s\n" % uname_kernel_version in caplog.records[-1].message
+    else:
+        is_loaded_kernel_latest()
+        assert "Kernel currently loaded is at the latest version." in caplog.records[-1].message
+
+
+@pytest.mark.parametrize(
+    ("repoquery_version", "return_code", "major_ver", "package_name", "unsupported_skip", "expected"),
+    (
+        (
+            "",
+            0,
+            8,
+            "kernel-core",
+            "1",
+            "Detected 'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' environment variable",
+        ),
+        ("", 0, 8, "kernel-core", "0", "Could not find any {} from repositories"),
+        ("", 0, 7, "kernel", "0", "Could not find any {} from repositories"),
+        ("", 0, 6, "kernel", "0", "Could not find any {} from repositories"),
+    ),
+)
+def test_is_loaded_kernel_latest_unsupported_skip(
+    repoquery_version, return_code, major_ver, package_name, unsupported_skip, expected, monkeypatch, caplog
+):
+    run_subprocess_mocked = mock.Mock(
+        spec=run_subprocess,
+        side_effect=run_subprocess_side_effect(
+            (
+                (
+                    "repoquery",
+                    "--quiet",
+                    "--qf",
+                    '"%{BUILDTIME}\\t%{VERSION}-%{RELEASE}"',
+                    package_name,
+                ),
+                (
+                    repoquery_version,
+                    return_code,
+                ),
+            ),
+        ),
+    )
+    monkeypatch.setattr(
+        checks.system_info,
+        "version",
+        value=systeminfo.Version(major=major_ver, minor=0),
+    )
+    monkeypatch.setattr(
+        checks,
+        "run_subprocess",
+        value=run_subprocess_mocked,
+    )
+    monkeypatch.setattr(os, "environ", {"CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK": unsupported_skip})
+
+    if unsupported_skip == "0":
+        with pytest.raises(SystemExit):
+            is_loaded_kernel_latest()
+        expected = expected.format(package_name)
+    else:
+        is_loaded_kernel_latest()
+
+    assert expected in caplog.records[-1].message
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -33,9 +33,20 @@ else:
 
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
 from convert2rhel import pkghandler, pkgmanager, utils
+from convert2rhel.pkghandler import (
+    _get_packages_to_update_dnf,
+    _get_packages_to_update_yum,
+    get_total_packages_to_update,
+)
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
 from convert2rhel.unit_tests import GetLoggerMocked, is_rpm_based_os
+
+
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
 
 
 class TestPkgHandler(unit_tests.ExtendedTestCase):
@@ -1391,6 +1402,84 @@ def test_call_yum_cmd_w_downgrades(monkeypatch, retcode, output):
     resolve_dep_errors.assert_not_called()
 
 
+@pytest.mark.parametrize(
+    ("version1", "version2", "expected"),
+    (
+        ("123-4.fc35", "123-4.fc35", 0),
+        ("123-5.fc35", "123-4.fc35", 1),
+        ("123-3.fc35", "123-4.fc35", -1),
+        ("4.6~pre16262021g84ef6bd9-3.fc35", "4.6~pre16262021g84ef6bd9-3.fc35", 0),
+        ("2:8.2.3568-1.fc35", "2:8.2.3568-1.fc35", 0),
+    ),
+)
+def test_compare_package_versions(version1, version2, expected):
+    assert pkghandler.compare_package_versions(version1, version2) == expected
+
+
+@pytest.mark.parametrize(
+    ("package_manager_type", "packages"),
+    (
+        (
+            "yum",
+            [
+                "convert2rhel.noarch-0.24-1.20211111151554764702.pr356.28.ge9ed160.el8",
+                "convert2rhel.src-0.24-1.20211111151554764702.pr356.28.ge9ed160.el8",
+            ],
+        ),
+        (
+            "dnf",
+            [
+                "dunst-1.7.1-1.fc35.x86_64",
+                "dunst-1.7.0-1.fc35.x86_64",
+                "java-11-openjdk-headless-1:11.0.13.0.8-2.fc35.x86_64",
+            ],
+        ),
+    ),
+)
+def test_get_total_packages_to_update(package_manager_type, packages, monkeypatch):
+    monkeypatch.setattr(pkgmanager, "TYPE", package_manager_type)
+    monkeypatch.setattr(pkghandler, "_get_packages_to_update_%s" % package_manager_type, value=lambda: packages)
+
+    assert get_total_packages_to_update() == packages
+
+
+@pytest.mark.skipif(pkgmanager.TYPE != "yum", reason="No yum module detected on the system, skipping it.")
+@pytest.mark.parametrize(("packages"), ((["package-1", "package-2", "package-3"],)))
+def test_get_packages_to_update_yum(packages, monkeypatch):
+    sys.path.insert(0, "/usr/share/yum-cli")
+    from cli import YumBaseCli  # pylint: disable=E0401
+
+    PkgName = namedtuple("PkgNames", ["name"])
+    PkgUpdates = namedtuple("PkgUpdates", ["updates"])
+    transaction_pkgs = []
+    for package in packages:
+        transaction_pkgs.append(PkgName(package))
+
+    pkg_lists_mock = mock.Mock(return_value=PkgUpdates(transaction_pkgs))
+
+    monkeypatch.setattr(YumBaseCli, "returnPkgLists", value=pkg_lists_mock)
+
+    assert _get_packages_to_update_yum() == packages
+
+
+@pytest.mark.skipif(pkgmanager.TYPE != "dnf", reason="No dnf module detected on the system, skipping it.")
+@pytest.mark.parametrize(("packages"), ((["package-1", "package-2", "package-3"],)))
+def test_get_packages_to_update_dnf(packages, monkeypatch):
+    from dnf import Base  # pylint: disable=E0401
+
+    dummy_mock = mock.Mock()
+    PkgName = namedtuple("PkgNames", ["name"])
+    transaction_pkgs = [PkgName(package) for package in packages]
+
+    monkeypatch.setattr(Base, "read_all_repos", value=dummy_mock)
+    monkeypatch.setattr(Base, "fill_sack", value=dummy_mock)
+    monkeypatch.setattr(Base, "upgrade_all", value=dummy_mock)
+    monkeypatch.setattr(Base, "resolve", value=dummy_mock)
+    monkeypatch.setattr(Base, "transaction", value=transaction_pkgs)
+
+    assert _get_packages_to_update_dnf() == packages
+
+
 YUM_PROTECTED_ERROR = """Error: Trying to remove "systemd", which is protected
 Error: Trying to remove "yum", which is protected"""
 
@@ -1460,7 +1549,6 @@ Problem: cannot install both python39-psycopg2-2.8.6-2.module+el8.4.0+9822+20bf1
    - package python39-psycopg2-debug-2.8.6-2.module_el8.4.0+680+7b309a77.x86_64 requires python39-psycopg2 = 2.8.6-2.module_el8.4.0+680+7b309a77, but none of the providers can be installed
    - cannot install the best update candidate for package python39-psycopg2-2.8.6-2.module_el8.4.0+680+7b309a77.x86_64
    - problem with installed package python39-psycopg2-debug-2.8.6-2.module_el8.4.0+680+7b309a77.x86_64"""
-
 
 # The following yum error is currently not being handled by the tool. The
 # tool would somehow need to decide, which of the two packages to remove and
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -41,6 +41,7 @@ from convert2rhel.pkghandler import (
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
 from convert2rhel.unit_tests import GetLoggerMocked, is_rpm_based_os
+from convert2rhel.unit_tests.conftest import all_systems, oracle7
 
 
 if sys.version_info[:2] <= (2, 7):
@@ -193,7 +194,13 @@ class TestPkgHandler(unit_tests.ExtendedTestCase):
 
         self.assertEqual(
             utils.run_subprocess.cmd,
-            ["yum", "install", "-y", "--releasever=8", "--setopt=module_platform_id=platform:el8"],
+            [
+                "yum",
+                "install",
+                "-y",
+                "--releasever=8",
+                "--setopt=module_platform_id=platform:el8",
+            ],
         )
 
     @unit_tests.mock(system_info, "version", namedtuple("Version", ["major", "minor"])(7, 0))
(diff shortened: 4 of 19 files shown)


convert2rhel/checks.py
```diff
@@ -449,3 +453,47 @@ def _bad_kernel_substring(kernel_release):
...
+    """Check if the installed kernel is behind or at the same version in yum repos"""
+    logger.task("Prepare: Checking if the installed kernel version is the most recent.")
+
+    # The latest kernel version on repo
+    repoquery_output, _ = run_subprocess('repoquery --qf "%{version}-%{release}.%{arch}" kernel')
+
+    # The installed kernel version
+    uname_output, _ = run_subprocess("uname -r")
+
+    match = LooseVersion(repoquery_output) > LooseVersion(uname_output)
```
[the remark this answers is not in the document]
abadger: I see that we `import rpm` in the pkghandler file so we already have a dependency on that.  rpm has a function to compare NEVR (Name epoch version release). `rpm.labelCompare((epoch1, version1, release1), (epoch2, version2, release2))`.  

convert2rhel is GPLv3 or later and rpmdevtools is GPLv2 or later (correct me if I'm wrong on either of those points :-) so you should be able to look at the source of rpmdev-rpmvercmp for some additional examples: https://pagure.io/rpmdevtools/blob/main/f/rpmdev-vercmp#_108

convert2rhel/checks.py
```diff
@@ -449,3 +453,47 @@ def _bad_kernel_substring(kernel_release):
...
+def check_package_updates():
+    """Ensures that the system packages installed are up-to-date."""
+    logger.task("Prepare: Checking if the packages are up-to-date")
+
+    yum_stdout, _ = run_subprocess("yum check-update -q", print_output=False)
+
+    # There is a whole blank line before outputting the packages to update
+    # and to output the real counting, we just -1 the counting.
+    packages_to_update_count = yum_stdout.count("\n") - 1
+
```
abadger, line 467: It would be slower but more robust to split the output and filter out the lines we don't want.  I also noticed that when I ran yum check-update -q locally, I got other, non package lines:
> vim-minimal.x86_64                             2:8.2.3568-1.fc33                        updates-testing
> Security: kernel-core-5.14.15-100.fc33.x86_64 is an installed security update
> Security: kernel-core-5.14.13-100.fc33.x86_64 is the currently running version

What do you think about:

``` python
lines = yum_stdout.split("\n")

# Filter lines that only have whitespace or are blank
lines = (line for line in yum_stdout.split("\n") if line.strip())
# Filter informational lines that don't have to do with available packages
lines = (line for line in lines if not line.endswith("is an installed security update"))
packages = [line for line in lines if not line.endswith("is the currently running version")]

packages_to_update_count = len(packages)
```

convert2rhel/checks.py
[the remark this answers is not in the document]
abadger: Note that arch probably doesn't make sense as part of greater than and less than comparisons.  But it would make sense for an equality comparison.  Perhaps you want to do something to make sure that you have the right arch prior to doing your version comparison? (Or you might want to assume correct arch due to having setup the repositories earlier?)

convert2rhel/unit_tests/checks_test.py
```diff
@@ -813,3 +825,86 @@ def test_custom_repos_are_invalid(self):
...
+    )
+    monkeypatch.setattr(
+        checks,
+        "run_subprocess",
+        value=run_subprocess_mocked,
+    )
+
+    if raise_system_exit:
+        with pytest.raises(SystemExit):
+            assert check_package_updates()
```
SpyTec, line 865: After the raises, it would be good to check if the amount of packages is accurate. `The system has %s packages to be updated`

convert2rhel/unit_tests/checks_test.py
```diff
@@ -813,3 +825,86 @@ def test_custom_repos_are_invalid(self):
...
         self.assertTrue("Unable to access the repositories passed through " in checks.logger.critical_msgs[0])
+
+
+@pytest.mark.parametrize(
+    ("package", "return_code", "raise_system_exit"),
+    (
+        (
+            "convert2rhel.noarch   0.24-1.20211019110627581338.pr343.11.g4ff8753.el8   copr:copr.fedorainfracloud.org:group_oamg:convert2rhel\n"
+            "convert2rhel.src      0.24-1.20211019110627581338.pr343.11.g4ff8753.el8   copr:copr.fedorainfracloud.org:group_oamg:convert2rhel\n"
+            "tzdata.noarch         2021c-1.el8                                         baseos\n",
```
SpyTec, line 836: Add this as well since Toshio mentioned it can appear
```
Security: kernel-core-5.14.15-100.fc33.x86_64 is an installed security update
Security: kernel-core-5.14.13-100.fc33.x86_64 is the currently running version
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -813,3 +825,86 @@ def test_custom_repos_are_invalid(self):
...
+    )
+    monkeypatch.setattr(
+        checks,
+        "run_subprocess",
+        value=run_subprocess_mocked,
+    )
+
+    if raise_system_exit:
+        with pytest.raises(SystemExit):
+            check_installed_kernel_version()
```
SpyTec, line 907: Check logs here too if possible to see that the current and latest kernel is correct as well

convert2rhel/checks.py
```diff
@@ -449,3 +453,47 @@ def _bad_kernel_substring(kernel_release):
         )
         return True
     return False
+
+
+def check_package_updates():
+    """Ensures that the system packages installed are up-to-date."""
+    logger.task("Prepare: Checking if the packages are up-to-date")
```
SpyTec, line 460:
```suggestion
    logger.task("Prepare: Checking if the system packages are up-to-date")
```

convert2rhel/checks.py
```diff
@@ -449,3 +453,47 @@ def _bad_kernel_substring(kernel_release):
...
+    else:
+        logger.info("System is up-to-date.")
+
+
+def check_installed_kernel_version():
+    """Check if the installed kernel is behind or at the same version in yum repos"""
+    logger.task("Prepare: Checking if the installed kernel version is the most recent.")
+
+    # The latest kernel version on repo
+    repoquery_output, _ = run_subprocess('repoquery --qf "%{version}-%{release}.%{arch}" kernel')
```
SpyTec, line 483: This could show multiple kernel versions, at least does for me on Fedora. Will this happen on CentOS or Oracle Linux?

SpyTec (dismissed):
Overall code looks good!

convert2rhel/checks.py
r0x0d: I agree with @SpyTec, the code looks way cleaner and easy to understand with the regex. I tested and got all the packages that needed to update in one go, and if we have to output to the user in the future about the packages, we can just iterate over the list and show the names to them.

convert2rhel/checks.py
```diff
@@ -30,6 +32,7 @@
 
 logger = logging.getLogger(__name__)
 
+PACKAGE_NAMES_RE = r"^[0-9]*:?([a-z][a-z0-9-_]*?)(-[0-9])*\."
```
abadger, line 35: One problem with regexes is that they are generally slower than other ways of checking.  Assigning regexes outside of any potential code loops like you're doing here is one way to mitigate that.  However, you need to not just assign the string here but also compile the regex (the speed savings is because you only have to compile the regex once, when the module is first imported).   So do something like this:
```suggestion
PACKAGE_NAMES_RE = re.compile(r"^[0-9]*:?([a-z][a-z0-9-_]*?)(-[0-9])*\.")
```
```

Comment (abadger) [reply convert2rhel/checks.py:35]: Hmm.. also... does this regex work for the data we're being given?  I was going to make a suggestion about the other problem with regexes being that they're hard to debug so we should comment new regexes heavily.  But when I tried running some yum check-output data through it so I could understand what this regex is doing, I came out with zero matches instead of the one match I was expecting.

``` python 
>>> import re
>>> import sh  # amoffat's sh library
>>> a = re.compile(r"^[0-9]*:?([a-z][a-z0-9-_]*?)(-[0-9])*\.")
>>> yum_stdout = sh.yum(['check-update', '-q'], _ok_code=[0, 100])
>>> yum_stdout

python3-pyOpenSSL.noarch              21.0.0-1.fc33              updates-testing
Security: kernel-core-5.14.15-100.fc33.x86_64 is an installed security update
Security: kernel-core-5.14.13-100.fc33.x86_64 is the currently running version

>>> re.findall(a, str(yum_stdout))
[]
```

convert2rhel/checks.py
abadger: <nod> Regexes can simplify surrounding code.  However, they mean you have a regex to maintain and understand. So do be mindful of that.

https://www.explainxkcd.com/wiki/index.php/1171:_Perl_Problems

convert2rhel/checks.py
r0x0d: @SpyTec, Could you provide the output for me? I think it may be getting kernels from other repositories... If so, this is something I haven't tested. Trying to do some more examples to catch those kernels. 

I'm thinking that we can use the same approach as here: https://github.com/oamg/convert2rhel/pull/353/files#diff-cf19bc90f65d9e26e1b6fa7dcd84a18dfba097f65706e211a4f4ac71b1b467f6R389

We could include the BUILDTIME variable and sort it out for the newest one, what do you think?

convert2rhel/checks.py
r0x0d: @abadger I've tested with the 2 ways, and now (yesterday I got the same output as you, an empty []) they are working as expected. I've changed to your suggestions (compiling the regex first) 

```python
('packages', [(u'binutils', u''), (u'convert2rhel', u''), (u'glibc', u''), (u'glibc-common', u''), (u'grub2', u''), (u'grub2-common', u''), (u'grub2-pc', u''), (u'grub2-pc-modules', u''), (u'grub2-tools', u''), (u'grub2-tools-extra', u''), (u'grub2-tools-minimal', u''), (u'kernel', u''), (u'kernel-tools', u''), (u'kernel-tools-libs', u''), (u'nfs-utils', u''), (u'nspr', u''), (u'nss', u''), (u'nss-softokn', u''), (u'nss-softokn-freebl', u''), (u'nss-sysinit', u''), (u'nss-tools', u''), (u'nss-util', u''), (u'openssl', u''), (u'openssl-libs', u''), (u'python-perf', u''), (u'sudo', u''), (u'tzdata', u'')])
```

And the log message

```bash
CRITICAL - The system has 27 packages to be updated. To proceed with the conversion, update the packages on your system by executing the following step:
```

convert2rhel/unit_tests/checks_test.py
r0x0d: @SpyTec, Those two lines are just informational and not telling the end-user that `kernel-core` needs to be updated? So our solution should gracefully ignore that, right?

convert2rhel/checks.py
abadger: Interesting.  Can you post your output from `yum check-update` or explain how the regex works?  I'm seeing that the unittests also appear to be failing in the same way as my local attempts are... zero packages are found in the output.

convert2rhel/checks.py
```diff
@@ -447,3 +452,57 @@ def _bad_kernel_substring(kernel_release):
...
+
+    # total_packages = packages.count()
+
+    lines = yum_stdout.split("\n")
+
+    # Filter lines that only have whitespace or are blank
+    lines = (line for line in yum_stdout.split("\n") if line.strip())
+    # Filter informational lines that don't have to do with available packages
+    lines = (line for line in lines if not line.endswith("is an installed security update"))
+    packages = [line for line ingit lines if not line.endswith("is the currently running version")]
```
abadger, line 473:
```suggestion
    packages = [line for line in lines if not line.endswith("is the currently running version")]
```

convert2rhel/checks.py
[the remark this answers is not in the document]
r0x0d: Do you guys think that comparing versions is a really important step on this? I mean, `repoquery` should output the latest version available at the current repositories, so if the `uname -r` doesn't match the output of `repoquery` it would mean that it's not updated, right? 

Don't know if I'm really on the correct train of thought here. If not, I can change to do version comparison with `rpm.labelCompare()`

convert2rhel/checks.py
bocekm: The regex would not handle well the situation with the security update @abadger mentioned - it would give us two packages to update even though there's none. So, I'm for the @abadger's approach.

bocekm (changes requested):

convert2rhel/checks.py
abadger: Here's output on my Fedora 33 system:

> [pts/15@camelback /srv/convert2rhel/c2r-vagrant]$ repoquery --qf "%{version}-%{release}.%{arch}" kernel 
> Last metadata expiration check: 3:01:15 ago on Thu 04 Nov 2021 08:32:35 AM PDT.
> 5.14.15-100.fc33.x86_64
> 5.8.15-301.fc33.x86_64


I think it might retrieve the latest package version for each repository?  In Fedora there's a repository for the release and separate repositories for updates so the kernel package could appear in all three of those.

I'm not sure if it's the same with rhel or if they simply have a monolithic repo.

convert2rhel/checks.py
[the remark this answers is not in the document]
abadger: I'm not familiar enough to give a definitive answer but here are some ideas of potential problems:

* What do we want to do if the repository is out-of-sync and reports an older version?
* What do we want to do in this function if the user has a non-standard but newer kernel installed?

convert2rhel/unit_tests/checks_test.py
```diff
@@ -813,3 +825,89 @@ def test_custom_repos_are_invalid(self):
         self.assertEqual(len(checks.logger.critical_msgs), 1)
         self.assertEqual(len(checks.logger.info_msgs), 0)
         self.assertTrue("Unable to access the repositories passed through " in checks.logger.critical_msgs[0])
+
+
+@pytest.mark.parametrize(
+    ("package", "return_code", "raise_system_exit"),
+    (
+        (
+            "convert2rhel.noarch   0.24-1.20211019110627581338.pr343.11.g4ff8753.el8   copr:copr.fedorainfracloud.org:group_oamg:convert2rhel\n"
```
abadger, line 834:
```suggestion
            "\n",
            "convert2rhel.noarch   0.24-1.20211019110627581338.pr343.11.g4ff8753.el8   copr:copr.fedorainfracloud.org:group_oamg:convert2rhel\n"
```

Also have an initial blank line since that's also a feature of the yum check-updates output

convert2rhel/checks.py
r0x0d: I see @abadger, gonna research a bit more on the `repoquery` output. I think it's querying on all repos the user may have enabled in his system, so it indeed could be a problem. The easiest way I can think of fixing this is sorting out the latest `kernel` as I did previously in another PR. 

Thanks for your output!

convert2rhel/checks.py
```diff
@@ -447,3 +452,57 @@ def _bad_kernel_substring(kernel_release):
...
+
+    total_packages = len(packages)
+
+    if total_packages > 0:
+        logger.critical(
```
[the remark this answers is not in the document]
r0x0d: We could change this to a `logger.warning()` instead of the critical. I think the `warning` will not cause an `SystemExit`, right?

convert2rhel/unit_tests/checks_test.py
SpyTec: Yeah exactly. We should still have it in the test to make sure we don't pick up on it though in the function

convert2rhel/checks.py
bocekm: @j-mracek, in the output of `yum check-update` shared by @abadger in https://github.com/oamg/convert2rhel/pull/356/files#r741452001 (`is an installed security update`/`is the currently running version`) we can see the same NEVRA at both lines. I suppose the NEVRAs can differ, right? Would in such a case such NEVRA appear with the "Security:" prefix or would it appear on a separate line without the prefix as well?

Also, is there any yum/dnf python API function that would give us the information provided by the `check-update` CLI command?

convert2rhel/unit_tests/checks_test.py
r0x0d: Got it. Added to the tests case.

convert2rhel/checks.py
```diff
@@ -447,3 +452,57 @@ def _bad_kernel_substring(kernel_release):
...
+            "To proceed with the conversion, update the packages on your system by executing the following step:\n\n"
+            "Run: yum update -y\n" % total_packages
+        )
+    else:
+        logger.info("System is up-to-date.")
+
+
+def check_installed_kernel_version():
+    """Check if the installed kernel is behind or at the same version in yum repos"""
+    logger.task("Prepare: Checking if the installed kernel version is the most recent.")
```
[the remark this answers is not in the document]
r0x0d: Changed to the more appropriate function name

convert2rhel/checks.py
r0x0d: Minor typo, fixed!

convert2rhel/unit_tests/checks_test.py
r0x0d: Added the assserts corresponding to this one.

convert2rhel/unit_tests/checks_test.py
r0x0d: Added a few more cases to the test as well to the addition of the two lines of the `Security:`log

convert2rhel/unit_tests/checks_test.py
r0x0d: Updated that assert as well.

convert2rhel/checks.py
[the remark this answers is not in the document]
r0x0d: In a recent call with @bocekm we were able to get this right. I've added the code for properly handling the `repoquery` and `uname` output, and to add the `rpm.labelCompare` function too.

convert2rhel/checks.py
r0x0d: As @bocekm stated in another thread here, we decided to go with the manual approach for that validation as it is handling better in some other cases for `yum check-update -q` command. 

And as far as the regex, I don't really know why the outputs were different. I'm not an expert on regex, but I think it was something along with the regex compilation method. Don't really know what caused the empty `[]` on the output.

convert2rhel/checks.py
j-mracek: I am little bit lost here, but I will try my best. There are multiple possibilities how to receive information provided by check-update command. Try use Query class with filters updates, and security. Please can you describe exactly what set of packages are you looking for? Then I can look in implementation.

r0x0d: @j-mracek We are looking for general packages updates. We are not aiming at a specific set of packages, but rather, if there is any package to be updated in the system. Since @abadger provided us with a different output from the ones I've got in my testing, we were wondering if `yum check-update` could give us any different messages that we are not aware of.

j-mracek: What about:
In case that you can directly play with goal in base
base.upgrade_all()
base.resolve()
Then iterate over base.transaction
for x in base.transaction:
    print(x, x.action_name)

kokesak:
After the first discussion with @r0x0d  we created a test case table with scenarios on how to test this PR.
https://docs.google.com/spreadsheets/d/17fw2qFEUI9g-M5_sf3uFnwIVhn7zswcj0eETvEOKSD0/edit#gid=0

This table will be updated as the development on this task progress.

convert2rhel/checks.py
r0x0d: @j-mracek For some reason, I was not able to manage to get your code snippet working. It doesn't output anything for me when I run it, maybe I'm doing something wrong? 

```python
import dnf

db = dnf.Base()

db.fill_sack()

db.upgrade_all()
db.resolve()

[print(package) for package in db.transaction]
```

I also tried like this: 

```python
import dnf
db = dnf.Base()

db.fill_sack()

query = db.sack.query()
installed = query.upgrades()

print("Installed dnf package:")
for pkg in list(installed):
    print(pkg, pkg.reponame)
```

But none of the above worked... Already checked the output of `dnf check-update` command and it reported I have 50 packages to update.

Also, @j-mracek, I've checked the code for dnf/yum regarding the command `check-update` and both seem to use a different method. Do you think we could also try to do to the same as the tool? Or is the solution you shared with us easier to implement?

Thanks in advance for your help!

convert2rhel/pkghandler.py
```diff
@@ -895,3 +894,51 @@ def compare_package_versions(version1, version2):
...
+
+def _get_packages_to_update_yum():
+    """Query all the packages with yum that has an update pending on the system."""
+    from cli import YumBaseCli
+
+    base = YumBaseCli()
+
+    packages = base.returnPkgLists(["updates"], None)
+
+    print("PACKAGES: ", packages)
```
r0x0d, line 922: It's not done yet. The print is just for debugging purposes, will remove it when it's done

convert2rhel/checks.py
j-mracek: Do you still need any assistance?

r0x0d: @j-mracek, for now, I think I'm all good! Thanks for your helping and patience!

convert2rhel/checks.py
```diff
@@ -447,3 +456,57 @@ def _bad_kernel_substring(kernel_release):
...
+    else:
+        logger.info("System is up-to-date.")
+
+
+def is_loaded_kernel_latest():
+    """Check if the loaded kernel is behind or of the same version as in yum repos."""
+    logger.task("Prepare: Checking if the loaded kernel version is the most recent.")
+
+    # The latest kernel version on repo
+    repoquery_output, _ = run_subprocess('repoquery -q --qf "%{BUILDTIME}\\t%{VERSION}-%{RELEASE}" kernel')
```
pirat89, line 483: @r0x0d @oamg/convert2rhel-developers   what in case the repository is not reachable? e.g.
* detached systems
* disabled repositories

in such a case you will not obtain any information to be able to compare the versions. (we already hit such a systems multiple times). unless you have written in requirements to have enabled base repositories for the particular system (for OL, CentOS, ..). If not, I suggest to have a fallback to check the latest installed kernel as well.

Additionally, for the installed packages, for EL 8+ you need to care about `kernel-core` rpms instead of `kernel`, as the `kernel` rpm is just a metapackage since EL 8 and it doesn't have to be installed at all (this is not problem when you check what's newest pkg in repository - if repo is enabled and accessible)

pirat89: to simulate such a situation, here is my output from centos 7 container
```
[root@1c732f612637 /]# repoquery -q --qf "%{BUILDTIME}\\t%{VERSION}-%{RELEASE}" --disablerepo="*" kernel
[root@1c732f612637 /]# echo $?
0
```

r0x0d: Hey, @pirat89! Thanks for your insights on this. Just out of curiosity, how did you guys handle those cases you mentioned? I don't recall having a programmatic set of enabling repositories in the tool, as far as I know, having at least the base repositories enabled is a user requirement (But I may be wrong on that). 

@SpyTec do you know any cases where the base repos for a system are disabled for our users?

convert2rhel/checks.py
```diff
@@ -447,3 +456,59 @@ def _bad_kernel_substring(kernel_release):
...
+def is_loaded_kernel_latest():
+    """Check if the loaded kernel is behind or of the same version as in yum repos."""
+    logger.task("Prepare: Checking if the loaded kernel version is the most recent.")
+
+    # For OL/CentOS 8 the `kernel` is just a meta package, instead, we check for `kernel-core`.
+    # But for 6 and 7 releases, the correct way to check is using `kernel`.
+    package_to_check = "kernel-core" if system_info.version.major >= 8 else "kernel"
+
+    # The latest kernel version on repo
+    repoquery_output, _ = run_subprocess('repoquery -q --qf "%{BUILDTIME}\\t%{VERSION}-%{RELEASE}" ' + package_to_check)
```
bocekm, line 487:
```suggestion
    repoquery_output, _ = run_subprocess('repoquery --quiet --qf "%{BUILDTIME}\\t%{VERSION}-%{RELEASE}" ' + package_to_check)
```
The `-q` means `no-op for rpmquery compatibility` on RHEL 6 - i.e. it does nothing.

bocekm (changes requested):

convert2rhel/checks.py
SpyTec: @pirat89 From our stand-up and GChat, we require repos to be enabled as mentioned in the docs. So not an issue we need to worry about for this PR

> Access to repositories providing packages from the original distribution is required, especially to the base repository, to perform a rollback. 

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/converting_from_an_rpm-based_linux_distribution_to_rhel/index#con_rollback_converting-from-a-linux-distribution-to-rhel

requirements/centos8.requirements.txt
```diff
@@ -1,5 +1,5 @@
 pylint==2.6.0
 astroid==2.4.2
-pytest==6.2.1
+pytest==6.2.5
```
r0x0d, line 3: pytest was throwing an error when running locally with python `3.10.0`, the fix was upgrading the version to `6.2.5`

convert2rhel/pkgmanager.py
```diff
@@ -14,10 +14,16 @@
...
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
+import sys
+
 
 try:
     from yum import *
 
+    # Yum's installation path for the CLI code
+    # https://github.com/rpm-software-management/yum/blob/4ed25525ee4781907bd204018c27f44948ed83fe/bin/yum#L26
+    sys.path.insert(0, "/usr/share/yum-cli")
```
r0x0d, line 25: This will be imported every time `yum` is detected, should I move it to here? https://github.com/oamg/convert2rhel/blob/11bab92119510bb40eb7e9c9bc9d0010f7122ff7/convert2rhel/pkghandler.py#L916 For now, this line is only used by this function.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1387,6 +1398,79 @@ def test_fix_default_kernel_with_no_incorrect_kernel(self):
...
+
+    dummy_mock = mock.Mock()
+
+    monkeypatch.setattr(Base, "read_all_repos", value=dummy_mock)
+    monkeypatch.setattr(Base, "fill_sack", value=dummy_mock)
+    monkeypatch.setattr(Base, "upgrade_all", value=dummy_mock)
+    monkeypatch.setattr(Base, "resolve", value=dummy_mock)
+    monkeypatch.setattr(Base, "transaction", value=packages)
+
+    assert _get_packages_to_update_dnf() == packages
```
r0x0d, line 1471: With this `pytest.mark.skipif` our coverage decreases a little because there will not be both of the package managers at the same time on a system. 

I know there is a way to combine multiple coverage runs to make it better, but don't know how much effort will be to adapt our CI to do that.

r0x0d: One thing that could be possible is splitting the `test_get_total_packages_to_update` test and reapplying these specific monkeypatchs to be compatible with either dnf or yum when running, so it could be something like this: 

* test_get_total_packages_to_update_with_yum
* test_get_total_packages_to_update_with_dnf

convert2rhel/checks.py
```diff
@@ -45,11 +51,14 @@
...
 def perform_pre_checks():
     """Early checks after system facts should be added here."""
+
     check_efi()
     check_tainted_kmods()
     check_readonly_mounts()
     check_rhel_compatible_kernel_is_used()
     check_custom_repos_are_valid()
+    check_package_updates()
+    is_loaded_kernel_latest()
```
r0x0d, line 61: This would be the correct order for these checks?

r0x0d (commented):
Reviewed.

convert2rhel/pkgmanager.py
bocekm: As it is now the `sys.path.insert(0, "/usr/share/yum-cli")` will be executed just once because the pkgmanager.py is loaded just once per convert2rhel execution. But moving it to the pkghandler.py makes sense from the point of view that it logically relates just to the one `_get_packages_to_update_yum()` function.

convert2rhel/checks.py
bocekm: Yes. But please move the following from main.py to run before the `perform_pre_checks()` call:
```
loggerinst.task("Prepare: Clear YUM/DNF version locks")
pkghandler.clear_versionlock()
```
The reason is that when [yum versionlock](https://access.redhat.com/solutions/98873) is in place for kernel, `yum update` will not report the kernel to be updated even when a newer version is available in repos.

bocekm (changes requested):

convert2rhel/unit_tests/pkghandler_test.py
SpyTec: Decrease in coverage is fine, code coverage is only a metric - not a goal

You should rename the method though, there are two tests methods with the same name 

```suggestion
@pytest.mark.skipif(pkgmanager.TYPE != "yum", reason="No yum module detected on the system, skipping it.")
@pytest.mark.parametrize(("packages"), ((["package-1", "package-2", "package-3"],)))
def test_get_packages_to_update_yum(packages, monkeypatch):
    from cli import YumBaseCli  # pylint: disable=E0401

    class PkgUpdates:
        def __init__(self, packages):
            self.updates = packages

    pkg_lists_mock = mock.Mock(return_value=PkgUpdates(packages))

    monkeypatch.setattr(YumBaseCli, "returnPkgLists", value=pkg_lists_mock)

    assert _get_packages_to_update_yum() == packages


@pytest.mark.skipif(pkgmanager.TYPE != "dnf", reason="No dnf module detected on the system, skipping it.")
@pytest.mark.parametrize(("packages"), ((["package-1", "package-2", "package-3"],)))
def test_get_packages_to_update_dnf(packages, monkeypatch):
    from dnf import Base  # pylint: disable=E0401

    dummy_mock = mock.Mock()

    monkeypatch.setattr(Base, "read_all_repos", value=dummy_mock)
    monkeypatch.setattr(Base, "fill_sack", value=dummy_mock)
    monkeypatch.setattr(Base, "upgrade_all", value=dummy_mock)
    monkeypatch.setattr(Base, "resolve", value=dummy_mock)
    monkeypatch.setattr(Base, "transaction", value=packages)

    assert _get_packages_to_update_dnf() == packages
```

SpyTec: Decrease in coverage is fine, code coverage is only a metric - not a goal

You should rename the method though, there are two tests methods with the same name 

```suggestion
@pytest.mark.skipif(pkgmanager.TYPE != "yum", reason="No yum module detected on the system, skipping it.")
@pytest.mark.parametrize(("packages"), ((["package-1", "package-2", "package-3"],)))
def test_get_packages_to_update_yum(packages, monkeypatch):
    from cli import YumBaseCli  # pylint: disable=E0401

    class PkgUpdates:
        def __init__(self, packages):
            self.updates = packages

    pkg_lists_mock = mock.Mock(return_value=PkgUpdates(packages))

    monkeypatch.setattr(YumBaseCli, "returnPkgLists", value=pkg_lists_mock)

    assert _get_packages_to_update_yum() == packages


@pytest.mark.skipif(pkgmanager.TYPE != "dnf", reason="No dnf module detected on the system, skipping it.")
@pytest.mark.parametrize(("packages"), ((["package-1", "package-2", "package-3"],)))
def test_get_packages_to_update_dnf(packages, monkeypatch):
    from dnf import Base  # pylint: disable=E0401

    dummy_mock = mock.Mock()

    monkeypatch.setattr(Base, "read_all_repos", value=dummy_mock)
    monkeypatch.setattr(Base, "fill_sack", value=dummy_mock)
    monkeypatch.setattr(Base, "upgrade_all", value=dummy_mock)
    monkeypatch.setattr(Base, "resolve", value=dummy_mock)
    monkeypatch.setattr(Base, "transaction", value=packages)

    assert _get_packages_to_update_dnf() == packages
```

convert2rhel/checks.py
pirat89: Hi guys, sorry for late response. I am not looking here so much during my holiday. I've read the discussion and it seems you should not have any problems here regarding the hard requirement to have enabled the base repository on the system. The unsupported envar seems as a good decision to bypass that requirement, keeping the possibility to handle problems by customers on their own.

@r0x0d  I spoke about cases we discovered in case of in-place upgrades between major versions of RHEL systems - probably I should mention that, sorry for the confusion. I am trying to inform you about problems we already had to deal with when I see you are dealing with similar problems :)

r0x0d (commented):
Review

r0x0d (commented):
Reviewed and fixed a few diffs.

r0x0d:

If this PR get merged before the #389, then, @abadger and @Andrew-ang9 you guys have to change the run_subprocess() calls I've added (Or I could push a commit on that branch too)

Otherwise, if this gets merged after the #389, then I have to rebase and modify the run_subprocess() call.

convert2rhel/checks.py
```diff
@@ -448,3 +457,81 @@ def _bad_kernel_substring(kernel_release):
...
+    """Check if the loaded kernel is behind or of the same version as in yum repos."""
+    logger.task("Prepare: Checking if the loaded kernel version is the most recent.")
+
+    # For OL/CentOS 8 the `kernel` is just a meta package, instead, we check for `kernel-core`.
+    # But for 6 and 7 releases, the correct way to check is using `kernel`.
+    package_to_check = "kernel-core" if system_info.version.major >= 8 else "kernel"
+
+    # The latest kernel version on repo
+    repoquery_output, _ = run_subprocess(
+        ["repoquery", "--quiet", "--qf", '"%{BUILDTIME}\\t%{VERSION}-%{RELEASE}"', package_to_check]
```
j-mracek, line 489: Why do you use CLI instead of API. What is the targeted systems and in which versions? (RHEL8, 7, Fedora ...)

j-mracek: Parsing outputs from DNF is sensitive to be contaminated by plugins.

convert2rhel/checks.py
```diff
@@ -448,3 +457,81 @@ def _bad_kernel_substring(kernel_release):
...
+        logger.info("System is up-to-date.")
+
+
+def is_loaded_kernel_latest():
+    """Check if the loaded kernel is behind or of the same version as in yum repos."""
+    logger.task("Prepare: Checking if the loaded kernel version is the most recent.")
+
+    # For OL/CentOS 8 the `kernel` is just a meta package, instead, we check for `kernel-core`.
+    # But for 6 and 7 releases, the correct way to check is using `kernel`.
+    package_to_check = "kernel-core" if system_info.version.major >= 8 else "kernel"
```
j-mracek, line 485: It means that the tool will be working in standard RHEL environment with default provided kernels.

convert2rhel/checks.py
j-mracek: _get_packages_to_update_dnf() uses DNF API - therefore repoquery can be replaced by  dnf query-

```
query = base.sack.query().filterm(name=package_to_check)
# to detect running kernel
running_kernel = base.sack.get_running_kernel()

# Then you can try to compare version of packages by direct comparison of  package object. Be aware that running kernel can be not the latest one installed

```

j-mracek: I think that using dnf API will allow to delete a lot of your code handling parsing output and comparison of versions.

r0x0d: IIRC, by the time I developed this code I was not aware if `yum` had an similar API to interact with repoquery as dnf has, that's why I was using the `repoquery` for that stuff, as it was available in every distro we handle (RHEL8 and RHEL7).

I'm currently looking over the `yum` repository again to see if I can find any APIs that could be used here to have the same effect.

kokesak:
Hi @r0x0d, I was just briefly looking at the test runs. There are many fails as of now, I suppose we will need to update tests so they correspond to this change.

I saw this error as one of the most one:
http://pastebin.test.redhat.com/1035993
```
CRITICAL - Could not find any kernel from repositories to compare against the loaded kernel.
```
Do we need to have some special repository on the machine or what this check exactly does?

convert2rhel/checks.py
```diff
@@ -448,3 +457,79 @@ def _bad_kernel_substring(kernel_release):
...
+    else:
+        logger.info("System is up-to-date.")
+
+
+def is_loaded_kernel_latest():
+    """Check if the loaded kernel is behind or of the same version as in yum repos."""
+    logger.task("Prepare: Checking if the loaded kernel version is the most recent.")
+
+    # For OL/CentOS 8 the `kernel` is just a meta package, instead, we check for `kernel-core`.
+    # But for 6 and 7 releases, the correct way to check is using `kernel`.
```
SpyTec, line 484:
```suggestion
    # For Oracle/CentOS Linux 8 the `kernel` is just a meta package, instead, we check for `kernel-core`.
    # But for 6 and 7 releases, the correct way to check is using `kernel`.
```

convert2rhel/pkghandler.py
```diff
@@ -933,3 +934,94 @@ def has_duplicate_repos_across_disablerepo_enablerepo_options():
...
+    evr2 = utils.string_to_version(version2)
+
+    return rpm.labelCompare(evr1, evr2)
+
+
+def get_total_packages_to_update():
+    """Return the total number of packages to update in the system
+
+    It uses both yum/dnf depending on weather they are installed on the system,
+    if it's OL/CentOS 7 or 6, it will use `yum`, otherwise it should use `dnf`.
```
SpyTec, line 976:
```suggestion
    if it's Oracle/CentOS Linux 7 or 6, it will use `yum`, otherwise it should use `dnf`.
```

SpyTec (commented):
Looked over again and code looks good

kokesak:
@r0x0d I found out the main reason it's failing. There is an issue when the system has two repo files that contains the same information. Eg. run `cp /etc/yum.repos.d/any-repo.repo /etc/yum.repos.d/any-repo-duplicate.repo`.
Here is the failing output from c2r:
https://privatebin.corp.redhat.com/?a3bbb2919317ace9#7EeFqdmJSAyYtYKPCcB26T7t5D7Ts3tGg2GcikXToMfC

Also I have a question regarding this message:
```
WARNING - The system has 1 packages not updated. If you wish to update then before continuing with the conversion, execute the following step:
            out: 
            out: Run: yum update -y
```
Sometimes we get this message, Is there any way that c2r could report the list of the non-updated packages?

r0x0d:
> @r0x0d I found out the main reason it's failing. There is an issue when the system has two repo files that contains the same information. Eg. run `cp /etc/yum.repos.d/any-repo.repo /etc/yum.repos.d/any-repo-duplicate.repo` then run the conversion. Here is the failing output from c2r: https://privatebin.corp.redhat.com/?a3bbb2919317ace9#7EeFqdmJSAyYtYKPCcB26T7t5D7Ts3tGg2GcikXToMfC
> 
> Also I have a question regarding this message:
> 
> ```
> WARNING - The system has 1 packages not updated. If you wish to update then before continuing with the conversion, execute the following step:
>             out: 
>             out: Run: yum update -y
> ```
> 
> Sometimes we get this message, Is there any way that c2r could report the list of the non-updated packages?

I will take a look on this error message, thanks for the debugging. 

And regarding showing the list of packages that needs an update, yes, of course. Will include that in a commit.

r0x0d:
@kokesak already added the ability to see what packages need to update in the log message. Now I'm taking a closer look at the repo thing.

r0x0d:
@kokesak I've fixed the scenario where it could have a duplicate repository defined in any repofiles under `/etc/yum.repos.d`. 

I think this is a corner case that should not happen very often, but nonetheless, I updated my parser for repoquery and now I'm throwing away that specific line!

r0x0d:
> @r0x0d Thanks for the fix. I did a little investigating on still failing OL8 issue. The error is this:
> 
> ```
>  out: [03/15/2022 18:15:13] TASK - [Prepare: Checking if the packages are up-to-date] *****************
> 18:15:13             out: Repository copr:copr.fedorainfracloud.org:group_oamg:convert2rhel is listed more than once in the configuration
> 18:15:13             out: Repository copr:copr.fedorainfracloud.org:group_oamg:convert2rhel is listed more than once in the configuration
> 18:15:14             out: Errors during downloading metadata for repository 'ol8_baseos_latest':
> 18:15:14             out:   - Curl error (6): Couldn't resolve host name for https://yum$ociregion.$ocidomain/repo/OracleLinux/OL8/baseos/latest/x86_64/repodata/repomd.xml [Could not resolve host: yum$ociregion.$ocidomain]
> 18:15:14             out: [03/15/2022 18:15:14] DEBUG - Traceback (most recent call last):
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/dnf/repo.py", line 574, in load
> 18:15:14             out:     ret = self._repo.load()
> 18:15:14             out:   File "/usr/lib64/python3.6/site-packages/libdnf/repo.py", line 397, in load
> 18:15:14             out:     return _repo.Repo_load(self)
> 18:15:14             out: libdnf._error.Error: Failed to download metadata for repo 'ol8_baseos_latest': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
> 18:15:14             out: 
> 18:15:14             out: During handling of the above exception, another exception occurred:
> 18:15:14             out: 
> 18:15:14             out: Traceback (most recent call last):
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 92, in main
> 18:15:14             out:     checks.perform_pre_checks()
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/checks.py", line 60, in perform_pre_checks
> 18:15:14             out:     check_package_updates()
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/checks.py", line 466, in check_package_updates
> 18:15:14             out:     packages_to_update = get_total_packages_to_update()
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/pkghandler.py", line 986, in get_total_packages_to_update
> 18:15:14             out:     packages = _get_packages_to_update_dnf()
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/pkghandler.py", line 1019, in _get_packages_to_update_dnf
> 18:15:14             out:     base.fill_sack()
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/dnf/base.py", line 399, in fill_sack
> 18:15:14             out:     self._add_repo_to_sack(r)
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/dnf/base.py", line 139, in _add_repo_to_sack
> 18:15:14             out:     repo.load()
> 18:15:14             out:   File "/usr/lib/python3.6/site-packages/dnf/repo.py", line 581, in load
> 18:15:14             out:     raise dnf.exceptions.RepoError(str(e))
> 18:15:14             out: dnf.exceptions.RepoError: Failed to download metadata for repo 'ol8_baseos_latest': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
> 18:15:14             out: 
> ```
> 
> I think there is some problem when expending variables `$ociregion.$ocidomain`. I'm not sure if the same variables are being used on OL7 images. I will try to check it, but the conversions are ok on OL7.
> 
> I tried to look here https://docs.oracle.com/en-us/iaas/oracle-linux/getting-started/install-software-on-instance.htm The variable in /etc/yum/vars/ocidomain on the tested system conatins "oracle.com", the ociregion is empty. I wasn't able to find out more

@kokesak, I have made a fix for this issue! Thanks for reporting this out. I did see this yesterday when testing out but hadn't the proper time to investigate. 

Now it should work fine for OracleLinux 8.

convert2rhel/pkghandler.py
```diff
@@ -933,3 +934,102 @@ def has_duplicate_repos_across_disablerepo_enablerepo_options():
...
+    system."""
+    from dnf import Base  # pylint: disable=E0401
+
+    packages = []
+    base = Base()
+    # Fix the case when we are trying to query packages in Oracle Linux 8
+    # when the DNF API gets called, those variables are not populated by default.
+    if system_info.id == "oracle":
+        base.conf.substitutions["ocidomain"] = "oracle.com"
+        base.conf.substitutions["ociregion"] = ""
```
r0x0d, line 1022: @SpyTec, this fix is for the issue that @kokesak raised (https://github.com/oamg/convert2rhel/pull/356#issuecomment-1069473404) 

Do you think it should be better for us to read the contents of /etc/yum/vars/{ocidomain,ociregion} instead of applying the defaults ourselves? I mean, the code could be running inside a machine that the user has changed the contents of the files, and then, our code would stop working. 

This only happens on OL8 as DNF does not expand those custom variables. (yum works fine, tho)

tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
```diff
@@ -0,0 +1,89 @@
...
+
+    with convert2rhel(
+        ("--no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug").format(
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        # TODO do we want to test the scenario where we have a package locked by this?
```
kokesak, line 26: We could also run full conversion when the versionlock is applied. I would run together with some non updated package.

tests/integration/tier1/set-locale/use_non_english_language.py
```diff
@@ -15,3 +15,9 @@ def test_use_non_english_language(shell):
     # set LANG to Chinese
     assert shell("localectl list-locales | grep zh_CN.utf8").returncode == 0
     assert shell("localectl set-locale LANG=zh_CN.utf8").returncode == 0
+
+    # When running tests on TF on CentOS7 we are getting an error
+    # about setting incompatible LC_CTYPE C.UTF-8).
+    # However the C.UTF-8 was added on RHEL-8 like distros.
+    if "centos-7" in os:
+        assert shell("localectl set-locale LC_CTYPE=zh_CN.utf8").returncode == 0
```
danmyway, line 23:
```suggestion
    # Testing farm is returning an error on CentOS7 mentioning 
    # setting incompatible LC_CTYPE C.UTF-8.
    # However the C.UTF-8 was added on RHEL-8 like distros.
    if "centos-7" in os:
        assert os.system("localectl set-locale LC_CTYPE=zh_CN.utf8") == 0
```

tests/integration/tier1/system-up-to-date/install_non_latest_kernel.py
```diff
@@ -0,0 +1,24 @@
...
+        shell("grub2-set-default 'CentOS Linux (3.10.0-1160.el7.x86_64) 7 (Core)'")
+    elif "oracle-7" in system_version:
+        assert shell("yum install kernel-3.10.0-1160.el7.x86_64 -y").returncode == 0
+        shell("grub2-set-default 'Oracle Linux Server 7.9, with Linux 3.10.0-1160.el7.x86_64'")
+    elif "centos-8" in system_version:
+        assert shell("yum install kernel-4.18.0-240.22.1.el8_3 -y").returncode == 0
+        shell("grub2-set-default 'CentOS Linux (4.18.0-240.22.1.el8_3.x86_64) 8'")
+    elif "oracle-8" in system_version:
+        assert shell("yum install kernel-4.18.0-240.22.1.el8_3 -y").returncode == 0
+        shell("grub2-set-default 'Oracle Linux Server (4.18.0-240.22.1.el8_3.x86_64) 8.3'")
```
danmyway, line 24:
```suggestion
    # Set default kernel
    if "centos-7" in system_version:
        assert os.system("yum install kernel-3.10.0-1160.el7.x86_64 -y") == 0
        os.system("grub2-set-default 'CentOS Linux (3.10.0-1160.el7.x86_64) 7 (Core)'")
    elif "oracle-7" in system_version:
        assert os.system("yum install kernel-3.10.0-1160.el7.x86_64 -y") == 0
        os.system("grub2-set-default 'Oracle Linux Server 7.9, with Linux 3.10.0-1160.el7.x86_64'")
    elif "centos-8" in system_version:
        assert os.system("yum install kernel-4.18.0-240.22.1.el8_3 -y") == 0
        os.system("grub2-set-default 'CentOS Linux (4.18.0-240.22.1.el8_3.x86_64) 8'")
    elif "oracle-8" in system_version:
        assert os.system("yum install kernel-4.18.0-240.22.1.el8_3 -y") == 0
        os.system("grub2-set-default 'Oracle Linux Server (4.18.0-240.22.1.el8_3.x86_64) 8.3'")
```

tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
```diff
@@ -0,0 +1,47 @@
...
+
+from envparse import env
+
+
+def set_latest_kernel(shell):
+    # We need to dynamically get the the name of the latest kernel
+    # that is in the repositories
+
+    # repoquery command requires 'yum-utils'
+    shell("yum install yum-utils -y")
```
danmyway, line 11:
```suggestion
    # Install 'yum-utils' required by the repoquery command
    os.system("yum install yum-utils -y")
```

tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Get the name of the kernel wich is the lates
+    latest_kernel = shell(
+        "repoquery --quiet --qf '%{BUILDTIME}\t%{VERSION}-%{RELEASE}' kernel 2>/dev/null | tail -n 1 | awk '{printf $NF}'"
+    ).output
+    shell("echo {}".format(latest_kernel))
```
danmyway, line 17:
```suggestion
    os.system("echo {}".format(latest_kernel))
```

tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Get full name of the kerenl
+    full_name = shell(
+        "grubby --info ALL | grep \"title=.*{}\" | tr -d '\"' | sed 's/title=//'".format(latest_kernel)
+    ).output
+
+    # Set the latest kernel as one we want to reboot to
+    shell("grub2-set-default '{}'".format(full_name.strip()))
```
danmyway, line 25:
```suggestion
    # Set the latest kernel as the one we want to reboot to
    os.system("grub2-set-default '{}'".format(full_name.strip()))
```

tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
```diff
@@ -0,0 +1,47 @@
+import platform
```
danmyway, line 1:
```suggestion
import platform
import os
```

danmyway: In case the `os.system` will be used instead of calling the `shell()` function

tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
```diff
@@ -0,0 +1,103 @@
...
+    # On CentOS-8 we cannot do downgrade as the repos contains only the latest package version.
+    # To workaround this we need to install package from older repo.
+    if "centos-8" in system_version:
+        assert shell("yum install -y {}".format(centos_8_pkg_url)).returncode == 0
+    else:
+        assert shell("yum install openldap wpa_supplicant -y").returncode == 0
+        assert shell("yum downgrade openldap wpa_supplicant -y").returncode == 0
+
+    assert shell("yum install -y yum-plugin-versionlock").returncode == 0
+    assert shell("yum versionlock wpa_supplicant").returncode == 0
```
danmyway, line 27:
```suggestion
    # Try to downgrade two packages
    # On CentOS-8 we cannot do the downgrade as the repos contain only the latest package version.
    # We need to install package from older repository as a workaround.
    if "centos-8" in system_version:
        assert os.system("yum install -y {}".format(centos_8_pkg_url)) == 0
    else:
        assert os.system("yum install openldap wpa_supplicant -y") == 0
        assert os.system("yum downgrade openldap wpa_supplicant -y") == 0

    assert os.system("yum install -y yum-plugin-versionlock") == 0
    assert os.system("yum versionlock wpa_supplicant") == 0
```

tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
```diff
@@ -0,0 +1,103 @@
...
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        # TODO do we want to test the scenario where we have a package locked by this?
+        c2r.expect("WARNING - YUM/DNF versionlock plugin is in use. It may cause the conversion to fail.")
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("n")
+
+    # Clean up
+    assert shell("yum versionlock clear").returncode == 0
```
danmyway, line 43:
```suggestion
    assert os.system("yum versionlock clear") == 0
```

tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_skip_kernel_check(shell, convert2rhel):
+    """
+    Test that it is possible to use env variable in some case to override
+    kernel check inhibitor. One of the way to allow this is to not have
+    any kernel packages present in repos.
+    """
+    shell("mkdir /tmp/my_tmp; mv /etc/yum.repos.d/* /tmp/my_tmp")
```
danmyway, line 52:
```suggestion
    os.systeml("mkdir /tmp/my_tmp; mv /etc/yum.repos.d/* /tmp/my_tmp")
```

tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
```diff
@@ -0,0 +1,103 @@
...
+        c2r.sendline("n")
+
+    # Clean up
+    assert shell("yum versionlock clear").returncode == 0
+
+
+def test_skip_kernel_check(shell, convert2rhel):
+    """
+    Test that it is possible to use env variable in some case to override
+    kernel check inhibitor. One of the way to allow this is to not have
```
danmyway, line 49:
```suggestion
    the kernel check inhibitor. One of the way to allow this is to not have
```

tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
```diff
@@ -0,0 +1,103 @@
...
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        c2r.expect("Detected 'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' environment variable")
+        c2r.sendcontrol("c")
+    assert c2r.exitstatus != 0
+
+    # Clean up
+    shell("mv /tmp/my_tmp/* /etc/yum.repos.d/")
```
danmyway, line 83:
```suggestion
    os.system("mv /tmp/my_tmp/* /etc/yum.repos.d/")
```

danmyway (changes requested):
Mainly minor typo and stylistic corrections.
I would try to use `os.system()` instead of calling `shell()` function, whenever the output is not needed.

plans/tier1.fmf
```diff
@@ -204,3 +204,26 @@ discover+:
...
+    script: pytest -svv tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
+  - name: reboot machine
+    how: ansible
+    playbook: tests/ansible_collections/roles/reboot/main.yml
+  - name: test conversion non updated package
+    how: shell
+    script: pytest -svv tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
+  - name: reboot after conversion
+    how: ansible
+    playbook: tests/ansible_collections/roles/reboot/main.yml
```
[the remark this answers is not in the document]
kokesak: Sadly no. We need to reboot the machine between some steps and the reboot may be tricky to implement correctly in the test itself.

tests/integration/tier0/backup-release/test_backup_release.py
```diff
@@ -24,6 +24,9 @@ def test_backup_os_release_no_envar(shell, convert2rhel):
     # Move all repos to other location so it is not used
```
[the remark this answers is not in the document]
kokesak: Thanks for catching the typoos in comments!

plans/tier1.fmf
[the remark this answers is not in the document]
danmyway: Oh, alright then!

tests/integration/tier0/backup-release/test_backup_release.py
```diff
@@ -24,6 +24,9 @@ def test_backup_os_release_no_envar(shell, convert2rhel):
...
     assert shell("mkdir /tmp/s_backup && mv /etc/yum.repos.d/* /tmp/s_backup/").returncode == 0
```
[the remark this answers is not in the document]
abadger: kokesak, danmyway, and I talked about this today and we decided it was okay to stick with shell.  subprocess is the encouraged way to call out to other programs ( First paragraph of https://docs.python.org/3/library/subprocess.html ).  The other functions in the python stdlib are considered legacy ways of calling out.

The one thing about the `shell()` function to be aware of is that (as the name implies) it sends the command to the shell.  This means that shell metacharacters are processed by the shell (`*`, `;`, `&&`, etc).  This can be a source of security bugs whenever user input is used but I don't believe that's a problem for our integration tests in general.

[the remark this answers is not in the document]
kokesak: @abadger thanks, we will stick with the `shell()` function.

danmyway (approved):
LGTM! Thank you!

convert2rhel/pkghandler.py
```diff
@@ -968,3 +969,109 @@ def get_pkg_names_from_rpm_paths(rpm_paths):
...
+
+    # Same thing for CentOS, but this time, the URL for the vault has changed
+    # and instead of just ussing $releasever (8.5, 8.4...), we need either
+    # using the correct $releasever (8.5.2111) or setting up the $contentdir
+    if system_info.id == "centos":
```
SpyTec, line 1062:
```suggestion
    elif system_info.id == "centos":
```

SpyTec (commented):
@r0x0d I skimmed and saw two comments about some pieces of code not being tested, which has to do with branches. The `sys.path.insert(0, "/usr/share/yum-cli")` was never asserted to be called, and `base.conf.substitutions` was never asserted either

r0x0d:
Hi, @abadger and @bocekm could one of you do a final review & approval in this PR? Thanks!

abadger (approved):

convert2rhel/pkghandler.py
```diff
@@ -968,3 +969,109 @@ def get_pkg_names_from_rpm_paths(rpm_paths):
...
+    """Query all the packages with dnf that has an update pending on the
+    system."""
+    from dnf import Base  # pylint: disable=E0401
+
+    packages = []
+    base = Base()
+    # Fix the case when we are trying to query packages in Oracle Linux 8
+    # when the DNF API gets called, those variables are not populated by default.
+    if system_info.id == "oracle":
+        base.conf.substitutions["ocidomain"] = "oracle.com"
```
bocekm, line 1056: Instead of hardcoding the value, it'd better to load the variables from the `/etc/dnf/vars/`. Oracle introduced this $ocidomain variable in
https://yum.oracle.com/repo/OracleLinux/OL8/baseos/latest/x86_64/getPackage/[oraclelinux-release-el8-1.0-16.0.1.el8.x86_64.rpm](https://yum.oracle.com/repo/OracleLinux/OL8/baseos/latest/x86_64/getPackage/oraclelinux-release-el8-1.0-16.0.1.el8.x86_64.rpm). 

You can change it in a subsequent PR.

Related comment: https://github.com/oamg/convert2rhel/pull/356#issuecomment-1069381290

r0x0d: Bugzilla from someone that had "the same problem": https://bugzilla.redhat.com/show_bug.cgi?id=1920735

danmyway (approved):

convert2rhel/pkghandler.py
r0x0d: Fixed in https://github.com/oamg/convert2rhel/pull/458/files#r864814726

PR #357, opened 2021-10-21
/etc/os-release file backup

r0x0d:
This PR aims to add the ability to make an backup of the file `/etc/os-release` during the conversion process

Ticket reference: https://issues.redhat.com/browse/OAMG-5457

### DIFF
--- a/convert2rhel/redhatrelease.py
+++ b/convert2rhel/redhatrelease.py
@@ -15,14 +15,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-import glob
 import logging
 import os
 import re
 
 from convert2rhel import pkgmanager, utils
 from convert2rhel.systeminfo import system_info
-from convert2rhel.toolopts import tool_opts
 
 
 loggerinst = logging.getLogger(__name__)
@@ -56,6 +54,14 @@ def get_system_release_content():
         return utils.get_file_content(filepath)
     except EnvironmentError as err:
         loggerinst.critical("%s\n%s file is essential for running this tool." % (err, filepath))
+
+
+def get_os_release_filepath():
+    """Return path of the OS Release file."""
+    release_filepath = "/etc/os-release"  # RHEL 6/7/8 based OSes
+    if os.path.isfile(release_filepath):
+        return release_filepath
+    loggerinst.critical("Error: Unable to find the %s file containing the information needed." % release_filepath)
 
 
 class YumConf(object):
@@ -108,3 +114,4 @@ class YumConf(object):
 
 # Code to be executed upon module import
 system_release_file = utils.RestorableFile(get_system_release_filepath())  # pylint: disable=C0103
+os_release_file = utils.RestorableFile(get_os_release_filepath())  # pylint: disable=C0103
--- a/convert2rhel/unit_tests/redhatrelease_test.py
+++ b/convert2rhel/unit_tests/redhatrelease_test.py
@@ -16,6 +16,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import glob
+import os
 import sys
 import unittest
 
@@ -31,13 +32,23 @@ from collections import namedtuple
 
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
 from convert2rhel import pkgmanager, redhatrelease, utils
-from convert2rhel.redhatrelease import YumConf
+from convert2rhel.redhatrelease import YumConf, get_os_release_filepath, get_system_release_filepath
 from convert2rhel.systeminfo import system_info
-from convert2rhel.toolopts import tool_opts
+
+
+YUM_CONF_WITHOUT_DISTROVERPKG = """[main]
+installonly_limit=3
+
+#  This is the default"""
+
+YUM_CONF_WITH_DISTROVERPKG = """[main]
+installonly_limit=3
+distroverpkg=centos-release
+
+#  This is the default"""
 
 
 class TestRedHatRelease(unittest.TestCase):
-
     supported_rhel_versions = [6, 7, 8]
 
     class DumbMocked(unit_tests.MockFunction):
@@ -142,13 +153,33 @@ def test_yum_patch(monkeypatch, modified, caplog):
         assert "Skipping patching, yum configuration file not modified" in caplog.text
 
 
-YUM_CONF_WITHOUT_DISTROVERPKG = """[main]
-installonly_limit=3
+@pytest.mark.parametrize(("is_file", "exception"), ((True, False), (False, True)))
+def test_get_system_release_filepath(is_file, exception, monkeypatch, caplog):
+    is_file_mock = mock.MagicMock(return_value=is_file)
+    monkeypatch.setattr(os.path, "isfile", value=is_file_mock)
 
-#  This is the default"""
+    if exception:
+        with pytest.raises(SystemExit):
+            get_system_release_filepath()
+        assert (
+            "Error: Unable to find the /etc/system-release file containing the OS name and version"
+            in caplog.records[-1].message
+        )
+    else:
+        assert get_system_release_filepath() == "/etc/system-release"
 
-YUM_CONF_WITH_DISTROVERPKG = """[main]
-installonly_limit=3
-distroverpkg=centos-release
 
-#  This is the default"""
+@pytest.mark.parametrize(("is_file", "exception"), ((True, False), (False, True)))
+def test_get_os_release_filepath(is_file, exception, monkeypatch, caplog):
+    is_file_mock = mock.MagicMock(return_value=is_file)
+    monkeypatch.setattr(os.path, "isfile", value=is_file_mock)
+
+    if exception:
+        with pytest.raises(SystemExit):
+            get_os_release_filepath()
+        assert (
+            "Error: Unable to find the /etc/os-release file containing the information needed."
+            in caplog.records[-1].message
+        )
+    else:
+        assert get_os_release_filepath() == "/etc/os-release"
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -81,6 +81,7 @@ def main():
         # backup system release file before starting conversion process
         loggerinst.task("Prepare: Backup System")
         redhatrelease.system_release_file.backup()
+        redhatrelease.os_release_file.backup()
         repo.backup_yum_repos()
 
         loggerinst.task("Prepare: Clear YUM/DNF version locks")


convert2rhel/redhatrelease.py
```diff
@@ -15,14 +15,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-import glob
```
r0x0d, line 18: Unused import line

convert2rhel/redhatrelease.py
```diff
@@ -58,6 +56,26 @@ def get_system_release_content():
...
+    """Return content of the file containing useful information
+    about the os release.
+    """
+
+    filepath = get_os_release_filepath()
+    try:
+        return utils.get_file_content(filepath)
+    except EnvironmentError as err:
+        loggerinst.critical("%s\n%s file is essential for running this tool." % (err, filepath))
+
```
abadger, line 77: A few things I noticed about this block of code:

* `get_os_release_content()` doesn't appear to be used by anything?  (I thought `RestorableFile` might use it in some manner but I'm not seeing anything.)  Do we want to implement it if it's not being used?
  * Is os-release's info being used or are we just implementing these functions so that we can make it a `RestorableFile`?  If the latter, maybe we should use  `loggerinst.info()` (to match what RestorableFile.backup() uses) instead of loggerinst.critical()?
* The code to handle os-release seems to be largely the same as for system-release with just the hardcoded path changed.  It feels like we should parametrize that so we don't have to create new functions (maybe using functools.partial to keep the current API?)

A bit separate from your PR:
* it feels like exceptions should be raised by these functions (and RestorableFile).  The functions don't know whether the program should continue or exit if the file does not exist (or whether to log a missing file as critical or merely an unexpected but probably unimportant bit of info) so raising exceptions allows the caller to make that decision.  Maybe that's a change we should make in the future.
* We're checking that these filepaths exist quite a few times (eveytime get_os_release_filepath() is called explicitly; whenever get_os_release_content() is called, get_os_release_filepath() is called implicitly; and whenever RestorableFile.backup() is called the filepath is checked there as well.  At some point we probably want to change this code so that the file path is checked once near startup and then rely on raising an exception if the file doesn't exist later on.  (no matter what, there is always a race between when the file is checked and we actually open the file for use.  So checking near startup allows us to exit early in common cases.  Raising and catching exceptions allow us to handle the error when the file is used.)

convert2rhel/unit_tests/redhatrelease_test.py
```diff
@@ -142,13 +158,30 @@ def test_yum_patch(monkeypatch, modified, caplog):
...
 
 
-YUM_CONF_WITHOUT_DISTROVERPKG = """[main]
-installonly_limit=3
+@unit_tests.mock(os.path, "isfile", lambda x: True)
+def test_get_os_release_filepath(caplog):
+    assert get_os_release_filepath() == "/etc/os-release"
 
-#  This is the default"""
+    os.path.isfile = lambda x: False
```
abadger, line 165: I would make a separate test for this case.  My rule of thumb (but it's only from my experience, let me know if you have a better guideline) is that you can assert multiple times in a single unittest but you should only call the function being tested a single time.  (There are exceptions to this, like if you're testing idempotence where you want to make sure that the state of the system does not change after the function is called a second time.  But I don't think that applies here.)

convert2rhel/unit_tests/redhatrelease_test.py
```diff
@@ -142,13 +158,30 @@ def test_yum_patch(monkeypatch, modified, caplog):
...
+    monkeypatch.setattr(redhatrelease, "get_os_release_filepath", get_os_release_filepath_mock)
+
+    if raise_error:
+        get_file_content_mock = mock.Mock(side_effect=EnvironmentError)
+        monkeypatch.setattr(redhatrelease.utils, "get_file_content", get_file_content_mock)
+        with pytest.raises(SystemExit):
+            assert get_os_release_content()
+    else:
+        monkeypatch.setattr(redhatrelease.utils, "get_file_content", mock.Mock(return_value=file_content))
+        assert "NAME=Test" in get_os_release_content()
```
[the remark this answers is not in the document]
abadger: * I would definitely make the happy path a separate test from the error path.

* I'm not sure how other people feel about unit tests vs functional tests vs integration tests ... I tend to favour functional tests over unit tests as they strike a balance between testing what will occur when the code is called for real and speed of execution plus ease of debugging.  If this was recast as a functional test, I would probably use pytest's facilities to create a temporary file with my desired content and then use monkeypatch on get_os_release_filepath to return the temporary filename.  That way:
  * Except for determining the filename, I would be running the actual code that would be in production with all of the potential places where one piece calling another could go wrong.
  * I could more easily construct my tests as setting up a system environment and then testing that the code runs on it.  ie: If I wanted to test that a directory instead of a file caused the expected exception, I could replace the temporary file with a directory.  No need to mock out anything further.
  * Since I'd still be monkeypatching get_os_release_filepath() in that example, I might also have one or two additional tests that showed that get_os_release_content was doing what I expected when get_os_release_filepath
was not mocked out.  That would probably be a few error checks and one happy path test where I'd have to monkeypatch utils.get_file_content instead.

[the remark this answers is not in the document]
SpyTec: Having separate happy tests could be good, though not a requirement since we don't do it in the rest of the codebase - so won't be a blocker

Unit tests in general is a bad term, testing as a unit doesn't mean mock everything and just hope for the best. Functional tests fit better due to the reasons you mentioned - but I'll still call these unit tests

100% anything that don't need to be mocked shouldn't. Creating temp files and mocking a path is much better than mocking a dependency. I did have some issues previously with creating a test file for some reason I don't remember right now

convert2rhel/redhatrelease.py
SpyTec: @abadger we have modified logger so critical level exits the program through a `sys.exit`, so there is an exception being thrown in `get_os_release_filepath`

abadger: tldr;
* Exceptions say that the function doesn't know what to do so it is asking its callers to make the decision
* Exceptions need to communicate enough information about errors so that the caller can make that decision
* SystemExit is special because it is about control flow, not errors.

True, but catching SystemExit is probably not a generally good thing.  (I see that we do it in `main()` and that doesn't seem too bad to me but let me try to explain what the difference is).

At a high level, exceptions are an alternate method for the function to communicate with its caller from the normal return value.  They should be raised when a function is saying, "Hey, I encountered this error condition and I don't know how to proceed!".  Then each caller is able to catch the exception if they know how to proceed in case of the error.  If they don't know how to continue, they let the exception propagate further up the call stack.  If nothing handles the exception, then the user is notified and program execution stops.

Most of the time, a caller needs some bits of information in order to decide whether (and how) to handle the error case.  For instance, MemoryError (raised when the system can't allocate any more RAM to the program) might be a case where the caller cannot do anything because it's likely they won't be able to allocate any new data structures to aid in recovery.  OTOH, an IOError due to file system permissions might lead the caller to print a message to the user to change permissions and exit or to ignore the error because the file operation was an optional feature and doesn't interfere with the normal program flow.

In order to make a decision about whether an exception can be handled or not, the caller needs to know certain information about the error.  Some things the caller knows from its own context (for instance, that it was trying to backup an optional file).  Other things, the caller needs to receive from the function (for instance, that the error was because the file did not exist vs a permissions error vs memoryerror vs user aborting the process [etc]).  With all of that information, the caller can decide whether to continue or not.

Raising SystemExit throughout the code causes problems for this.  SystemExit is implemented as an exception but it's really a control flow structure (which is why the standard way to raise SystemExit is to call `sys.exit()` rather than `raise SystemExit()`).  It's message to the caller is "I've unilaterally decided that we need to exit so please unwind the stack and end now."  It's not really meant to be caught by callers.  For that reason, it doesn't give much information to the caller that would be useful to decide whether the caller should catch and process it or not. 

Tangent: We catch SystemExit in main().  Is that a bad thing?  I would say sort of but it's allowable.  We catch SystemExit as a way to inject an exit handler.  We want to make sure certain things happen when the program exits prematurely no matter what the reason for the exit is.  So we try to catch any and all reasons that the system could exit prematurely and run those handlers based on context that we know about in main() (at what point in the convert2rhel run we were at).  Using atexit() to do this might fit into the language's design better ( https://docs.python.org/3/library/atexit.html ) but I think this limited use of SystemExit isn't horrible.  Python makes it the most obvious way to create an exit handler so it doesn't feel too abusive to do it the way we are.

Back to the main topic... How does this apply to get_os_release_filepath(), RestorableFile, etc?  These functions and objects are checking whether the filepath that they are given exists and if so, they (indirectly) call sys.exit().  So they're saying "Hello caller, I've decided that the program needs to exit now.  Please do so".  When these were used only with mandatory files (needed  for the operation of convert2rhel), that choice didn't raise any red flags.  If you had raised a normal exception, the caller would have made the same decision that we're currently making in the callee.

With this PR, though, we're introducing optional RestorableFiles.  We want to make sure the file exists in its original state if we have to rollback but if they don't exist to begin with we're fine with that too.  That reveals the flaw with raising SystemExit in these functions.  We no longer want to exit in all cases.  Further, RestorableFile doesn't know which cases are mandatory and which are optional.  So we need the caller to make the decision on whether to exit or not.  What information does the function need to return to the caller?  The caller knows whether the file is optional or mandatory from its own context but it needs us to tell it what went wrong (FileNotFound on an optional file can be ignored.  PermissionError might point at some large, unexpected problem like chattr or a filesystem mounted read-only.  MemoryError is likely unrecoverable).  So instead of raising SystemExit, it's better if these functions raise an exception which contains the information the caller needs to make its decision to recover and continue or abort.

convert2rhel/unit_tests/redhatrelease_test.py
r0x0d: @abadger generally, I follow the same rule (always separate the good from the bad tests), but for this one, I just followed the convention of other tests in the codebase. The only place I saw using the separate good/bad tests was on class tests, which I believe is not the case on this one. I kinda like the way of using good/bad tests in the `pytest.mark.parametrize`, it's concise and clean, and if you have to add new tests cases for that, it's not that difficult after all.

convert2rhel/unit_tests/redhatrelease_test.py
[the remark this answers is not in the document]
r0x0d: I think my question for that test is no longer valid... Since the review, we found out the function had no purpose as it was not being used by anyone else, so, this test got deleted as well. But the discussion about different tests types is really good. Maybe in the future, we can think about how to improve more our tests with the help of QE folks and get something more consistent too.

convert2rhel/unit_tests/redhatrelease_test.py
bocekm: We try to call the functions under test just once in a test. In case you need to catch the Exception, we typically do it the way you have it now - calling it either with or without the `with pytest.raises(<Exc>):` statement, depending on a condition provided by `parametrize`. So it's not called twice within one test run.

We started using `parametrize` quite late so you can see that some functions under test are tested by multiple test functions instead of just one test function with `parametrize`. I myself like the parametrize approach more as it prevents code duplication.

@abadger, let us know if you know of a more elegant solution.

convert2rhel/unit_tests/redhatrelease_test.py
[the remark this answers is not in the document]
bocekm: I'm on the fence here because if you write a functional test for a high level function (in terms of abstraction) you could end up mocking lots of functions many levels lower.

I think this is worth discussing on a separate meeting, to get ourselves on the same page when it comes to testing.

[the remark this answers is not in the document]
r0x0d: Agreed. Our testes are really divergent on some files, maybe we could set up that separate meeting in the near future to discuss better on what we want to do (or we could leave it when we port c2r to leapp as well).

convert2rhel/unit_tests/redhatrelease_test.py
abadger: I love parametrize too!  However, I love it so much that I have had colleagues point out that it can easily lead to tests that are hard to read and debug (it makes what the test does declarative but that also means that reasoning about what a test is doing [to debug it or to add a new test case] is hard to follow and the relevant information is located in multiple locations).  I haven't completely cured myself of overusing parametrization but there are a few rules that seem to work well:

* Use a new test if it would require adding a conditional to the current test's code.
* Use a new test if it you'd have to create different mock objects for it in the current test.
* Use a new test if the output that you check for has to follow a different pattern (type, format string, regex, exception, etc) for the new parameter.

Over parametrization often goes along with this antipattern (story format): https://www.informit.com/articles/article.aspx?p=1392524  (more traditional): https://martinfowler.com/bliki/FlagArgument.html  Instead of creating a different function whose name can clue you in on what is happening, you end up with a long data structure of True|False values that aren't obvious on their own what is happening.  Looking up the meaning of the values is possible but has a cost when debugging or adding a new test case.

Mentioning debugging brings up a difference (in emphasis) that I see between test code and regular code.  All code should be as simple to read as possible, but it's especially important for test code.  When your unit tests fail, you want it to be obvious what the test case is doing.  It's not a good feeling to have a failing test and spend most of the time trying to decide if the test case is wrong.

For the question of preventing code duplication, I think that fixtures are what's appropriate in the general case (and parametrization is appropriate in specific cases).  pytest fixtures are very flexible and you can push large sections of boilerplate into them (although, again, be careful that you don't make it so someone has to spend more time determining if your test is correct than looking for the bug in their code).

I think it's debatable whether duplication is appropriate to worry about here as there's only two lines in common but to see what it looks like:

``` python
@pytest.fixture    
def monkeypatch_isfile(request, monkeypatch):    
    isfile_returns = request.node.get_closest_marker("isfile_returns")    
    if isfile_returns is None:    
        isfile_returns = False    
    else:    
        isfile_returns = isfile_returns.args[0]    
    
    is_file_mock = mock.MagicMock(return_value=isfile_returns)    
    monkeypatch.setattr(os.path, "isfile", value=is_file_mock)    
    return is_file_mock    

@pytest.mark.isfile_returns(True)    
def test_get_os_release_filepath(monkeypatch_isfile, caplog):    
    assert get_os_release_filepath() == "/etc/os-release"    
    
@pytest.mark.isfile_returns(False)    
def test_get_os_release_filepath_file_not_found(monkeypatch_isfile, caplog):    
    with pytest.raises(SystemExit):    
        get_os_release_filepath()    
    assert (    
        "Error: Unable to find the /etc/os-release file containing the information needed."    
        in caplog.records[-1].message    
    )
```

SpyTec (approved):

convert2rhel/unit_tests/redhatrelease_test.py
bocekm: Great write up, @abadger. I agree with it in full. Some duplication is indeed better than a test that hard to decipher. The fixture approach is something I haven't considered and I like it.

The Clean Code Tip story is great. I fear our codebase that smells like it a bit on couple of places.

bocekm: I 'd say the tests as @r0x0d has them now are short and easily readable/understandable. Or would you like him to change them somehow?

abadger: It is okay with me.  I would like to see good and bad test cases split up in the future but I think maybe that could be a task I take on for all the tests in the future?  Would you be okay with that sort of mass cleanup from me?

convert2rhel/unit_tests/redhatrelease_test.py
```diff
@@ -31,13 +32,23 @@
...
+YUM_CONF_WITHOUT_DISTROVERPKG = """[main]
+installonly_limit=3
+
+#  This is the default"""
+
+YUM_CONF_WITH_DISTROVERPKG = """[main]
+installonly_limit=3
+distroverpkg=centos-release
+
+#  This is the default"""
```
SpyTec, line 48: Are both the default?

r0x0d: I didn't change those strings, just moved them to the top of the file for better organizing. They were already like this

SpyTec: Ah I see it now, no worries

SpyTec (approved):

convert2rhel/redhatrelease.py
bocekm: Thanks for the insights. I wasn't aware of `atexit` and it looks like a better alternative to our exit mechanism indeed.

You raise good questions. We don't read the content of the /etc/os-release file. We just want to make sure we back it up if it exists. If it doesn't, it's ok, we shouldn't have a problem with that. So I agree with you on using `loggerinst.info()` instead of `loggerinst.critical()`. Perhaps with a message like _"Skipping backing up /etc/os-release - file not available."_.

convert2rhel/unit_tests/redhatrelease_test.py
bocekm: Sure, I'm definitely not against that.

convert2rhel/redhatrelease.py
bocekm: I'm getting on CentOS Linux 6:
```
Backing up None.
[12/03/2021 20:20:16] DEBUG - Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/convert2rhel/main.py", line 94, in main
    redhatrelease.os_release_file.backup()
  File "/usr/lib/python2.6/site-packages/convert2rhel/utils.py", line 527, in backup
    if os.path.isfile(self.filepath):
  File "/usr/lib64/python2.6/genericpath.py", line 29, in isfile
    st = os.stat(path)
TypeError: coercing to Unicode: need string or buffer, NoneType found

No changes were made to the system.
```

bocekm: The

bocekm: Fixed by https://github.com/oamg/convert2rhel/pull/384.

PR #358, opened 2021-10-21
Halt when unable to download backup

Andrew-ang9:
I changed what happens when the conversion can't rollback the rpm files. it now prompts the user after letting them know that the conversion couldn't rollback the packages and shows them the packages that couldn't be rolledback and then prompts the user to see if they want to continue of not.

Fixes OAMG-5463

### DIFF
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -207,10 +207,17 @@ class TestUtils(unittest.TestCase):
 
     @unit_tests.mock(system_info, "version", namedtuple("Version", ["major", "minor"])(7, 0))
     @unit_tests.mock(utils, "run_cmd_in_pty", RunSubprocessMocked(ret_code=1))
-    def test_download_pkg_failed_download(self):
+    @unit_tests.mock(os, "environ", {"CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK": "1"})
+    def test_download_pkg_failed_download_overridden(self):
         path = utils.download_pkg("kernel")
 
         self.assertEqual(path, None)
+
+    @unit_tests.mock(system_info, "version", namedtuple("Version", ["major", "minor"])(7, 0))
+    @unit_tests.mock(utils, "run_cmd_in_pty", RunSubprocessMocked(ret_code=1))
+    @unit_tests.mock(os, "environ", {})
+    def test_download_pkg_failed_download_exit(self):
+        self.assertRaises(SystemExit, utils.download_pkg, "kernel")
 
     @unit_tests.mock(system_info, "version", namedtuple("Version", ["major", "minor"])(7, 0))
     @unit_tests.mock(utils, "run_cmd_in_pty", RunSubprocessMocked(ret_code=0))
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -471,10 +471,24 @@ def download_pkg(
 
     output, ret_code = run_cmd_in_pty(cmd, print_output=False)
     if ret_code != 0:
-        loggerinst.warning(
-            "Couldn't download the %s package using yumdownloader.\n"
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
-        )
+        loggerinst.warning("Output from the yumdownloader call:\n%s" % (output))
+
+        if not "CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK" in os.environ:
+            loggerinst.critical(
+                "Couldn't download the %s package. This means we will not be able to do a"
+                " complete rollback and may put the system in a broken state.\n"
+                "Check to make sure that the %s repositories are enabled"
+                " and the package is updated to its latest version.\n"
+                "If you would rather ignore this check set the environment variable"
+                " 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK'." % (pkg, system_info.name)
+            )
+        else:
+            loggerinst.warning(
+                "Couldn't download the %s package. This means we will not be able to do a"
+                " complete rollback and may put the system in a broken state.\n"
+                "'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' environment variable detected, continuing conversion."
+                % (pkg)
+            )
         return None
 
     path = get_rpm_path_from_yumdownloader_output(cmd, output, dest)
--- a/tests/integration/tier0/backup-release/test_backup_release.py
+++ b/tests/integration/tier0/backup-release/test_backup_release.py
@@ -1,14 +1,15 @@
+import os
 import platform
 
 from envparse import env
 
 
-def test_backup_os_release(shell, convert2rhel):
+def test_backup_os_release_no_envar(shell, convert2rhel):
     """
-    This test case renoves all the repos on the system and validates that
-    the /etc/os-release package is being backed up and restored during rollback.
-    Ref ticket: OAMG-5457. Note that after the test, the $releaserver
-    variable is unset.
+    In this scenario there is no variable `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK` set.
+    This means the conversion is inhibited in early stage.
+    This test case removes all the repos on the system which prevents the backup of some files.
+    Satellite is being used in all of test cases.
     """
 
     # OL distros may not have wget installed
@@ -25,10 +26,35 @@ def test_backup_os_release(shell, convert2rhel):
 
     assert shell("find /etc/os-release").returncode == 0
     with convert2rhel(
-        ("--no-rpm-va -k {} -o {} --debug --keep-rhsm").format(
+        ("-y --no-rpm-va -k {} -o {} --debug --keep-rhsm").format(
             env.str("SATELLITE_KEY"),
             env.str("SATELLITE_ORG"),
         )
+    ) as c2r:
+        c2r.expect("set the environment variable 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK.")
+        assert c2r.exitstatus != 0
+
+    assert shell("find /etc/os-release").returncode == 0
+
+
+def test_backup_os_release_with_envar(shell, convert2rhel):
+    """
+    In this scenario the variable `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK` is set.
+    This test case removes all the repos on the system and validates that
+    the /etc/os-release package is being backed up and restored during rollback.
+    Ref ticket: OAMG-5457. Note that after the test, the $releaserver
+    variable is unset.
+    """
+
+    assert shell("find /etc/os-release").returncode == 0
+
+    os.environ["CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK"] = "1"
+
+    with convert2rhel(
+        ("--no-rpm-va -k {} -o {} --debug --keep-rhsm").format(
+            env.str("SATELLITE_KEY"),
+            env.str("SATELLITE_ORG"),
+        ),
     ) as c2r:
         c2r.expect("Continue with the system conversion?")
         c2r.sendline("y")


convert2rhel/utils.py
```diff
@@ -471,10 +471,17 @@ def download_pkg(
 
     output, ret_code = run_cmd_in_pty(cmd, print_output=False)
     if ret_code != 0:
-        loggerinst.warning(
-            "Couldn't download the %s package using yumdownloader.\n"
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
-        )
+        loggerinst.warning("you are using an env that overides an inhibitor and 
+            you understand that if you continue you might lose the rollback."
+            \n"Output from the yumdownloader call:\n%s" % (pkg, output)
```
r0x0d, line 476: You need to close the quotation marks on every physical line break in strings, otherwise, Python would understand you left the string open (by the initial quotation mark) and suddenly wanted to work on a new line, thus, creating an error... 

It's okay if you do something like this: 

```python
print("This gonna be a very very very long"
       " message for the user, which I will break a line for better readability.")
       
# Outputs
# This gonna be a very very very long message for the user, which I will break a line for better readability.
```

It didn't actually break a line in the message output, thus, you can create very long strings with easier reading without *actually* including a mandatory `\n` on your strings.

```suggestion
        loggerinst.warning("you are using an env that overides an inhibitor and" 
            " you understand that if you continue you might lose the rollback.\n"
            "Output from the yumdownloader call:\n%s" % (pkg, output)
```

convert2rhel/utils.py
```diff
@@ -471,10 +471,17 @@ def download_pkg(
 
     output, ret_code = run_cmd_in_pty(cmd, print_output=False)
     if ret_code != 0:
-        loggerinst.warning(
-            "Couldn't download the %s package using yumdownloader.\n"
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
-        )
+        loggerinst.warning("you are using an env that overides an inhibitor and
```
r0x0d, line 474: The `you` at the beginning of the message should be capitalized.

convert2rhel/utils.py
```diff
@@ -471,10 +471,17 @@ def download_pkg(
...
+            you understand that if you continue you might lose the rollback."
+            \n"Output from the yumdownloader call:\n%s" % (pkg, output)
+            )
+            cont = user_prompt("\nContinue with the system conversion? [y/n]: ")
+ 
+            if user_prompt == "y":
+                break
+
+            if user_prompt == "n":
+                sys.exit("User canceled the conversion\n")
```
r0x0d, line 484: Here you could just replace this part of the code with an actual call to the function [ask_to_continue](https://github.com/oamg/convert2rhel/blob/874e25a4fad8e40a31cec6b0006414f4b419394b/convert2rhel/utils.py#L232), since it does exactly what you want to do and you can just reuse the function.

r0x0d (changes requested):
The code looks code! Just some minor enhancements to look even better!

r0x0d:
Just a quick tip. The pipeline for `pre-commit` and `pylint` failed, you should be able to fix those pipelines by following the "[Linting](https://github.com/oamg/convert2rhel/blob/main/CONTRIBUTING.md#linting)" guide and the "[Work with pre-commit hooks](https://github.com/oamg/convert2rhel/blob/main/CONTRIBUTING.md#work-with-pre-commit-hooks)" guide too.

SpyTec (changes requested):
Code looks good! But we want to do it in an opposite way, by default we should halt conversion and only allow the user to go past it by using an environment variable. We shouldn't ask the user during the conversion as they may end up with a broken system

Instead use the `os` package and check that the environment variable `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK` exists in `os.environ`, and if it doesn't we simply halt conversion with a message

convert2rhel/utils.py
```diff
@@ -472,10 +472,19 @@ def download_pkg(
...
         loggerinst.warning(
-            "Couldn't download the %s package using yumdownloader.\n"
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
+            " you are using an env that overrides an inhibitor and"
+            " you understand that if you continue you might lose the rollback."
+            " \nOutput from the yum downloader call:\n%s" % (pkg, output)
         )
-        return None
+
+        if not CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK in os.environ:  # template for what needs to be done
```
r0x0d, line 480: Think on this line you forgot to add some `""` in the `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK`, as it seems you're doing an existence check here. 

Maybe something like this: 

```python
if not "CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK" in os.environ:  # template for what needs to be done
```

One thing that crossed my mind right now is: 

* Should we actually validate the value for `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK` environment variable? 

Because if we do, we could add another condition to this `if-statement` to validate if the value for that variable is the right one.

convert2rhel/utils.py
```diff
@@ -472,10 +472,19 @@ def download_pkg(
...
+            loggerinst.critical("")
```
r0x0d, line 481: I believe you missed the logger critical message here. As it seems, the message that was supposed to go here is in the line below, `loggerinst.info()`.

convert2rhel/utils.py
```diff
@@ -472,10 +472,19 @@ def download_pkg(
...
+        loggerinst.info(
+            "could not complete the rollback because one or more files are not updated, Check to see"
+            "if all packages are updated and try the conversion again"
+        )
+
+        return none
```
r0x0d, line 487: This should be a `None`, with capitalized `N`

```suggestion
        return None
```

convert2rhel/utils.py
```diff
@@ -472,10 +472,19 @@ def download_pkg(
     output, ret_code = run_cmd_in_pty(cmd, print_output=False)
     if ret_code != 0:
         loggerinst.warning(
-            "Couldn't download the %s package using yumdownloader.\n"
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
+            " you are using an env that overrides an inhibitor and"
```
r0x0d, line 475:
```suggestion
            "You are using an env that overrides an inhibitor and"
```

r0x0d (changes requested):

convert2rhel/utils.py
```diff
@@ -472,9 +472,18 @@ def download_pkg(
...
+            " you understand that if you continue you might lose the rollback."
+            " \nOutput from the yum downloader call:\n%s %s" % (pkg, output)
```
SpyTec, line 477: I'd remove this warning log and just use the logging from the if-statement below

convert2rhel/utils.py
```diff
@@ -472,9 +472,18 @@ def download_pkg(
...
         )
+
+        if not "CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK" in os.environ:  # template for what needs to be done
+            loggerinst.critical("")
+        loggerinst.info(
+            "could not complete the rollback because one or more files are not updated, Check to see"
+            "if all packages are updated and try the conversion again"
+        )
```
SpyTec, line 485: Rollback isn't happening here. It's just that when the rollback happens and they have the override variable active the rollback won't have the package backed up. I'd also change this to a warning

convert2rhel/unit_tests/utils_test.py
```diff
@@ -207,11 +207,19 @@ def test_download_pkg_success_with_all_params(self):
...
         path = utils.download_pkg("kernel")
 
         self.assertEqual(path, None)
 
+    @unit_tests.mock(system_info, "version", namedtuple("Version", ["major", "minor"])(7, 0))
+    @unit_tests.mock(utils, "run_cmd_in_pty", RunSubprocessMocked(ret_code=1))
+    @unit_tests.mock(os, "environ", {})
+    def test_download_pkg_failed_download_exit(self):
+
+        self.assertRaises(SystemExit, utils.download_pkg, "kernel")
```
SpyTec, line 221:
```suggestion
        self.assertRaises(SystemExit, utils.download_pkg, "kernel")
```

SpyTec (changes requested):

convert2rhel/utils.py
Andrew-ang9: I'm not quite sure which would be better to do in this situation, but I will check with Eric or Michael to see what would be best!

thanks!

convert2rhel/utils.py
```diff
@@ -473,8 +473,16 @@ def download_pkg(
...
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
+            " \nOutput from the yum downloader call:\n%s %s" % (pkg, output)
         )
+
+        if not "CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK" in os.environ:
+            loggerinst.critical(
+                "Without thoes packages we would be unable to rollback if needed."
+                " Check to make sure that your packages are up to date."
+                " If you would rather ignore this check set the envirament variable 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' to 1."
+            )
```
SpyTec, line 484:
```suggestion
                "This means we will not be able to do a complete rollback and may put the system in an broken state."
                " Check to make sure that your packages are up to date."
                " If you would rather override this check set the 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' environment variable.\n"
            )
        else:
            loggerinst.warning(
                "'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' environment variable detected, continuing conversion."
                " This may put the system in an unsupported broken state on a rollback."
            )
```

convert2rhel/utils.py
```diff
@@ -473,8 +473,16 @@ def download_pkg(
     if ret_code != 0:
         loggerinst.warning(
             "Couldn't download the %s package using yumdownloader.\n"
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
+            " \nOutput from the yum downloader call:\n%s %s" % (pkg, output)
```
SpyTec, line 476:
```suggestion
            " Output from the yumdownloader call:\n%s" % (pkg, output)
```

SpyTec (changes requested):

convert2rhel/utils.py
SpyTec: We don't really need to care about validating the data. Just as long as the env is set that's good enough

convert2rhel/utils.py
Andrew-ang9: I have reviewed this suggestion and I have added it to the most recent commit.

convert2rhel/utils.py
Andrew-ang9: I have reworked this and merged this into the code!

convert2rhel/utils.py
Andrew-ang9: the new code has changed the warning message and doesn't use this message anymore

convert2rhel/utils.py
Andrew-ang9: the new code doesn't require any immediate action from the user so a prompts isn't needed.

convert2rhel/utils.py
Andrew-ang9: I have merged the changes from this comment

convert2rhel/utils.py
Andrew-ang9: I have fixed the capitalization on this

convert2rhel/utils.py
Andrew-ang9: the new code doesn't use this message anymore.

convert2rhel/utils.py
Andrew-ang9: I have went back and fixed the typos in this comment

convert2rhel/unit_tests/utils_test.py
Andrew-ang9: I have gotten rid of the extra line in this comment

convert2rhel/utils.py
Andrew-ang9: In the new code for this I have totally changed the message this warning outputs

r0x0d (commented):
Think this is a minor typo in the tests name function

bocekm (changes requested):

convert2rhel/utils.py
```diff
@@ -473,8 +473,20 @@ def download_pkg(
...
+        if not "CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK" in os.environ:
+            loggerinst.critical(
+                "This means we will not be able to do a complete rollback and may put the system in a broken state."
+                " Check to make sure that your packages are up to date."
+                " If you would rather ignore this check set the environment variable 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' to 1."
+            )
+        else:
+            loggerinst.warning(
+                "'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' environment variable detected, continuing conversion."
+                " This may put the system in an unsupported broken state on a rollback."
```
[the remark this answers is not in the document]
Andrew-ang9: I will add these changes in my next push to this PR

convert2rhel/utils.py
```diff
@@ -473,8 +473,20 @@ def download_pkg(
...
             "Couldn't download the %s package using yumdownloader.\n"
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
+            " Output from the yumdownloader call:\n%s" % (pkg, output)
         )
+
+        if not "CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK" in os.environ:
+            loggerinst.critical(
+                "This means we will not be able to do a complete rollback and may put the system in a broken state."
+                " Check to make sure that your packages are up to date."
+                " If you would rather ignore this check set the environment variable 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' to 1."
```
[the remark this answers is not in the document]
Andrew-ang9: I will add these changes in my next push to this PR

convert2rhel/utils.py
```diff
@@ -473,8 +473,20 @@ def download_pkg(
     if ret_code != 0:
         loggerinst.warning(
             "Couldn't download the %s package using yumdownloader.\n"
```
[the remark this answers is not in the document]
Andrew-ang9: Yeah now that you mention it, I do see that there seems to be no association between the two, I will make the changes that you have suggested in the other comments and push up!

convert2rhel/utils.py
```diff
@@ -473,8 +473,25 @@ def download_pkg(
...
```
bocekm, line 475:
```suggestion
```
Please remove this duplicate msg.

bocekm (changes requested):
An integration test is needed before merge.

convert2rhel/utils.py
Andrew-ang9: sounds good!

convert2rhel/utils.py
```diff
@@ -471,10 +471,24 @@ def download_pkg(
...
+        loggerinst.warning(" Output from the yumdownloader call:\n%s" % (output))
+
+        if not "CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK" in os.environ:
+            loggerinst.critical(
+                "Couldn't download the %s package. This means we will not be able to do a"
+                " complete rollback and may put the system in a broken state.\n"
+                "Check to make sure that the %s repositories are enabled"
+                " and the package is updated to it's latest version.\n"
+                "If you would rather ignore this check set the environment variable"
+                " variable 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' to 1." % (pkg, system_info.name)
```
SpyTec, line 483:
```suggestion
                " and the package is updated to its latest version.\n"
                "If you would rather ignore this check set the environment variable"
                " 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK'." % (pkg, system_info.name)
```

convert2rhel/utils.py
```diff
@@ -471,10 +471,24 @@ def download_pkg(
 
     output, ret_code = run_cmd_in_pty(cmd, print_output=False)
     if ret_code != 0:
-        loggerinst.warning(
-            "Couldn't download the %s package using yumdownloader.\n"
-            "Output from the yumdownloader call:\n%s" % (pkg, output)
-        )
+        loggerinst.warning(" Output from the yumdownloader call:\n%s" % (output))
```
SpyTec, line 474:
```suggestion
        loggerinst.warning("Output from the yumdownloader call:\n%s" % (output))
```

SpyTec (changes requested):

SpyTec (dismissed):

SpyTec:
Code LGTM. Just need an integration test now

r0x0d (commented):
double `variable` in message

kokesak:
The failing test was re-run: http://artifacts.osci.redhat.com/testing-farm/68bcaaff-2470-470f-b03f-255ebf3f9120 (PASSED)

bocekm (approved):

r0x0d (approved):

PR #359, opened 2021-10-25
Add archive to old log files

r0x0d:
This PR aims to add the ability to archive old log files before a new run. 

For example: 
* /var/log/convert2rhel/archive/convert2rhel-1635162445070567607.log
* /var/log/convert2rhel/archive/convert2rhel-1635162478219820043.log

The name chosen for the new file is based on the previous log file name (eg; convert2rhel.log), but with an appended timestamp to it. This timestamp is relative to the last modified date (Which means, the last command sent to the log file).

Beware, this PR is also linked to the [sosreport/sos - PR#2734](https://github.com/sosreport/sos/pull/2734) as we have to include the archived logs in the sosreport tool.

Ticket reference: https://issues.redhat.com/browse/OAMG-5458

### DIFF
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -27,6 +27,7 @@ FILE      (5)     CUSTOM LABEL - Prints only to file handler (using date/time)
 """
 import logging
 import os
+import shutil
 import sys
 
 from convert2rhel.toolopts import tool_opts
@@ -46,8 +47,8 @@ class LogLevelFile(object):
     label = "FILE"
 
 
-def initialize_logger(log_name, log_dir=LOG_DIR):
-    """Initialize custom logging levels, handlers, and so on. Call this method
+def setup_logger_handler(log_name, log_dir):
+    """Setup custom logging levels, handlers, and so on. Call this method
     from your application's main start point.
         log_name = the name for the log file
         log_dir = path to the dir where log file will be presented
@@ -86,6 +87,42 @@ def initialize_logger(log_name, log_dir=LOG_DIR):
     handler.setFormatter(formatter)
     handler.setLevel(LogLevelFile.level)
     logger.addHandler(handler)
+
+
+def archive_old_logger_files(log_name, log_dir):
+    """Archive the old log files to not mess with multiple runs outputs.
+    Every time a new run begins, this method will be called to archive the previous logs
+    if there is a `convert2rhel.log` file there, it will be archived using
+    the same name for the log file, but having an appended timestamp to it.
+        log_name = the name for the log file
+        log_dir = path to the dir where log file will be presented
+
+    For example:
+        /var/log/convert2rhel/archive/convert2rhel-1635162445070567607.log
+        /var/log/convert2rhel/archive/convert2rhel-1635162478219820043.log
+
+    This way, the user can track the logs for each run individually based on the timestamp.
+    """
+
+    current_log_file = os.path.join(log_dir, log_name)
+    archive_log_dir = os.path.join(log_dir, "archive")
+
+    if not os.path.exists(current_log_file):
+        # No log file found, that means it's a first run or it was manually deleted
+        return
+
+    stat = os.stat(current_log_file)
+
+    # Get the last time the file was modified
+    last_modified_at = stat.st_mtime
+
+    # Create the directory if it don't exist
+    if not os.path.exists(archive_log_dir):
+        os.makedirs(archive_log_dir)
+
+    file_name, suffix = tuple(log_name.rsplit(".", 1))
+    archive_log_file = "%s/%s-%s.%s" % (archive_log_dir, file_name, last_modified_at, suffix)
+    shutil.copyfile(current_log_file, archive_log_file)
 
 
 def _task(self, msg, *args, **kwargs):
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -19,21 +19,10 @@ import logging
 import os
 import sys
 
-from convert2rhel import (
-    breadcrumbs,
-    cert,
-    checks,
-    grub,
-    logger,
-    pkghandler,
-    redhatrelease,
-    repo,
-    special_cases,
-    subscription,
-    systeminfo,
-    toolopts,
-    utils,
-)
+from convert2rhel import breadcrumbs, cert, checks, grub
+from convert2rhel import logger as logger_module
+from convert2rhel import pkghandler, redhatrelease, repo, special_cases, subscription, systeminfo, toolopts, utils
+from convert2rhel.grub import logger
 
 
 loggerinst = logging.getLogger(__name__)
@@ -47,6 +36,26 @@ class ConversionPhase(object):
     POST_PONR_CHANGES = 3
 
 
+def initialize_logger(log_name, log_dir):
+    """
+    Entrypoint function that aggregates other calls for initialization logic
+    and setup for logger handlers.
+
+    .. warning::
+        Setting log_dir underneath a world-writable directory (including
+        letting it be user settable) is insecure.  We will need to write
+        some checks for all calls to `os.makedirs()` if we allow changing
+        log_dir.
+    """
+
+    try:
+        logger_module.archive_old_logger_files(log_name, log_dir)
+    except (IOError, OSError) as e:
+        print("Warning: Unable to archive previous log: %s" % e)
+
+    logger_module.setup_logger_handler(log_name, log_dir)
+
+
 def main():
     """Perform all steps for the entire conversion process."""
 
@@ -54,8 +63,9 @@ def main():
     utils.require_root()
 
     process_phase = ConversionPhase.INIT
+
     # initialize logging
-    logger.initialize_logger("convert2rhel.log")
+    initialize_logger("convert2rhel.log", logger_module.LOG_DIR)
 
     # prepare environment
     utils.set_locale()
--- a/convert2rhel/unit_tests/logger_test.py
+++ b/convert2rhel/unit_tests/logger_test.py
@@ -29,7 +29,7 @@ def test_logger_handlers(tmpdir, caplog, read_std, is_py2, capsys):
     # initializing the logger first
     log_fname = "convert2rhel.log"
     tool_opts.debug = True  # debug entries > stdout if True
-    logger_module.initialize_logger(log_name=log_fname, log_dir=str(tmpdir))
+    logger_module.setup_logger_handler(log_name=log_fname, log_dir=str(tmpdir))
     logger = logging.getLogger(__name__)
 
     # emitting some log entries
@@ -49,7 +49,7 @@ def test_logger_handlers(tmpdir, caplog, read_std, is_py2, capsys):
 
 def test_tools_opts_debug(tmpdir, read_std, is_py2):
     log_fname = "convert2rhel.log"
-    logger_module.initialize_logger(log_name=log_fname, log_dir=str(tmpdir))
+    logger_module.setup_logger_handler(log_name=log_fname, log_dir=str(tmpdir))
     logger = logging.getLogger(__name__)
     tool_opts.debug = True
     logger.debug("debug entry 1")
@@ -73,9 +73,35 @@ def test_tools_opts_debug(tmpdir, read_std, is_py2):
 def test_logger_custom_logger(tmpdir, caplog):
     """Test CustomLogger."""
     log_fname = "convert2rhel.log"
-    logger_module.initialize_logger(log_name=log_fname, log_dir=str(tmpdir))
+    logger_module.setup_logger_handler(log_name=log_fname, log_dir=str(tmpdir))
     logger = logging.getLogger(__name__)
     logger.task("Some task")
     logger.file("Some task write to file")
     with pytest.raises(SystemExit):
         logger.critical("Critical error")
+
+
+@pytest.mark.parametrize(
+    ("log_name", "path_exists"),
+    (
+        ("convert2rhel.log", True),
+        ("convert2rhel.log", False),
+    ),
+)
+def test_archive_old_logger_files(log_name, path_exists, tmpdir, caplog):
+    tmpdir = str(tmpdir)
+    archive_dir = os.path.join(tmpdir, "archive")
+    log_file = os.path.join(tmpdir, log_name)
+    test_data = "test data\n"
+
+    if path_exists:
+        open(log_file, "w").write(test_data)
+
+    logger_module.archive_old_logger_files(log_name, tmpdir)
+
+    if path_exists:
+        assert "archive" in os.listdir(tmpdir)
+        archive_files = os.listdir(archive_dir)
+        assert len(archive_files) == 1
+        with open(os.path.join(archive_dir, archive_files[0])) as archive_f:
+            assert archive_f.read() == test_data
--- a/convert2rhel/unit_tests/main_test.py
+++ b/convert2rhel/unit_tests/main_test.py
@@ -17,7 +17,18 @@
 
 
 import os
+import sys
 import unittest
+
+
+if sys.version_info[:2] <= (2, 7):
+    import mock  # pylint: disable=import-error
+else:
+    from unittest import mock  # pylint: disable=no-name-in-module
+
+import pytest
+
+from convert2rhel import logger as logger_module
 
 
 try:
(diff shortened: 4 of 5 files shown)


r0x0d:
Update: The PR is now merged on their end: https://github.com/sosreport/sos/pull/2734

convert2rhel/logger.py
```diff
@@ -88,6 +90,35 @@ def initialize_logger(log_name, log_dir=LOG_DIR):
...
+    current_log_file = "%s/convert2rhel.log" % LOG_DIR
+
+    if os.path.exists(current_log_file):
+        stat = os.stat(current_log_file)
+
+        # Get the last time the file was modified as nanoseconds
+        last_modified_at = stat.st_mtime_ns
+
+        # Create the directory if it don't exist
+        if not os.path.isdir(ARCHIVE_LOG_DIR):
```
abadger, line 115: This is a directory check but the logic is really about existence.  Maybe this should be `os.path.exists()` instead?

Additionally, maybe we want to check if it is a directory afterwards and fail with an exception that we can catch at the higher level to emit a nice error message (Like `"%s exists but is not a directory.  Old log files not archived." % ARCHIVE_LOG_DIR`)  (We can also catch whatever exceptions shutil.copyfile  emits rather than checking whether we're using a directory explicitly.  I'm not sure, though, how many other things copyfile could fail on that we'd have to look through.)

Lastly, just a future note:  Using `os.makedirs()` like this is sometimes insecure.  As long as we're hardcoding the path to something that isn't writable by other users, we should be okay but we'd need to worry about it if we make this user settable (as the user could set it to something like `/tmp/convert2rhel/` which would lead to symlink attacks)

convert2rhel/unit_tests/logger_test.py
```diff
@@ -79,3 +88,15 @@ def test_logger_custom_logger(tmpdir, caplog):
...
+@unit_tests.mock(os.path, "exists", lambda x: True)
+@unit_tests.mock(os.path, "isdir", lambda x: False)
+@unit_tests.mock(os, "makedirs", lambda x: None)
+@unit_tests.mock(shutil, "copyfile", lambda x, y: None)
+@mock.patch("os.stat")
+def test_archive_old_logger_files(os_stat):
+    # Random value for st_mtime_ns mock
+    os_stat.return_value.st_mtime_ns = 123456
+
+    assert logger_module.archive_old_logger_files()
```
abadger, line 102: Maybe we should assert that copyfile was called with filenames that we expect?

convert2rhel/main.py
```diff
@@ -54,6 +54,10 @@ def main():
     utils.require_root()
 
     process_phase = ConversionPhase.INIT
+
+    # Archive the old log file before starting a new run
+    logger.archive_old_logger_files()
```
abadger, line 59: It feels like at this level, we just want to call `initialize_logger()` and that would take care of all tasks related to setting up logging for us to run.  However, the present initialize_logger seems (appropriately) focused on setup that relates to the logging stdlib library.  Maybe we need a new function that sits in the middle?  What do you think of organizing it something like this:

``` python
# In main.py:
def main():
    logger.initialize_logger("convert2rhel.log")

# In logger.py:
def initialize_logger(log_name, log_dir=LOG_DIR):
    archive_old_logger_files(log_name)
    setup_logging_library(log_name, log_dir)

def setup_logging_library(log_name, log_dir=LOG_DIR):
    # Contents of the old initialize_logger
```

convert2rhel/logger.py
abadger: Addtion to the insecure note.... the way we're constructing the path to the logfile means that if either the `log_dir` or `log_name` become user settable we'd have issues.  We may want to write some mitigating code in logging.py as the log_dir and log_name are both passed into this set of functions.  We're relying on main.py to hardcode the values.  Someone changing how the functions are called from main.py may not realize the security implications that raises here.

convert2rhel/logger.py
```diff
@@ -88,6 +90,35 @@ def initialize_logger(log_name, log_dir=LOG_DIR):
...
+    the same name for the log file, but having an appended timestamp to it.
+
+    For example:
+        /var/log/convert2rhel/archive/convert2rhel-1635162445070567607.log
+        /var/log/convert2rhel/archive/convert2rhel-1635162478219820043.log
+
+    This way, the user can track the logs for each run individually based on the timestamp.
+    """
+
+    current_log_file = "%s/convert2rhel.log" % LOG_DIR
```
abadger, line 106: Looks like we're hardcoding convert2rhel.log here but `initialize_logger()` allows it to be set by the caller.  So this needs to be set by the caller as well. If you like my proposed change: https://github.com/oamg/convert2rhel/pull/359/files#r741280923 it's natural to keep those in sync as the function that's the entrypoint will be able to pass the value to both helper functions.

abadger (changes requested):
I left one structural request:
* https://github.com/oamg/convert2rhel/pull/359/files#r741280923

and a pair of minor things:
* https://github.com/oamg/convert2rhel/pull/359/files#r741267961
* https://github.com/oamg/convert2rhel/pull/359/files#r741266062

Let me know what you think

convert2rhel/logger.py
r0x0d: Great catch, @abadger. Just to be sure on that exception you mentioned. At this point of code, we would have something like this? 

```python

# In logger.py:
if not os.path.exists(ARCHIVE_LOG_DIR):
    ...
else:
    raise FileNotFoundError
```

And in an earlier stage, let's say, on `logger.py` we would have something like 

```python
def initialize_logger(log_name, log_dir=LOG_DIR):
    
    try:
        archive_old_logger_files(log_name=log_name)
    except FileNotFoundError:
        ...

    setup_logger(log_name=log_name, log_dir=log_dir)
 ```
 
 Or do you have any better ideas?

File: convert2rhel/logger.py (hunk 4 above)
Comment (r0x0d) [reply convert2rhel/logger.py:106]: You're right! It's indeed way better to just keep them in sync and avoid duplication of code! I just wonder about this line https://github.com/oamg/convert2rhel/pull/359/files/6beb9a275de4891162a42a7dc2cbe9d5020c180c#diff-f82efabf7bb126100afd0472e3d8a5902da6576c6ddf75e3184cce9eb647c52aR118, since it will be used by another tool to collect the archived logs, I wonder If we can reuse the `log_name` to create the archived log file? I mean, they already accepted the PR on `main` branch https://github.com/sosreport/sos/pull/2734, but I can always open another one to fix the path if we need some changes. Do you have any thoughts on that one?

Comment (abadger) [reply convert2rhel/logger.py:106]: Yep, I agree that we should use log_name to both lookup the current log file and as part of the name of the archived log.

I don't think we need another change to sosreport right now.  We're hardcoding log_name as convert2rhel.log when we call initialize_logger() in main.py here: https://github.com/oamg/convert2rhel/pull/359/files/6beb9a275de4891162a42a7dc2cbe9d5020c180c#diff-7d31913eebe96d308744cc192ecf12ed870f425d573064d6e5dad3b53381ecc4R62

If we allow user's to customize the log name in the future, sosreport will need to be changed for both the current log name and the archived log name.

File: convert2rhel/logger.py [hunk 5]
```diff
@@ -47,7 +49,14 @@ class LogLevelFile(object):
 
 
 def initialize_logger(log_name, log_dir=LOG_DIR):
-    """Initialize custom logging levels, handlers, and so on. Call this method
+    """Entrypoint function that aggregates other calls for initialization logic
+    and setup for logger handlers."""
```
abadger, line 53:
```suggestion
    """
    Entrypoint function that aggregates other calls for initialization logic
    and setup for logger handlers.
    
    .. warning::
        Setting log_dir underneath a world-writable directory (including
        letting it be user settable) is insecure.  We will need to write
        some checks for all calls to `os.makedirs()` if we allow changing
        log_dir.
    """
```

convert2rhel/logger.py
```diff
@@ -88,6 +97,41 @@ def initialize_logger(log_name, log_dir=LOG_DIR):
...
+        log_name = the name for the log file
+
+    For example:
+        /var/log/convert2rhel/archive/convert2rhel-1635162445070567607.log
+        /var/log/convert2rhel/archive/convert2rhel-1635162478219820043.log
+
+    This way, the user can track the logs for each run individually based on the timestamp.
+    """
+
+    current_log_file = "%s/%s" % (LOG_DIR, log_name)
```
abadger, line 114: Sorry I didn't catch this earlier.  I just noticed that when we call `setup_logger_handler`, we pass in `log_dir`.  `log_dir` is then used to set what directory the current_log_file is output to here: https://github.com/oamg/convert2rhel/pull/359/files#diff-f82efabf7bb126100afd0472e3d8a5902da6576c6ddf75e3184cce9eb647c52aL83 

I recommend passing log_dir into this function too and setting both current_log_file and archive_log_dir based on that. ie:

``` python
current_log_file = os.path.join(log_dir, log_name)
archive_log_dir = os.path.join(log_dir, 'archive')
```

You can then also remove the toplevel ARCHIVE_LOG_DIR constant since we would be setting archive_log_dir based on log_dir.

abadger (changes requested):
I found one thing in the code that I had missed before.  Sorry about that!

I'm looking at the test right now, too, to see if there's a way to do it using a pytest tempdir instead of mocking os.exists and so forth.  I think the test will be easier to understand if we do that instead of mocking the stdlib functions.

convert2rhel/main.py
```diff
@@ -54,6 +54,7 @@ def main():
     utils.require_root()
 
     process_phase = ConversionPhase.INIT
+
     # initialize logging
     logger.initialize_logger("convert2rhel.log")
```
abadger, line 60: Question:  What do we want to happen when we cannot create the archive?  We explicitly raise IOError here, for instance:

https://github.com/oamg/convert2rhel/pull/359/files#diff-f82efabf7bb126100afd0472e3d8a5902da6576c6ddf75e3184cce9eb647c52aR128

Options I can see are:
* Choose to ignore it, possibly with a warning to the user.  We have the current log. Archiving it is just a convenience for the user.
* Fail with a message instructing the user on how to ix it.  We consider this historical data important enough that we do not want to continue if we cannot protect it.

convert2rhel/unit_tests/logger_test.py
```diff
@@ -73,9 +102,37 @@ def test_tools_opts_debug(tmpdir, read_std, is_py2):
...
+    monkeypatch.setattr(os.path, "isdir", value=is_dir_mock)
+    monkeypatch.setattr(shutil, "copyfile", value=shutil_mock)
+
+    os_stat.return_value.st_mtime_ns = 1234
+
+    if exception:
+        with pytest.raises(IOError):
+            logger_module.archive_old_logger_files(log_name)
+    else:
+        assert logger_module.archive_old_logger_files(log_name) is None
```
abadger, line 138: I suggest using pytest's tmpdir facility for this test instead of mocks.  This is what it would look like (Note: test won't pass/work until log_dir is added to `archive_old_logger_files()`: https://gist.github.com/abadger/19b31a1cd186c1dddd7351cba77b90a3

Using tmpdir pointed out that the mock based test would not have been able to test the case where current_log exists but the archive directory does not since checking for the current_log and archive directory are both done using os.path.exists(). (You could modify the mock based tests to return different values based on what path is passed in but that would make the mock based test even more complex)

convert2rhel/logger.py
r0x0d: Nice catch, @abadger! Didn't see that either. Just fixed

convert2rhel/main.py
r0x0d: But one thing that's crossed my mind right now is: 

* What are the odds of the current `archive` directory actually being a `file`? 

The way I see it, this directory will be created by us in a specific directory, at least for now. Since we don't handle user-provided paths, is it really necessary for us to do this checking right now? I mean, I see a scenario where this would be really applicable for user-provided paths, but since it's already "hardcoded" in our codebase, is it really necessary for us to do this checking?

bocekm: To @abadger's questions - I'd just print a warning if we are unable to archive an old log as archiving logs is not a critical part of the conversion process, especially when we generate the so called breadcrumbs file https://github.com/oamg/convert2rhel/pull/324 which records the basic information about all executions of convert2rhel.

@r0x0d, instead of checking whether the path is a dir you can add a try/catch block around the `shutil.copyfile` call (catching IOError and OSError to cover both py2 and py3 respectively) and log a warning upon catching an exception. That way you'd catch more issues than just the not-a-dir one, e.g. permission-related issues.

abadger: We don't necessarily have to check that this specific error condition exists but I think we should catch certain errors at the higher level.  And when we do that, we may want to give the user a good error message.   Let me explain with some pseudo-code:

```
# main.py
try:
    logger.initialize_logger("convert2rhel.log")
except IOError as e:
    print("Unable to initialize the logger: %s" % e)
    print("Stopping the run")
    sys.exit(1)
except ArchiveCreationError as e:
    print("Unable to backup previous log: %s" % e)
    print("Please take care of the issue and re-run convert2rhel")
    sys.exit(1)
```

If we don't have the try, except, then any error will give the user a traceback along with whatever the raw exception message is.  If we raise IOError (and other things that we're calling can also raise IOError), then we can print a generic message and handle all of those the same.  If we raise a different error than what other code could raise, then we can handle different errors in different ways (custom messages, choice to continue for some exceptions and exiting in others, etc).

So, I think we should have a try: except in main.py to catch error conditions having to do with creating the archive and give the user a nicer error message but whether we want to check specifically for `archive` being a `file` depends on what we want to do when that error occurs.  If we want to treat all errors the same, with the same generic message, then we do not need to do the `isdir()` check.  If we want to differentiate the two, then we do need the isdir() check and we need to either raise a unique exception or make sure the IOError we raise is what we want the user to read.

Does that make sense?

bocekm: Catching exceptions at a higher level sounds good. I would probably vote rather for adding the try/except under the `initialize_logger()` to have: a) all things logger contained in the logger.py, and 2) to keep main.py as clean and concise as possible (ideally no conditions or low level code). 

I don't think we need to differentiate between the archive dir path not being a dir and not being able to copy a file into the dir. Because the copying would fail if the path is not a dir. So I'd just remove the isdir() check.


E.g.
```
def initialize_logger(log_name, log_dir=LOG_DIR):
    ...
    try:
        archive_old_logger_files(log_name)
    except IOError as e:
        print("Unable to initialize the logger: %s" % e)
        print("Stopping the run")
        sys.exit(1)
    except ArchiveCreationError as e:
        print("Unable to backup previous log: %s" % e)
        print("Please take care of the issue and re-run convert2rhel")
        sys.exit(1)

        setup_logger_handler(log_name, log_dir)
```

abadger: We talked about this in chat and decided that `initialize_logger()` is the right place to do the checks (as Michal said, so that we can keep the `main()` function clean and concise) but we don't want to have `print()`s in logger.py.  That's so we can maintain a separation between code which interacts with the user and code which is about having the computer perform actions (it's the difference between code for a library and code for an application).  The way we decided to reconcile those two ideas is to use @bocekm's  version of `initialize_logger()` but move it into main.py.

@r0x0d One thing to note is when you add the exception handling to initialize_logger, leave out the block for ArchiveCreationError.  Since we aren't checking with isdir(), we won't have an opportunity to raise that.

convert2rhel/unit_tests/logger_test.py
```diff
@@ -73,9 +81,36 @@ def test_tools_opts_debug(tmpdir, read_std, is_py2):
...
+@pytest.mark.parametrize(
+    ("log_name", "path_exists", "is_dir"),
+    (
+        ("convert2rhel.log", True, True),
+        ("convert2rhel.log", True, True),
+        ("convert2rhel.log", False, False),
+        ("convert2rhel.log", True, False),
+    ),
+)
+def test_archive_old_logger_files(log_name, path_exists, is_dir, tmpdir, caplog):
```
abadger, line 101:
```suggestion
@pytest.mark.parametrize(
    ("log_name", "path_exists"),
    (
        ("convert2rhel.log", True),
        ("convert2rhel.log", False),
    ),
)
def test_archive_old_logger_files(log_name, path_exists, tmpdir, caplog):
```

Small nitpick, is_dir is no longer used (because we are no longer checking in the code whether archive is a directory.)

Other than that, this looks good!

abadger (dismissed):

convert2rhel/unit_tests/main_test.py
```diff
@@ -19,6 +19,12 @@
 import os
 import unittest
 
+from unittest import mock
```
abadger, line 22:
```suggestion
import sys
import unittest

if sys.version_info[:2] <= (2, 7):
    import mock  # pylint: disable=import-error
else:
    from unittest import mock  # pylint: disable=no-name-in-module
```

r0x0d:
@abadger Hopefully fixed all the tests and errors from CI. Could you do another quick review just in case I missed something out?

abadger (approved):
Looks good!

r0x0d:
@SpyTec and @bocekm already squashed the commits into a single commit.

abadger (approved):

convert2rhel/unit_tests/logger_test.py
```diff
@@ -73,9 +73,37 @@ def test_tools_opts_debug(tmpdir, read_std, is_py2):
...
+
+@pytest.mark.parametrize(
+    ("log_name", "path_exists"),
+    (
+        ("convert2rhel.log", True),
+        ("convert2rhel.log", True),
+        ("convert2rhel.log", False),
+        ("convert2rhel.log", True),
+    ),
+)
```
SpyTec, line 92: Since 3 of the runs are identical I'm guessing each test has side effects that affect the other runs. Ideally we'd have two parameterized tests here, one for path exists and one where it doesn't. And to test that multiple runs create archives I'd instead have a loop within the test itself

```suggestion
@pytest.mark.parametrize(
    ("log_name", "path_exists"),
    (
        ("convert2rhel.log", True),
        ("convert2rhel.log", False),
    ),
)
```

SpyTec (changes requested):
Overall code looks good! One minor thing about the test

SpyTec (approved):

bocekm:
I'd prefer the timestamps to be human readable, e.g. like this https://github.com/oamg/convert2rhel/blob/main/convert2rhel/breadcrumbs.py#L115 but with additional seconds.
I know I'm late to the party, so if nobody volunteers and I don't find the time for it, it's ok too.

PR #360, opened 2021-11-03
Add support for podman to the Makefile

abadger:
Docker doesn't work everywhere because the version of docker sometimes
needs cgroups v1 but the system is using cgroups v2.  In those places,
podman seems to work fine.

* Give users the ability to specify what docker command to use (ie:
  make install DOCKER=podman)
* Autodetect in the Makefile whether there's a working podman or docker
  if the user did not specify one.  Note that this uses some GNU make
  extensions ($warning).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,18 @@ VENV ?= .venv3
 PRE_COMMIT ?= pre-commit
 SHOW_CAPTURE ?= no
 
+# Let the user specify DOCKER at the CLI, otherwise try to autodetect a working podman or docker
+ifndef DOCKER
+  DOCKER := $(shell podman run alpine echo podman 2> /dev/null)
+  ifndef DOCKER
+    DOCKER := $(shell docker run alpine echo docker 2> /dev/null)
+    ifndef DOCKER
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,6 +19,18 @@ VENV ?= .venv3
 PRE_COMMIT ?= pre-commit
 SHOW_CAPTURE ?= no
 
+# Let the user specify DOCKER at the CLI, otherwise try to autodetect a working podman or docker
+ifndef DOCKER
+  DOCKER := $(shell podman run alpine echo podman 2> /dev/null)
+  ifndef DOCKER
+    DOCKER := $(shell docker run alpine echo docker 2> /dev/null)
+    ifndef DOCKER
+      DUMMY := $(warning Many of the make targets require a working podman or docker.  Please install one of those and check that `poodman run alpine echo hello` or `docker run alpine echo hello` work)
```
SpyTec, line 28:
```suggestion
      DUMMY := $(warning Many of the make targets require a working podman or docker.  Please install one of those and check that `podman run alpine echo hello` or `docker run alpine echo hello` work)
```

SpyTec (dismissed):
Small typo otherwise fine!

Makefile
abadger: Thanks @SpyTec !

SpyTec (approved):

PR #368, opened 2021-11-16
Fix tests after RHEL8.5 release

kokesak:
Inside the kmod integration test we call `modprobe -r` command which can sometimes remove unused dependency modules. This can result in multi-line output string thus the assert in the test can fail. This PR replaces the output for return-code assert.


kokesak:
At this moment only C8 test for satellite fails due to missing repos on the satellite server. MR to fix this: https://gitlab.cee.redhat.com/satellite/sat-infra/-/merge_requests/227

mkluson (approved):

Issue #366, opened 2021-11-10
RHEL 8.5 doesn't work


T5wxnxWgbJF6:
I tried convert from a centos 8.4 to a rhel 8.4. it looks like he now tries to update to rhel 8.5 (which was released yesterday) and fails with the following error:

`('[%s] %s - %s', '11/10/2021 10:53:00', 'DEBUG', "Calling command 'yum install -y --setopt=module_platform_id=platform:el8 /usr/share/convert2rhel/subscription-manager/subscription-manager-1.28.21-3.el8.x86_64.rpm /usr/share/convert2rhel/subs
cription-manager/subscription-manager-rhsm-certificates-1.28.21-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/python3-subscription-manager-rhsm-1.28.21-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/dnf-plugin-s
ubscription-manager-1.28.21-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/python3-syspurpose-1.28.21-3.el8.x86_64.rpm'")
Failed loading plugin "spacewalk": No module named 'up2date_client'
Last metadata expiration check: 0:34:02 ago on Wed Nov 10 10:18:58 2021.
Error:
 Problem 1: conflicting requests
  - nothing provides python3-cloud-what = 1.28.21-3.el8 needed by subscription-manager-1.28.21-3.el8.x86_64
 Problem 2: conflicting requests
  - nothing provides python3-cloud-what = 1.28.21-3.el8 needed by python3-subscription-manager-rhsm-1.28.21-3.el8.x86_64
(try to add '--skip-broken' to skip uninstallable packages)
CRITICAL - Failed to install subscription-manager packages. See the above yum output for details.
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 91, in main
    pre_ponr_conversion()
  File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 182, in pre_ponr_conversion
    subscription.replace_subscription_manager()
  File "/usr/lib/python3.6/site-packages/convert2rhel/subscription.py", line 204, in replace_subscription_manager
    install_rhel_subscription_manager()
  File "/usr/lib/python3.6/site-packages/convert2rhel/subscription.py", line 246, in install_rhel_subscription_manager
[...]

PR #370, opened 2021-11-19
✨ Updated Versions of Ansible Preflight Check

nwerker:
Updated compability of Ansible Preflight Check to Support CentOS / Oracle Linux 8.5 since CentOS 8.5 was recently released. I tested the playbook and it worked perfectly with 8.5, even fixed some bugs that are currently open as issue, for instance #366.


bocekm (approved):

bocekm:
Thank you, @nwerker!

PR #371, opened 2021-11-22
Allow OL testing

kokesak:
Allow CI to run on OracleLinux images


kokesak:
The packit testing is blocked for security reasons at this moment.

.packit.yaml
```diff
@@ -45,7 +45,9 @@ jobs:
       epel-8-x86_64:
         distros: [centos-8]
       # TODO enable when OL7/8 images are working in artemis
```
kokesak, line 47: Remove this `todo` comment

tests/integration/tier1/remove-excluded-pkgs/main.fmf
```diff
@@ -1,10 +1,6 @@
 summary: remove-excluded-pkgs
 
 tier: 0
-adjust:
-  enabled: false
-  when: >
-    distro != centos-8
 
 /good:
```
mkluson, line 5: Can we remove this sub test, since there is only one test case?

tests/integration/tier0/resolve-broken-ol8-rollback/main.fmf
```diff
@@ -1,13 +1,6 @@
...
 tier: 0
-adjust:
-  enabled: false
-  when: >
-    distro != centos-7 and
-    distro != centos-8 and
-    distro != oracle-7 and
-    distro != oracle-8
 
 /good:
```
mkluson, line 5: Can we remove this sub test, since there is only one test case?

tests/integration/tier0/inhibit-if-oracle-system-uses-not-standard-kernel/test_oracle_bad_kernel.py
```diff
@@ -9,21 +9,6 @@
 from envparse import env
 
 
-@pytest.mark.good_tests
```
mkluson, line 12: Why do you remove this case?

tests/integration/tier1/remove-excluded-pkgs/main.fmf
kokesak: Good catch, I removed it.

tests/integration/tier0/resolve-broken-ol8-rollback/main.fmf
kokesak: Good catch, I removed it.

tests/integration/tier0/inhibit-if-oracle-system-uses-not-standard-kernel/test_oracle_bad_kernel.py
kokesak: The reason is that the "good" case does the conversion with suitable kernel and nothing else. It's the same test case as the basic ones in the `tier1`.

mkluson (approved):

PR #373, opened 2021-11-24
Avoid mutable types as default args.

abadger:
This change removes two places where we use a set as a default value.
Mutable containers should not be used as default arguments as they are
created when the function is defined (basically, the first time the
module is imported) and then that one instance is used anytime the
method is invoked without that argument specified.  This means that
changes to the container will persist to the next time the function is
invoked.

<strike>This change seems to fix a bug we're seeing on oracle7 where we assemble
a list of packages to downgrade that includes gcc & gcc-c++ but then we find
that systemd is a protected package.  At that point, the list of
packages to downgrade seems to revert to just systemd and we build up
the list again.  This time, when we see that gcc is going to be
downgraded but gcc-c++ requires it, we do not decide to downgrade
gcc-c++.  I haven't tracked down precisely why this is but I suspect that
gcc-c++ is getting placed in the default set for the `excluded_pkgs`
argument so that when we iterate over the pakcages the second time we
never re-add gcc-c++ to the list of pakcages to downgrade.</strike>

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -144,7 +144,7 @@ def call_yum_cmd(
     return stdout, returncode
 
 
-def get_problematic_pkgs(output, excluded_pkgs=set()):
+def get_problematic_pkgs(output, excluded_pkgs=frozenset()):
     """Parse the YUM/DNF output to find what packages are causing a transaction failure."""
     loggerinst.info("Checking for problematic packages")
     problematic_pkgs = {
@@ -201,7 +201,7 @@ def get_pkgs_to_distro_sync(problematic_pkgs):
     )
 
 
-def resolve_dep_errors(output, pkgs=set()):
+def resolve_dep_errors(output, pkgs=frozenset()):
     """Recursive function. If there are dependency errors in the yum output,
     try to resolve them by yum downgrades.
     """


bocekm (approved):
Thanks, @abadger!

SpyTec (approved):

kokesak:
I run the test suite manually:
- Centos8: http://artifacts.osci.redhat.com/testing-farm/1f7d3440-c61b-4c75-9593-ec6d6eeaa0ea/
  - There was a test issue in `channged_yum_conf` test, so I rerun the test with a fix:
  http://artifacts.osci.redhat.com/testing-farm/c25540df-80e8-46e1-a463-ad216f57a61f/ 

-  Centos7: http://artifacts.osci.redhat.com/testing-farm/54dd04b6-8099-4a09-bf45-76fd0f264e34/
   - rerun of the failed (due to infrastructure issues):
   http://artifacts.osci.redhat.com/testing-farm/1dea961e-6a55-4f73-8e32-550cfdd36550/
   http://artifacts.osci.redhat.com/testing-farm/ef80b04b-5323-40a9-aaec-864ce22518b4/
 
There is still (known) issue with `satellite`, this one can be waived.

I also run the tests on the OL, however not all tests are completely functional yet. I just reviewed them manually
- OL7: http://artifacts.osci.redhat.com/testing-farm/5ab7da71-c832-4a2c-b836-e27423dd4e60/
- OL8: http://artifacts.osci.redhat.com/testing-farm/59657cf8-8014-4b7c-8f4a-3f86bda84804/

Overall, LGTM we can merge this PR.

abadger:
Thanks @kokesak @bocekm and @SpyTec for the reviews and testing!  Now merged.

PR #374, opened 2021-11-25
Fix detection of gcc-c++ and other package names.

abadger:
The regular expression that was being used to parse package names out of
yum output was missing gcc-c++, NetworkManager, and other package names.
Using a simpler regex and a function that parses the package names from
the rest of the NVR should fix that.

This fixes a user visible problem where packages like gcc-c++ can cause
convert2rhel to fail to downgrade packages.  This is because gcc-c++
would be missed by the regex and thus we'd be unable to figure out that
we need to downgrade gcc-c++ when at the same time as we downgrade gcc.

Resolves [OAMG-6136](https://issues.redhat.com/browse/OAMG-6136).

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -36,6 +36,28 @@ MAX_YUM_CMD_CALLS = 3
 _VERSIONLOCK_FILE_PATH = "/etc/yum/pluginconf.d/versionlock.list"  # This file is used by the dnf plugin as well
 versionlock_file = utils.RestorableFile(_VERSIONLOCK_FILE_PATH)  # pylint: disable=C0103
 
+#
+# Regular expressions used to find package names in yum output
+#
+
+# This regex finds package NEVRs + arch (name epoch version release and
+# architechture) in a string.  It separates the epoch from the rest of the
+# NEVR, leaving us to parse the name from the remaining NVR.
+# Note that the regex requires at least two dashes but the NVR can contain
+# more than that.  For instance: gcc-c++-4.8.5-44.0.3.el7.x86_64
+PKG_NEVR = r"\b([0-9]+:)?(\S+-\S+-\S+)\b"
+
+# It would be better to construct this dynamically but we don't have lru_cache
+# in Python-2.6 and modifying a global after your program initializes isn't a
+# good idea.
+_KNOWN_PKG_MESSAGE_KEYS = (
+    "%s",
+    "Error: Package: %s",
+    "multilib versions: %s",
+    "problem with installed package %s",
+)
+_PKG_REGEX_CACHE = dict((k, re.compile(k % PKG_NEVR, re.MULTILINE)) for k in _KNOWN_PKG_MESSAGE_KEYS)
+
 
 class PkgWFingerprint(object):
     """Tuple-like storage for a package object and a fingerprint with which the package was signed."""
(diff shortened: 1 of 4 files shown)


convert2rhel/pkghandler.py
```diff
@@ -36,6 +36,28 @@
...
+#
+# Regular expressions used to find package names in yum output
+#
+
+# This regex finds package NEVRs + arch (name epoch version release and
+# architechture) in a string.  It separates the epoch from the rest of the
+# NEVR, leaving us to parse the name from the remaining NVR.
+# Note that the regex requires at least two dashes but the NVR can contain
+# more than that.  For instance: gcc-c++-4.8.5-44.0.3.el7.x86_64
+PKG_NEVR = r"\b([0-9]+:)?(\S+-\S+-\S+)\b"
```
SpyTec, line 48: Since first group isn't used should we make it non-capturing?
```suggestion
PKG_NEVR = r"\b(?:[0-9]+:)?(\S+-\S+-\S+)\b"
```

SpyTec: This RegEx also changes the group from the system name to the whole name. Currently we get a group `systemd-python` but with this change we'd get `systemd-python-219-78.0.7.el7_9.3.x86_64`

SpyTec: We could also do this so we do not need to do split later
```suggestion
PKG_NEVR = r"\b(?:[0-9]+:)?(\S+)-(\S+)-(\S+)\b"
```

abadger: Yeah, I think that would work and then the function can be reduced to doing less work.  I have time this morning so let me try to whip that up.

abadger: Updated the code to use almost this RegEx.  I modified it slightly so that it will also capture the epoch in case we need to retrieve the epoch, version, and release in the future.

abadger:
Updated to use @SpyTec 's regex and added some simple unittests for the new function.  Probably should add to the unittests for `get_problematic_pkgs()` as well so that package names like gcc-c++ and NetworkManager are tested there.

bocekm:
The unit test can test with:
```
gcc-c++
NetworkManager
ImageMagick-c++
devtoolset-1.1-libstdc++-devel
389-ds-base
```

abadger:
Added those packages to the unittest for `find_pkg_name()`.  Restructured the unittest for `get_problematic_pkgs()` and will construct some output that uses those pkg names to add to that test.

abadger:
Okay, unless CI finds some new problems, I'm done with this PR now.  The latest commit just adds to the unittests.  One interesting note is that all of the package names @bocekm mentioned here: https://github.com/oamg/convert2rhel/pull/374#issuecomment-980310434 now succeed when parsed from the `Error: Package:` string but they all fail when parsed as part of `Requires:`.  (Fixing that will be tracked by #378)

bocekm (approved):
Tested manually on OL7 with gcc-c++ installed:
`Found packages causing dependency errors: set([u'nss-softokn', u'gcc', u'gcc-c++', u'libstdc++-devel'])`

bocekm:
@kokesak verified the change as well.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1453,3 +1450,176 @@ def test_fix_default_kernel_with_no_incorrect_kernel(self):
...
+        # misparses a Requires in the test data (It turns
+        # python2_hawkey into python2).  Testing it in its own function
+        # for now.
+        # (
+        #     YUM_REQUIRES_ERROR,
+        #     {
+        #         "protected": set(),
+        #         "errors": set(("gcc-c++", "libreport-anaconda", "abrt-cli", "mod_ldap", "redhat-lsb-trialuse")),
+        #         "multilib": set(),
+        #         "required": set(("gcc", "libreport-plugin-rhtsupport", "python2-hawkey", "redhat-lsb-core")),
```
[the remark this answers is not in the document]
abadger: <nod> It would depend on what the actual packages are.  You would be correct if there are two separate packages, python2-hawkey and python2_hawkey.  In my mind I was thinking that the package is python2-hawkey and it has a virtual provide/capability named python2_hawkey.  (Like this in the spec file:

```
Name: python2-hawkey
# [..]
Provides: python2_hawkey = %{version}-%{release}
```
This situation would arise when a package has been renamed or the packager thinks people would sometimes use the alternate name for some other reason.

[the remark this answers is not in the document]
bocekm: I was referring to the test yum output https://github.com/abadger/convert2rhel/blob/d24e424da77636af37e38881d797f5326da8f383/convert2rhel/unit_tests/pkghandler_test.py#L1381 where we have `Requires: python2_hawkey >= 0.7.0`.

Such `Provides:` in reality (in hawkey spec file) does not exist. We came up with the `Requires: python2_hawkey` for the sake of testing. However we failed to write the test properly as you revealed. We should have added 
`self.assertIn("python2_hawkey", error_pkgs["required"])` to the test in
https://github.com/oamg/convert2rhel/pull/325/files#diff-c3ce6b0e1e7ddae0678f9eb7c71f9c5f8aecbfd5b7db1deb7d8947ccf1492a08R314.

PR #375, opened 2021-11-29
Archived logfile name to include UTC timestamp

r0x0d:
This change provides a more human-readable string to append in the archive log instead of the unix-timestamp previously used in the archive log file names.

Based on the @bocekm comment: https://github.com/oamg/convert2rhel/pull/359#issuecomment-980469344

### DIFF
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -30,6 +30,8 @@ import os
 import shutil
 import sys
 
+from time import gmtime, strftime
+
 from convert2rhel.toolopts import tool_opts
 from convert2rhel.utils import format_msg_with_datetime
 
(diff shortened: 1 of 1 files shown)


convert2rhel/logger.py
```diff
@@ -113,15 +115,18 @@ def archive_old_logger_files(log_name, log_dir):
 
     stat = os.stat(current_log_file)
 
-    # Get the last time the file was modified
-    last_modified_at = stat.st_mtime
+    # Convert it to a local time
+    last_modified_at = localtime(int(stat.st_mtime))
+
+    # Format time to a human-readable format
+    formatted_time = strftime("%Y%m%dT%H%M%S%Z", last_modified_at)
```
bocekm, line 122:
```suggestion
    # Get the last modified time in UTC
    last_modified_at = time.gmtime(stat.st_mtime)

    # Format time to a human-readable format
    formatted_time = strftime("%Y%m%dT%H%M%SZ", last_modified_at)
```

 Looking at the https://en.wikipedia.org/wiki/ISO_8601 the above format complies.

bocekm (changes requested):
I'm getting: `20211129T120624CET`

I'd definitely prefer UTC/Zulu time instead of local.

I can see that [in the breadcrumbs file we're finishing the string](https://github.com/oamg/convert2rhel/blob/main/convert2rhel/breadcrumbs.py#L115) with `%Z` even though it could have also been just simply `Z` for Zulu. With that I'd say we should make the format in our breadcrumbs file the same as in LEAPP: https://github.com/oamg/leapp-repository/pull/719/files#diff-fb57b9cae633d0fcde3ed254e6a76573bf2537c17a3209fa9b224bc127f92cedR24. But that would be for a separate PR.

convert2rhel/logger.py
```diff
@@ -113,15 +115,18 @@ def archive_old_logger_files(log_name, log_dir):
 
     stat = os.stat(current_log_file)
 
-    # Get the last time the file was modified
-    last_modified_at = stat.st_mtime
+    # Get the last modified time in UTC
+    last_modified_at = gmtime(stat.st_mtime)
+
+    # Format time to a human-readable format
+    formatted_time = strftime("%Y%m%dT%H%M%SZ", last_modified_at)
```
r0x0d, line 122: @bocekm you think we should include the extra milliseconds in the format string too? 
```suggestion
    formatted_time = strftime("%Y%m%dT%H%M%S%fZ", last_modified_at)
```

r0x0d (commented):
Reviewed

convert2rhel/logger.py
bocekm: No need for ms there :)

bocekm (approved):

PR #376, opened 2021-11-29
Testing on behalf of SpyTec

BluABK:
Test that tries to prove that actions don't run for outside contributors.

Edit and PR made on behalf of SpyTec.


bocekm:
Thanks, @BluABK! The test served it's purpose, closing.

PR #380, opened 2021-12-01
Added more information to the docstring about releasever

Andrew-ang9:
I went in and added to the docstring giving more information on what the releasever does. I've gotten the mock up for what is to be used.

Fixes for OAMG-4673

### DIFF
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -87,7 +87,7 @@ class SystemInfo(object):
         self.default_rhsm_repoids = None
         # List of repositories enabled through subscription-manager
         self.submgr_enabled_repos = []
-        # Gets the release version of the Linux based system, see the docstring in the “get_releasever()”
+        # Value to use for substituting the $releasever variable in the url of RHEL repositories
         self.releasever = None
         # List of kmods to not inhbit the conversion upon when detected as not available in RHEL
         self.kmods_to_ignore = []
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -87,6 +87,7 @@ class SystemInfo(object):
         self.default_rhsm_repoids = None
         # List of repositories enabled through subscription-manager
         self.submgr_enabled_repos = []
+        # Gets the release version of the Linux based system, see the docstring in the “get_releasever()”
         self.releasever = None
         # List of kmods to not inhbit the conversion upon when detected as not available in RHEL
         self.kmods_to_ignore = []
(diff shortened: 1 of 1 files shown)


convert2rhel/systeminfo.py
```diff
@@ -87,6 +87,7 @@ def __init__(self):
         self.default_rhsm_repoids = None
         # List of repositories enabled through subscription-manager
         self.submgr_enabled_repos = []
+        # Gets the release version of the Linux based system, see the docstring in the “get_releasever()”
```
abadger, line 90:
```suggestion
        # Holds the release version of the Linux based system, see the docstring in the “get_releasever()”
```

I would say holds since it's a variable.  I'd use `get` for a description of a function.  (variables are generally nouns and functions are verbs)

convert2rhel/systeminfo.py
```diff
@@ -204,7 +205,13 @@ def _get_repofile_pkgs(self):
...
 
     def _get_releasever(self):
-        """Get the release version to be passed to yum through --releasever.
+        """Releasever is used to referance the Version of Oracle Linux or Centos
+        to a variable to be converted to when going through the conversion that will
+        be put in a repo baseurl. By default the RELEASE_VER_MAPPING is used but if it
+        the var can’t be found then the releasever will be used, to find the version
+        specific config files in the convert2rhel/convert2rhel/data .
+
+        Get the release version to be passed to yum through --releasever.
```
abadger, line 214: * Similar to a git commit message, docstrings usually have a one line summary at the top, then a blank line, and then a longer explanation.  So I'd put Line 214 back at the top.

* I think the releasever is actually the version of RHEL that we're converting too rather than the release of Oracle or CentOS that we're converting from.

* The language around whether RELEASE_VER_MAPPING or the config file takes precedence is a bit confusing.  I think what it should say is:
> By default, the value is found by mapping from the current distro's version to a compatible version of RHEL via the RELEASE_VER_MAPPING.  This can be overridden by the user by specifying it in the config file.  The version specific config files are located in convert2rhel/convert2rhel/data.

* Note that some of that was taken from the existing docstring (the two lines after this comment).  It's probably best to merge the information together like this rather than having two separate sections that say almost the same thing.

bocekm (changes requested):

convert2rhel/systeminfo.py
Andrew-ang9: Ok I have pushed a edited version of the docstring to take in account this comment!

r0x0d (approved):

SpyTec (approved):

bocekm (approved):

PR #382, opened 2021-12-02
Change timestamp format for breadcrumbs

r0x0d:
This PR changes how the timestamp output is presented in the
/etc/migrations-results is presented on the file.

Before, it used `gmlocal()` which outputs the current time in GMT, now, this
file will have `utcnow()` timestamp output to be more equal on how leapp do
their breadcrumbs.

with gmlocal():
            "activity_started": "20211201T1810GMT",
            "activity_ended": "20211201T1811GMT",

with utcnow():
            "activity_started": "2021-12-02T11:56:07.425477Z",
            "activity_ended": "2021-12-02T11:56:59.478299Z",

Reference ticket: https://issues.redhat.com/browse/OAMG-6139

PR #383, opened 2021-12-02
Fix minor typos in breadcrumbs.py

r0x0d:
Found some typos on `breadcrumbs.py` file. Didn't want to commit them while fixing another problem for #382.

### DIFF
--- a/convert2rhel/breadcrumbs.py
+++ b/convert2rhel/breadcrumbs.py
@@ -69,7 +69,7 @@ class Breadcrumbs(object):
         self._set_started()
 
     def finish_success(self):
-        """Set data accessible after sucessfull conversion and generate the JSON file."""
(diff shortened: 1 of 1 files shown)


hosekadam (approved):
LGTM

convert2rhel/breadcrumbs.py
```diff
@@ -69,7 +69,7 @@ def collect_early_data(self):
         self._set_started()
 
     def finish_success(self):
-        """Set data accessible after sucessfull conversion and generate the JSON file."""
+        """Set data accessible after successfully conversion and generate the JSON file."""
```
SpyTec, line 72: This should be `Set data accessible after a successful conversion` or `Set data accessible after successfully converting`. Now it's a mix of a past and future action

PR #385, opened 2021-12-04
added changes to auto add subs when there is only one

Andrew-ang9:
Changes what happens to when the user only has one subscription instead of asking the user to choose that one subscription, it automatically adds that one subscription and lets the user know that was done.

Also has the unit test for that addition, testing both when the user only has one subscription and when the user has more then one subscription.

Ticket reference: https://issues.redhat.com/browse/OAMG-4479


convert2rhel/subscription.py
```diff
@@ -281,18 +281,30 @@ def attach_subscription():
...
+
+        else:
+            # Let the user choose the subscription appropriate for the conversion
+            loggerinst.info("Manually select subscription appropriate for the conversion")
+
+            print_avail_subs(subs_list)
+            sub_num = utils.let_user_choose_item(len(subs_list), "subscription")
+            pool = "--pool " + subs_list[sub_num].pool_id
+            tool_opts.pool = pool
+            loggerinst.info("Attaching subscription with pool ID %s to the system ..." % subs_list[sub_num].pool_id)
```
abadger, line 307: Just a suggestion for a small cleanup:  the code for setting the pool is almost a duplicate in both branches of the if statement.  it should be possible to move that outside the if statement so you only have to maintain one copy of it like this:

``` python 
        if len(subs_list) == 1:
            sub_num = 1
            # Remove pool = "--pool" [..]
             # and tool_opts.pool = [..]
            loggerinst.info([...])
        else:
            [..]
            # Remove pool = "--pool" [..]
             # and tool_opts.pool = [..]

     # Setting pool and tool_opts can now be done here, outside of the if-statement:
     pool = "--pool " + subs_list[sub_num].pool_id
     tool_opts.pool = pool
     _, ret_code = utils.run_subprocess("subscription-manager attach %s" % pool)
```

Andrew-ang9: gotcha, I will add these changes and re-push!

SpyTec (changes requested):

convert2rhel/subscription.py
```diff
@@ -281,19 +281,28 @@ def attach_subscription():
...
 
-        print_avail_subs(subs_list)
-        sub_num = utils.let_user_choose_item(len(subs_list), "subscription")
-        pool = "--pool " + subs_list[sub_num].pool_id
-        tool_opts.pool = pool
-        loggerinst.info("Attaching subscription with pool ID %s to the system ..." % subs_list[sub_num].pool_id)
+        if len(subs_list) == 1:
+            sub_num = 0
+            loggerinst.info(
+                "This %s was the only subscription that you have, It will automaticlly be attached for the conversion."
```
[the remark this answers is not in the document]
Andrew-ang9: sounds good I'll add these changes and git them pushed!

convert2rhel/subscription.py
```diff
@@ -281,19 +281,28 @@ def attach_subscription():
...
         if len(subs_list) == 0:
             loggerinst.warning("No subscription available for the conversion.")
             return False
 
-        print_avail_subs(subs_list)
-        sub_num = utils.let_user_choose_item(len(subs_list), "subscription")
-        pool = "--pool " + subs_list[sub_num].pool_id
-        tool_opts.pool = pool
-        loggerinst.info("Attaching subscription with pool ID %s to the system ..." % subs_list[sub_num].pool_id)
+        if len(subs_list) == 1:
```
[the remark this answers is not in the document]
Andrew-ang9: sounds good I'll add these changes and git them pushed!

convert2rhel/subscription.py
```diff
@@ -281,19 +281,28 @@ def attach_subscription():
...
+                "This %s was the only subscription that you have, It will automaticlly be attached for the conversion."
+                % subs_list[0].pool_id
+            )
 
+        else:
+            # Let the user choose the subscription appropriate for the conversion
+            loggerinst.info("Manually select subscription appropriate for the conversion")
+            print_avail_subs(subs_list)
+            sub_num = utils.let_user_choose_item(len(subs_list), "subscription")
+            loggerinst.info("Attaching subscription with pool ID %s to the system ..." % subs_list[sub_num].pool_id)
```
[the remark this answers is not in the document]
Andrew-ang9: sounds good I'll add these changes and git them pushed!

convert2rhel/subscription.py
```diff
@@ -281,19 +281,28 @@ def attach_subscription():
...
 
-        print_avail_subs(subs_list)
-        sub_num = utils.let_user_choose_item(len(subs_list), "subscription")
-        pool = "--pool " + subs_list[sub_num].pool_id
-        tool_opts.pool = pool
-        loggerinst.info("Attaching subscription with pool ID %s to the system ..." % subs_list[sub_num].pool_id)
+        elif len(subs_list) == 1:
+            sub_num = 0
+            loggerinst.info(
+                " %s is the only subscription avalible, it will automaticlly be selected for the conversion."
```
r0x0d, line 293:
```suggestion
                "%s is the only subscription avalible, it will automaticlly be selected for the conversion."
```

r0x0d, line 293:
```suggestion
                "%s is the only subscription available, it will automatically be selected for the conversion."
```

r0x0d, line 293:
```suggestion
                "%s is the only subscription available, it will automatically be selected for the conversion."
```

r0x0d (changes requested):

convert2rhel/subscription.py
Andrew-ang9: Thanks for pointing out theses typo's I will correct them and push it!

r0x0d (dismissed):

convert2rhel/subscription.py
```diff
@@ -281,19 +281,28 @@ def attach_subscription():
...
 
+        else:
+            # Let the user choose the subscription appropriate for the conversion
+            loggerinst.info("Manually select subscription appropriate for the conversion")
+            print_avail_subs(subs_list)
+            sub_num = utils.let_user_choose_item(len(subs_list), "subscription")
+            loggerinst.info("Attaching subscription with pool ID %s to the system ..." % subs_list[sub_num].pool_id)
+
+    pool = "--pool " + subs_list[sub_num].pool_id
+    tool_opts.pool = pool
```
SpyTec, line 305: Should be indented one more. Also throws an error if you try to convert.
```suggestion
        pool = "--pool " + subs_list[sub_num].pool_id
        tool_opts.pool = pool
```

Since no tests failed either due to the pool being incorrect/not working you should add another test to verify that the `tool_opts.pool` is set correctly if `tool_opts.auto_attach` or `tool_opts.pool` is set (it should throw an error but don't make assertions on this since it's related to the bug and not implementation). In this case it's better to write the test first, see if it fails, then indent to see if the tests pass

SpyTec (changes requested):

SpyTec (approved):
LGTM! Needs integration tests

SpyTec (approved):

tests/integration/tier0/check-user-response/test_user_response.py
```diff
@@ -52,3 +56,48 @@ def test_check_user_response_organization(convert2rhel):
...
+        c2r.expect_exact("Registering the system using subscription-manager ...")
+        if "oracle-7" in booted_os or "centos-7" in booted_os:
+            c2r.sendcontrol("c")
+            c2r.sendcontrol("d")
+        else:
+            c2r.sendcontrol("c")
+
+        assert c2r.expect("Rollback: RHSM-related actions") == 0
+
+        assert c2r.exitstatus != 0
```
r0x0d, line 79: This test is currently blocked because of a problem we have with the account `RHSM_SINGLE_SUB_USERNAME`. Once this is done, we can properly test that feature out. 

## More on the blocker: 

We need an account that only has 1 subscription attached to it, but, in every account, we created with the internal test tool we have, the account got another extra subscription to it. Since this is happening, we can't properly test this feature out and need to wait on this blocker to be fixed.

Currently we have an issue created about this for the internal service that handles this stuff and a ticket open for the IT to see if they can remove that extra subscription from our test account.

r0x0d: This issue is not happening anymore since we discovered a way of adding only one subscription to an account. I have tested it locally and this issue seems to be fixed.

tests/integration/tier0/check-user-response/test_user_response.py
```diff
@@ -1,8 +1,12 @@
+import platform
+
 from envparse import env
 
 
-def test_check_user_response_user_and_password(convert2rhel):
+booted_os = platform.platform()
```
kokesak, line 6: The `booted_os` variable is used in this test? If not then remove it please (the import of `platform` as well) :D

tests/integration/tier0/check-user-response/test_user_response.py
```diff
@@ -52,3 +56,20 @@ def test_check_user_response_organization(convert2rhel):
...
+        "-y --no-rpm-va --serverurl {} --username {} --password {} --debug".format(
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_SINGLE_SUB_USERNAME"),
+            env.str("RHSM_SINGLE_SUB_PASSWORD"),
+        )
+    ) as c2r:
+        c2r.expect(
+            f"{single_pool_id} is the only subscription available, it will automatically be selected for the conversion."
+        )
+        c2r.sendcontrol("c")
```
kokesak, line 73: Maybe I would use the `d` character so it's consistent in the whole test file.

kokesak (changes requested):
I had a small comments abut the test. Also I noticed there is like 4 commits, maybe it would be better to squash them into one.

tests/integration/tier0/check-user-response/test_user_response.py
r0x0d: done!

tests/integration/tier0/check-user-response/test_user_response.py
r0x0d: done!

danmyway (approved):
LGTM! 
Thank you!

kokesak (approved):

r0x0d:
> I had a small comments abut the test. Also I noticed there is like 4 commits, maybe it would be better to squash them into one.

We can perform the squash on merge right now, so I think we don't need to worry about it. But I will check with the other guys to be sure

convert2rhel/subscription.py
```diff
@@ -304,29 +304,36 @@ def attach_subscription():
     pool = ["subscription-manager", "attach"]
     if tool_opts.auto_attach:
         pool.append("--auto")
-        tool_opts.pool = "-a"
```
bocekm, line 307: I don't see any comment in this PR mentioning this change. But I agree with it. The `tool_opts.pool` is not used anywhere later on. It is a leftover after removing this functionality: https://github.com/oamg/convert2rhel/pull/240.

bocekm (approved):

convert2rhel/subscription.py
bocekm: Aha, I see it mentioned in a message of the commit I'm going to squash soon: https://github.com/oamg/convert2rhel/pull/385/commits/7312775d3734e9ebd5c0d021d1edb4ce442c23fd

> Previous to this change, if a pool had been specified on the
convert2rhel command line and attach_subscription() was called more than
once, the subscription-manager command line would be constructed
incorrectly.  It would have "--pool" added to the front of the pool id
each time attach-subscription was called leading to something like this:
"subscription-manager attach --pool --pool POOLID".

> To fix this, we now treat tool_opts.pool as read-only and do not modify
it on each call to attach_subscription().  This goes along with our
overall move to treat global singletons as read-only whenever possible.

PR #386, opened 2021-12-06
Set LANG env var to C the same as LC_ALL

bocekm:
Not setting the LANG was causing the RHSM registration to fail because subscription-manager considers LANG to have priority over LC_ALL. That goes against [POSIX which specifies that LC_ALL overrides LANG](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_02). So even though we were setting LC_ALL to C which should warrant only ASCII output, the output of subscription-manager contains non-ASCII characters.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2022854

### DIFF
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -586,11 +586,13 @@ def get_rpm_header(rpm_path, _open=open):
 
 
 def set_locale():
-    """Set the POSIX default locale for the main process as well as the child processes.
+    """Set the C locale, also known as the POSIX locale, for the main process as well as the child processes.
 
     The reason is to get predictable output from the executables we call, not influenced by non-default locale.
+    We need to be setting not only LC_ALL but LANG as well because subscription-manager considers LANG to have priority
+    over LC_ALL even though it goes against POSIX which specifies that LC_ALL overrides LANG.
     """
-    os.environ.update({"LC_ALL": "C"})
+    os.environ.update({"LC_ALL": "C", "LANG": "C"})
 
 
 changed_pkgs_control = ChangedRPMPackagesController()  # pylint: disable=C0103


abadger:
+1 from me.  One note about the commit message.  I think this is probably a
problem with the dmidecode package, not rhel in general.

abadger (dismissed):
The code looks correct to me!

+1

abadger:
Okay, this file in subscription-manager is the source of the bug: https://github.com/candlepin/subscription-manager/blob/main/src/subscription_manager/i18n.py

It seems like the file is intended to allow subscription-manager to change its locale on the fly when it is run as a server.  There are several problems with how that's being done... I'm not sure if it's the combination of those that produces problems or if any one of the issues would be enough to cause problems in one place or another.  For the specific bug we're running into here, I believe that the following are the salient points:

* The code sets the starting langauge from LANG, not from LC_ALL.  This goes against the POSIX spec since LC_ALL is supposed to override any other language setting.  (And GNU adds the LANGUAGE variable which I believe overrides even LC_ALL).
* The code in rhsmlib/facts/dmiinfo.py is localizing its logging output using subscription_manager.i18n.ugettext:  https://github.com/candlepin/subscription-manager/blob/main/src/rhsmlib/facts/dmiinfo.py#L128
  * This is the only rhsmlib/facts code to do so.  It may be that facts modules should not localize any of their output and it's a bug that this one does so.
  * The effect of this is that even though LC_ALL is set to the C locale which can only display ascii characters [1]_ there are messages being generated and attempted to be logged by dmiinfo.py which contain non-ascii characters because LANG has been set to a non-C locale.  When python sees that message, it raises an exception because the text cannot be transformed using only ascii.

I think that the right way for subscription-manager to deal with this is to have a strict separation between messages which are going to be displayed to the client by way of the server and messages which are displayed in the logs or at the terminal (and possibly a desktop GUI?  I see mention of glade files in the comments in subscription_manager.i18n which are data files which define GUIs in gtk).  The logs and terminal (and GUI) should use the standard method to set the locale and *not* change it while the program is running.  For the messages which need to respond to the client side language choices, subscription_manager's i18n machinery should instead create a cache of its own gettext objects that target each language that is requested.  These would be separate from the standard locale mechanism. When a client's request needs to be served, these gettext objects would be used to lookup the translated messages for the user on the remote system rather than the standard ones associated with the user's locale.

[1]_ Later versions of RHEL and Fedora have a C.utf-8 locale so that won't always be the case.  Later versions of python3 see that C is the specified locale and change the encoding to utf-8 instead of ascii which also avoids this problem.

That's probably more information than is necessary for our bug report, but it's nice to trace down exactly where the problem is.

kokesak:
The integration test with verification results is under #387, @bocekm you can merge them both.

bocekm:
@abadger, thanks for the investigation. Agreed, the POSIX is clear on that LC_ALL shall override LANG, while in the subscription-manager code they have it the opposite: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_02.

I'm going to update the commit message. Will you report a bug for the subscription-manager?

SpyTec (approved):

pirat89:
@abadger  thanks for detailed explanation! \o

PR #389, opened 2021-12-07
Port run_subprocess to take the command as a list instead of a string.

abadger:
Security hardening: change the run_subprocess() function to take the command as a list instead of a string.

This is complete and ready for review. @Andrew-ang9 and I have been working on this together.

Implements: https://issues.redhat.com/browse/OAMG-6183

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -105,7 +105,7 @@ def check_tainted_kmods():
         system76_io 16384 0 - Live 0x0000000000000000 (OE)  <<<<<< Tainted
         system76_acpi 16384 0 - Live 0x0000000000000000 (OE) <<<<< Tainted
     """
-    unsigned_modules, _ = run_subprocess("grep '(' /proc/modules")
+    unsigned_modules, _ = run_subprocess(["grep", "(", "/proc/modules"])
     module_names = "\n  ".join([mod.split(" ")[0] for mod in unsigned_modules.splitlines()])
     if unsigned_modules:
         logger.critical(
@@ -162,10 +162,10 @@ def check_custom_repos_are_valid():
     # Without clearing the metadata cache, the `yum makecache` command may return 0 (everything's ok) even when
     # the baseurl of a repository is not accessible. That would happen when the repository baseurl is changed but yum
     # still uses the previous baseurl stored in its cache.
-    call_yum_cmd(command="clean", args="metadata", print_output=False)
+    call_yum_cmd(command="clean", args=["metadata"], print_output=False)
 
     output, ret_code = call_yum_cmd(
-        command="makecache", args="-v --setopt=*.skip_if_unavailable=False", print_output=False
+        command="makecache", args=["-v", "--setopt=*.skip_if_unavailable=False"], print_output=False
     )
     if ret_code != 0:
         logger.critical(
@@ -210,10 +210,10 @@ def get_loaded_kmods():
     kernel modules in case of different kernel release
     """
     logger.debug("Getting a list of loaded kernel modules.")
-    lsmod_output, _ = run_subprocess("lsmod", print_output=False)
+    lsmod_output, _ = run_subprocess(["lsmod"], print_output=False)
     modules = re.findall(r"^(\w+)\s.+$", lsmod_output, flags=re.MULTILINE)[1:]
     return set(
-        _get_kmod_comparison_key(run_subprocess("modinfo -F filename %s" % module, print_output=False)[0])
+        _get_kmod_comparison_key(run_subprocess(["modinfo", "-F", "filename", module], print_output=False)[0])
         for module in modules
     )
 
@@ -236,20 +236,25 @@ def _get_kmod_comparison_key(path):
 
 def get_rhel_supported_kmods():
     """Return set of target RHEL supported kernel modules."""
-    repoquery_repoids_args = " ".join("--repoid " + repoid for repoid in system_info.get_enabled_rhel_repos())
+    basecmd = [
+        "repoquery",
+        "--releasever=%s" % system_info.releasever,
+    ]
+    if system_info.version.major == 8:
+        basecmd.append("--setopt=module_platform_id=platform:el8")
+
+    for repoid in system_info.get_enabled_rhel_repos():
+        basecmd.extend(("--repoid", repoid))
+
+    cmd = basecmd[:]
+    cmd.append("-f")
+    cmd.append("/lib/modules/*.ko*")
     # Without the release package installed, dnf can't determine the modularity
     #   platform ID.
-    setopt_arg = "--setopt=module_platform_id=platform:el8" if system_info.version.major == 8 else ""
     # get output of a command to get all packages which are the source
     # of kmods
-    kmod_pkgs_str, _ = run_subprocess(
-        ("repoquery --releasever={releasever} {setopt_arg} {repoids_args} -f /lib/modules/*.ko*").format(
-            releasever=system_info.releasever,
-            setopt_arg=setopt_arg,
-            repoids_args=repoquery_repoids_args,
-        ),
-        print_output=False,
-    )
+    kmod_pkgs_str, _ = run_subprocess(cmd, print_output=False)
+
     # from these packages we select only the latest one
     kmod_pkgs = get_most_recent_unique_kernel_pkgs(kmod_pkgs_str.rstrip("\n").split())
     if not kmod_pkgs:
@@ -264,16 +269,13 @@ def get_rhel_supported_kmods():
             "Comparing the loaded kernel modules with the modules available in the following RHEL"
             " kernel packages available in the enabled repositories:\n {0}".format("\n ".join(kmod_pkgs))
         )
+
     # querying obtained packages for files they produces
-    rhel_kmods_str, _ = run_subprocess(
-        ("repoquery --releasever={releasever} {setopt_arg} {repoids_args} -l {pkgs}").format(
-            releasever=system_info.releasever,
-            setopt_arg=setopt_arg,
-            repoids_args=repoquery_repoids_args,
-            pkgs=" ".join(kmod_pkgs),
-        ),
-        print_output=False,
-    )
+    cmd = basecmd[:]
+    cmd.append("-l")
+    cmd.extend(kmod_pkgs)
+    rhel_kmods_str, _ = run_subprocess(cmd, print_output=False)
+
     return get_rhel_kmods_keys(rhel_kmods_str)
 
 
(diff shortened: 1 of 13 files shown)


convert2rhel/checks.py
```diff
@@ -243,11 +243,14 @@ def get_rhel_supported_kmods():
...
-        ("repoquery --releasever={releasever} {setopt_arg} {repoids_args} -f /lib/modules/*.ko*").format(
-            releasever=system_info.releasever,
-            setopt_arg=setopt_arg,
-            repoids_args=repoquery_repoids_args,
-        ),
+        [
+            "repoquery",
+            "--releasever=%s" % system_info.releasever,
+            setopt_arg,
+            repoquery_repoids_args,
```
abadger, line 250: I think setopt_arg and reppoquery_repoids_args here need to be modified a bit:
* setopt_arg might be set to the empty string: https://github.com/oamg/convert2rhel/pull/389/files#diff-cf19bc90f65d9e26e1b6fa7dcd84a18dfba097f65706e211a4f4ac71b1b467f6R242  I'm not sure if repoquery will ignore that or attempt to do something with it that fails.  It would be better to only add setopt_arg if it exists.
* repoquery_repoids_args is being set to a string with spaces for each repository like this:
  `'--repoid fedora --repoid fedora-updates'`.  We need each of those to be separate entries in the list.

Given those constraints, I think we need to build up the command as a list outside of the run_subprocess() rather than trying to construct the list inside of the function call.  Something like we did in pkghandler: https://github.com/oamg/convert2rhel/pull/389/files#diff-361f5dc14483ec43e37529967acee396da5714dbf6562472ccece5db4bba5e0dR134

r0x0d:
@abadger, a quick question... Should this PR cover the same security hardening for: https://github.com/oamg/convert2rhel/blob/120ad4f6356f50e861acc15f826132c4c9d0e7e3/convert2rhel/utils.py#L160

As it seems, this command is not used in many places, only here: https://github.com/oamg/convert2rhel/blob/120ad4f6356f50e861acc15f826132c4c9d0e7e3/convert2rhel/utils.py#L452-L472

If this method could benefit from the same process, I guess it could be good to add the same mitigations for this method as well.

convert2rhel/pkghandler.py
```diff
@@ -124,12 +124,14 @@ def call_yum_cmd(
       file for the particular system OR passed through a CLI option by the user
 
     YUM/DNF typically expands the $releasever variable used in repofiles. However it fails to do so after we remove the
-    release packages (centos-release, oraclelinux-release, etc.) and before the redhat-release package is installed.
+    e packages (centos-release, oraclelinux-release, etc.) and before the redhat-release package is installed.
```
r0x0d, line 127:
```suggestion
    packages (centos-release, oraclelinux-release, etc.) and before the redhat-release package is installed.
```

convert2rhel/subscription.py
```diff
@@ -126,13 +126,13 @@ def register_system():
...
     loggerinst.info("Building subscription-manager command ... ")
-    registration_cmd = "subscription-manager register --force"
+    registration_cmd = ["subscription-manager", "register", "--force"]
     if tool_opts.activation_key:
         # Activation key has been passed
         # -> username/password not required
         # -> organization required
         loggerinst.info("    ... activation key detected: %s" % tool_opts.activation_key)
-        registration_cmd += ' --activationkey="%s"' % tool_opts.activation_key
+        registration_cmd.append("--activationkey=%s" % tool_opts.activation_key)
```
r0x0d, line 135: The parameter is missing the quotes for %s

```suggestion
        registration_cmd.append('--activationkey="%s"' % tool_opts.activation_key)
```

convert2rhel/subscription.py
```diff
@@ -153,7 +153,7 @@ def get_registration_cmd():
                 # Hint user for which username they need to enter pswd
                 loggerinst.info("Username: " + username)
             password = utils.prompt_user("Password: ", password=True)
-        registration_cmd += ' --username=%s --password="%s"' % (username, password)
+        registration_cmd.extend(["--username=%s" % username, "--password=%s" % password])
```
r0x0d, line 156:
```suggestion
        registration_cmd.extend(["--username=%s" % username, '--password="%s"' % password])
```

convert2rhel/subscription.py
```diff
@@ -167,16 +167,16 @@ def get_registration_cmd():
         org = utils.prompt_user("Organization: ")
     if "org" in locals():
         # TODO: test how this option works with org name with spaces
-        registration_cmd += " --org=%s" % org
+        registration_cmd.append("--org=%s" % org)
     if tool_opts.serverurl:
         loggerinst.debug("    ... using custom RHSM URL")
-        registration_cmd += ' --serverurl="%s"' % tool_opts.serverurl
+        registration_cmd.append("--serverurl=%s" % tool_opts.serverurl)
```
r0x0d, line 173:
```suggestion
        registration_cmd.append('--serverurl="%s"' % tool_opts.serverurl)
```

r0x0d (changes requested):
The changes look good! Just a minor comment on a few changes that looked different from the original ones. 

There is a `newfile.txt` in the PR, maybe it was created by mistake.

convert2rhel/subscription.py
abadger: I don't think quotes are needed.  When we pass in a string and then split it into arguments using `shlex.split()` we have to be careful that anything that could contain spaces is enclosed in quotes.  Otherwise `shlex.split()` will split it into two arguments.  When we pass a list of arguments, instead, no additional splitting will occur, each element of the list will be a single argument:
``` python
>>> import subprocess, shlex
>>> open('test case', 'w').close()
>>> # Oops, we forgot to quote the filename
>>> subprocess.Popen(shlex.split('cat test case')).wait()
cat: test: No such file or directory
cat: case: No such file or directory
1
>>> # This is how we have to construct the command as a string:
>>> subprocess.Popen(shlex.split('cat "test case"')).wait()
0
>>> # When we use a list, we don't have to quote:
>>> subprocess.Popen(['cat', 'test case']).wait()
0
```
Not having to quote is part of why this is a more secure API for this function.  It's easy to forget to quote something when constructing a string to pass in (especially when dealing with variables). The code will then work until someone constructs a string with a space in it (possibly maliciously).  Passing a list instead means that we do not have to remember to quote.

convert2rhel/utils.py
```diff
@@ -119,24 +119,22 @@ def restart_system():
...
-def run_subprocess(cmd="", print_cmd=True, print_output=True):
+def run_subprocess(cmd, print_cmd=True, print_output=True):
     """Call the passed command and optionally log the called command (print_cmd=True) and its
     output (print_output=True). Switching off printing the command can be useful in case it contains
     a password in plain text.
+
+    The cmd is specified as a list starting with the command and followed by a list of arguments.
+    Example: ["dnf", "repoquery", "kernel"]
     """
     if print_cmd:
```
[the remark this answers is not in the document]
abadger: If we do, then we probably want to check whether this is a string rather than checking if it is not a list to preserve duck-typing (subprocess.Popen should work with anything that is a Sequence [a tuple, for instance.])

A question, though:
* Are we guarding against str being passed in because str used to be passed in?
  * If so, we probably want to limit the time that we perform this check.
  * If not, what is the difference between this function and other functions which take lists and would error if given a string?

convert2rhel/subscription.py
bocekm: I agree that the quotes are not needed. Frankly I needed to try it out myself with the _=_ sign to verify it 🤫 The following Popen call:
```python
>>> subprocess.Popen(['subscription-manager', 'register', '--serverurl=host.example.com; rm file']).wait()
Unable to reach the server at host.example.com; rm file:443/subscription
69
```
... is the same as if running the following in shell:
```shell
# subscription-manager register "--serverurl=host.example.com; rm file"
Unable to reach the server at host.example.com; rm file:443/subscription
```

The following code construct would not make me think twice about the functionality. But that would change the command called and that's I believe not the intent of this PR.
```suggestion
        registration_cmd.extend(["--activationkey", "%s" % tool_opts.activation_key])
```

convert2rhel/utils.py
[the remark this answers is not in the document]
r0x0d: I think you're right, we should compare against `str` and not `list`, which makes more sense. 

Regarding the question, I thought about checking against this because of what you mentioned in your gist. Someone could forget to use `shlex.quote()` when passing an `str` to our `run_subprocess()` and thus, we would be in the same spot as of right now. 

Let me know what you think about that, it's just a suggestion (don't know either if I understood correctly what I read 😅)

convert2rhel/subscription.py
r0x0d: Oh, that's cool! Thought that we had to preserve the quotes for these cases (Imagined that this could cause some unreliable calls to subscription-manager). Either way, the other comments about quotes can be ignored safely then!

convert2rhel/checks.py
abadger: @Andrew-ang9 Could you take a look at fixing this one?

abadger:
@r0x0d I've incorporated your suggestions and modified  run_cmd_in_pty().  I'll let @Andrew-ang9 modify setopt_arg and reppoquery_repoids_args in checks.py and then this should be ready for another review.

convert2rhel/checks.py
Andrew-ang9: Yeah I can take care of that!

abadger:
> 

I think that all of the things you've mentioned have now been fixed.  Let me know if there's more problems!

(And you were right about newfile... I meant to delete it after we tested that andrew could push to my repo).

SpyTec (dismissed):

convert2rhel/pkghandler.py
```diff
@@ -158,10 +160,10 @@ def call_yum_cmd(
         repos_to_enable = system_info.get_enabled_rhel_repos()
 
     for repo in repos_to_enable:
-        cmd += " --enablerepo=%s" % repo
+        cmd.append("--enablerepo=%s" % repo)
 
     if args:
```
r0x0d, line 165: I think this check is not entirely necessary here, at the beginning we are already saying that if the `args is None`, then `args = []`.

What do you think, @abadger?

r0x0d (dismissed):

convert2rhel/pkghandler.py
abadger: I agree!  Removing it now.

abadger:
Sorry, I messed up my rebase.  Give me a moment to fix that.

abadger:
Alright, fixed.

r0x0d (approved):

SpyTec (approved):

kokesak:
@SpyTec yes, I was just checking it. I wil try to re-run it again. If it fails again I think it's ok to merge

abadger:
Thanks for the reviews @r0x0d @SpyTec and @kokesak !

PR #390, opened 2021-12-09
Prevent user from not selecting an response

r0x0d:
Add a loop to prevent the user from not responding to the question, this will cause
a loop of the same question being asked every time the user just hit enter
without answering anything to the tool.

Jira reference: https://issues.redhat.com/browse/OAMG-3143


convert2rhel/utils.py
```diff
@@ -247,11 +247,17 @@ def ask_to_continue():
...
-    loggerinst.info("\n")
+
+    while True:
+        if password:
+            response = getpass.getpass(color_question)
+        else:
+            response = moves.input(color_question)
+        loggerinst.info("\n")
+
+        if response != "" or len(response) > 0:
```
abadger, line 258: Is there a time when the empty string test and the length test would return different values?

Checking whether something is Falsey (like the following) is more pythonic but I don't know if there's a reason that you are avoiding it? 

``` python
if not response:
    break
```

r0x0d: Right! I think I just left this piece of code for when I was testing out. A simple 

```python
if response: 
   break
```

would work properly!

abadger:
I was struck by a worry: `prompt_user()` is a generic utility function.  So I worried that there might be a time when it was valid for the user to specify an empty string.  A quick grep doesn't show anything in our current code so that seems safe for now.  I'm not sure if we should address the possibility that a user supplied string should be allowed to be empty now or later, if we actually have code that can legitimately be empty.

Looking at the consumers of the code did make me see that the username and password can come from other places, than `prompt_user()`, though.  That makes me think that it might be important to validate the string is non-empty at the caller so that we can check the value from tool_opts as well as from user_prompt.  What do you think of that?

r0x0d:
> I was struck by a worry: `prompt_user()` is a generic utility function. So I worried that there might be a time when it was valid for the user to specify an empty string. A quick grep doesn't show anything in our current code so that seems safe for now. I'm not sure if we should address the possibility that a user supplied string should be allowed to be empty now or later, if we actually have code that can legitimately be empty.
> 
> Looking at the consumers of the code did make me see that the username and password can come from other places, than `prompt_user()`, though. That makes me think that it might be important to validate the string is non-empty at the caller so that we can check the value from tool_opts as well as from user_prompt. What do you think of that?

I think you're right in this one... It might be better to remove the loop from the `prompt_user()` function as it could be used by others callers and they might want it to be empty. Thought that only adding the loop there, would be sufficient to cover any cases in the future as well (Well, at least the cases we don't want the string to be empty).  

```python
while True:
   username = utils.prompt_user("Username: ")
   if username:
      break
```

convert2rhel/utils.py
```diff
@@ -246,12 +246,21 @@ def ask_to_continue():
 
 
 def prompt_user(question, password=False):
+    """Prompt the user with a question and return his response to the caller.
+
+    :param question: The question to prompt to the user.
+    :param password: If the question is for a password, then the output is blanked out.
```
r0x0d, line 252: @abadger did I make it right referencing the parameters this way?

convert2rhel/subscription.py
```diff
@@ -127,50 +127,40 @@ def get_registration_cmd():
...
+        username = tool_opts.username
+        while not username:
             username = utils.prompt_user("Username: ")
-        if tool_opts.password:
-            loggerinst.info("    ... password detected")
-            password = tool_opts.password
-        else:
-            if tool_opts.username:
-                # Hint user for which username they need to enter pswd
-                loggerinst.info("Username: " + username)
```
SpyTec, line 154: It would be preferential to keep the logs here whenever we detect that a value exists in toolopts. Likewise a log should be added if we detect an org in toolopts

The reason being that we're being explicit in the logic of what's going on. Rather than just logs in the beginning that we found specific values in the parsing of the CLI args.

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -105,6 +105,17 @@ class PromptUserMocked(unit_tests.MockFunction):
...
+    class PromptUserLoopMocked(unit_tests.MockFunction):
+        def __init__(self):
+            self.called = 0
+
+        def __call__(self, *args, **kwargs):
+            if self.called >= 1:
+                return "test"
+
+            self.called += 1
+            return ""
```
SpyTec, line 117: This doesn't do anything in the tests right now, and also stops keeping track if it's been called more than once. One way to test that it loops for each prompt is to keep a list of calls per argument. It should be enough to have a list for first argument like `"Organization: "` and make sure it fails on first call, making it easier to make assertions in the code.

Ideally that should be a separate test though just testing that it loops as it becomes very connected with the code. For the other tests we can use a `DumbCallableObject`

convert2rhel/utils.py
SpyTec: We don't use parameters too much in the code, but the style is correct! :+1:

SpyTec (changes requested):

r0x0d (commented):
Reviewed

SpyTec (approved):
LGTM. Just needs integration tests

tests/integration/tier0/check-user-response/test_user_response.py
```diff
@@ -0,0 +1,50 @@
...
+        c2r.expect_exact("Organization: ")
+        c2r.sendline()
+        substitute_org = "foo"
+        if c2r.expect_exact("Organization: ") == 0:
+            c2r.sendline(substitute_org)
+        else:
+            c2r.sendcontrol("d")
+        c2r.expect_exact("Registering the system using subscription-manager ...")
+        c2r.sendcontrol("d")
+    assert c2r.exitstatus != 0
```
r0x0d, line 50: Hey, @danmyway, just a question... It's not easier to separate this into another test case? like 

```python
def test_check_user_response_user_and_password():
   ...
   
def test_check_user_response_organization(): 
   ...
```

tests/integration/tier0/check-user-response/test_user_response.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+    ) as c2r:
+        c2r.expect_exact(" ... activation key not found, username and password required")
+        c2r.expect_exact("Username: ")
+        c2r.sendline()
+        if c2r.expect_exact(["Username", "Password"]) == 0:
+            # Provide username, expect password prompt
+            c2r.sendline(env.str("RHSM_USERNAME"))
+        else:
+            c2r.sendcontrol("d")
```
kokesak, line 20: I don't quite understand why do you use `ctrl + d` here. Is there some reason for it?

tests/integration/tier0/check-user-response/test_user_response.py
```diff
@@ -0,0 +1,50 @@
...
+        c2r.expect_exact(" ... activation key detected: ")
+        c2r.expect_exact("Organization: ")
+        c2r.sendline()
+        substitute_org = "foo"
+        if c2r.expect_exact("Organization: ") == 0:
+            c2r.sendline(substitute_org)
+        else:
+            c2r.sendcontrol("d")
+        c2r.expect_exact("Registering the system using subscription-manager ...")
+        c2r.sendcontrol("d")
```
kokesak, line 49: I would use `c` character to terminate the running process. Even though `ctrl d` may work I don't think it's safe to use it.

r0x0d: I think in OL/CentOS 7 there is a bug where you have to hit Ctrl + C multiple times to exit the process... Don't know if it's a bug in how python handles it or it is a bug in our code, but I experienced it in here https://github.com/oamg/convert2rhel/pull/411

r0x0d: image

The `^C` is for Ctrl + C and the `EOFError` at the end is for the actual Ctrl + D. 

(This is from CentOS 7)

r0x0d: image

Works fine on CentOS 8 (Which is python3 by default)

kokesak: Oh I see now. If sending Ctrl + C doesn't work I would make a comment in code about it so it's clear why we use Ctrl + D instead.

r0x0d: Totally agree! Can even link this thread to the comment in code

danmyway: Yeah, for some reason Ctrl + C doesn't work as consistently as Ctrl + D does.
Anyway I will add the comment mentioning this.

tests/integration/tier0/check-user-response/test_user_response.py
danmyway: Good point, I will change that.

danmyway:
Rebased. @mkluson, @kokesak can you please add your review, after the checks are done?

tests/integration/tier0/check-user-response/test_user_response.py
```diff
@@ -0,0 +1,50 @@
...
+    # check for user prompt enforcing input, then continue with registration
+    with convert2rhel(
+        "-y --no-rpm-va --serverurl {}".format(
+            env.str("RHSM_SERVER_URL"),
+        )
+    ) as c2r:
+        c2r.expect_exact(" ... activation key not found, username and password required")
+        c2r.expect_exact("Username")
+        c2r.sendline()
+        assert c2r.expect_exact(["Username", "Password"]) == 0
```
kokesak, line 16: Just a question, are the asserts here necessary?

danmyway: I believe so, as I would like to check if the assertion wouldn't raise `assert 1 == 0` which is the unwanted behavior.

kokesak: As far as I know the `expect()` returns the index of the first match it gets. So in this case you are checking that the "Username" will be displayed and not the "Password" right?

kokesak: If that's the case I would add a comment above this line, explaining that we want to check the presence of the "Username" and raise am error if the "Password" would appear instead.

kokesak:
@r0x0d it seems like there is an issue when using the activation key or satellite method to subscribe the machine. All tests are failing on this. I thought it was some infrastructure error but it's still happening. Can you please try it yourself and verify if there isn't any bug in code?

Here is pastebin of the failed attempt https://privatebin.corp.redhat.com/?f0cd92b51b18652f#46eU5mZw2u1oGsSMX2i8BxP3cHqmYs4AKsfbTK7qQxuq
The credentials used are correct.

tests/integration/tier0/check-user-response/test_user_response.py
danmyway: Exactly. Alright will do that.

r0x0d:
> @r0x0d it seems like there is an issue when using the activation key or satellite method to subscribe the machine. All tests are failing on this. I thought it was some infrastructure error but it's still happening. Can you please try it yourself and verify if there isn't any bug in code?
> 
> Here is pastebin of the failed attempt https://privatebin.corp.redhat.com/?f0cd92b51b18652f#46eU5mZw2u1oGsSMX2i8BxP3cHqmYs4AKsfbTK7qQxuq The credentials used are correct.

Sure, let me try.

kokesak (approved):
LGTM. Thanks for the good work.

r0x0d (commented):
LGTM.

PR #393, opened 2021-12-15
Untrack previously installed submgr dependencies

r0x0d:
After performing a rollback on the system, the python3-syspurpose package is removed from the system causing some packages to report an error when `yum check` is executed after the rollback.

Jira reference: https://issues.redhat.com/browse/OAMG-5749
Bugzilla reference (If any):


convert2rhel/subscription.py
```diff
@@ -247,8 +247,15 @@ def install_rhel_subscription_manager():
     else:
         loggerinst.info("Packages installed:\n%s" % "\n".join(rpms_to_install))
         pkg_names = get_installed_submgr_pkg_names(rpms_to_install)
-        utils.changed_pkgs_control.track_installed_pkgs(pkg_names)
+        if "python3-syspurpose" in pkg_names:
+            loggerinst.debug(
+                "Untracking python3-syspurpose from installed packages"
+                "as it could be a system package installed before this process."
+            )
+            pkg_names.remove("python3-syspurpose")
```
r0x0d, line 255: I'm wondering, should we check beforehand if the packages to remove are already installed on the system? Maybe with `rpm -q python3-syspurpose`? 

If we find out that the mentioned package is already present in the system before installing `subscription-manager`, what should we do? Leave it as it is? Proceed with the reinstallation of the package for the correct version for `subscription-manager`?

convert2rhel/subscription.py
```diff
@@ -242,13 +246,33 @@ def install_rhel_subscription_manager():
         # When using the original system repos, we need YUM/DNF to expand the $releasever by itself
         set_releasever=False,
     )
-    if ret_code:
-        loggerinst.critical("Failed to install subscription-manager packages. See the above yum output for details.")
-    else:
```
SpyTec, line 247: Just a nitpick. We could keep `if ret_code:` and simply remove the else statement since critical is a SysExit

convert2rhel/subscription.py
```diff
@@ -223,13 +223,17 @@ def remove_original_subscription_manager():
 
 
 def install_rhel_subscription_manager():
-    loggerinst.info("Installing subscription-manager RPMs.")
+    loggerinst.info("Checking for subscription-manager RPMs.")
```
SpyTec, line 226: We're still installing subscription-manager in the method. We just have more checks to track and untrack the now installed packages

convert2rhel/subscription.py
```diff
@@ -242,13 +246,33 @@ def install_rhel_subscription_manager():
...
-    else:
-        loggerinst.info("Packages installed:\n%s" % "\n".join(rpms_to_install))
-        pkg_names = get_installed_submgr_pkg_names(rpms_to_install)
-        utils.changed_pkgs_control.track_installed_pkgs(pkg_names)
+    if not ret_code:
+        loggerinst.info("\nPackages installed:\n%s" % "\n".join(rpms_to_install))
+
+        # Remove any packages that are present in the system before the install
+        [pkg_names.remove(rpm) for rpm in rpms_to_untrack if rpm in pkg_names]
+        loggerinst.debug("Untracked pacakges: %r" % rpms_to_untrack)
```
SpyTec, line 254:
```suggestion
        loggerinst.debug("Untracked packages: %r" % rpms_to_untrack)
```

convert2rhel/subscription.py
```diff
@@ -242,13 +246,33 @@ def install_rhel_subscription_manager():
...
         loggerinst.debug("Tracking installed packages: %r" % pkg_names)
+        utils.changed_pkgs_control.track_installed_pkgs(pkg_names)
+    else:
+        loggerinst.critical("Failed to install subscription-manager packages. See the above yum output for details.")
+
+
+def check_for_pkg_presence(pkg_names):
+    """Returns a list of packages that are already present in the system before
+    installing any packages needed by subscrpition-manager
```
SpyTec, line 263:
```suggestion
    """Check that packages are installed on the system.
```

convert2rhel/subscription.py
```diff
@@ -242,13 +246,33 @@ def install_rhel_subscription_manager():
...
+
+    :param pkg_names: List of package names
+    """
+    rpms_present = []
+    # Check for already installed packages that subscription-manager uses.
+    # If found any, then add it to a list an consider it as an untracked pkg.
```
SpyTec, line 269:
```suggestion
    # Check for already installed packages
    # If a package is installed, add it to a list which is returned
```

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -293,12 +293,43 @@ def test_remove_original_subscription_manager(self):
...
+    @unit_tests.mock(pkghandler, "call_yum_cmd", lambda a, b, enable_repos, disable_repos, set_releasever: (None, 0))
+    @unit_tests.mock(logging.Logger, "info", LogMocked())
     @unit_tests.mock(os.path, "isdir", lambda x: True)
     @unit_tests.mock(os, "listdir", lambda x: ["filename"])
+    @unit_tests.mock(subscription, "check_for_pkg_presence", lambda x: ["test"])
+    @unit_tests.mock(subscription, "get_installed_submgr_pkg_names", lambda x: ["test"])
+    @unit_tests.mock(utils.changed_pkgs_control, "track_installed_pkgs", lambda x: None)
+    def test_install_rhel_subscription_manager(self):
+        subscription.install_rhel_subscription_manager()
+        self.assertTrue("\nPackages installed:\n" in logging.Logger.info.msg)
```
SpyTec, line 305: No assertions are being made about if packages get untracked from the tracked packages in `utils`. `check_for_pkg_presence` could return correct packages but we don't make any assertions that the return value in `rpms_to_untrack` is being used. Same applies to `pkg_names`

If I were to do this I'd expect a test to fail
```python
# rpms_to_untrack = check_for_pkg_presence(pkg_names)
rpms_to_untrack = []
# or 
rpms_to_untrack = ["should-fail"]
```

convert2rhel/subscription.py
```diff
@@ -223,13 +223,17 @@ def remove_original_subscription_manager():
...
     rpms_to_install = [os.path.join(SUBMGR_RPMS_DIR, filename) for filename in os.listdir(SUBMGR_RPMS_DIR)]
 
     if not rpms_to_install:
-        loggerinst.warn("No RPMs found in %s." % SUBMGR_RPMS_DIR)
+        loggerinst.warning("No RPMs found in %s." % SUBMGR_RPMS_DIR)
         return
 
+    pkg_names = get_installed_submgr_pkg_names(rpms_to_install)
+    rpms_to_untrack = check_for_pkg_presence(pkg_names)
+
```
SpyTec, line 235: Can be moved below the submgr install

SpyTec (changes requested):

convert2rhel/subscription.py
r0x0d: Actually, these calls need to be before the submr install, we need to track any packages that are present on the system before the installation of submr, if, for example, `python3-syspurpose` or `python3-cloud-what` were already present on the system (and may be used by other packages other than submr, we need to untrack them and let those packages be present on the system after the conversion)

convert2rhel/unit_tests/subscription_test.py
r0x0d: You're right! Actually, this reminds me that the `rpms_to_untrack` was not being asserted on the code either!

convert2rhel/subscription.py
r0x0d: Yeah, but the `loggerinst.info("Installing subscription-manager RPMs.")` is just at the line 236, since below that log is the actual install, I just changed this one because it makes more sense since we are checking for the availability of the rpms first, and then, installing the submr

convert2rhel/subscription.py
SpyTec: You're right. I'd add a comment instead then so we don't accidentally move it in the future to be closer

convert2rhel/subscription.py
```diff
@@ -242,13 +246,33 @@ def install_rhel_subscription_manager():
...
+    """Returns a list of packages that are already present in the system before
+    installing any packages needed by subscrpition-manager
+
+    :param pkg_names: List of package names
+    """
+    rpms_present = []
+    # Check for already installed packages that subscription-manager uses.
+    # If found any, then add it to a list an consider it as an untracked pkg.
+    for pkg in pkg_names:
+        _, ret_code = utils.run_subprocess("rpm -q %s" % pkg)
```
SpyTec, line 271:
```suggestion
        _, ret_code = utils.run_subprocess("rpm -q %s" % pkg, print_output=False)
```

convert2rhel/subscription.py
SpyTec: Reason for this change is because the function is unrelated to subscription-manager and is very generic

convert2rhel/subscription.py
SpyTec: Reason for this change is because the function is unrelated to subscription-manager and is very generic

SpyTec (changes requested):

convert2rhel/subscription.py
r0x0d: Sure! Good idea. Will add this command right now

r0x0d:

If this PR get merged before the https://github.com/oamg/convert2rhel/pull/389, then, @abadger and @Andrew-ang9 you guys have to change the `run_subprocess()` calls I've added 

* https://github.com/r0x0d/convert2rhel/blob/untrack-python3-syspurpose/convert2rhel/subscription.py#L271

Otherwise, if this gets merged after the #389, then I have to rebase and modify the `run_subprocess()` call.

convert2rhel/unit_tests/subscription_test.py
r0x0d: I've updated the test with some assertions, but don't know if they are the right ones. I believe with the assertations I just updated, we can validate that the `rpms_to_untrack` was rightfully called?

convert2rhel/subscription.py
```diff
@@ -244,11 +253,33 @@ def install_rhel_subscription_manager():
...
+def check_for_pkg_presence(pkg_names):
+    """Check that packages are installed on the system.
+
+    :param pkg_names: List of package names
+    """
+    rpms_present = []
+    # Check for already installed packages.
+    # If a package is installed, add it to a list which is returned.
+    for pkg in pkg_names:
+        _, ret_code = utils.run_subprocess("rpm -q %s" % pkg, print_output=False)
```
abadger, line 278:
```suggestion
        _, ret_code = utils.run_subprocess(["rpm", "-q", pkg], print_output=False)
```

mkluson (approved):

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,81 @@
...
+        c2r.sendline("y")
+        c2r.expect("The tool allows rollback of any action until this point.")
+        c2r.sendline("n")
+        c2r.expect("Calling command 'subscription-manager unregister'")
+        c2r.expect("System unregistered successfully.")
+
+    # check that packages still are in place
+    assert shell("rpm -qi usermode").returncode == 0
+    assert shell("rpm -qi rhn-setup").returncode == 0
+    if "centos-7" in platform.platform():
```
kokesak, line 38: We should extend the condition for oracle linux here as well.

kokesak (changes requested):
Looks good, Since we are testing on all distros (cent7/8 and ol7/8) I would add the equivalent checks to the OL distros that we have for CentOS.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,81 @@
...
+def test_check_untrack_pkgs_force(convert2rhel, shell):
+    # Terminate the c2r process forcefully,
+    # so it performs rollback.
+    with convert2rhel(f"-y --no-rpm-va") as c2r:
+        c2r.expect("Username")
+        # Due to inconsistent behaviour of 'Ctrl+c'
+        # use 'Ctrl+d' instead
+        c2r.sendcontrol("d")
+        assert c2r.exitstatus != 0
+    if "centos-8" in platform.platform():
```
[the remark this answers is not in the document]
danmyway: Sure, will do

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,81 @@
...
+        assert_not_installed(shell, "python3-syspurpose")
+    elif "oracle-7" in platform.platform():
```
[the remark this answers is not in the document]
danmyway: Okay, good point, thanks!

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Will do!

convert2rhel/subscription.py
```diff
@@ -244,11 +253,33 @@ def install_rhel_subscription_manager():
...
+    if rpms_to_untrack and pkg_names:
+        for rpm in rpms_to_untrack:
+            loggerinst.debug("Untracking package: %s" % rpm)
+            pkg_names.remove(rpm)
+
+    loggerinst.debug("Tracking installed packages: %r" % pkg_names)
+    utils.changed_pkgs_control.track_installed_pkgs(pkg_names)
+
+
+def check_for_pkg_presence(pkg_names):
```
bocekm, line 269: 1. The function name `check_for_pkg_presence` does not imply clearly what it does. I'd rename it to e.g. `filter_installed_pkgs()`.
2. The function would be better suited for `pkghandler.py` rather than this `subscription.py` module.
3. The line `_, ret_code = utils.run_subprocess(["rpm", "-q", pkg], print_output=False)` can be moved to a separate function `is_pkg_installed()`
    - this new `is_pkg_installed()` can be used instead of `pkghandler.get_installed_pkg_objects()` within `subscription.py` where simple true/false would suffice (i.e. no need to call yum/dnf python API to retrieve pkg objects)

bocekm (changes requested):

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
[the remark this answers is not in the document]
danmyway: Actually, let me reassess a bit. It seems that `python-syspurpose` was removed in [Oracle Linux 7.2](https://docs.oracle.com/en/operating-systems/oracle-linux/7/relnotes7.9/ol7-pkg-appendix.html) and `python3-syspurpose` was removed in [Oracle Linux 8.2](https://docs.oracle.com/en/operating-systems/oracle-linux/8/relnotes8.2/ol8-pkg-appendix.html#ol8-packages-bin-removed-baseos), so there is no need to extend the condition.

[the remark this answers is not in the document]
kokesak: In that case we can maybe add centos8 base repository (https://vault.centos.org/centos/8/BaseOS/x86_64/os/) to OL8 and install the package `python-3syspurpose` from there. Alternatively we can use just the url as the package will not be updated in the future - [python3-syspurpose](https://vault.centos.org/centos/8/BaseOS/x86_64/os/Packages/python3-syspurpose-1.28.21-3.el8.x86_64.rpm).

It feels like a corner case, but since we are already running the test on all platforms, we can add this I think.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_check_untrack_pkgs_graceful(convert2rhel, shell):
+    username = "foo"
+    password = "bar"
+    # Provide c2r with incorrect username and password,
+    # so it fails the registration and performs rollback.
+    with convert2rhel(f"-y --no-rpm-va --username {username} --password {password}") as c2r:
+        assert c2r.exitstatus != 0
+    if "centos-8" in platform.platform():
```
kokesak, line 63: We are calling `platform.platform()` many times in all test cases. I would create one module-level variable that would be then used in all if conditions. We can avoid calling the function multiple times.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,81 @@
...
+        c2r.sendline("y")
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("y")
+        c2r.expect("The tool allows rollback of any action until this point.")
+        c2r.sendline("n")
+        c2r.expect("Calling command 'subscription-manager unregister'")
+        c2r.expect("System unregistered successfully.")
+
+    # check that packages still are in place
+    assert shell("rpm -qi usermode").returncode == 0
```
kokesak, line 36: We can use the function `assert_not_installed` here as well earlier test case as well.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,81 @@
...
+    assert shell("rpm -qi rhn-setup").returncode == 0
+    if "centos-7" in platform.platform():
+        assert shell("rpm -qi centos-release").returncode == 0
+    elif "centos-8" in platform.platform():
+        assert shell("rpm -qi centos-linux-release").returncode == 0
+
+
+def assert_not_installed(shell, pkg=""):
```
kokesak, line 44: I would rename the function to `assert_is_installed` instead. I cannot see the reason why there is `not` in the name.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Good point

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Will do

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Yeah, that makes sense

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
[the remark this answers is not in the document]
danmyway: You're right, will do the same thing for OL7

convert2rhel/subscription.py
```diff
@@ -223,18 +223,27 @@ def remove_original_subscription_manager():
...
-        loggerinst.warn("No RPMs found in %s." % SUBMGR_RPMS_DIR)
+        loggerinst.warning("No RPMs found in %s." % SUBMGR_RPMS_DIR)
         return
 
+    # These commands have to be here, before the call for installing subscription-manager this is intended, otherwise
+    # `pkghandler.filter_installed_pkgs()` would return every single package that is listed in `rpms_to_install` and
+    # we don't want this to happen. We want only the packages that were here before the instalation of
+    # subscription-manager.
+    pkg_names = get_installed_submgr_pkg_names(rpms_to_install)
+    rpms_to_untrack = pkghandler.filter_installed_pkgs(pkg_names)
```
bocekm, line 238:
```suggestion
    # These functions have to be called before installation of the subscription-manager packages, otherwise
    # `pkghandler.filter_installed_pkgs()` would return every single package that is listed in `rpms_to_install` and
    # we don't want this to happen. We want to know about the packages that were installed before the installation of
    # subscription-manager.
    pkg_names = get_pkg_names_from_rpm_paths(rpms_to_install)
    pkgs_to_not_track = pkghandler.filter_installed_pkgs(pkg_names)
```

convert2rhel/subscription.py
```diff
@@ -244,11 +253,17 @@ def install_rhel_subscription_manager():
...
+    if rpms_to_untrack and pkg_names:
+        for rpm in rpms_to_untrack:
+            loggerinst.debug("Untracking package: %s" % rpm)
+            pkg_names.remove(rpm)
+
+    loggerinst.debug("Tracking installed packages: %r" % pkg_names)
+    utils.changed_pkgs_control.track_installed_pkgs(pkg_names)
 
 
 def get_installed_submgr_pkg_names(rpm_paths):
```
bocekm, line 269:
```suggestion
def get_pkg_names_from_rpm_paths(rpm_paths):
```

Please move this function to `pkghandler.py` or `utils.py` as it's generic.

convert2rhel/subscription.py
```diff
@@ -244,11 +253,17 @@ def install_rhel_subscription_manager():
...
+    loggerinst.info("\nPackages installed:\n%s" % "\n".join(rpms_to_install))
+
+    # Untrack any packages that are present in the system before the install
+    if rpms_to_untrack and pkg_names:
+        for rpm in rpms_to_untrack:
+            loggerinst.debug("Untracking package: %s" % rpm)
+            pkg_names.remove(rpm)
+
+    loggerinst.debug("Tracking installed packages: %r" % pkg_names)
+    utils.changed_pkgs_control.track_installed_pkgs(pkg_names)
```
bocekm, line 266:
```suggestion

    track_installed_submgr_pkgs()
    
def track_installed_submgr_pkgs(installed_pkg_names, pkgs_to_not_track):
    """Tracking newly installed subscription-manager pkgs to be able to remove them during a rollback if needed."""
    pkgs_to_track = []
    for installed_pkg in installed_pkg_names:
        if installed_pkg not in pkgs_to_not_track:
            pkgs_to_track.append(pkg)
        else:
            # Don't track packages that were present on the system before the installation
            loggerinst.debug("Skipping tracking previously installed package: %s" % installed_pkg)

    loggerinst.debug("Tracking installed packages: %r" % pkgs_to_track)
    utils.changed_pkgs_control.track_installed_pkgs(pkgs_to_track)
```

bocekm (changes requested):
Some code readability suggestions. No functionality change requested.

convert2rhel/subscription.py
r0x0d: Done.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: :heavy_check_mark:

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: :heavy_check_mark:

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: :heavy_check_mark:

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,81 @@
...
+
+def test_proper_rhsm_clean_up(shell, convert2rhel):
+    """Test that c2r does not remove usermod, rhn-setup and os-release during rollback.
+    Also checks that the system was successfully unregistered.
+    """
+
+    # Ensure usermode and rhn-setup packages are presented
+    assert shell("yum install -y usermode rhn-setup").returncode == 0
+
+    # run c2r until subscribing the system and then emulate pressing Ctrl + C
```
danmyway, line 15: @kokesak To be honest, I'm not sure about this comment saying 'emulate pressing Ctrl + C' since there is no emulation as such.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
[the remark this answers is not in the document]
danmyway: :heavy_check_mark:

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: :heavy_check_mark:

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
kokesak: Good catch, it's from the old test. Please remove the ctrl+c part

danmyway: Sure, thanks!

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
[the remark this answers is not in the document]
danmyway: :heavy_check_mark:

danmyway:
Review issues resolved, please review @mkluson, @kokesak

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,116 @@
...
+
+from envparse import env
+
+
+plt = platform.platform()
+# TODO: in integration test for PR#411. Add 'oracle-logos' and 'centos-logos'
+#  to (ol7pkgs, ol8pkgs) and (cos7pkgs, cos8pkgs) respectively
+#  as these missing packages are a result of a different cause
+#  but still missing after a rollback
+ol7pkgs = ["oracle-release-el7", "usermode", "rhn-setup", "python-syspurpose"]
```
kokesak, line 11: I would the variable so it's in uppercase and words are separated by underscore.
```suggestion
OL_7_PKGS = ["oracle-release-el7", "usermode", "rhn-setup", "python-syspurpose"]
```

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,116 @@
...
+ol8pkgs = ["oracle-release-el8", "usermode", "rhn-setup", "python3-syspurpose"]
+cos7pkgs = ["centos-release", "usermode", "rhn-setup", "python-syspurpose"]
+cos8pkgs = ["centos-linux-release", "usermode", "rhn-setup", "python3-syspurpose"]
```
kokesak, line 14: The same would apply to these as well.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,116 @@
+import platform
+
+from envparse import env
+
+
+plt = platform.platform()
```
kokesak, line 6: It's nitch picking but I would rename it to something that would represent the value inside, Eg. something like `system_release`

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,116 @@
...
+#  but still missing after a rollback
+ol7pkgs = ["oracle-release-el7", "usermode", "rhn-setup", "python-syspurpose"]
+ol8pkgs = ["oracle-release-el8", "usermode", "rhn-setup", "python3-syspurpose"]
+cos7pkgs = ["centos-release", "usermode", "rhn-setup", "python-syspurpose"]
+cos8pkgs = ["centos-linux-release", "usermode", "rhn-setup", "python3-syspurpose"]
+# The packages 'python-syspurpose' and 'python3-syspurpose' were removed in Oracle Linux 7.9
+# and Oracle Linux 8.2 respectively.
+# Installing and checking for these packages (just to make sure)
+# from CentOS repositories to cover all distros consistently.
+ol7repobypass = "http://mirror.centos.org/centos/7/os/x86_64/Packages/python-syspurpose-1.24.42-1.el7.centos.x86_64.rpm"
```
kokesak, line 19:
```suggestion
OL_7_REPO_BYPASS = "http://mirror.centos.org/centos/7/os/x86_64/Packages/python-syspurpose-1.24.42-1.el7.centos.x86_64.rpm"
```

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def define_pkg(shell, osys="", pkgs=None):
+    if osys in plt:
+        for pkg in pkgs:
+            print(f"PREP: Setting up {pkg}")
+            query = shell(f"rpm -q {pkg}")
+            if f"{pkg} is not installed" in query.output:
+                if pkg == "python3-syspurpose" and osys == "oracle-8":
+                    shell(f"yum install -y {ol8repobypass}")
```
kokesak, line 32: I would rather put assert here. In case the URL stops working or there is some error we need to know. If the `shell` command returns non-zero return code the pytest won't report a failure.
```suggestion
                    assert shell(f"yum install -y {ol8repobypass}").returncode == 0
```

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,116 @@
...
+                elif pkg == "python-syspurpose" and osys == "oracle-7":
+                    shell(f"yum install -y {ol7repobypass}")
```
kokesak, line 34: The same as above

kokesak (changes requested):
Looks good, I would just change mentioned things.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Sure, good point

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: After the disscussion with @kokesak and @mkluson it has been decided to abandon checking for python-syspurpose and python3-syspurpose on Oracle Linux completely as these are not available on convertable versions of OL

kokesak (changes requested):

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -0,0 +1,106 @@
...
+# and Oracle Linux 8.2 respectively.
+
+
+def install_pkg(shell, pkgs=None):
+    for pkg in pkgs:
+        print(f"PREP: Setting up {pkg}")
+        assert shell(f"yum install -y {pkg}").returncode == 0
+
+
+def define_booted_os(shell):
```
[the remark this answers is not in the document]
danmyway: Sure, good point!
Done.

danmyway:
Ready for review.
@kokesak, @mkluson - when you'll have time, please.

kokesak (approved):
Looks good. Thanks for the good work!

danmyway:
@bocekm, @SpyTec could you please check for the requested changes, so we might merge? Thank you!

bocekm (approved):

SpyTec (approved):

PR #396, opened 2022-01-03
Improve README for Unit Tests

r0x0d:
Improving the README.md docs for Unit Tests since the old one was just
pointing out to a script that didn't exist anymore.

Jira reference: https://issues.redhat.com/browse/OAMG-6081
Bugzilla reference (If any):

### DIFF
--- a/convert2rhel/unit_tests/README.md
+++ b/convert2rhel/unit_tests/README.md
@@ -1,0 +1,161 @@
+# Unit testing in Convert2RHEL
+
+In this file, you will understand a bit more the key concepts for working with
+unit tests in Convert2RHEL, such as:
+
+* Writing unit tests
+* Running the tests (Container / Locally)
+* Coverage
+
+## Getting started
+
+To start interacting with the project's unit tests, you first need to do a
+simple setup on your machine, this setup is basically to install some
+dependencies on your `virtualenv` and build the docker images.
+
+The only thing you have to do is run the following command in your terminal
+(You must be inside the project's root directory):
+
+```bash
+# Install dependencies, pre-commit and container images
+make install
+```
+
+This command can take a while to finish, depending on your internet connection.
+After finishing, the command will create some hidden files in the root
+directory, such as `.install`, `.pre-commit`, `.images` and so on... Don't
+worry, this is just a way to tell `make` to not run the same commands over and
+over if you try.
+
+## Writing unit tests
+
+Unit tests basically consist of testing small units of code, either a function
+call or a class definition to assert the code is doing what is expected.
+Usually, unit tests tend to mock all the external dependencies and other
+function calls that are not part of the testing code, but this is not something
+entirely necessary for every case. Mocking of external dependencies should be
+done sparingly for things that are unreliable (e.g. mocking a filepath, API
+calls, or some system-related commands).
+
+For instance, take the following test pseudo-code as an example of something
+that needs to be mocked:
+
+```python
+import mock
+import package_handler
+
+def test_check_for_yum_updates(monkeypatch):
+    packages_to_update_mock = mock.Mock(return_value=["package-1", "package-2"])
+    monkeypatch.setattr(package_handler, "get_packages_to_update", value=packages_to_update_mock)
+
+    assert package_handler.get_packages_to_update() is not None
+    assert packages_to_update_mock.assert_called_once_with(["package-1", "package-2"])
+```
+
+And this other example, of a test that don't need any mocks for external
+dependencies:
+
+```python
+import os
+import file_handler
+
+def test_archive_old_files(tmpdir):
+    tmpdir = str(tmpdir)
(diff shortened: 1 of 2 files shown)


convert2rhel/unit_tests/README.md
```diff
@@ -0,0 +1,170 @@
...
+output of the `package_handler` (Because in every test run, the result may be
+different depending on the system, because we are relying on an external
+dependency), so we mock it to be sure we have full control over the results,
+and for the second one, we know it's such a simple function that just move old
+files to another directory, so it's better to provide the files and some tmp
+path, still having full control of the output, no mocking necessary.
+
+## Running the tests
+
+The tests can be run in two different ways, `Locally` or with `Docker`.
```
SpyTec, line 101:
```suggestion
The tests can be run in two different ways, in a container and locally
```

convert2rhel/unit_tests/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Running the tests locally has the advantage of using a debugger to follow the
+execution of the code, and it's easier to generate coverage reports from the
+tests.
```
SpyTec, line 105: Code coverage can be retrieved from the `make coverage`. If we want to we can also extend that command to generate a report.  
```suggestion
Running the tests locally has the advantage of using a debugger to follow the
execution of the code.
```

convert2rhel/unit_tests/README.md
```diff
@@ -0,0 +1,170 @@
...
+And, if you want to run the tests in versions individually, you can just do
+this:
+
+```bash
+# Running the tests only in CentOS 7
+make tests7
+
+# Running the tests only in CentOS 8
+make tests8
+```
```
SpyTec, line 151: I'd move this above the Locally section, since it should be the first thing we use due to the possibility of breaking your system or leaving traces from tests with bad mocking

Also we should refer CentOS as to CentOS Linux to not be confused with Stream. Likewise, we should just say containers instead of Docker or Podman as it's interchangeable and less confusing if you use either tool.

convert2rhel/unit_tests/README.md
```diff
@@ -0,0 +1,170 @@
...
+# Run pytest and gather code coverage
+pytest --cov
+
+# Run a specific test (or group of tests based on name)
+pytest -k test_my_thing
+```
+
+### Docker
+
+To run the tests using a docker container, you can do so with the below command.
```
SpyTec, line 133:
```suggestion
### Container

To run the tests using a container, you can do so with the below command.
```

convert2rhel/unit_tests/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+This command will run the full tests suit in both CentOS 7 and 8.
```
SpyTec, line 135:
```suggestion
This command will run the full tests suit in both CentOS Linux 7 and 8.
```

convert2rhel/unit_tests/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+```bash
+# Running tests in both CentOS 7 and 8
```
SpyTec, line 138:
```suggestion
# Running tests in both CentOS Linux 7 and 8
```

convert2rhel/unit_tests/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+And, if you want to run the tests in versions individually, you can just do
+this:
+
+```bash
+# Running the tests only in CentOS 7
+make tests7
+
+# Running the tests only in CentOS 8
+make tests8
```
SpyTec, line 150:
```suggestion
# Running the tests only in CentOS Linux 7
make tests7

# Running the tests only in CentOS Linux 8
make tests8
```

convert2rhel/unit_tests/README.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+## Checking for test coverage
+
+After running the tests, you can check for the coverage in a nicer HTML
+visualization instead of the terminal output.
+
+For this, run those command in your terminal:
```
SpyTec, line 158: We'd need to run `pytest --cov` to get the coverage report right?

r0x0d: You're right, I forgot to add the statement to run pytest with --cov enabled.

convert2rhel/unit_tests/README.md
r0x0d: Makes sense! Will do the changes

convert2rhel/unit_tests/README.md
r0x0d: Uh, we don't have `coverage` in our `Makefile` right now. But sure, we can do a make command just for that, I think it could be easier for people to remember, after all, running `make coverage` is more probably to be remembered than the whole command sequence for that.

SpyTec (approved):

PR #410, opened 2022-01-20
tests: increase VM timeout from 10 to 15m

SpyTec:
The current integration tests for os7 are very flaky and fails most of the time. Increasing it from 10m to 15m should hopefully address most cases for now


SpyTec:
Closing as the issue is unrelated

PR #411, opened 2022-01-20
Add checks for handling SIGINT events

r0x0d:
The changes proposed here are intended to solve a problem that involves
the way we deal with SIGINT signals that are captured and processed by
subprocesses or third-party libraries.

This commit solves two problems, the first being regarding the
`subscription-manager` subprocess managing the SIGINT signals and the
second, this time being `librpm` intercepting such signals internally in
its C code.

Jira reference (If any): https://issues.redhat.com/browse/OAMG-5756
Bugzilla reference (If any):

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -225,3 +225,68 @@ For building the copr locally, you will need:
 ```bash
 make copr-build
 ```
+
+### Avoiding calling yum API in some specific cases
+
+We found out a bug that happened during the handling of a SIGINT
+signal (or pressing CTRL + C), the problem was caused when a user pressed those
+combination of keys during any step of the conversion before the point of no
+return, thus, starting the process of rollback and failing in the first call
+`yum` call to check wether or not some packages were installed on the system.
+
+The problem is that when we enter in that state of rollback after a SIGINT,
+`librpm` is handling the signal on their code (without any possibility to
+change this) and when the librpm detects that a SIGINT was raised, it
+understands that the execution of process should stop and a exit code 1 is sent
+to the user. One way to bypass that is not calling `yum` methods that actually
+trigger `librpm` in the process, instead, prefer calling the `yum` binary or
+any other binary that achieve the same output you desire. The reason for this
+change is that `librpm` won't trigger their handling of SIGINT if we call the
+binary directly with `call_yum_cmd` or `run_subporcess` for example, as it
+understands it's a new process and that process wasn't in the middle of the
+SIGINT raising.
+
+For example, checkout the
+[PR#411](https://github.com/oamg/convert2rhel/pull/411/files#diff-51e9ff11d39778d3b26778b293fc350430a82e2feed0dab2938da7c0069ffdc6R84)
+that introduced the change, as well, some exaplantion on why the code was
+changed from calling the yum API to calling the `rpm` command.
+
+Another example is this minimal code reproduction from
+[@abadger](https://github.com/abadger) where you can see the issue.
+
+And lastly, here's a reference ticket that originated this issue
+[OAMG-5756](https://issues.redhat.com/browse/OAMG-5756).
+
+```python
+#!/usr/bin/python2 -tt
+import time
+from yum import YumBase
+from rpmUtils.miscutils import checkSignals
+
+yb = YumBase()
+try:
+    #yb.doConfigSetup(init_plugins=False)
+    yb.runTransaction(False)
+except (BaseException, SystemExit, Exception) as e:
+    print('We were able to catch an exception from runTransaction!')
+    print(type(e))
+    print(e)
+print('Sleeping')
+time.sleep(3)
+# Hit Ctrl-C
+# If KeyboardInterrupt is raised, chances are that this did not cause the
+# problem
+try:
+    checkSignals()
+except (BaseException, SystemExit, Exception) as e:
+    print('We were able to catch an exception from checkSignals!')
+    print(type(e))
+    print(e)
+finally:
+    print('In the finally block')
+# If the previous try: except  exits immediately without message, then the
+# issue occurred.
+# If it raises a KeyboardInterrupt traceback then we're okay.
+# If it prints out anything then we're either okay or encountering a different
+# behaviour.
+```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -228,34 +228,32 @@ make copr-build
 
 ### Avoiding calling yum API in some specific cases
 
-We found out a bug that happened during the handling of a SIGINT
-signal (or pressing CTRL + C), the problem was caused when a user pressed those
+We found a bug that happened during the handling of a SIGINT
+signal (or pressing CTRL + C). The bug occurred when a user pressed that
 combination of keys during any step of the conversion before the point of no
-return, thus, starting the process of rollback and failing in the first call
-`yum` call to check wether or not some packages were installed on the system.
+return. The Ctrl-C started the rollback process and failed in the first
+`yum` API call to check whether or not some packages were installed on the system.
 
-The problem is that when we enter in that state of rollback after a SIGINT,
-`librpm` is handling the signal on their code (without any possibility to
-change this) and when the librpm detects that a SIGINT was raised, it
-understands that the execution of process should stop and a exit code 1 is sent
-to the user. One way to bypass that is not calling `yum` methods that actually
-trigger `librpm` in the process, instead, prefer calling the `yum` binary or
-any other binary that achieve the same output you desire. The reason for this
-change is that `librpm` won't trigger their handling of SIGINT if we call the
-binary directly with `call_yum_cmd` or `run_subporcess` for example, as it
-understands it's a new process and that process wasn't in the middle of the
-SIGINT raising.
+The problem is that when we enter rollback after a SIGINT, `librpm` handles the
+signal in its code (without any possibility to change this). The `librpm` handler
+detects that a SIGINT was raised, decides that the execution of the process
+should stop, and sends 1 as the program's exit code. One way to bypass this
+problem is not to call `yum` methods that actually trigger `librpm` in the
+process, instead, calling the `yum` binary or any other binary that achieves
+the same goal. By calling `yum` via `call_yum_cmd()` or `run_subprocess()`,
+a separate process is started and the SIGINT is handled by that separate
+process rather than the main convert2rhel process.
 
-For example, checkout the
+For example, checkout
 [PR#411](https://github.com/oamg/convert2rhel/pull/411/files#diff-51e9ff11d39778d3b26778b293fc350430a82e2feed0dab2938da7c0069ffdc6R84)
-that introduced the change, as well, some exaplantion on why the code was
+which introduced the fix for this bug, as well, some explanation on why the code was
 changed from calling the yum API to calling the `rpm` command.
 
-Another example is this minimal code reproduction from
-[@abadger](https://github.com/abadger) where you can see the issue.
-
-And lastly, here's a reference ticket that originated this issue
+Here's the ticket that originated this issue:
 [OAMG-5756](https://issues.redhat.com/browse/OAMG-5756).
+
+Lastly, a minimal reproducer from
+[@abadger](https://github.com/abadger) where you can see the issue.
 
 ```python
 #!/usr/bin/python2 -tt
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -81,8 +81,16 @@ def unregister_system():
         loggerinst.info("Skipping due to the use of --keep-rhsm.")
         return
 
-    submgr_installed = pkghandler.get_installed_pkg_objects("subscription-manager")
-    if not submgr_installed:
+    # We are calling run_subprocess with rpm here because of a bug in
+    # Oracle/CentOS Linux 7 in which the process always exit with 1 in case of a
+    # rollback when KeyboardInterrupt is raised, so, to avoid many changes and
+    # different conditionals, we are doing a simple call for rpm to verify if
+    # subscription-manager is installed on the system.  This is the current line
+    # on `rpm` that causes the process to exit with any interaction with the yum
+    # library
+    # https://github.com/rpm-software-management/rpm/blob/rpm-4.11.x/lib/rpmdb.c#L640
+    _, ret_code = utils.run_subprocess(["rpm", "--quiet", "-q", "subscription-manager"])
+    if ret_code != 0:
         loggerinst.info("The subscription-manager package is not installed.")
         return
     unregistration_cmd = ["subscription-manager", "unregister"]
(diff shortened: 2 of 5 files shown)


r0x0d:

If this PR get merged before the #389, then, @abadger and @Andrew-ang9 you guys have to change the run_subprocess() calls I've added (Or I could push a commit on that branch too)

    https://github.com/r0x0d/convert2rhel/blob/untrack-python3-syspurpose/convert2rhel/subscription.py#L271

Otherwise, if this gets merged after the #389, then I have to rebase and modify the run_subprocess() call.

convert2rhel/subscription.py
```diff
@@ -81,8 +81,16 @@ def unregister_system():
...
-    submgr_installed = pkghandler.get_installed_pkg_objects("subscription-manager")
-    if not submgr_installed:
+    # We are calling run_subprocess with rpm here because of a bug in OL/CentOS 7
+    # in which the process always exit with 1 in case of a rollback when
+    # KeyboardInterrupt is raised, so, to avoid many changes and different
+    # conditionals, we are doing a simple call for rpm to verify if
+    # subscription-manager is installed on the system.
+    # This is the current line on `rpm` that causes the process to exit with
+    # any interaction with the yum library
+    # https://github.com/rpm-software-management/rpm/blob/rpm-4.11.x/lib/rpmdb.c#L640
```
abadger, line 91: Great comment explaining what's going on.  Does this prohibition from using librpm to query or modify the packageset apply to all code run before the point of no return?  (Or some other point in time?)  If so, do we need to document it somewhere else as well?  (CONTRIBUTOR.md?)

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -35,7 +35,7 @@
 from convert2rhel import pkghandler, pkgmanager, utils
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
-from convert2rhel.unit_tests import GetLoggerMocked, is_rpm_based_os
+from convert2rhel.unit_tests import GetLoggerMocked, is_rpm_based_os, run_subprocess_side_effect
```
abadger, line 38: Does this need to be here?  I don't see that it is used inside of this file.

abadger (dismissed):
A few minor comments but otherwise it looks good to me.  If documenting when we can use rpm is needed, it could be done in this PR or another PR; up to you.

convert2rhel/subscription.py
r0x0d: I don't think it's too tied to the point of no return, but rather to the event of happening a SIGINT, it could be before or after the point of no return, as I tested with some really early code to SIGINT and it happened as it normally would in a later stage... But I got the point, this could affect any operations is being made after the point of no return. 

But yeah, strongly agree that we could benefit from a more detailed explanation somewhere, I'm just wondering if the CONTRIBUTING guide is the correct place for it, maybe a `TROUBLESHOOTING.md`?

convert2rhel/unit_tests/pkghandler_test.py
r0x0d: Good catch, this was left out when I was trying to remember how to mock the run_subporcess! Already removed.

convert2rhel/subscription.py
r0x0d: (I can add the explanation on this PR, so It become more complete as well)

convert2rhel/subscription.py
```diff
@@ -81,8 +81,16 @@ def unregister_system():
...
+    _, ret_code = utils.run_subprocess("rpm --quiet -q subscription-manager")
```
hosekadam, line 92:
```suggestion
    _, ret_code = utils.run_subprocess(["rpm", "--quiet", "-q", "subscription-manager"])
```

r0x0d: Thanks, @hosekadam!

CONTRIBUTING.md
```diff
@@ -248,3 +248,65 @@ For building the copr locally, you will need:
 ```bash
 make copr-build
 ```
+
+### Avoiding calling `yum` API in some specific cases
```
r0x0d, line 252: @abadger I've added a small statement about this case in the `CONTRIBUTING.md`. Let me know if it's good enough as I don't know if I did a good job explaining it.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -104,3 +101,111 @@ def test_check_untrack_pkgs_force(convert2rhel, shell):
...
+        terminate_and_assert_good_rollback(c2r)
+
+
+def test_terminate_pool(convert2rhel):
+    # Send termination signal on each user prompt asking about continuation.
+    # There is one less prompt on Oracle Linux 8 and CentOS 7
+    if "oracle-8" in booted_os or "centos-7" in booted_os:
+        x = 2
+    else:
+        x = 3
```
mkluson, line 181: This is not math, use more descriptive name for the variable :-D

danmyway: Good point :+1: 
Fixed

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -8,10 +8,10 @@
 #  to (OL_7_PKGS, OL_8_PKGS) and (COS_7_PKGS, COS_8_PKGS) respectively
 #  as these missing packages are a result of a different cause
 #  but still missing after a rollback.
```
kokesak, line 10: Please remove this TODO section

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -104,3 +101,111 @@ def test_check_untrack_pkgs_force(convert2rhel, shell):
         assert c2r.exitstatus != 0
 
     post_rollback_check(shell)
+
+
+def terminate_and_assert_good_rollback(c2r):
```
kokesak, line 106: Move this function more above among the others functions defintions.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -104,3 +101,111 @@ def test_check_untrack_pkgs_force(convert2rhel, shell):
...
+        # Assert the rollback finished all tasks by going through its last task
+        assert c2r.expect("Rollback: Removing installed RHSM certificate") == 0
+        assert c2r.exitstatus != 1
+    else:
+        c2r.sendcontrol("c")
+        assert c2r.expect("Rollback: Removing installed RHSM certificate") == 0
+        assert c2r.exitstatus != 1
+
+
+def test_terminate_registration_asap(convert2rhel):
```
kokesak, line 123: I do not like the word asap used here. I would use something like:
`test_terminate_on_registration_start`

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -104,3 +101,111 @@ def test_check_untrack_pkgs_force(convert2rhel, shell):
...
+
+
+def test_terminate_username(convert2rhel):
+    # Send termination signal on the user prompt for Username.
+    with convert2rhel("-y --no-rpm-va") as c2r:
+        c2r.expect("Username:")
+        terminate_and_assert_good_rollback(c2r)
+
+
+def test_terminate_password(convert2rhel):
```
kokesak, line 143: `test_terminate_on_password`

kokesak (changes requested):

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Good catch! Thanks

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Good point, done.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
```diff
@@ -104,3 +101,111 @@ def test_check_untrack_pkgs_force(convert2rhel, shell):
...
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+        )
+    ) as c2r:
+        c2r.expect("Registering the system using subscription-manager")
+        terminate_and_assert_good_rollback(c2r)
+
+
+def test_terminate_username(convert2rhel):
```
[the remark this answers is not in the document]
danmyway: Sounds better, thank you for suggestion.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Sounds better, thank you for suggestion.

tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py
danmyway: Done.

danmyway: Looks better, will do.
Thanks.

kokesak (dismissed):
LGTM! Thanks :D

kokesak:
@r0x0d can this be merged? It's OK from the qe-side.

r0x0d:
> @r0x0d can this be merged? It's OK from the qe-side.

Awesome! Gonna do a one more round of review and if it's okay, will merge it! 

Thanks for the hard work, guys

r0x0d (commented):
LGTM

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -488,37 +500,49 @@ def test_download_rhsm_pkgs(version, downloaded_pkgs, monkeypatch):
...
-            assert "System unregistration failed" in caplog.text
-        else:
-            assert "System unregistered successfully." in caplog.text
-    else:
-        assert "The subscription-manager package is not installed." in caplog.text
+
+def test_unregister_system_keep_rhsm(monkeypatch, caplog):
+    monkeypatch.setattr(tool_opts, "keep_rhsm", value=True)
+    subscription.unregister_system()
+    assert "Skipping due to the use of --keep-rhsm." in caplog.records[-1].message
```
abadger, line 546: When I have a lot of pytest test functions in a single file, I like to use classes to organize them a bit.

So something like
``` python
class TestUnregisteringSystem:
    def test_unregister_system[...]
    def test_unregister_system_submgr_not_found[...]
    def test_unregister_system_keep_rhsm[...]
    def test_unregister_system_skipped[...]
```

might be nice to do.  This is an optional request, though.  It doesn't change the functioning in any way, it's purely stylistic.

CONTRIBUTING.md
```diff
@@ -225,3 +225,68 @@ For building the copr locally, you will need:
...
+`librpm` is handling the signal on their code (without any possibility to
+change this) and when the librpm detects that a SIGINT was raised, it
+understands that the execution of process should stop and a exit code 1 is sent
+to the user. One way to bypass that is not calling `yum` methods that actually
+trigger `librpm` in the process, instead, prefer calling the `yum` binary or
+any other binary that achieve the same output you desire. The reason for this
+change is that `librpm` won't trigger their handling of SIGINT if we call the
+binary directly with `call_yum_cmd` or `run_subporcess` for example, as it
+understands it's a new process and that process wasn't in the middle of the
+SIGINT raising.
```
abadger, line 247:
```suggestion
The problem is that when we enter rollback after a SIGINT, `librpm` handles the
signal in its code (without any possibility to change this). The `librpm` handler
detects that a SIGINT was raised, decides that the execution of the process
should stop, and sends 1 as the program's exit code. One way to bypass this
problem is not to call `yum` methods that actually trigger `librpm` in the
process, instead, calling the `yum` binary or any other binary that achieves
the same goal. By calling `yum` via `call_yum_cmd()` or `run_subprocess()`,
a separate process is started and the SIGINT is handled by that separate
process rather than the main convert2rhel process.
```

CONTRIBUTING.md
```diff
@@ -225,3 +225,68 @@ For building the copr locally, you will need:
...
+
+For example, checkout the
+[PR#411](https://github.com/oamg/convert2rhel/pull/411/files#diff-51e9ff11d39778d3b26778b293fc350430a82e2feed0dab2938da7c0069ffdc6R84)
+that introduced the change, as well, some exaplantion on why the code was
+changed from calling the yum API to calling the `rpm` command.
```
abadger, line 252:
```suggestion
For example, checkout
[PR#411](https://github.com/oamg/convert2rhel/pull/411/files#diff-51e9ff11d39778d3b26778b293fc350430a82e2feed0dab2938da7c0069ffdc6R84)
which introduced the fix for this bug, as well, some explanation on why the code was
changed from calling the yum API to calling the `rpm` command.
```

CONTRIBUTING.md
```diff
@@ -225,3 +225,68 @@ For building the copr locally, you will need:
...
+
+Another example is this minimal code reproduction from
+[@abadger](https://github.com/abadger) where you can see the issue.
+
+And lastly, here's a reference ticket that originated this issue
+[OAMG-5756](https://issues.redhat.com/browse/OAMG-5756).
+
```
abadger, line 259: Switch the order of these because the example code is given inline, right after these two.
```suggestion
Here's the ticket that originated this issue:
[OAMG-5756](https://issues.redhat.com/browse/OAMG-5756).

Lastly, a minimal reproducer from
[@abadger](https://github.com/abadger) where you can see the issue.

```

convert2rhel/subscription.py
```diff
@@ -81,8 +81,16 @@ def unregister_system():
...
 
-    submgr_installed = pkghandler.get_installed_pkg_objects("subscription-manager")
-    if not submgr_installed:
+    # We are calling run_subprocess with rpm here because of a bug in
+    # Oracle/CentOS Linux 7 in which the process always exit with 1 in case of a
+    # rollback when KeyboardInterrupt is raised, so, to avoid many changes and
+    # different conditionals, we are doing a simple call for rpm to verify if
+    # subscription-manager is installed on the system.  This is the current line
+    # on `rpm` that causes the process to exit with any interaction with the yum
+    # library
```
abadger, line 90:
```suggestion
    # We are calling run_subprocess with rpm here because of a bug in
    # Oracle/CentOS Linux 7 in which the process always exits with 1 in case of a
    # rollback when KeyboardInterrupt is raised.  To avoid many changes and
    # different conditionals to handle that, we are doing a simple call to rpm to verify if
    # subscription-manager is installed on the system.  This is the current line
    # in `rpm` that causes the process to exit with any interaction with the yum
    # library
```

abadger (changes requested):
Okay, here are the grammar and spelling changes.

convert2rhel/unit_tests/subscription_test.py
r0x0d: Think this is a nice way of grouping them. Adding the changes to the next commit. Thanks!

abadger (approved):
Looks good to me!

+1 to merge.

PR #414, opened 2022-01-25
Made grammatical and style changes to text

mportman12:
Make grammatical and style changes to the Convert2RHEL error messages.

### DIFF
--- a/convert2rhel/grub.py
+++ b/convert2rhel/grub.py
@@ -324,7 +324,7 @@ def _copy_grub_files(required, optional):
     Return False when any required file has not been copied or is missing.
     """
     if systeminfo.system_info.id != "centos":
-        logger.debug("Skipping copying grub files - only related to CentOS Linux.")
+        logger.debug("Skipping copying GRUB files - only related to CentOS Linux.")
         return True
 
     # TODO(pstodulk): check behaviour for efibin from a different dir or with a different name for the possibility of
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -174,7 +174,7 @@ def call_yum_cmd(
 
 
 def get_problematic_pkgs(output, excluded_pkgs=frozenset()):
-    """Parse the YUM/DNF output to find what packages are causing a transaction failure."""
+    """Parse the YUM/DNF output to find which packages are causing a transaction failure."""
     loggerinst.info("Checking for problematic packages")
     problematic_pkgs = {
         "protected": set(),
@@ -281,7 +281,7 @@ def resolve_dep_errors(output, pkgs=frozenset()):
 
 
 def get_installed_pkgs_by_fingerprint(fingerprints, name=""):
-    """Return list of those installed packages (just names) which are signed
+    """Return list of names of installed packages that are signed
     by the specific OS GPG keys. Fingerprints of the GPG keys are passed as a
     list in the fingerprints parameter. The packages can be optionally
     filtered by name.
@@ -291,7 +291,7 @@ def get_installed_pkgs_by_fingerprint(fingerprints, name=""):
 
 
 def get_installed_pkgs_w_fingerprints(name=""):
-    """Return a list of objects, each holding one of the installed packages (yum.rpmsack.RPMInstalledPackage in case
+    """Return a list of objects that hold one of the installed packages (yum.rpmsack.RPMInstalledPackage in case
     of yum and hawkey.Package in case of dnf) and GPG key fingerprints used to sign it. The packages can be
     optionally filtered by name.
     """
@@ -328,7 +328,7 @@ def get_pkg_signature(pkg_obj):
 def get_rpm_header(pkg_obj):
     """The dnf python API does not provide the package rpm header:
       https://bugzilla.redhat.com/show_bug.cgi?id=1876606.
-    So instead, we're getting the header directly from the rpm db.
+    The header is instead fetched directly from the rpm db.
     """
     ts = rpm.TransactionSet()
     rpm_hdr_iter = ts.dbMatch("name", pkg_obj.name)
@@ -377,9 +377,7 @@ def _get_installed_pkg_objects_dnf(name):
 
 
 def get_third_party_pkgs():
-    """Get all the third party packages (non-Red Hat and non-original OS
-    signed) which are going to be kept untouched.
-    """
+    """Get all the third party packages (non-Red Hat and non-original OS-signed) that are going to be kept untouched."""
     third_party_pkgs = get_installed_pkgs_w_different_fingerprint(
         system_info.fingerprints_orig_os + system_info.fingerprints_rhel
     )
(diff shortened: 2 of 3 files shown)


convert2rhel/pkghandler.py
```diff
@@ -671,6 +672,7 @@ def get_kernel_availability():
     """Return a tuple - a list of installed kernel versions and a list of
     available kernel versions.
     """
+# is tuple spelled correctly? This is a math term.
```
abadger, line 675: Yes, it's a python data type (a read-only list): https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences  Thanks for verifying!

convert2rhel/pkghandler.py
```diff
@@ -517,13 +516,15 @@ def remove_excluded_pkgs():
...
-    in the repofiles (e.g. $releasever)
+    in the repofiles (e.g. $releasever).
 
     We can't be removing them together with the other excluded packages - it wouldn't
     be possible to access and install subscription-manager dependencies. At the same time
     we can't install subscription-manager before removing the excluded packages
     as there would be a conflict with one of the excluded packages (rhn-client-tools).
     """
+    # The above comment is very confusing. What does it mean to not remove the packages with the other excluded packages? 
+    # It's not clear when the above comment is referring to the excluded packages and when it's referring to nonexcluded packages that need to be removed.
```
abadger, line 527: Michal wrote this when he last updated this docstring:

commit b59a7e5fbf9c355487d3026ae78451109c3c64e2
Author: Michal Bocek <mbocek@redhat.com>
Date:   Fri Jan 29 19:24:42 2021 +0100

    Remove repofile pkgs after submgr installation
    
    -> We need the original system repositories available in order to have
    access to the dependencies of subscription-manager when we get to
    installing it with yum/dnf. For that reason we need to make sure the
    packages containg the system yum/dnf repofiles are removed only after
    installing subscription-manager.
    
    Remove release pkgs only after installing submgr and subscribing the
    system -> we were removing the release pkgs before installing the
    subscription-manager pkgs and we were passing the --releasever option to
    yum with a value for RHEL repos, not the original system repos. That
    caused the submgr installation to fail. And on CentOS 8, subscription
    of the system fails if there's no release package installed
    (centos-linux-release, redhat-release).

I'm going to take a guess at what this means but also going to CC @bocekm to erify that I am correct:

When convert2rhel runs, it removes some excluded packages (what precisely are excluded packages?).  Some of those packages contain non-RHEL repo files (files which tell yum and dnf where to look for packages to install on the system).  In order to install subscription-manager, we need those repo files to be present so that yum or dnf can find the packages that subscription-manager depends on.  So we have to remove these particular packages after subscription-manager is installed.

However, the rhn-client-tools excluded package must be removed before we install subscription-manager because the two packages conflict.

To meet both requirements, we first remove the excluded packages which do not contain repo files.  Then we install subscription-manager.  We then subscribe the system (due to the way subscribing systems works in CentOS 8).  Only after those three steps do we call this function which removes the excluded packages which do contain repo files.

@bocekm, did I get that right?

bocekm: You got it right, @abadger.

The [summary of the issue](https://github.com/oamg/convert2rhel/pull/107#issuecomment-726353163) I fixed by introducing the `remove_repofile_pkgs()` function:
`The problem is that we first uninstall the excluded pkgs like oraclelinux-release or centos-release, which provide the standard system repofiles. So after we uninstall them and we get to install subscription-manager, yum has no repos to get the subscription-manager deps from.`

BTW, here's an example of list of excluded pkgs we're removing at the start of the conversion: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/data/8/x86_64/configs/centos-8-x86_64.cfg#L10. The excluded pkgs are those that either cause transaction failures or are vendor-specific (like logos).

mportman12: Thanks, updated the message to make it more clear which packages are being referred to. I do wonder how much info does the customer need - do we need to explain why exactly the non-RHEL packages in question weren't removed earlier?

bocekm: Thanks, it looks good. I don't think we necessarily need to tell them more than what we do now. But feel free to suggest change to any message printed to the user.

By the way, most of your changes in this PR are targeting strings in triple apostrophes `"""` which are not being printed to the user. These are just documentation strings for the code developers. Feel free to update these as well. Tthough it won't have an effect to what the user sees in the terminal when executing convert2rhel. The messages printed to the user are in `logger.` or `loggerinst.<loglevel>("<message>")`.

bocekm (approved):

convert2rhel/grub.py
```diff
@@ -324,7 +324,7 @@ def _copy_grub_files(required, optional):
     Return False when any required file has not been copied or is missing.
     """
     if systeminfo.system_info.id != "centos":
-        logger.debug("Skipping copying grub files - only related to CentOS Linux.")
+        logger.debug("Skipping copying GRUB files - only related to CentOS Linux.")
```
r0x0d, line 327: @mportman12, just a quick question... In the commend above this, we have the word `grub` lowercased (on line 314, more specifically) this shouldn't be uppercase like in this line?

convert2rhel/pkghandler.py
```diff
@@ -173,8 +173,8 @@ def call_yum_cmd(
     return stdout, returncode
 
 
-def get_problematic_pkgs(output, excluded_pkgs=frozenset()):
-    """Parse the YUM/DNF output to find what packages are causing a transaction failure."""
+def get_problematic_pkgs(output, excluded_pkgs=set()):
```
r0x0d, line 176: @bocekm, When I rebased this and resolved the conflicts, this line changed from `frozenset()`to `set()` this is right? I didn't saw this PR earlier, so I don't remember if this is intended.

r0x0d: Ok! gonna revert to the original. Thanks

r0x0d (approved):

convert2rhel/pkghandler.py
mportman12: @bocekm Oh, good to know... That would have saved me some time.

convert2rhel/grub.py
mportman12: Good catch, fixed.

r0x0d:
I can fix the merge conflicts after the changes are done.

bocekm (approved):

bocekm (changes requested):

convert2rhel/pkghandler.py
```diff
@@ -173,8 +173,12 @@ def call_yum_cmd(
     return stdout, returncode
 
 
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
r0x0d: Gotcha! Will do in a sec.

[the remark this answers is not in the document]
r0x0d: Done.

r0x0d (dismissed):

bocekm (dismissed):

bocekm (dismissed):

bocekm (approved):

PR #415, opened 2022-01-26
checks to see if C2R is the latest version

Andrew-ang9:
this adds to the code a case for if the users version of convert 2 RHEL isn't the most recent version, then it will let the user know to either go and update to the current version or to set an envar to bypass it. 

Jira ticket OAMG-4773


SpyTec (changes requested):
Also needs unit tests

convert2rhel/checks.py
```diff
@@ -57,6 +60,71 @@ def perform_pre_ponr_checks():
     ensure_compatibility_of_kmods()
 
 
+def check_convert2rhel_latest():
+    """Make sure that we a running the latest version of convert2rhel"""
+
+    # get lastest avalible version
```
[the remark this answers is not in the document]
Andrew-ang9: yeah after I submitted this PR I noticed that I had left the debug comments in, so I will definitely get those out! I am working on the unit test so it should be added to this PR soon! Thank you for the feedback!

convert2rhel/checks.py
```diff
@@ -21,11 +21,13 @@
 import os
 import re
 
+import conver2rhel
```
r0x0d, line 24:
```suggestion
```
This import is not necessary

r0x0d, line 24: I think a better way of importing this would be like 

```suggestion
from convert2rhel import __version__ as convert2rhel_version
```

This way we don't trigger warnings from `unused module` and don't mess with importing the whole module (if in the future we happen to have more things than usual there)

convert2rhel/checks.py
```diff
@@ -57,6 +60,71 @@ def perform_pre_ponr_checks():
     ensure_compatibility_of_kmods()
 
 
+def check_convert2rhel_latest():
+    """Make sure that we a running the latest version of convert2rhel"""
```
r0x0d, line 64:
```suggestion
    """Make sure that we are running the latest version of convert2rhel"""
```

convert2rhel/checks.py
```diff
@@ -57,6 +60,71 @@ def perform_pre_ponr_checks():
...
+
+    repo_content = (
+        "[convert2rhel]\n"
+        "name=Convert2RHEL Repository\n"
+        "baseurl=https://ftp.redhat.com/redhat/convert2rhel/$releasever/os/\n"
+        "gpgcheck=0\n"
+        "enabled=1\n"
+    )
+
+    repo_dir = os.path.join(utils.TMP_DIR, "convert2rhel_repo")
```
r0x0d, line 76: It's missing the import of the `utils` module at the top of the file

```python
```

convert2rhel/checks.py
```diff
@@ -57,6 +60,71 @@ def perform_pre_ponr_checks():
...
+        ("repoquery --releasever={releasever} --setopt=reposdir={repodir} convert2rhel").format(
+            releasever=system_info.releasever,
+            # releasever="8",
+            repodir=repo_dir,
+        ),
+        print_output=False,
+    )
+
+    #   - if cannot access url: print warning and continue
+    PKG_NEVR = r"\b(\S+)-(?:([0-9]+):)?(\S+)-(\S+)\b"
```
r0x0d, line 92: @abadger or @SpyTec I may be wrong, but this regex should be compiled first and preferably outside the function isn't it?

convert2rhel/checks.py
```diff
@@ -57,6 +60,71 @@ def perform_pre_ponr_checks():
...
+
+        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
+
+        if ver_compare == 1:
+            latest_version = package_version[1:]
+
+    # get version of currently running convert2rhel
+
+    # if current version is not latest:
+    ver_compare = rpm.labelCompare(("0", convert2rhel.__version__, "0"), ("0", latest_version[1], "0"))
```
r0x0d, line 109:
```suggestion
    ver_compare = rpm.labelCompare(("0", convert2rhel_version, "0"), ("0", latest_version[1], "0"))
```

Following my suggestion from the import

convert2rhel/checks.py
```diff
@@ -57,6 +60,71 @@ def perform_pre_ponr_checks():
...
+            logger.warning("This is not the latest version of convert2rhel.")
+
+        # - If we are converting rhel7 or rhel8:
+        # - Stop conversion
+        else:
+            logger.critical(
+                "You are currently running %s, the latest version of convert2rhel is %s. "
+                % "\n upgrading to the latest version for the conversion is only supported."
+                " You can use the envar to continue, but using the envar wouldn't be supported"(
+                    convert2rhel.__version__, latest_version
```
r0x0d, line 123:
```suggestion
                    convert2rhel_version, latest_version
```

Following my suggestion of importing the __version__

r0x0d (changes requested):
Just a few things I spotted on the code, it's not entirely

convert2rhel/checks.py
SpyTec: It could be, but not strictly necessary. It's a performance enhancement that makes more sense when you have to iterate through something and check with a RegEx multiple times, or expect the function to be called multiple times. I'm indifferent to whichever solution is done in this case

r0x0d: Gotcha! Will take a note of that for future reviews. Thanks!

convert2rhel/checks.py
```diff
@@ -74,13 +77,14 @@ def check_convert2rhel_latest():
...
     repo_dir = os.path.join(utils.TMP_DIR, "convert2rhel_repo")
+    utils.mkdir_p(repo_dir)
     # repo_dir = os.path.join("/tmp/","convert2rhel_repo")
     repo_path = os.path.join(repo_dir, "convert2rhel.repo")
     store_content_to_file(filename=repo_path, content=repo_content)
 
     # repoquery --setopt=reposdir=$(pwd)/test_repo --releasever=8 convert2rhel
     raw_output_convert2rhel_versions, _ = run_subprocess(
-        ("repoquery --releasever={releasever} --setopt=reposdir={repodir} convert2rhel").format(
+        "repoquery --releasever={releasever} --setopt=reposdir={repodir} convert2rhel".format(
```
r0x0d, line 87: Just a friendly reminder to not forget to rebase this PR later and get the latest changes from `main`, which includes now the work done in #389

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+            )
+
+        else:
+            logger.info(
+                "You are currently running %s, the latest version of convert2rhel is %s.\n"
+                "Continuing conversion." % (convert2rhel_version, latest_version)
+            )
+
+    # - If we are converting rhel7 or rhel8:
+    elif ver_compare == 1:
```
r0x0d, line 133: We could add a comment somewhere here to specify what is the meaning of those values being compared to the `ver_compare`, 

Like this, for example: https://www.programcreek.com/python/example/104387/rpm.labelCompare

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+        logger.info(
+            "You are currently running %s, the latest version of convert2rhel is %s.\n"
+            "Continuing conversion." % (convert2rhel_version, latest_version)
+        )
+
+    else:
+        logger.info(
+            "You are currently running %s, the latest version of convert2rhel is %s.\n"
+            "Continuing conversion." % (convert2rhel_version, latest_version)
+        )
```
r0x0d, line 144: I'm wondering, @Andrew-ang9, should this code all be inside the for-loop? Right now, it's outside of the loop and we are comparing the current convert2rhel version against the ("0", "0.0", "0") tuple in `latest_version`

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+
+    # repoquery --setopt=reposdir=$(pwd)/test_repo --releasever=8 convert2rhel
+    raw_output_convert2rhel_versions, _ = run_subprocess(
+        "repoquery --releasever={releasever} --setopt=reposdir={repodir} convert2rhel".format(
+            releasever=system_info.releasever,
+            # releasever="8",
+            repodir=repo_dir,
+        ),
+        print_output=False,
+    )
```
r0x0d, line 92: And also, as referenced in the description in the Jira Issue, we should be checking if this fails (maybe with the return_code) because the user might not have an internet connection at the time of running this. 

```python
    raw_output_convert2rhel_versions, return_code = run_subprocess(
        "repoquery --releasever={releasever} --setopt=reposdir={repodir} convert2rhel".format(
            releasever=system_info.releasever,
            # releasever="8",
            repodir=repo_dir,
        ),
        print_output=False,
    )
    
    if return_code != 0: 
        loggerinst.warnin("Couldn't check if this is the latest version of convert2rhel")
```

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+    utils.mkdir_p(repo_dir)
+    # repo_dir = os.path.join("/tmp/","convert2rhel_repo")
+    repo_path = os.path.join(repo_dir, "convert2rhel.repo")
+    store_content_to_file(filename=repo_path, content=repo_content)
+
+    # repoquery --setopt=reposdir=$(pwd)/test_repo --releasever=8 convert2rhel
+    raw_output_convert2rhel_versions, _ = run_subprocess(
+        "repoquery --releasever={releasever} --setopt=reposdir={repodir} convert2rhel".format(
+            releasever=system_info.releasever,
+            # releasever="8",
```
r0x0d, line 88: This comment can be removed, as I believe this was only to debug the command execution.

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+        "baseurl=https://ftp.redhat.com/redhat/convert2rhel/$releasever/os/\n"
+        "gpgcheck=0\n"
+        "enabled=1\n"
+    )
+
+    repo_dir = os.path.join(utils.TMP_DIR, "convert2rhel_repo")
+    utils.mkdir_p(repo_dir)
+    # repo_dir = os.path.join("/tmp/","convert2rhel_repo")
+    repo_path = os.path.join(repo_dir, "convert2rhel.repo")
+    store_content_to_file(filename=repo_path, content=repo_content)
```
r0x0d, line 82: @abadger do you think replacing this with the [tempfile](https://docs.python.org/3/library/tempfile.html) is a more secure way? Since we are using this as a temporary file and not something that will be persistent across the whole conversion.

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+                "Only the latest version is supported for conversion. If you want to ignore"
+                " this you can set 'CONVERT2RHEL_UNSUPPORTED_VERSION' to continue"
+                % (convert2rhel_version, latest_version)
+            )
+
+        else:
+            logger.info(
+                "You are currently running %s, the latest version of convert2rhel is %s.\n"
+                "Continuing conversion." % (convert2rhel_version, latest_version)
+            )
```
r0x0d, line 130: What do you think about changing this to be a simple warning to the user if they are on the RHEL6 since we don't officially support it? 

Maybe something like this: 

```suggestion
    if system_info.version.major <= 6:
        logger.warning(
            "You're running an unsupported operating system version for convert2rhel. "
            "If you want to ignore  this, then you can set 'CONVERT2RHEL_UNSUPPORTED_VERSION' to continue"
        )
```

The message is just a placeholder, feel free to improve it if you think it's a good change.

r0x0d (commented):
Looking good. Only a few comments I spotted reviewing this one more time. 

Also, if I'm not mistaken, we could reorder the version testing, to make it a bit easier to read and not confusing.

convert2rhel/checks.py
abadger: I would continue to create the files in `utils.TMPDIR` although you are right that using `tempfile.mkdtemp()` inside of `utils.TMPDIR` might be a little better than hardcoding `"convert2rhel_rep"`.

First of all, you're right that this data is very temporary in nature.  However, in general, that's not the reason that you should use `/tmp.` or `/var/tmp`.  Those should really only be used if you can't depend on some other directory (the user's home directory or, in our case, '/var/lib/convert2rhel/`) or you have to share the file with some other user.  Anytime you operate in `/tmp` or `/var/tmp` it is less secure because those directories are world writable which exposes you to all sorts of race conditions which can be used to attack the system.  The functions in the tempfile stdlib module should be secure against most of those attacks except for a denial of service attack.  Here's an accessible writeup (it doesn't go into the intricacies of what O_CREAT and O_EXCL actually do and do not protect against but it is accurate and pretty easy to understand):  https://owasp.org/www-community/vulnerabilities/Insecure_Temporary_File

Using `tempfile.mkdtemp()` to create repo_dir could be a little better because it generates a name with some random characters in it.  That way if `/var/lib/convert2rhel/convert2rhel.repo` already existed, we wouldn't collide  with whatever was using it.  Using it would look something like this on Line 78:
``` python
repo_dir = tempfile.mkdtemp(prefix="convert2rhel_repo.", dir=utils.TMP_DIR)`
```
If we did that, I we'd need to make sure that utils.TMP_DIR is created at some point before the `mkdtemp()` call.  I can't remember off the top of my head if we're doing that anywhere or if we need to do it here.

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+    PKG_NEVR = r"\b(\S+)-(?:([0-9]+):)?(\S+)-(\S+)\b"
+    convert2rhel_versions = re.findall(PKG_NEVR, raw_output_convert2rhel_versions, re.MULTILINE)
+
+    latest_version = ("0", "0.00", "0")
+    #   - get latest version from https://ftp.redhat.com/redhat/convert2rhel/.
+    for package_version in convert2rhel_versions:
+        if package_version[0] != "convert2rhel":
+            continue
+
+        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
```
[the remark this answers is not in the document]
abadger: I think there's some missing logic after this (still within the for loop).  I think it should check `ver_compare` to see if `package_version` evaluated as later than `latest_version`.  If so, set `latest_version = package_version[1:]`.

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+
+    # if current version is not latest:
+    ver_compare = rpm.labelCompare(("0", convert2rhel_version, "0"), ("0", latest_version[1], "0"))
+    # - If envar is detected
+    if "CONVERT2RHEL_UNSUPPORTED_VERSION" in os.environ:
```
abadger, line 109: I think the decision tree here needs to be reworked.  Here's how I would approach the problem:

1. What are the conditions that I need to check?
2. Which conditions match up with which outcomes?
3. What data do I have to form each condition?

let's take those one at a time:

## 1. What conditions do I want to check?
In the jira ticket, Michal and Eric list the following things:

- Check whether there's a newer packaged version than what we currently have available
- Stop the conversion in case of conversions to RHEL 7 and 8, warn when converting to RHEL 6
- If the user has set an environment variable, then continue instead of stopping

## 2. Matching the outcomes with the checks

* If the package version is later than the running version
  * If system version <= 6.x
    * warn
  * else #  7 or 8 (or soon, 9)
    * if environment variable is set
      * warn or info and continue
    * else # env var not set
      * error out with an error message that tells why and about the env var
* else # the running version is the same or later than the package version
  * Continue 

Notice that in this case we have a maximum of three levels of nesting in the if-else blocks.  This matches up with our three conditions.  The maximum nesting happens when we find this: [1st level of nesting] if the packaged version is later than the running version and [2nd level of nesting]  if we're converting something > 6 and [3rd level] whether the env var was set or unset.

# 3. What data do I have for each condition?

So the toplevel condition needs to compare the package version with the running version.  In your code (once it's fixed), you should have package version in the `latest_version` variable and running version in the `convert2rhel_version` variable.  You compare them with `rpm.labelCompare()` to set ver_compare (outside of the loop).  So the toplevel condition will be `if ver_compare == SOMETHING`. (I don't know off the top of my head whether 1 or -1 will mean the running version or package version is later.  You should be able to test that at the python shell, though.)

The second level condition needs to test the system version.  You've found out where to get that, if `system_info.version.major`.

The third level condition tests whether the env var is set.  You've figured out that condition too: `if "CONVERT2RHEL_UNSUPPORTED_VERSION" in os.environ`

## 4. Put it together.

So the decision tree that you have here is a little different than what I came up with in step 2.  What you have right now is:
* if env var:
  *  continue
* else if package version is later
  * if system is <= 6
    * stop conversion
  * else: # system is 7, 8, 9, etc
    * continue
* elif running version is later than the package version:
  * continue with message
* else running version is equal to package version
  * continue with message that is almost the same as the previous one

So there's a few conditions that should move to different nesting levels and a few conditions that are backwards.  For the nesting levels, one thing that is helpful is to look at what you are comparing at each level.  If you are comparing different things (your topmost level compares envvar, ver_compare,  and ver_compare) then ask yourself (1) if you may have made a mistake or (2) if it would be easier to understand if you wrote it a different way.

For conditions being reversed, that happens to everyone from time to time.  You just have to either double check your work or write a simple test case and then see if the code does the opposite of what the test case tells you you wanted to happen.

convert2rhel/checks.py
abadger: I think @rodolfo is partially right here.  I notice that you are setting `ver_compare` from `rpm.labelCompare()` both inside of the loop and outside of the loop.  I think what may have happened is you needed to do something like this pseudo-code inside of the loop:
``` python
for package_version in packages:
    ver_compare = rpm.labelCompare(package_version, latest_version)
    if ver_compare > 0:
        latest_version = package_version
```

and then you needed to do something like this pseudo-code after the loop:
``` python
ver_compare = rpm.labelCompare(running_version, latest_version)
if ver_compare <= 0:
    # Do all the error checking
```
But when you saw you were checking ver_comp again, you merged the conditions inside the loop with the conditions outside of the loop (which is how `latest_version = package_version[1:]` ended up on line 134.)

In this case, what you have to remember, is even though ver_compare is the same variable name inside and outside of the loop, what it contains is different.  Inside of the loop, it is a comparison between another package version and the latest package version found so far. (your iterating through each package version, doing the comaprison, trying to find the latest package version that exists).  Outside of the loop, it is a comparison between the latest package version and the running version.

convert2rhel/checks.py
r0x0d: Bravo! I was looking at this part yesterday for almost 1h trying to figure out how to point out a better way of working with this decision tree, but couldn't find any way of writing it down.

Andrew-ang9: That makes a lot more since now looking at it that way, in the past few days revisited this code and was trying to see if there was a better way to handle this! Thank you guys I will make changes to the code and push them when i'm done!

convert2rhel/checks.py
```diff
@@ -57,6 +60,71 @@ def perform_pre_ponr_checks():
...
+        # - If we are converting rhel7 or rhel8:
+        # - Stop conversion
+        else:
+            logger.critical(
+                "You are currently running %s, the latest version of convert2rhel is %s. "
+                % "\n upgrading to the latest version for the conversion is only supported."
+                " You can use the envar to continue, but using the envar wouldn't be supported"(
+                    convert2rhel.__version__, latest_version
+                )
+            )
```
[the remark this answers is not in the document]
Andrew-ang9: I've taken care of the debug comments.

convert2rhel/checks.py
```diff
@@ -57,6 +60,71 @@ def perform_pre_ponr_checks():
...
+    repo_dir = os.path.join(utils.TMP_DIR, "convert2rhel_repo")
+    # repo_dir = os.path.join("/tmp/","convert2rhel_repo")
+    repo_path = os.path.join(repo_dir, "convert2rhel.repo")
+    store_content_to_file(filename=repo_path, content=repo_content)
+
+    # repoquery --setopt=reposdir=$(pwd)/test_repo --releasever=8 convert2rhel
+    raw_output_convert2rhel_versions, _ = run_subprocess(
+        ("repoquery --releasever={releasever} --setopt=reposdir={repodir} convert2rhel").format(
+            releasever=system_info.releasever,
+            # releasever="8",
```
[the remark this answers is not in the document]
Andrew-ang9: I've taken care of the debug comments.

convert2rhel/checks.py
Andrew-ang9: I have fixed this spelling error.

convert2rhel/checks.py
Andrew-ang9: I've added the Imports that where needed for this PR now.

convert2rhel/checks.py
Andrew-ang9: I have ported the changes that where need to follow the guidelines that where set in the PR #389

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+            continue
+
+        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
+
+    # if current version is not latest:
+    ver_compare = rpm.labelCompare(("0", convert2rhel_version, "0"), ("0", latest_version[1], "0"))
+    # - If envar is detected
+    if "CONVERT2RHEL_UNSUPPORTED_VERSION" in os.environ:
+        logger.warning(
+            "You are currently running %s, the latest version of convert2rhel is %s.\n"
```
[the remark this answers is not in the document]
Andrew-ang9: I have re-edited all of the messages and will make sure that spelling and format is good!

convert2rhel/checks.py
Andrew-ang9: I have added the check to see if this check fails.

convert2rhel/checks.py
Andrew-ang9: this comment was removed!

convert2rhel/checks.py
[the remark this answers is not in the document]
Andrew-ang9: I have used this and the other comment to refactor these lines of code

convert2rhel/checks.py
Andrew-ang9: I have went through and changed this code!

convert2rhel/checks.py
Andrew-ang9: I have made sure that TMP_DIR is being created in utils, and so I have made this change as well!

convert2rhel/unit_tests/checks_test.py
```diff
@@ -146,6 +156,23 @@ def test_pre_ponr_checks(monkeypatch):
...
+
+    # set package_version
+    def fake_run_subprocess(cmd="", print_cmd=True, print_output=True):
+        return package_version, 0
+
+    monkeypatch.setattr(checks, "run_subprocess", fake_run_subprocess)
+    monkeypatch.setattr(utils, "TMP_DIR", str(tmpdir))
+    # call the function
+    results = checks.check_convert2rhel_latest()
+    # asster that it's all correct
```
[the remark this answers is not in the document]
Andrew-ang9: I have gotten rid of this line as it is not needed!

convert2rhel/unit_tests/checks_test.py
```diff
@@ -146,6 +156,23 @@ def test_pre_ponr_checks(monkeypatch):
     ensure_compatibility_of_kmods_mock.assert_called_once()
 
 
+@pytest.mark.parametrize("local_version, package_version", (("0.24", "convert2rhel-0:0.22-1.el7.noarch"),))
+def test_convert2rhel_latest(local_version, package_version, monkeypatch, tmpdir):
```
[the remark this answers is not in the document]
Andrew-ang9: I have added two other tests that see if the messages actually are used!

convert2rhel/checks.py
```diff
@@ -66,6 +73,81 @@ def perform_pre_ponr_checks():
...
 
 
+def check_convert2rhel_latest():
+    """Make sure that we are running the latest version of convert2rhel"""
+
+    repo_content = (
+        "[convert2rhel]\n"
+        "name=Convert2RHEL Repository\n"
+        "baseurl=https://ftp.redhat.com/redhat/convert2rhel/$releasever/os/\n"
+        "gpgcheck=0\n"
```
abadger, line 83: This probably won't matter because we're not actually downloading the package (and so we'll never have the opportunity to check the gpg signatures) but it is good practice to use `gpgcheck=1` everywhere just in case we install the package in some future version of the code.

```suggestion
        "gpgcheck=1\n"
```

convert2rhel/checks.py
Andrew-ang9: done!

abadger (approved):
Looks good.  Assuming this passes tests, you can go ahead and merge!

Good work!

convert2rhel/checks.py
```diff
@@ -67,6 +72,81 @@ def perform_pre_ponr_checks():
...
+    if ver_compare < 0:
+        if "CONVERT2RHEL_UNSUPPORTED_VERSION" in os.environ:
+            logger.warning(
+                "You are currently running %s and the latest version of convert2rhel is %s.\n"
+                "'CONVERT2RHEL_UNSUPPORTED_VERSION' environment detected, continuing conversion"
+                % (convert2rhel_version, latest_version[1])
+            )
+
+        else:
+            if system_info.version.major[0] <= "6":
```
danmyway, line 129:
```suggestion
            if system_info.version.major <= 6:
```

danmyway (changes requested):

convert2rhel/checks.py
danmyway: There is a bug comparing int to str.

convert2rhel/checks.py
```diff
@@ -57,6 +62,88 @@ def perform_pre_ponr_checks():
...
+    latest_version = ("0", "0.00", "0")
+    #   - get latest version from https://ftp.redhat.com/redhat/convert2rhel/.
+    for package_version in convert2rhel_versions:
+        if package_version[0] != "convert2rhel":
+            continue
+
+        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
+
+    # if current version is not latest:
+    ver_compare = rpm.labelCompare(("0", convert2rhel_version, "0"), ("0", latest_version[1], "0"))
```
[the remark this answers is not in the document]
SpyTec: This could still be changed to reduce confusion

convert2rhel/checks.py
```diff
@@ -67,6 +73,84 @@ def perform_pre_ponr_checks():
...
+                    "Only the latest version is supported for conversion. If you want to ignore"
+                    " this you can set 'CONVERT2RHEL_UNSUPPORTED_VERSION' to continue"
+                    % (convert2rhel_version, latest_version[1])
+                )
+
+    else:
+        logger.info(
+            "You are currently running %s and the latest version of convert2rhel is %s.\n"
+            "Continuing conversion." % (convert2rhel_version, latest_version[1])
+        )
```
SpyTec, line 151: Since this will be either the same version or we have a newer one installed, I would instead opt for a debug log message saying "Latest available convert2rhel version is installed"

Andrew-ang9: ok, sounds good I'll go ahead and get that changed!

Andrew-ang9: @SpyTec also me and Toshio weren't sure where to put the cert files, at the moment we have put them in the data dir. let me know if there is a better place to put it?

convert2rhel/checks.py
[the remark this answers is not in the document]
Andrew-ang9: I have changed the format of the latest_version var in this line!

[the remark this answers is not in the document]
Andrew-ang9: I have just ran the conversion with these's changes and have realized that how the version compare is set up now it is expecting latest_version to have three segments like this ("0", "0.00", "0") which is why I had it set up like that. Should I write a comment as to why it needs to be in that order or should I try and see if I can change it to work with the format of just the version (0.00)

convert2rhel/checks.py
SpyTec: I believe that the current place is fine. It is in the right directory, but the question is more about should it go to a subdirectory. Given it's a single file I would say no and keep it where it is

convert2rhel/checks.py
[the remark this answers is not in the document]
SpyTec: @Andrew-ang9 At this stage, since you are accessing `latest_version[1]` we still have the same format as before when you first initialized the variable `latest_version = ("0", "0.00", "0")`. Unless the `latest_version` variable format is changed when you assign it the `package_version[1:]` it should work fine doing it like this and it will cause less confusion

What is the variable format that you are seeing? What is it that isn't working?

convert2rhel/checks.py
Andrew-ang9: ok, sounds good! Thank you for going over that for me!

Andrew-ang9: I have made changes to this making it a debug message, and changing the message to be clear as well!

danmyway:
All tests are passing except for cve check on centos-8.4 which is stuck in execute phase for nearly 20 hours now.
@kokesak, @abadger, @r0x0d, @SpyTec whenever you have time for a final review, please do so.

kokesak (approved):
The test looks great @danmyway, I like the approach. 

The only way to improve this would be to add some clean-up after the last test case (restore the original c2r version and delete the env variable), but it's up to you if you want to add this.

danmyway:
Rebased to upstream.

danmyway (approved):

convert2rhel/checks.py
```diff
@@ -67,6 +73,84 @@ def perform_pre_ponr_checks():
...
+        # rpm.lableCompare(pkg1, pkg2) compare two kernel version strings and return
+        # -1 if str2 is greater then str1, 0 if they are equal, 1 if str1 is greater the str2
+        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
+        if ver_compare > 0:
+            latest_version = package_version[1:]
+
+    # After the for loop latest_version will have the epoch ,version, and release ex:("0" "0.26" "1.el7") information from convert2rhel yum repo.
+    # the release for latest_verion will be "1.el7" and the relase for convert2rhel_version will be hard coded as "0" below,
+    # therefore when the versions are the same the latest_version's release field will cause it to evaluate as later
+    ver_compare = rpm.labelCompare(("0", convert2rhel_version, "0"), ("0", latest_version[1], "0"))
```
bocekm, line 126:
```suggestion
    ver_compare = rpm.labelCompare(("0", installed_convert2rhel_version, "0"), ("0", latest_available_version[1], "0"))
```
Naming of variables could be better. It's difficult for the code reader to tell the difference between convert2rhel_version, latest_version, convert2rhel_versions without looking where are they being set.

bocekm (commented):
I know I'm late to the party. I don't want to block merging of the PR. My comments can serve as a basis for a code cleanup in a subsequent PR.

convert2rhel/checks.py
```diff
@@ -67,6 +73,84 @@ def perform_pre_ponr_checks():
     ensure_compatibility_of_kmods()
 
 
+def check_convert2rhel_latest():
+    """Make sure that we are running the latest version of convert2rhel"""
```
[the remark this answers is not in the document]
bocekm: Also, add this log message here:
`logger.task("Prepare: Checking if this is the latest version of convert2rhel")`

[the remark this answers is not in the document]
bocekm: And, skip the check if the internet connection is not available to prevent log messages like:
```
19:19:29         out: [07/20/2022 19:19:29] DEBUG - Calling command 'repoquery --releasever=7 --setopt=reposdir=/var/lib/convert2rhel/convert2rhel_repo.DiUuQs convert2rhel'
19:19:30         out: WARNING - Couldn't check if this is the latest version of convert2rhel
19:19:30         out: repoquery failed 1, failure: repodata/repomd.xml from convert2rhel: [Errno 256] No more mirrors to try.
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
19:19:30         out: https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/repodata/repomd.xml: [Errno 14] curl#7 - "Failed connect to cdn.redhat.com:443; Connection refused"
```
Thanks, @r0x0d, for spotting that.

PR #417, opened 2022-01-27
Update grub images and config after a conversion

r0x0d:
This change introduces a new function call at the end of the conversion
to ensure that the grub images and configuration are properly installed
on the system after the conversion is finished.

Jira reference (If any): https://issues.redhat.com/browse/OAMG-4209
Bugzilla reference (If any):


convert2rhel/grub.py
```diff
@@ -572,6 +581,31 @@ def post_ponr_set_efi_configuration():
         _log_critical_error(e.message)
 
 
+def update_grub_after_conversion():
+    """Update grub images and config after conversion."""
```
bocekm, line 585:
```suggestion
    """Update GRUB2 images and config after conversion.
    
    This is mainly a protective measure to prevent issues in case the original distribution GRUB tooling
    generates images that expect different format of a config file. To be on the safe side we
    rather re-generate both the GRUB config file and the GRUB image.

    We opted for doing that only for GRUB2 and not for GRUB Legacy intentionally for the reason of
    RHEL 6 having transitioned to the Extended Life-cycle Support phase.
    """
```

bocekm (changes requested):

convert2rhel/grub.py
```diff
@@ -572,6 +581,31 @@ def post_ponr_set_efi_configuration():
...
+    grub2_BIOS_config_file = (
+        GRUB2_BIOS_CONFIG_FILE if not is_efi() else os.path.join(RHEL_EFIDIR_CANONICAL_PATH, "grub.cfg")
+    )
+
+    # Generate the grub.cfg file
+    _, ret_code = utils.run_subprocess(["/usr/sbin/grub2-mkconfig", "-o", grub2_BIOS_config_file], print_output=False)
```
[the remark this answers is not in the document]
r0x0d: Do you think that a simple if statement for not running this under RHEL 6 would be sufficient? Because we have the `grub-install` available under RHEL 6, but no `grub-mkconfig` or any other useful command at all...

The way I see it, we have the following options:: 

* Using an if-statment to no turn under RHEL 6 `if systeminfo.system_info.version.major <= 6: return` and prevent it from running the entire code if RHEL 6 is detected
* Ignoring the `grub2-mkconfig` using the same if, but actually let the code run for everything else and just adapt it to to run either grub2 or grub depending on the version 

What do you think it's best? @bocekm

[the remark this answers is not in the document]
r0x0d: I've updated to ignore if it's RHEL6

convert2rhel/grub.py
```diff
@@ -572,6 +581,31 @@ def post_ponr_set_efi_configuration():
...
+
+    if ret_code != 0:
+        logger.warning("GRUB2 config file generation failed.")
+        return
+
+    if not is_efi():
+        logger.debug("Detected BIOS setup, proceeding to install the new grub images.")
```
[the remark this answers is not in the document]
bocekm: Please update all the comments and log messages to use GRUB2 instead of grub, grub2, or gurb2.

bocekm (changes requested):

convert2rhel/grub.py
```diff
@@ -25,6 +25,15 @@
...
 logger = logging.getLogger(__name__)
 
+GRUB2_BIOS_ENTRYPOINT = "/boot/grub2"
+"""The path entrypoint path of the BIOS gurb2"""
+
+GRUB2_BIOS_CONFIG_FILE = os.path.join(GRUB2_BIOS_ENTRYPOINT, "grub.cfg")
+"""The path to the configuration file for grub2 in BIOS"""
+
+GRUB2_BIOS_ENV_FILE = os.path.join(GRUB2_BIOS_ENTRYPOINT, "grubenv")
+"""The path to the env file for grub2 in BIOS"""
```
SpyTec, line 35:
```suggestion
GRUB2_BIOS_ENTRYPOINT = "/boot/grub2"
"""The entrypoint path of the BIOS GRUB2"""

GRUB2_BIOS_CONFIG_FILE = os.path.join(GRUB2_BIOS_ENTRYPOINT, "grub.cfg")
"""The path to the configuration file for GRUB2 in BIOS"""

GRUB2_BIOS_ENV_FILE = os.path.join(GRUB2_BIOS_ENTRYPOINT, "grubenv")
"""The path to the env file for GRUB2 in BIOS"""
```

convert2rhel/grub.py
```diff
@@ -572,6 +581,51 @@ def post_ponr_set_efi_configuration():
...
+    generates images that expect different format of a config file. To be on the safe side we
+    rather re-generate the GRUB2 config file and install the GRUB2 image.
+    We opted for doing that only for GRUB2 and not for GRUB Legacy intentionally for the reason of
+    RHEL 6 having transitioned to the Extended Life-cycle Support phase.
+    """
+
+    if systeminfo.system_info.version.major == 6:
+        logger.warning(
+            "Convert2RHEL does not install updated GRUB Legacy bootloader image on RHEL 6. Install the image manually "
+            "by following https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/installation_guide/sect-grub-installing. "
```
SpyTec, line 597:
```suggestion
            "by following https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/installation_guide/sect-grub-installing."
```

convert2rhel/grub.py
```diff
@@ -572,6 +581,51 @@ def post_ponr_set_efi_configuration():
...
+
+    utils.RestorableFile(GRUB2_BIOS_CONFIG_FILE).backup()
+    utils.RestorableFile(GRUB2_BIOS_ENV_FILE).backup()
+
+    grub2_config_file = GRUB2_BIOS_CONFIG_FILE if not is_efi() else os.path.join(RHEL_EFIDIR_CANONICAL_PATH, "grub.cfg")
+
+    _, ret_code = utils.run_subprocess(["/usr/sbin/grub2-mkconfig", "-o", grub2_config_file], print_output=False)
+
+    if ret_code != 0:
+        logger.warning("GRUB2 config file generation failed.")
```
SpyTec, line 609: Might be beneficial to add another debug log for the output for both `grub2-mkconfig` commands

convert2rhel/grub.py
```diff
@@ -572,6 +581,51 @@ def post_ponr_set_efi_configuration():
...
+        # Relevant bugzilla for this: https://bugzilla.redhat.com/show_bug.cgi?id=1917213
+        logger.debug("Detected BIOS setup, proceeding to install the new GRUB2 images.")
+        blk_device = get_grub_device()
+
+        _, ret_code = utils.run_subprocess(["/usr/sbin/grub2-install", blk_device], print_output=False)
+
+        if ret_code != 0:
+            logger.warning("Couldn't install the new images with GRUB2.")
+            return
+
```
SpyTec, line 627: Add a logger saying GRUB2 was successful at the end of the method

convert2rhel/main.py
```diff
@@ -114,12 +114,15 @@ def main():
...
         post_ponr_conversion()
 
-        loggerinst.task("Final: rpm files modified by the conversion")
+        loggerinst.task("Final: RPM files modified by the conversion")
         systeminfo.system_info.modified_rpm_files_diff()
 
         loggerinst.task("Final: Remove temporary folder %s" % utils.TMP_DIR)
         utils.remove_tmp_dir()
 
+        loggerinst.task("Final: Updating grub2 configuration")
```
SpyTec, line 123:
```suggestion
        loggerinst.task("Final: Updating GRUB2 configuration")
```

SpyTec (changes requested):

convert2rhel/grub.py
```diff
@@ -572,6 +581,56 @@ def post_ponr_set_efi_configuration():
...
+    if systeminfo.system_info.version.major == 6:
+        logger.warning(
+            "Convert2RHEL does not install updated GRUB Legacy bootloader image on RHEL 6. Install the image manually "
+            "by following https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/installation_guide/sect-grub-installing."
+        )
+        return
+
+    utils.RestorableFile(GRUB2_BIOS_CONFIG_FILE).backup()
+    utils.RestorableFile(GRUB2_BIOS_ENV_FILE).backup()
```
r0x0d, line 602: @SpyTec now this crossed my mind. We are backing up those 2 files here but not restoring them in a case of rollback. Should we add this to `rollback_changes()` function? The `restore()` function has some conditionals to check if the file was backed-up or not, so I think it will be okay adding that there. 

The only thing is that we would probably doing this: 

```python
from convert2rhel.grub import GRUB2_BIOS_CONFIG_FILE, GRUB2_BIOS_ENV_FILE
utils.RestorableFile(GRUB2_BIOS_CONFIG_FILE).restore()
utils.RestorableFile(GRUB2_BIOS_ENV_FILE).restore()
```

Or even 

```python
# convert2rhel/grub.py 

def update_grub_after_conversion():
    ....
    grub2_bios_config_file.backup()
    grub2_bios_env_file.backup()
    ....
    
grub2_bios_config_file = utils.RestorableFile(GRUB2_BIOS_CONFIG_FILE).backup()
grub2_bios_env_file = utils.RestorableFile(GRUB2_BIOS_ENV_FILE).backup()

# convert2rhel/main.py 

def rollback_changes(): 
   ....
   grub.grub2_bios_config_file.restore()
   grub.grub2_bios_env_file.restore()
   ....
```

What do you think? It's better to do the rollback on the `rollback_changes()` function, or add a condition on `update_grub_after_conversion()` that in case of failure, we do a quick rollback of those 2 files?

SpyTec: The backup was something Michal suggested in an earlier review and he gave this comment

> I don't think we need to back the files up before the package replacement (and create two new singletons). The reason to back them up is just to give the customer a way to restore them manually post-conversion. Our rollback feature won't take care of that.

So I wouldn't say we need to restore it. But it allows the customer to restore it themselves if they need it

r0x0d: Oh, that's right! My bad. It's been some time without reading the comments here and I totally forgot about it hehe.

bocekm (approved):

r0x0d:
Rebasing to make the integration tests pass (hopefully)

plans/tier1.fmf
```diff
@@ -238,3 +238,25 @@ discover+:
...
+    script: pytest -svv tests/integration/tier1/changed-grub-file/test_valid_changed_grub.py
+  - name: reboot after conversion
+    how: ansible
+    playbook: tests/ansible_collections/roles/reboot/main.yml
+
+/changed_grub_invalid:
+  discover+:
+    test: checks-after-conversion
+  prepare+:
+  - name: modify /etc/default/grub with valid changes
```
mkluson, line 257: s/valid/invalid/

tests/integration/tier1/changed-grub-file/test_invalid_changed_grub.py
```diff
@@ -0,0 +1,28 @@
...
+import fileinput
+
+from envparse import env
+
+
+targetline = "GRUB_CMDLINE_LINUX"
+
+
+def test_modify_grub_invalid(convert2rhel):
+    # shutil.copy("/etc/default/grub", "/tmp/grub")
```
mkluson, line 12: some leftover

plans/tier1.fmf
danmyway: Oh, my bad! Thank you!

tests/integration/tier1/changed-grub-file/test_invalid_changed_grub.py
danmyway: Deleted. Thank you!

tests/integration/tier1/changed-grub-file/test_valid_changed_grub.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+targetline = "GRUB_CMDLINE_LINUX"
+
+
+def test_modify_grub_valid(convert2rhel):
+    for line in fileinput.FileInput("/etc/default/grub", inplace=True):
+        blockcmt = "\n# comment added by test\n"
+        inlinecmt_post = "# comment added by test"
+        whitespace = "     "
```
kokesak, line 15: I would move these variable definitions before the for loop, so they are not "re-defined" each loop.

kokesak (commented):
Also please rebase the PR to upsteam.

tests/integration/tier1/changed-grub-file/test_valid_changed_grub.py
danmyway: Sure, thanks for noting that!

bocekm (changes requested):
The `grub2-tools` package needs to be added to the `packaging/convert2rhel.spec` as a new dependency to cover the use of the `grub2-install` and `grub2-mkconfig` utils:
```
%if 0%{?rhel} && 0%{?rhel} >= 7
Requires:       grub2-tools
%endif
```
Note: I already checked that the utils are part of the grub2-tools pkg on CentOS Linux 7 and 8.

r0x0d:
> The `grub2-tools` package needs to be added to the `packaging/convert2rhel.spec` as a new dependency to cover the use of the `grub2-install` and `grub2-mkconfig` utils:
> 
> ```
> %if 0%{?rhel} && 0%{?rhel} >= 7
> Requires:       grub2-tools
> %endif
> ```
> 
> Note: I already checked that the utils are part of the grub2-tools pkg on CentOS Linux 7 and 8.

Are we supporting grub only in convert2rhel? Because if the user has something installed on the system to manage the boot loader that is not grub, we will be installing a dependency that is not essential to them.

Probably we don't have any use cases that will require other backends other than grub, but this is something that we should keep in mind for the future, I guess...

r0x0d:
> We test conversions with grub only as of now. However anyone is free to use any other bootloader. Red Hat supports a conversion of a system with any configuration and workload unless excluded in the official product documentation.
> 
> It seems to me that adding grub2-tools as a new convert2rhel dep is the most elegant solution ideal for most people. Yes, the downside is that the dep is unnecessary bloat in case GRUB2 is not being used. What are the other options of installing the dep though? I can imagine installing the dep during the convert2rhel execution only if GRUB2 is detected. But it seems to me like adding unnecessary complexity to the conversion process that can be prone to issues and worsening user experience.

That's true, I agree. But it's a good thing that we are aware that other bootloaders can be present on the user system, and not only grub2. Maybe when the time come we can revive this thread in a cabal session to discuss more about how that would be done properly.

Also, adding support for other backends would require some big changes in the convert2rhel code… We would need a spike first to understand all the cases we need to change and how we want to do it. 

For now, adding it as a dependency is the only solution (that I can think of too) to merge the PR, I will do it in a few.

kokesak (dismissed):
The integration test looks good, nice job @danmyway!

danmyway:
All check have passed, @bocekm, @SpyTec whenever you're ready for the final review, please do so.

bocekm (approved):
No need to block merging of the PR on my suggestions.

tests/integration/tier1/changed-grub-file/test_invalid_changed_grub.py
```diff
@@ -0,0 +1,26 @@
...
+
+import fileinput
+
+from envparse import env
+
+
+targetline = "GRUB_CMDLINE_LINUX"
+
+
+def test_modify_grub_invalid(convert2rhel):
```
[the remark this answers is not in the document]
danmyway: I agree.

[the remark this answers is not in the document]
danmyway: The docstring addition committed.

PR #419, opened 2022-01-28
Add EXTRA_ARGS to Makefile

r0x0d:
This small improvement is to allow us to pass different parameters to the pytest
runner without having to modify anything in the Makefile when necessary.

For example, we can do `make tests EXTRA_ARGS="-k test_efibootinfo"` to test only
this specific test inside the centos7 and 8 containers.

Jira reference (If any): https://issues.redhat.com/browse/OAMG-6434
Bugzilla reference (If any):

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ PIP ?= pip3
 VENV ?= .venv3
 PRE_COMMIT ?= pre-commit
 SHOW_CAPTURE ?= no
+PYTEST_ARGS ?=
 
 # Let the user specify DOCKER at the CLI, otherwise try to autodetect a working podman or docker
 ifndef DOCKER
@@ -49,7 +50,7 @@ install: .install .images .env .pre-commit
 	cp .env.example .env
 
 tests-locally: install
-	. $(VENV)/bin/activate; pytest
+	. $(VENV)/bin/activate; pytest $(PYTEST_ARGS)
 
 lint-locally: install
 	. $(VENV)/bin/activate; ./scripts/run_lint.sh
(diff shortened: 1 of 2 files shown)


r0x0d:
Let me know if you guys think that this could be useful for everyone (and if that could be useful to other commands too)

Makefile
```diff
@@ -18,6 +18,7 @@ PIP ?= pip3
 VENV ?= .venv3
 PRE_COMMIT ?= pre-commit
 SHOW_CAPTURE ?= no
+EXTRA_ARGS ?=
```
SpyTec, line 21:
```suggestion
PYTEST_ARGS ?=
```

SpyTec (commented):
Love this! Thought about looking into this but never committed to it, glad someone did!

I would just change the env to be specifically about pytest. Expanding the testing section in READMEs would be benefitial too. Could you also create a Jira issue so it's documented?

r0x0d:
> Love this! Thought about looking into this but never committed to it, glad someone did!
> 
> I would just change the env to be specifically about pytest. Expanding the testing section in READMEs would be benefitial too. Could you also create a Jira issue so it's documented?

Sure! I will make the new Jira issue and update the description. Thanks for the review!

convert2rhel/unit_tests/README.md
```diff
@@ -117,6 +117,26 @@ make tests7
 make tests8
 ```
 
+If you're willing to pass extra arguments for the pytest execution inside the container
```
abadger, line 120:
```suggestion
If you want to pass extra arguments for the pytest execution inside the container
```

SpyTec (approved):

PR #424, opened 2022-02-04
Stop if the booted kernel is not owned by any package

Andrew-ang9:
adds a unit test to catch where the tag "{{NAME}}" has an extra set of curly brace.

This error was fixed in the PR https://github.com/oamg/convert2rhel/pull/389

if you'd like to test this use the commit hash 31a304c63f815efe77fa32170a7f3d253a559310 to use the istant with the error
and use commit hash 9a1c5f50e44d2a098e2458263580d748e7ba6282 for the fix for this error

Jira ticket: [RHELC-126](https://issues.redhat.com/browse/RHELC-126)

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -437,15 +437,23 @@ def _bad_kernel_version(kernel_release):
 
 def _bad_kernel_package_signature(kernel_release):
     """Return True if the booted kernel is not signed by the original OS vendor, i.e. it's a custom kernel."""
-    kernel_pkg = run_subprocess(
-        ["rpm", "-qf", "--qf", "%{NAME}", "/boot/vmlinuz-%s" % kernel_release], print_output=False
-    )[0]
+    vmlinuz_path = "/boot/vmlinuz-%s" % kernel_release
+    kernel_pkg, return_code = run_subprocess(["rpm", "-qf", "--qf", "%{NAME}", vmlinuz_path], print_output=False)
     logger.debug("Booted kernel package name: {0}".format(kernel_pkg))
+
+    os_vendor = system_info.name.split()[0]
+    if return_code == 1:
+        logger.warning(
+            "The booted kernel %s is not owned by any installed package."
+            " It needs to be owned by a package signed by %s." % (vmlinuz_path, os_vendor)
+        )
+
+        return True
+
     kernel_pkg_obj = get_installed_pkg_objects(kernel_pkg)[0]
     kernel_pkg_gpg_fingerprint = get_pkg_fingerprint(kernel_pkg_obj)
     bad_signature = system_info.cfg_content["gpg_fingerprints"] != kernel_pkg_gpg_fingerprint
     # e.g. Oracle Linux Server -> Oracle
-    os_vendor = system_info.name.split()[0]
     if bad_signature:
         logger.warning("Custom kernel detected. The booted kernel needs to be signed by %s." % os_vendor)
         return True
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -705,41 +705,60 @@ def test_bad_kernel_substring(kernel_release, exp_return, monkeypatch):
 
 
 @pytest.mark.parametrize(
-    ("kernel_release", "kernel_pkg", "kernel_pkg_fingerprint", "exp_return"),
+    ("kernel_release", "kernel_pkg", "kernel_pkg_fingerprint", "get_installed_pkg_objects", "exp_return"),
     (
         (
             "4.18.0-240.22.1.el8_3.x86_64",
             "kernel-core",
             "05b555b38483c65d",
+            "yajl.x86_64",
             False,
         ),
         (
             "4.18.0-240.22.1.el8_3.x86_64",
             "kernel-core",
             "somebadsig",
+            "somepkgobj",
             True,
         ),
     ),
 )
 @centos8
-def test_bad_kernel_fingerprint(
+def test_bad_kernel_package_signature(
     kernel_release,
     kernel_pkg,
     kernel_pkg_fingerprint,
+    get_installed_pkg_objects,
     exp_return,
     monkeypatch,
     pretend_os,
 ):
-    run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(kernel_pkg, ""))
+    run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(kernel_pkg, 0))
     get_pkg_fingerprint_mocked = mock.Mock(spec=get_pkg_fingerprint, return_value=kernel_pkg_fingerprint)
+    monkeypatch.setattr(system_info, "name", "CentOS Linux")
     monkeypatch.setattr(checks, "run_subprocess", run_subprocess_mocked)
+    get_installed_pkg_objects_mocked = mock.Mock(spec=get_installed_pkg_objects, return_value=[kernel_pkg])
     monkeypatch.setattr(
         checks,
         "get_installed_pkg_objects",
-        mock.Mock(return_value=[kernel_pkg]),
+        get_installed_pkg_objects_mocked,
     )
     monkeypatch.setattr(checks, "get_pkg_fingerprint", get_pkg_fingerprint_mocked)
     assert _bad_kernel_package_signature(kernel_release) == exp_return
+    run_subprocess_mocked.assert_called_with(
+        ["rpm", "-qf", "--qf", "%{NAME}", "/boot/vmlinuz-%s" % kernel_release], print_output=False
+    )
+
+
+def test_kernel_not_installed(caplog, monkeypatch):
+    run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(" ", 1))
+    monkeypatch.setattr(checks, "run_subprocess", run_subprocess_mocked)
+    assert _bad_kernel_package_signature("4.18.0-240.22.1.el8_3.x86_64")
+    log_message = (
+        "The booted kernel /boot/vmlinuz-4.18.0-240.22.1.el8_3.x86_64 is not owned by any installed package."
+        " It needs to be owned by a package signed by CentOS."
+    )
+    assert log_message in caplog.text
 
 
 class TestReadOnlyMountsChecks(unittest.TestCase):


convert2rhel/unit_tests/checks_test.py
```diff
@@ -695,39 +695,27 @@ def test_bad_kernel_substring(kernel_release, exp_return, monkeypatch):
...
 ):
-    run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(kernel_pkg, ""))
+    kernel_release = run_subprocess(["uname", "-r"], print_output=False)[0].rstrip()
     get_pkg_fingerprint_mocked = mock.Mock(spec=get_pkg_fingerprint, return_value=kernel_pkg_fingerprint)
-    monkeypatch.setattr(checks, "run_subprocess", run_subprocess_mocked)
-    monkeypatch.setattr(
-        checks,
-        "get_installed_pkg_objects",
-        mock.Mock(return_value=[kernel_pkg]),
-    )
```
SpyTec, line 730: By removing this the tests fail as it gets out of range
```
>       kernel_pkg_obj = get_installed_pkg_objects(kernel_pkg)[0]
E       IndexError: list index out of range
```

SpyTec (changes requested):

abadger:
This now looks good to me (but I have been helping Andrew to write it so another pair of eyes is welcome).

convert2rhel/checks.py
```diff
@@ -422,10 +422,14 @@ def _bad_kernel_version(kernel_release):
...
     """Return True if the booted kernel is not signed by the original OS vendor, i.e. it's a custom kernel."""
-    kernel_pkg = run_subprocess(
+    kernel_pkg, return_code = run_subprocess(
         ["rpm", "-qf", "--qf", "%{NAME}", "/boot/vmlinuz-%s" % kernel_release], print_output=False
-    )[0]
+    )
     logger.debug("Booted kernel package name: {0}".format(kernel_pkg))
+    # add a case for when the kernel package isn't returning any package info
+    if return_code is 0:
+        logger.critical("The running kernel is not from a rpm")
```
r0x0d, line 431: Should we test that this branch was actually used? I don't think that the current unit tests are handling this

r0x0d (commented):
Looking good, one minor comment about the code.

convert2rhel/checks.py
bocekm: The user won't know what to do to fix this. The message they should see is here: https://github.com/oamg/convert2rhel/blob/v0.25/convert2rhel/checks.py#L380. That means you need to change this from critical to warning and return True from the function right away so that the linked message gets displayed.

Also I suggest making the message more meaningful to the user, similar to this message https://github.com/oamg/convert2rhel/blob/v0.25/convert2rhel/checks.py#L434.

```suggestion
    vmlinuz_path = "/boot/vmlinuz-%s" % kernel_release
    ...
    if return_code is 0:
        logger.warning("The running kernel %s is not owned by any installed package. It needs to be owned by a package signed by %s." % {vmlinuz_path,os_vendor})
        return True
```

bocekm (changes requested):

convert2rhel/unit_tests/checks_test.py
Andrew-ang9: thank you for bringing that to my attention, I have gotten that fixed!

convert2rhel/checks.py
Andrew-ang9: thank you guys, I will work on these changes!

convert2rhel/checks.py
```diff
@@ -422,10 +422,21 @@ def _bad_kernel_version(kernel_release):
...
 def _bad_kernel_package_signature(kernel_release):
     """Return True if the booted kernel is not signed by the original OS vendor, i.e. it's a custom kernel."""
-    kernel_pkg = run_subprocess(
+    kernel_pkg, return_code = run_subprocess(
         ["rpm", "-qf", "--qf", "%{NAME}", "/boot/vmlinuz-%s" % kernel_release], print_output=False
-    )[0]
+    )
     logger.debug("Booted kernel package name: {0}".format(kernel_pkg))
+
+    vmlinuz_path = "/bootvmlinuz-%s" % kernel_release
```
SpyTec, line 430:
```suggestion
    vmlinuz_path = "/boot/vmlinuz-%s" % kernel_release
```

convert2rhel/checks.py
```diff
@@ -422,10 +422,21 @@ def _bad_kernel_version(kernel_release):
...
+
+    if return_code == 0:
+        logger.warning(
+            "The booted kernel %s is not owned by any installed package."
+            " It needs to be owned by a package signed by %s." % (vmlinuz_path, os_vendor)
```
SpyTec, line 435: `os_vendor` is undefined. Can see how to fetch it from here 
https://github.com/oamg/convert2rhel/blob/120ad4f6356f50e861acc15f826132c4c9d0e7e3/convert2rhel/checks.py?rgh-link-date=2022-02-16T15%3A42%3A26Z#L432-L434

SpyTec (changes requested):

convert2rhel/checks.py
Andrew-ang9: Thank you for pointing out this typo, I'll go ahead and get this fixed!

convert2rhel/checks.py
Andrew-ang9: I'll implement this definition to my code to get this fixed, thank you for the feedback!

convert2rhel/unit_tests/checks_test.py
```diff
@@ -735,6 +735,17 @@ def test_bad_kernel_fingerprint(
     )
 
 
+def test_kernel_not_installed(caplog, monkeypatch):
+    run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(" ", 1))
+    monkeypatch.setattr(checks, "run_subprocess", run_subprocess_mocked)
+    _bad_kernel_package_signature("4.18.0-240.22.1.el8_3.x86_64") == True
```
r0x0d, line 741: I believe you are missing a `assert` call here to actually validate that the return is actually `True`. 

```suggestion
    assert _bad_kernel_package_signature("4.18.0-240.22.1.el8_3.x86_64") == True
```

Andrew-ang9: you are right about that, thank you for catching that for me! I'll get that fixed and pushed.

convert2rhel/checks.py
```diff
@@ -422,10 +422,19 @@ def _bad_kernel_version(kernel_release):
...
     """Return True if the booted kernel is not signed by the original OS vendor, i.e. it's a custom kernel."""
-    kernel_pkg = run_subprocess(
-        ["rpm", "-qf", "--qf", "%{NAME}", "/boot/vmlinuz-%s" % kernel_release], print_output=False
-    )[0]
+    vmlinuz_path = "/boot/vmlinuz-%s" % kernel_release
+    kernel_pkg, return_code = run_subprocess(["rpm", "-qf", "--qf", "%{NAME}", vmlinuz_path], print_output=False)
     logger.debug("Booted kernel package name: {0}".format(kernel_pkg))
+
+    if return_code == 1:
+        os_vendor = system_info.name.split()[0]
```
r0x0d, line 430: You have the same line that gets the os_vendor on line 442, I believe that you can remove that call from line 442 and move your `os_vendor` variable to be above that if statement `if return_code == 1`, like this: 

```suggestion
        # e.g. Oracle Linux Server -> Oracle
        os_vendor = system_info.name.split()[0]
        if return_code == 1:
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -730,6 +730,20 @@ def test_bad_kernel_fingerprint(
...
+
+def test_kernel_not_installed(caplog, monkeypatch):
+    run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(" ", 1))
+    monkeypatch.setattr(checks, "run_subprocess", run_subprocess_mocked)
+    assert _bad_kernel_package_signature("4.18.0-240.22.1.el8_3.x86_64")
+    log_message = (
+        "The booted kernel /boot/vmlinuz-4.18.0-240.22.1.el8_3.x86_64 is not owned by any installed package."
+        " It needs to be owned by a package signed by CentOS."
+    )
+    assert log_message in caplog.text
```
r0x0d, line 746: Just a side note, I tried to run this test locally and it failed to me. 

```bash
(.venv3) [r0x0d@fedora convert2rhel]$ pytest -sk test_kernel_not_installed
=============================================================== test session starts ================================================================
platform linux -- Python 3.10.4, pytest-7.0.1, pluggy-1.0.0
rootdir: /home/r0x0d/Workspace/r0x0d/convert2rhel, configfile: pytest.ini, testpaths: convert2rhel/unit_tests
plugins: cov-3.0.0
collected 344 items / 343 deselected / 1 selected                                                                                                  

convert2rhel/unit_tests/checks_test.py F

===================================================================== FAILURES =====================================================================
____________________________________________________________ test_kernel_not_installed _____________________________________________________________

caplog = <_pytest.logging.LogCaptureFixture object at 0x7f2ab0015510>, monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f2ab0015600>

    def test_kernel_not_installed(caplog, monkeypatch):
        run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(" ", 1))
        monkeypatch.setattr(checks, "run_subprocess", run_subprocess_mocked)
>       assert _bad_kernel_package_signature("4.18.0-240.22.1.el8_3.x86_64")

convert2rhel/unit_tests/checks_test.py:741: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

kernel_release = '4.18.0-240.22.1.el8_3.x86_64'

    def _bad_kernel_package_signature(kernel_release):
        """Return True if the booted kernel is not signed by the original OS vendor, i.e. it's a custom kernel."""
        vmlinuz_path = "/boot/vmlinuz-%s" % kernel_release
        kernel_pkg, return_code = run_subprocess(["rpm", "-qf", "--qf", "%{NAME}", vmlinuz_path], print_output=False)
        logger.debug("Booted kernel package name: {0}".format(kernel_pkg))
    
        if return_code == 1:
>           os_vendor = system_info.name.split()[0]
E           AttributeError: 'NoneType' object has no attribute 'split'

convert2rhel/checks.py:430: AttributeError
---------------------------------------------------------------- Captured log call -----------------------------------------------------------------
FILE     convert2rhel.checks:checks.py:427 Booted kernel package name:
================================================================= warnings summary =================================================================
../../../../../usr/lib64/python3.10/subprocess.py:956
../../../../../usr/lib64/python3.10/subprocess.py:956
../../../../../usr/lib64/python3.10/subprocess.py:956
  /usr/lib64/python3.10/subprocess.py:956: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
    self.stdout = io.open(c2pread, 'rb', bufsize)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================= short test summary info ==============================================================
FAILED convert2rhel/unit_tests/checks_test.py::test_kernel_not_installed - AttributeError: 'NoneType' object has no attribute 'split'
================================================== 1 failed, 343 deselected, 3 warnings in 0.33s ===================================================
```

Maybe `system_info` is being initialized in other tests in the `checks_test.py` and thus giving the impression that this one is passing normally in the CI?

r0x0d: What do you think @abadger?

r0x0d (commented):
Pretty good work, @Andrew-ang9! I just left some minor comments for you that might improve the readability of the code in general!

convert2rhel/unit_tests/checks_test.py
abadger: I didn't trace the unittest code but I think @r0x0d is correct.  Monkeypatching system_info.name inside of this test function is probably what we should do.

r0x0d: Another easy approach to fix this is to call the @centos8 decorator in the test function, for example. This way we have the system_info mocked and ready to go.

Andrew-ang9: I have made the change to this comment as well!

SpyTec (dismissed):

convert2rhel/unit_tests/checks_test.py
Andrew-ang9: I have taken care of this comment! thank you both for the input!

danmyway:
@SpyTec, @bocekm I suppose we will not need an integration test for this, right?

bocekm (dismissed):

abadger:
I agree with @bocekm.  The additonal case is likely to be encountered extremely infrequently.

abadger (approved):

danmyway:
Alright, thank you, @bocekm and @abadger !
I will keep it in backlog just in case.
If you could rebase to upstream @Andrew-ang9, please, I think we're ready for merge then.

PR #427, opened 2022-02-08
Add integration test to verify requiring root

danmyway:
Run as a basic user and check for the program to exit as it is required to run as root.

Relates to https://github.com/oamg/convert2rhel/pull/81.

### DIFF
--- a/tests/integration/tier0/check-privileges/test_user_privileges.py
+++ b/tests/integration/tier0/check-privileges/test_user_privileges.py
@@ -1,0 +1,17 @@
+import os
+
+
+def test_check_user_privileges(shell):
+    user = "testuser"
+    # Create non-root user if not created already
+    os.system(f"useradd '{user}'")
+    # Set user to non-root entity 'testuser' and run c2r
+    result = shell("runuser -l testuser -c 'convert2rhel'")
+    # Check the program exits as it is required to be run by root
+    assert result.returncode != 0
+    # Check the program exits for the correct reason
+    assert (
+        result.output == "The tool needs to be run under the root user.\n" "\n" "No changes were made to the system.\n"
+    )
+    # Delete testuser (if present)
+    os.system(f"userdel -r '{user}'")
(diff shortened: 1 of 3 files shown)


tests/integration/tier0/check-privileges/test_user_privileges.py
```diff
@@ -0,0 +1,8 @@
+def test_check_user_privileges(shell):
+    # Create non-root user 'testuser' if not created already
+    shell("if cat /etc/passwd/ | grep -q 'testuser'; then :; else useradd testuser; fi")
```
mkluson, line 3: Could you implement the logic in python and call shell only where it is necessary

mkluson (dismissed):

tests/integration/tier0/check-privileges/test_user_privileges.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+
+def test_check_user_privileges(shell):
+    user = "testuser"
+    # Create non-root user if not created already
+    os.system(f"useradd '{user}'")
+    # Set user to non-root entity 'testuser' and run c2r
+    # the program should exit as it is required to be run by root
+    assert shell("runuser -l testuser -c 'convert2rhel'").returncode != 0
```
kokesak, line 10: I would like to make sure that we know exactly why convert2rhel ended with non zero return value. Something like this:
```suggestion
    result = shell("runuser -l testuser -c 'convert2rhel'")
    assert result.returncode != 0
    assert result.output = "convert2rhel requires root privileges" 
```
I don't know the exact string, but something that would check that the correct behavior.

danmyway: Good point, I will change that.

mkluson (approved):

kokesak (approved):
LGTM!

kokesak (approved):
All tests passed! Thanks, Good job @danmyway

PR #428, opened 2022-02-09
Split ansible playbook into setup and executing stages.

abadger:
We may start pointing end users at these playbooks to install and run
convert2rhel.  An end user may want to setup the environment separately
from running convert2rhel so we need to split the two steps into
separate playbooks.

The idea of modifying CentOS 8.5 systems to use vault repositories prior
to running convert2rhel was the prompt for doing this.

Note, @tabowling, this PR is only for the split, not the other things that we talked about doing in the meeting.  IIRC, those were:

* Use activation key instead of rhsm username and password
  * Point users at documentation on setting an activation key
  * Add the ability to give the activation key in a file instead of at the command line.
* If the user is on CentOS 8.5, make sure their repositories are pointed at vault rather than mirrors.centos.org
  * Change their repository files if necessary

/CC @tabowling @bocekm

### DIFF
--- a/ansible/run-convert2rhel.yml
+++ b/ansible/run-convert2rhel.yml
@@ -1,0 +1,46 @@
+- name: Run setup tasks
+  import_playbook: setup-convert2rhel.yml
+
+- hosts: all
+  become: yes
+  strategy: free
+
+  # Modify these to be your red hat subscription manager username and password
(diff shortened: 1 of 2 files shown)


bocekm:
Thanks, @abadger. Integration tests don't cover this so it can be merged even when the tests are failing.

PR #447, opened 2022-02-23
[RHELC-45] Fix an error logged when removing an RHSM cert

Andrew-ang9:
RHSM was throwing an logging error when removing the current certs on the system, this error was not supposed to be displayed at this moment in the conversion because it might've not been installed yet.

If you want to test this fix, you can follow the steps in the Jira ticket and observe that the error message doesn't appear anymore.

Jira: [RHELC-45](https://issues.redhat.com/browse/RHELC-45)

### DIFF
--- a/convert2rhel/cert.py
+++ b/convert2rhel/cert.py
@@ -74,7 +74,7 @@ class SystemCert(object):
             loggerinst.info("Certificate %s removed" % self._target_cert_path)
         except OSError as err:
             if err.errno == 2:
-                """Resolves RHSM error when removing certs, as the system migth not have intalled any certs yet"""
-                loggernist.debug("No RHSM certificates found to be removed")
+                # Resolves RHSM error when removing certs, as the system might not have installed any certs yet
+                loggerinst.debug("No RHSM certificates found to be removed")
             else:
                 loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
--- a/convert2rhel/cert.py
+++ b/convert2rhel/cert.py
@@ -73,4 +73,7 @@ class SystemCert(object):
             os.remove(self._target_cert_path)
             loggerinst.info("Certificate %s removed" % self._target_cert_path)
         except OSError as err:
-            loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
+            if err.errno == 2:
+                pass
+            else:
+                loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
--- a/convert2rhel/cert.py
+++ b/convert2rhel/cert.py
@@ -74,6 +74,7 @@ class SystemCert(object):
             loggerinst.info("Certificate %s removed" % self._target_cert_path)
         except OSError as err:
             if err.errno == 2:
-                pass
+                """Resolves RHSM error when removing certs, as the system migth not have intalled any certs yet"""
+                loggernist.debug("No RHSM certificates found to be removed")
             else:
                 loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
--- a/convert2rhel/cert.py
+++ b/convert2rhel/cert.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import errno
 import logging
 import os
 import shutil
@@ -73,7 +74,7 @@ class SystemCert(object):
             os.remove(self._target_cert_path)
             loggerinst.info("Certificate %s removed" % self._target_cert_path)
         except OSError as err:
-            if err.errno == 2:
+            if err.errno == errno.ENOENT:
                 # Resolves RHSM error when removing certs, as the system might not have installed any certs yet
                 loggerinst.debug("No RHSM certificates found to be removed")
             else:
--- a/convert2rhel/cert.py
+++ b/convert2rhel/cert.py
@@ -76,6 +76,6 @@ class SystemCert(object):
         except OSError as err:
             if err.errno == errno.ENOENT:
                 # Resolves RHSM error when removing certs, as the system might not have installed any certs yet
-                loggerinst.debug("No RHSM certificates found to be removed")
+                loggerinst.info("No RHSM certificates found to be removed.")
             else:
                 loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -90,12 +90,12 @@ class TestCert(unittest.TestCase):
         shutil.rmtree(unit_tests.TMP_DIR)
 
 
-def test_remove_cert(caplog, system_cert):
-    cert_filename = system_cert._target_cert_path
+def test_remove_cert(caplog, system_cert_with_target_path):
+    cert_filename = system_cert_with_target_path._target_cert_path
     with open(cert_filename, "wb") as cert_file:
         cert_file.write(b"some content")
 
-    system_cert.remove()
+    system_cert_with_target_path.remove()
 
     assert "/filename removed" in caplog.messages[-1]
 
@@ -113,12 +113,14 @@ def test_remove_cert(caplog, system_cert):
         (OSError(13, "[Errno 13] Permission denied: '/tmpdir/certfile'"), "Permission denied:"),
     ),
 )
-def test_remove_cert_error_conditions(error_condition, expected_text_in_logs, caplog, monkeypatch, system_cert):
+def test_remove_cert_error_conditions(
+    error_condition, expected_text_in_logs, caplog, monkeypatch, system_cert_with_target_path
+):
     def fake_os_remove(path):
         raise error_condition
 
     monkeypatch.setattr(os, "remove", fake_os_remove)
 
-    system_cert.remove()
+    system_cert_with_target_path.remove()
 
     assert expected_text_in_logs != caplog.messages[-1]
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -133,8 +133,6 @@ def test_remove_cert_error_conditions(error_condition, expected_text_in_logs, ca
 
     monkeypatch.setattr(os, "remove", fake_os_remove)
 
-    # Execute the tested code
     remove_cert_setup["sys_cert_instance"].remove()
 
-    # Assert that the test passed
     assert expected_text_in_logs != caplog.messages[-1]
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -98,24 +98,27 @@ def test_remove_cert(caplog, system_cert_with_target_path):
 
     system_cert_with_target_path.remove()
 
-    assert "Certificate /filename removed" in caplog.messages[-1]
+    assert "Certificate %s removed" % cert_file_path in caplog.messages[-1]
 
 
 @pytest.mark.parametrize(
     (
         "error_condition",
-        "expected_text_in_logs",
+        "text_not_expected_in_logs",
     ),
     (
         (
             OSError(2, "No such file or directory"),
             "No such file or directory",
         ),
-        (OSError(13, "[Errno 13] Permission denied: '/tmpdir/certfile'"), "Permission denied:"),
+        (
+            OSError(13, "[Errno 13] Permission denied: '/tmpdir/certfile'"),
+            "OSError(13): Permission denied: '/tmpdir/certfile'",
+        ),
     ),
 )
 def test_remove_cert_error_conditions(
-    error_condition, expected_text_in_logs, caplog, monkeypatch, system_cert_with_target_path
+    error_condition, text_not_expected_in_logs, caplog, monkeypatch, system_cert_with_target_path
 ):
     def fake_os_remove(path):
         raise error_condition
@@ -124,4 +127,5 @@ def test_remove_cert_error_conditions(
 
     system_cert_with_target_path.remove()
 
-    assert expected_text_in_logs != caplog.messages[-1]
+    for message in caplog.messages:
+        assert text_not_expected_in_logs not in message
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -90,26 +90,12 @@ class TestCert(unittest.TestCase):
         shutil.rmtree(unit_tests.TMP_DIR)
 
 
-@pytest.fixture
-def remove_cert_setup(monkeypatch, tmpdir):
-    tmp_file = tmpdir / "filename"
-
-    monkeypatch.setattr(cert.SystemCert, "_get_cert", value=mock.Mock(return_value=("anything", "anything")))
-    monkeypatch.setattr(cert.SystemCert, "_get_target_cert_path", value=mock.Mock(return_value=str(tmp_file)))
-
-    sys_cert = cert.SystemCert()
-
-    return {
-        "cert_path_obj": tmp_file,
-        "sys_cert_instance": sys_cert,
-    }
-
-
 def test_remove_cert(caplog, remove_cert_setup):
-    cert_file = remove_cert_setup["cert_path_obj"]
-    cert_file.write(b"some content")
+    cert_file = system_cert._target_cert_path
+    with open(cert_filename, 'wb') as  cert_file:
+        cert_file.write(b'some content')
 
-    remove_cert_setup["sys_cert_instance"].remove()
+    system_cert[sys_cert].remove()
 
     assert "/filename removed" in caplog.messages[-1]
 
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -98,7 +98,7 @@ def test_remove_cert(caplog, system_cert_with_target_path):
 
     system_cert_with_target_path.remove()
 
-    assert "/filename removed" in caplog.messages[-1]
+    assert "Certificate /filename removed" in caplog.messages[-1]
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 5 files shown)


convert2rhel/cert.py
```diff
@@ -75,4 +75,7 @@ def remove(self):
             os.remove(self._target_cert_path)
             loggerinst.info("Certificate %s removed" % self._target_cert_path)
         except OSError as err:
-            loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
+            if err.errno == 2:
```
r0x0d, line 78: We could put a comment above this check to indicate that the `err.errno == 2` means that the file was not found

convert2rhel/cert.py
```diff
@@ -75,4 +75,7 @@ def remove(self):
...
+                pass
```
r0x0d, line 79: Instead of just calling `pass` here, shouldn't be better to also add a logging? I mean, the rollback message prior to this function is: 

```console
TASK - [Rollback: Removing installed RHSM certificate] ********************
```

we could output a log like this: 

```python
loggerinst.warning("No RHSM certificates found to be removed.")
pass
```

This way the user knows that is something happening in the function and not only "trying to remove something and silently failing". 

The visualization would be something like: 

```console
TASK - [Rollback: Removing installed RHSM certificate] ********************
WARNING - No RHSM certificates found to be removed.
....
```

convert2rhel/unit_tests/cert_test.py
```diff
@@ -98,30 +98,49 @@ def test_install_cert(self):
...
+
+    monkeypatch.setattr(cert.SystemCert, "_get_cert", value=mock.Mock(return_value=("anything", "anything")))
+    monkeypatch.setattr(cert.SystemCert, "_get_target_cert_path", value=mock.Mock(return_value=str(tmp_file)))
+
+    sys_cert = cert.SystemCert()
+
+    return {
+        "cert_path_obj": tmp_file,
+        "sys_cert_instance": sys_cert,
+    }
```
r0x0d, line 113: What do you think about moving this fixture to a `conftest.py` file? I think it can be well organized to separate the fixture from an actual test code.

r0x0d (changes requested):

convert2rhel/cert.py
Andrew-ang9: yeah that sounds like a better way to handle this thank you!

convert2rhel/unit_tests/cert_test.py
abadger: I think it's good to put a fixture into conftest.py when we think we can reuse it in other unit tests but we shouldn't put it there if it can only be used in a single unit test file.  In this case, I see that there are unittests in main_test.py that (once they are ported to pytest) could make use of this so it makes sense to move it to me.

convert2rhel/cert.py
Andrew-ang9: with this in this Jira ticket one of the acceptance criteria's is not logging an error when the RHSM cert is not installed during rollback. so we don't need to log any messages for this error!

convert2rhel/cert.py
SpyTec: @Andrew-ang9 the conversation is resolved but it was never changed. Adding a comment stating the reason why it's skipped will be ideal, as currently if we encounter this further down the line we'd have no clue what is going on and would have to look up the return codes of certificate

And although we shouldn't log an error, adding a log at debug level is preferable rather than passing

convert2rhel/unit_tests/cert_test.py
```diff
@@ -98,30 +98,35 @@ def test_install_cert(self):
         shutil.rmtree(unit_tests.TMP_DIR)
 
 
+def test_remove_cert(caplog, remove_cert_setup):
+    cert_file = system_cert._target_cert_path
+    with open(cert_filename, 'wb') as  cert_file:
+        cert_file.write(b'some content')
+
+    system_cert[sys_cert].remove()
```
SpyTec, line 106: Is `sys_cert` supposed `cert_file`?

convert2rhel/unit_tests/cert_test.py
```diff
@@ -98,30 +98,35 @@ def test_install_cert(self):
...
-    sys_cert = cert.SystemCert()
-    sys_cert.remove()
-    assert expected_text_in_logs in caplog.messages[-1]
+def test_remove_cert_error_conditions(error_condition, expected_text_in_logs, caplog, monkeypatch, remove_cert_setup):
+    def fake_os_remove(path):
+        raise error_condition
+
+    monkeypatch.setattr(os, "remove", fake_os_remove)
+
+    remove_cert_setup["sys_cert_instance"].remove()
```
SpyTec, line 130: Where does `sys_cert_instance` come from in this case?

SpyTec (changes requested):

convert2rhel/unit_tests/cert_test.py
Andrew-ang9: sys_cert was some left over code from the old remove_cert_setup() as it returned a dictionary with "sys_cert_instance": sys_cert as one of it's elements, I have already fixed this in https://github.com/oamg/convert2rhel/pull/447/commits/204a0b5aa81e9b95bdebeb71da918e2559cf283f this commit!
Thank you for pointing this out to me!

convert2rhel/unit_tests/cert_test.py
Andrew-ang9: sys_cert_instance is also leftover code from the old function remove_cert_setup and is apart of the dicionary that is returned from that old function. I have already fixed this in https://github.com/oamg/convert2rhel/pull/447/commits/204a0b5aa81e9b95bdebeb71da918e2559cf283f this commit!

convert2rhel/unit_tests/cert_test.py
```diff
@@ -98,30 +98,35 @@ def test_install_cert(self):
         shutil.rmtree(unit_tests.TMP_DIR)
 
 
+def test_remove_cert(caplog, remove_cert_setup):
+    cert_file = system_cert._target_cert_path
+    with open(cert_filename, 'wb') as  cert_file:
```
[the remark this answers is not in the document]
Andrew-ang9: done!

convert2rhel/unit_tests/conftest.py
```diff
@@ -57,7 +57,19 @@ def pkg_root(is_py2):
...
+@pytest.fixture
+def system_cert(monkeypatch, tmpdir):
+    tmp_file = tmpdir / "filename"
+
+    monkeypatch.setattr(cert.SystemCert, "_get_cert", value=mock.Mock(return_value=("anything", "anything")))
+    monkeypatch.setattr(cert.SystemCert, "_get_target_cert_path", value=mock.Mock(return_value=str(tmp_file)))
+
+    sys_cert = cert.SystemCert()
+
+    return sys_cert
```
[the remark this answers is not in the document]
Andrew-ang9: I have changed all to old references to remove_cert_setup and also changed the name of the function! and I also changed it to where the fixture now gets the tmp_file from the test function now!

convert2rhel/unit_tests/cert_test.py
```diff
@@ -98,30 +98,38 @@ def test_install_cert(self):
...
+    monkeypatch.setattr(os, "remove", fake_os_remove)
 
-    monkeypatch.setattr(cert.SystemCert, "_get_cert", value=mock.Mock(return_value=("anything", "anything")))
-    monkeypatch.setattr(cert.SystemCert, "_get_target_cert_path", value=mock.Mock(return_value=str(tmp_file)))
+    system_cert_with_target_path.remove()
 
-    sys_cert = cert.SystemCert()
-    sys_cert.remove()
-    assert expected_text_in_logs in caplog.messages[-1]
+    assert expected_text_in_logs != caplog.messages[-1]
```
r0x0d, line 135: I was wondering about these tests, I think that we could tweak it a little bit and instead of checking against a specific message, we could check for the message we want to achieve. 

Let me give a more concrete example here: 

Right now, we are asserting that a certain message does not appear after removing the certificates, although this is usually an okay case, I believe for this test specific we should do the inverse and check that we have the message we want. This way I feel that we can be more assertive and be sure that the test/code is working properly. 

## What are we doing right now

```python
@pytest.mark.parametrize(
    (
        "create_file",
        "error_condition",
        "expected_text_in_logs",
    ),
    (
        (True, "/filename removed"),
        (False, "No such file or directory"),
        (
            OSError(2, "No such file or directory"),
            "No such file or directory",
        ),
        (OSError(13, "[Errno 13] Permission denied: '/tmpdir/certfile'"), "Permission denied:"),
    ),
)
...
assert expected_text_in_logs != caplog.messages[-1]
```
The parameter `expected_text_in_logs` doesn't have a really good usage here, since it could be anything like `"aaa"` and the tests would pass without a problem (And, as I said, this can be okay in some scenarios). But we are not necessarily asserting that the message being sent to the user is the correct one, right?  We are just assuming that whatever it outputs is different from the `expected_text_in_logs` message. 

## What we could do (suggestion) 

I believe to change this, we could only do a minimal change to the test, and instead of checking that the `expected_text_in_logs` is not present in the log, we can check that the `expected_text_in_logs` is actually present. This way, we can be sure that, for example, if the cert was not located in the system during the removal, we are outputting the right message to the user with a proper message. 

So, my suggestion is something like this: 

```python
@pytest.mark.parametrize(
    (
        "create_file",
        "error_condition",
        "expected_text_in_logs",
    ),
    (
        (True, "/filename removed"),
        (False, "No such file or directory"),
        (
            OSError(2, "No such file or directory"),
            "No RHSM certificates found to be removed",
        ),
        (OSError(13, "[Errno 13] Permission denied: '/tmpdir/certfile'"), "OSError(13): Permission denied: '/tmpdir/certfile'"),
    ),
)
...
assert expected_text_in_logs in caplog.messages[-1]
```

Also, the test becomes more complete since the issue in Jira says that the message `ERROR - OSError(2): No such file or directory` should not appear in that cert removal. 

----

What do you think about that? It's just a suggestion that I thought it was valid to point out!

convert2rhel/unit_tests/conftest.py
```diff
@@ -60,6 +66,25 @@ def setup_logger(tmpdir):
...
+        temporary_filename = mark.args[0]
+
+    tmp_file = tmpdir / temporary_filename
+
+    monkeypatch.setattr(cert.SystemCert, "_get_cert", value=mock.Mock(return_value=("anything", "anything")))
+    monkeypatch.setattr(cert.SystemCert, "_get_target_cert_path", value=mock.Mock(return_value=str(tmp_file)))
+
+    sys_cert = cert.SystemCert()
+
+    return sys_cert
```
r0x0d, line 85: I'm just curious about this fixture and why we are going the route to create a new marker instead of using a parametrized fixture here... Don't get me wrong, I think this is cool and I learned something different today... 

If using a marker is indeed the easiest option here (Or actually, the better option) what do you think about making a little comment above the line `72` regarding the use of `pytest.mark`? I believe not many people will take a look at this and guess what's going on on the fly. Also, could be good to point out the [pytest docs](https://docs.pytest.org/en/7.1.x/example/markers.html) to reference what a custom marker is, this way you don't need to type in all of the information by hand. 

_Note_: I can see the point here about using a custom marker to make it easier in the tests by just calling a `@pytest.mark.cert_filename` and it's done, but also, I think that a parametrized fixture would be simpler to use. Anyway, just curious about this one!

r0x0d (commented):
In an overall, pretty good code, @Andrew-ang9! I liked some of the things you made in the unit tests/codebase. 

I just left some comments and a few suggestions for you, but let me know if you want to discuss my comments/suggestions later!

convert2rhel/unit_tests/conftest.py
Andrew-ang9: Hey @r0x0d I'm not exactly sure why it is a better reason to use the marker over using paramertrized, @abadger helped me out with this code and I think he could supply a better answer then I can.

abadger: I like to limit use of parametrize to just things that need to run the test multiple times with different data each time.  For constant data, I pass the data another way (and it looks like mark is the way that pytest recommends).

The problem with using parametrize for constant data is that making sure that parametrize's data covers every permutation of the data is difficult.  Throwing a constant into that will leave people wondering whether the constant should vary just like the other pieces of data.  IMHO, it's better to move this information into something other than the parametrize data.

kokesak:
Once you will have time, please rebase @Andrew-ang9

convert2rhel/cert.py
```diff
@@ -73,4 +73,8 @@ def remove(self):
             os.remove(self._target_cert_path)
             loggerinst.info("Certificate %s removed" % self._target_cert_path)
         except OSError as err:
-            loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
+            if err.errno == 2:
```
vinzenz, line 76: I'd for readability for this:
```suggestion
            if err.errno == errno.ENOENT:
```

Needs of course, `errno` to be imported.

Andrew-ang9: Thank you for the help with this Vinzenz! I will implement this into the code!

convert2rhel/unit_tests/cert_test.py
Andrew-ang9: My applogies for taking so long to reply on this comment! Thank you so much for sharing this very insightful comments on this PR! It is very helpful in making sure that I understand what it is that you are saying for this comment!

danmyway:
Rebased with integration test included.

tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks.py
```diff
@@ -71,6 +72,7 @@ def test_c2r_latest_newer(convert2rhel):
 
     # Clean up
     os.system(f"cp /tmp/__init__.py {PATH_TO_VERSION}")
+    os.system(f"cat {PATH_TO_VERSION}")
```
kokesak, line 75: This is some debugging leftover I presume?

tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks.py
```diff
@@ -86,6 +88,7 @@ def test_c2r_latest_older_inhibit(convert2rhel):
...
```
kokesak, line 91: The same here

tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks.py
```diff
@@ -112,4 +115,16 @@ def test_c2r_latest_older_unsupported_version(convert2rhel):
...
```
kokesak, line 118: and here?

kokesak (changes requested):

tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks.py
Andrew-ang9: Ok, thank you for spotting those, I'll make sure to remove those lines of clean up code!

danmyway: Oh, good catch! You're right, it's gone now

tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks.py
```diff
@@ -113,3 +114,14 @@ def test_c2r_latest_older_unsupported_version(convert2rhel):
...
+
+def test_rhsm_error_logged(convert2rhel):
+    """
+    Test if the OSError for RHSM certificate being removed
+    is not being logged in cases the certificate is not installed yet.
+    """
+    with convert2rhel("--debug --no-rpm-va") as c2r:
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("n")
+        assert c2r.expect("DEBUG - No RHSM certificates found to be removed") == 0
```
kokesak, line 127: Looks good. Just in case I would probably add a check that would make sure that the `ERROR - OSError(2): No such file or directory` message is not shown (for regression reasons). Maybe the easiest way to do it would be to look into the `/var/log/convert2rhel/convert2rhel.log` file and check that this line is not present.

kokesak (changes requested):

tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks.py
danmyway: You are right, thank you for noting that! 
Regression check included now

kokesak (dismissed):
The integration tests are done. Nice job @danmyway !

danmyway:
All tests have passed, good job @Andrew-ang9 !
Are we ready for final approval?

r0x0d (dismissed):

SpyTec (dismissed):

SpyTec (dismissed):

bocekm (changes requested):

convert2rhel/cert.py
```diff
@@ -73,4 +74,8 @@ def remove(self):
             os.remove(self._target_cert_path)
             loggerinst.info("Certificate %s removed" % self._target_cert_path)
         except OSError as err:
-            loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
+            if err.errno == errno.ENOENT:
+                # Resolves RHSM error when removing certs, as the system might not have installed any certs yet
+                loggerinst.debug("No RHSM certificates found to be removed")
```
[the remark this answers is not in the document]
bocekm: Oh, I see why. It should be `loggerinst.info`, not `loggerinst.debug` so that it's printed even without using the `--debug` option. It's like that for all other TASK-type messages for the reason that we want to have at least some kind of status message printed under each TASK before getting to another TASK.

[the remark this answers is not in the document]
bocekm: I did the change under https://github.com/oamg/convert2rhel/pull/447/commits/adaf87ba45573fc33b8f4416bee2a871e9f15a90.

PR #449, opened 2022-02-24
Fix lazy evaluation of string logs for loglevel.FILE

abadger:
For log messages, you can do lazy evaluation of the format string to
occassionally boost performance when a log message is not used.  For
instance:

``` python
  class Hypothetical(object):
      def __str__(self):
          return self.calculate_fibonacci_10000()
  log.debug("10,000th fibonacci number is: %s", Hypothetical())
```
The format string will not be interpolated (and thus the 10,000th
fibonacci number will not be calculated) unless we are actually going to
log the debug() output.

The code in commit b870460 broke that because it calculated its own
format string and args for the message, interfering with whatever was
passed in via function parameter.  This change fixes that by moving the
formatting pieces of b870460 into the CustomFormatter class.  By that
time, the message string has already been interpolated so any changes to
the message string will not interfere.  Additionally, in the formatter
class, those pieces can be performed by standard logging.Formatter
features so we don't need the custom code to do that anymore.

OAMG-6571 #in-progress

### DIFF
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -33,7 +33,6 @@ import sys
 from time import gmtime, strftime
 
 from convert2rhel.toolopts import tool_opts
-from convert2rhel.utils import format_msg_with_datetime
 
 
 LOG_DIR = "/var/log/convert2rhel"
@@ -153,7 +152,7 @@ def _debug(self, msg, *args, **kwargs):
         if tool_opts.debug:
             self._log(logging.DEBUG, msg, args, **kwargs)
         else:
-            self._log(LogLevelFile.level, format_msg_with_datetime(msg, "debug"), args, **kwargs)
+            self._log(LogLevelFile.level, msg, args, **kwargs)
 
 
 class bcolors:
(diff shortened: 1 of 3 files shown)


convert2rhel/unit_tests/logger_test.py
```diff
@@ -33,52 +33,57 @@ def test_logger_handlers(tmpdir, caplog, read_std, is_py2, capsys):
...
     except AssertionError:
         if not is_py2:
-            assert "debug entry 1" in stdouterr_err
+            assert "debug entry 1: data" in stdouterr_err
         else:
             # this workaround is not working for py2 - passing
             pass
     tool_opts.debug = False
-    logger.debug("debug entry 2")
+    logger.debug("debug entry 2: %s", "data")
```
abadger, line 68: Note: Without the change to logger.py, this is the specific logger call that was broken (calling `logger.debug()` with a lazily evaluated string when `tool_opts.debug = False`)

convert2rhel/logger.py
```diff
@@ -183,7 +182,7 @@ def format(self, record):
             new_fmt = fmt_orig if self.color_disabled else bcolors.OKGREEN + fmt_orig + bcolors.ENDC
             self._fmt = new_fmt
             self.datefmt = "%m/%d/%Y %H:%M:%S"
-        elif record.levelno in [logging.INFO, LogLevelFile.level]:
+        elif record.levelno in [logging.INFO]:
```
abadger, line 185: Note, after removing this, `self._fmt` and `self.datefmt` will be set by the `else:` clause (the same as `logging.DEBUG`)

SpyTec:
@abadger could you create a Jira issue about it?

abadger:
Jira issue created: OAMG-6571

SpyTec:
OAMG-6571 #in-progress

SpyTec:
OAMG-6571 #comment Testing smart commits

PR #451, opened 2022-02-25
[RHELC-411, RHELC-413] Add the config file option

hosekadam:
Add the possibility to give secrets (password or activation key) via configuration file. Configuration file can be 1) user specified and passed via command line or is in 2) ~/convert2rhel.ini, 3) /etc/convert2rhel.ini. The numbers shows priority, 1) is the highest.  The structure of the file is copied to /etc/convert2rhel.ini during installation.

Secrets passed by CLI have precedence and are merged with the secrets in config files. If password is passed via CLI and activation key is in config file, the activation key isn't used. If it would be, during the activation process would be used the activation key and the precedence of secrets from CLI wouldn't be preserved. If there are both, the activation key has priority.

Also contains warnings if user specifies more than one authentication method. And warning if some method has lower priority than others. The priority is: password file > CLI > config file.

[RHELC-411](https://issues.redhat.com/browse/RHELC-411)
[RHELC-413](https://issues.redhat.com/browse/RHELC-413)

### DIFF
--- a/convert2rhel/toolopts.py
+++ b/convert2rhel/toolopts.py
@@ -15,14 +15,21 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import copy
 import logging
 import optparse
+import os
 import sys
+
+from six.moves import configparser
 
 from convert2rhel import __version__, utils
 
 
 loggerinst = logging.getLogger(__name__)
+
+# Paths for configuration files
+CONFIG_PATHS = ["~/.convert2rhel.ini", "/etc/convert2rhel.ini"]
 
 
 class ToolOpts(object):
@@ -30,6 +37,7 @@ class ToolOpts(object):
         self.debug = False
         self.username = None
         self.password_file = None
+        self.config_file = None
         self.password = None
         self.no_rhsm = False
         self.enablerepo = []
@@ -49,6 +57,15 @@ class ToolOpts(object):
         # set True when credentials (username & password) are given through CLI
         self.credentials_thru_cli = False
 
+    def set_opts(self, supported_opts):
+        """Set ToolOpts data using dict with values from config file.
+
+        :param supported_opts: Supported options in config file
+        """
+        for key, value in supported_opts.items():
+            if value is not None and hasattr(self, key):
+                setattr(self, key, value)
+
 
 class CLI(object):
     def __init__(self):
@@ -63,6 +80,7 @@ class CLI(object):
             "  convert2rhel [-h]\n"
             "  convert2rhel [--version]\n"
             "  convert2rhel [-u username] [-p password | -f pswd_file] [--pool pool_id | -a] [--disablerepo repoid]"
+            "  convert2rhel [-c | --config-file conf_file]"
             " [--enablerepo repoid] [--serverurl url] [--keep-rhsm] [--no-rpm-va] [--debug] [--restart]"
             " [--disable-colors] [-y]\n"
             "  convert2rhel [--no-rhsm] [--disablerepo repoid]"
@@ -159,7 +177,7 @@ class CLI(object):
             help="File containing"
             " password for the subscription-manager in the plain"
             " text form. It's an alternative to the --password"
-            " option.",
+            " option. Deprecated, use --config-file instead.",
         )
         group.add_option(
             "-k",
@@ -178,6 +196,13 @@ class CLI(object):
             " organizations is possible to obtain by running"
             " 'subscription-manager orgs'. From the listed pairs"
             " Name:Key, use the Key here.",
+        )
+        group.add_option(
+            "-c",
+            "--config-file",
+            help="Configuration file containing"
+            " password or activation key for subscription-manager."
+            " Example of this file in /etc/convert2rhel.ini",
         )
         group.add_option(
             "-a",
@@ -263,8 +288,19 @@ class CLI(object):
         parsed_opts, _ = self._parser.parse_args()
 
         global tool_opts  # pylint: disable=C0103
+
         if parsed_opts.debug:
             tool_opts.debug = True
+
+        # Processing the configuration file
+        conf_file_opts = options_from_config_files(parsed_opts.config_file)
+        ToolOpts.set_opts(tool_opts, conf_file_opts)  # pylint: disable=E0601
+        config_opts = copy.copy(tool_opts)
+        tool_opts.config_file = parsed_opts.config_file
+        # corner case: password on CLI or in password file and activation-key in the config file
+        # password from CLI has precedence and activation-key must be deleted (unused)
+        if config_opts.activation_key and (parsed_opts.password or parsed_opts.password_from_file):
+            tool_opts.activation_key = None
 
         if parsed_opts.disable_colors:
             tool_opts.disable_colors = True
@@ -279,6 +315,7 @@ class CLI(object):
             tool_opts.password = parsed_opts.password
 
         if parsed_opts.password_from_file:
+            loggerinst.warning("Deprecated. Use -c | --config-file instead.")
             tool_opts.password_file = parsed_opts.password_from_file
             tool_opts.password = utils.get_file_content(parsed_opts.password_from_file)
 
@@ -325,6 +362,19 @@ class CLI(object):
         if parsed_opts.org:
             tool_opts.org = parsed_opts.org
 
+        # Checks of multiple authentication sources
+        if tool_opts.password and tool_opts.activation_key:
+            loggerinst.warning("Set only one of password or activation key. Activation key take precedence.")
+
+        if parsed_opts.password and parsed_opts.password_from_file:
+            loggerinst.warning("Password file argument take precedence over the password argument.")
+
+        if (config_opts.activation_key or config_opts.password) and (parsed_opts.activationkey or parsed_opts.password):
+            loggerinst.warning("Command line authentication method take precedence over method in configuration file.")
+
+        if (config_opts.activation_key or config_opts.password) and parsed_opts.password_from_file:
+            loggerinst.warning("Password file take precedence over the config file.")
+
         if tool_opts.username and tool_opts.password:
             tool_opts.credentials_thru_cli = True
 
@@ -336,6 +386,62 @@ def warn_on_unsupported_options():
             "See help (convert2rhel -h) for more information."
         )
         utils.ask_to_continue()
+
+
+def options_from_config_files(cfg_path=None):
+    """Parse the convert2rhel.ini configuration file.
+
+    This function will try to parse the convert2rhel.ini configuration file and
+    return a dictionary containing the values found in the file.
+
+    .. note::
+       This function will parse the configuration file following a specific
+       order, which is:
+       1) Path provided by the user in cfg_path (Highest priority).
+       2) ~/.convert2rhel.ini (The 2nd highest priority).
+       3) /etc/convert2rhel.ini (The lowest priority).
+
+    :param cfg_path: Path of a custom configuration file
+    :type cfg_path: str
+
+    :return: Dict with the supported options alongside their values.
+    :rtype: dict[str | None, str | None]
+    """
+    headers = ["subscription_manager"]  # supported sections in config file
+    config_file = configparser.ConfigParser()
+    paths = [os.path.expanduser(path) for path in CONFIG_PATHS]
+    # Create dict with all supported options, all of them set to None
+    # needed for avoiding problems with files priority
+    # The name of supported option MUST correspond with the name in ToolOpts()
+    # Otherwise it won't be used
+    supported_opts = {"password": None, "activation_key": None}
+
+    if cfg_path:
+        cfg_path = os.path.expanduser(cfg_path)
+        if not os.path.exists(cfg_path):
+            raise OSError(2, "No such file or directory: '%s'" % cfg_path)
+        paths.insert(0, cfg_path)  # highest priority
+
+    for path in paths:
+        if os.path.exists(path):
+            if not oct(os.stat(path).st_mode)[-4:].endswith("00"):
+                loggerinst.critical("The %s file must only be accessible by the owner (0600)" % path)
+            config_file.read(path)
+
+            for header in config_file.sections():
+                if header in headers:
+                    for option in config_file.options(header):
+                        if option.lower() in supported_opts:
+                            # Solving priority
+                            if supported_opts[option.lower()] is None:
+                                supported_opts[option] = config_file.get(header, option)
+                                loggerinst.debug("Found %s in %s" % (option, path))
+                        else:
+                            loggerinst.warning("Unsupported option %s in %s" % (option, path))
+                elif header not in headers and header != "DEFAULT":
+                    loggerinst.warning("Unsupported header %s in %s." % (header, path))
+
+    return supported_opts
 
 
 # Code to be executed upon module import
--- a/convert2rhel/unit_tests/toolopts_test.py
+++ b/convert2rhel/unit_tests/toolopts_test.py
@@ -193,7 +193,7 @@ def test_config_file(argv, content, output, message, monkeypatch, tmp_path, capl
     # Re-init needed delete the set data
     convert2rhel.toolopts.tool_opts.__init__()
     path = os.path.join(str(tmp_path), "convert2rhel.ini")
-    with open(path, "w") as file:  # pylint: disable=unspecified-encoding
+    with open(path, "w") as file:
         file.write(content)
     os.chmod(path, 0o600)
 
@@ -228,7 +228,7 @@ def test_multiple_auth_src_combined(argv, content, message, output, caplog, monk
     """Test combination of password file or configuration file and CLI arguments."""
     convert2rhel.toolopts.tool_opts.__init__()
     path = os.path.join(str(tmp_path), "convert2rhel.file")
-    with open(path, "w") as file:  # pylint: disable=unspecified-encoding
+    with open(path, "w") as file:
         file.write(content)
     os.chmod(path, 0o600)
     # The path for file is the last argument
@@ -257,10 +257,10 @@ def test_multiple_auth_src_combined(argv, content, message, output, caplog, monk
 def test_multiple_auth_src_files(argv, content, message, output, caplog, monkeypatch, tmp_path):
     """Test combination of password file, config file and CLI."""
     path0 = os.path.join(str(tmp_path), "convert2rhel.password")
-    with open(path0, "w") as file:  # pylint: disable=unspecified-encoding
+    with open(path0, "w") as file:
         file.write(content[0])
     path1 = os.path.join(str(tmp_path), "convert2rhel.ini")
-    with open(path1, "w") as file:  # pylint: disable=unspecified-encoding
+    with open(path1, "w") as file:
         file.write(content[1])
     # Set the paths
     argv[-3] = path0
@@ -320,7 +320,7 @@ def test_options_from_config_files_default(content, output, monkeypatch, tmp_pat
     """Test config files in default path."""
     path = os.path.join(str(tmp_path), "convert2rhel.ini")
     if content:
-        with open(path, "w") as file:  # pylint: disable=unspecified-encoding
+        with open(path, "w") as file:
             file.write(content)
         os.chmod(path, 0o600)
 
@@ -360,13 +360,13 @@ def test_options_from_config_files_default(content, output, monkeypatch, tmp_pat
 def test_options_from_config_files_specified(content, output, content_lower_priority, monkeypatch, tmp_path, caplog):
     """Test user specified path for config file."""
     path = os.path.join(str(tmp_path), "convert2rhel.ini")
-    with open(path, "w") as file:  # pylint: disable=unspecified-encoding
+    with open(path, "w") as file:
         file.write(content)
     os.chmod(path, 0o600)
 
     path_lower_priority = os.path.join(str(tmp_path), "convert2rhel_lower.ini")
     content_lower_priority = "[subscription_manager]\npassword = low_prior_pass"
-    with open(path_lower_priority, "w") as file:  # pylint: disable=unspecified-encoding
+    with open(path_lower_priority, "w") as file:
         file.write(content_lower_priority)
     os.chmod(path_lower_priority, 0o600)
 
--- a/convert2rhel/unit_tests/toolopts_test.py
+++ b/convert2rhel/unit_tests/toolopts_test.py
@@ -18,6 +18,7 @@
 # Required imports:
 
 
+import os
 import sys
 import unittest
 
@@ -144,3 +145,255 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
         convert2rhel.toolopts.CLI()
 
     assert convert2rhel.toolopts.tool_opts.no_rhsm == no_rhsm_value
+
+
+@pytest.mark.parametrize(
+    ("argv", "content", "output", "message"),
+    (
+        (
+            mock_cli_arguments([""]),
+            "[subscription_manager]\npassword=conf_pass",
+            {"password": "conf_pass", "activation_key": None},
+            None,
+        ),
+        (
+            mock_cli_arguments(["-p", "password"]),
+            "[subscription_manager]\nactivation_key=conf_key",
+            {"password": "password", "activation_key": None},
+            "Command line authentication method take precedence over method in configuration file.",
+        ),
+        (
+            mock_cli_arguments(["-k", "activation_key"]),
+            "[subscription_manager]\nactivation_key=conf_key",
+            {"password": None, "activation_key": "activation_key"},
+            "Command line authentication method take precedence over method in configuration file.",
+        ),
+        (
+            mock_cli_arguments(["-k", "activation_key"]),
+            "[subscription_manager]\npassword=conf_pass",
+            {"password": "conf_pass", "activation_key": "activation_key"},
+            "Command line authentication method take precedence over method in configuration file.",
+        ),
+        (
+            mock_cli_arguments(["-k", "activation_key", "-p", "password"]),
+            "[subscription_manager]\npassword=conf_pass\nactivation_key=conf_key",
+            {"password": "password", "activation_key": "activation_key"},
+            "Command line authentication method take precedence over method in configuration file.",
+        ),
+        (
+            mock_cli_arguments([""]),
+            "[subscription_manager]\npassword=conf_pass\nactivation_key=conf_key",
+            {"password": "conf_pass", "activation_key": "conf_key"},
+            "Set only one of password or activation key. Activation key take precedence.",
+        ),
+    ),
+)
+def test_config_file(argv, content, output, message, monkeypatch, tmp_path, caplog):
+    # After each test there were left data from previous
+    # Re-init needed delete the set data
+    convert2rhel.toolopts.tool_opts.__init__()
+    path = os.path.join(str(tmp_path), "convert2rhel.ini")
+    with open(path, "w") as file:  # pylint: disable=unspecified-encoding
+        file.write(content)
+    os.chmod(path, 0o600)
+
+    monkeypatch.setattr(sys, "argv", argv)
+    monkeypatch.setattr(convert2rhel.toolopts, "CONFIG_PATHS", value=[path])
+    convert2rhel.toolopts.CLI()
+
+    assert convert2rhel.toolopts.tool_opts.activation_key == output["activation_key"]
+    assert convert2rhel.toolopts.tool_opts.password == output["password"]
+    if message:
+        assert message in caplog.text
+
+
+@pytest.mark.parametrize(
+    ("argv", "content", "message", "output"),
+    (
+        (
+            mock_cli_arguments(["--password", "pass", "-f"]),
+            "pass_file",
+            "Password file argument take precedence over the password argument.",
+            {"password": "pass_file", "activation_key": None},
+        ),
+        (
+            mock_cli_arguments(["--password", "pass", "--config-file"]),
+            "[subscription_manager]\nactivation_key = key_cnf_file",
+            "Command line authentication method take precedence over method in configuration file.",
+            {"password": "pass", "activation_key": None},
+        ),
+    ),
+)
+def test_multiple_auth_src_combined(argv, content, message, output, caplog, monkeypatch, tmp_path):
+    """Test combination of password file or configuration file and CLI arguments."""
+    convert2rhel.toolopts.tool_opts.__init__()
+    path = os.path.join(str(tmp_path), "convert2rhel.file")
+    with open(path, "w") as file:  # pylint: disable=unspecified-encoding
+        file.write(content)
+    os.chmod(path, 0o600)
+    # The path for file is the last argument
+    argv.append(path)
+
+    monkeypatch.setattr(sys, "argv", argv)
+    monkeypatch.setattr(convert2rhel.toolopts, "CONFIG_PATHS", value=[""])
+    convert2rhel.toolopts.CLI()
+
+    assert message in caplog.text
+    assert convert2rhel.toolopts.tool_opts.activation_key == output["activation_key"]
+    assert convert2rhel.toolopts.tool_opts.password == output["password"]
+
+
+@pytest.mark.parametrize(
+    ("argv", "content", "message", "output"),
+    (
+        (
+            mock_cli_arguments(["--password", "pass", "-f", "file", "--config-file", "file"]),
+            ("pass_file", "[subscription_manager]\nactivation_key = pass_cnf_file"),
+            "Password file take precedence over the config file.",
+            {"password": "pass_file", "activation_key": None},
+        ),
+    ),
+)
+def test_multiple_auth_src_files(argv, content, message, output, caplog, monkeypatch, tmp_path):
+    """Test combination of password file, config file and CLI."""
+    path0 = os.path.join(str(tmp_path), "convert2rhel.password")
+    with open(path0, "w") as file:  # pylint: disable=unspecified-encoding
+        file.write(content[0])
+    path1 = os.path.join(str(tmp_path), "convert2rhel.ini")
+    with open(path1, "w") as file:  # pylint: disable=unspecified-encoding
+        file.write(content[1])
+    # Set the paths
+    argv[-3] = path0
+    argv[-1] = path1
+    os.chmod(path1, 0o600)
+
+    monkeypatch.setattr(sys, "argv", argv)
+    monkeypatch.setattr(convert2rhel.toolopts, "CONFIG_PATHS", value=[""])
+    convert2rhel.toolopts.CLI()
+
+    assert message in caplog.text
+    assert convert2rhel.toolopts.tool_opts.activation_key == output["activation_key"]
+    assert convert2rhel.toolopts.tool_opts.password == output["password"]
+
+
+@pytest.mark.parametrize(
+    ("argv", "message", "output"),
+    (
+        (
+            mock_cli_arguments(["--password", "pass", "--activationkey", "key"]),
+            "Set only one of password or activation key.",
+            {"password": "pass", "activation_key": "key"},
+        ),
+    ),
+)
+def test_multiple_auth_src_cli(argv, message, output, caplog, monkeypatch):
+    """Test both auth methods in CLI."""
+    monkeypatch.setattr(sys, "argv", argv)
+    monkeypatch.setattr(convert2rhel.toolopts, "CONFIG_PATHS", value=[""])
+    convert2rhel.toolopts.CLI()
+
+    assert message in caplog.text
+    assert convert2rhel.toolopts.tool_opts.activation_key == output["activation_key"]
+    assert convert2rhel.toolopts.tool_opts.password == output["password"]
+
+
+@pytest.mark.parametrize(
+    ("content", "output"),
+    (
+        (
+            "[subscription_manager]\npassword = correct_password",
+            {"password": "correct_password", "activation_key": None},
+        ),
+        (
+            "[subscription_manager]\nactivation_key = correct_key\nPassword = correct_password",
+            {"password": "correct_password", "activation_key": "correct_key"},
+        ),
+        (
+            "[subscription_manager]\nincorrect_option = incorrect_content",
+            {"password": None, "activation_key": None},
+        ),
+        ("[INVALID_HEADER]\nactivation_key = correct_key", {"password": None, "activation_key": None}),
+        (None, {"password": None, "activation_key": None}),
+    ),
+)
+def test_options_from_config_files_default(content, output, monkeypatch, tmp_path, caplog):
+    """Test config files in default path."""
+    path = os.path.join(str(tmp_path), "convert2rhel.ini")
+    if content:
+        with open(path, "w") as file:  # pylint: disable=unspecified-encoding
+            file.write(content)
+        os.chmod(path, 0o600)
+
+    paths = ["/nonexisting/path", path]
+    monkeypatch.setattr(convert2rhel.toolopts, "CONFIG_PATHS", value=paths)
+    opts = convert2rhel.toolopts.options_from_config_files()
+
+    assert opts["password"] == output["password"]
+    assert opts["activation_key"] == output["activation_key"]
+    if content:
+        if "INVALID_HEADER" in content:
+            assert "Unsupported header" in caplog.text
+        if "incorrect_option" in content:
+            assert "Unsupported option" in caplog.text
+
+
+@pytest.mark.parametrize(
+    ("content", "output", "content_lower_priority"),
+    (
+        (
+            "[subscription_manager]\nactivation_key = correct_key\nPassword = correct_password",
+            {"password": "correct_password", "activation_key": "correct_key"},
+            "[subscription_manager]\npassword = low_prior_pass",
+        ),
+        (
+            "[subscription_manager]\nactivation_key = correct_key\nPassword = correct_password",
+            {"password": "correct_password", "activation_key": "correct_key"},
+            "[INVALID_HEADER]\nincorrect_option = incorrect_option",
+        ),
+        (
+            "[subscription_manager]\nactivation_key = correct_key\nPassword = correct_password",
+            {"password": "correct_password", "activation_key": "correct_key"},
+            "[subscription_manager]\nincorrect_option = incorrect_option",
+        ),
+    ),
+)
+def test_options_from_config_files_specified(content, output, content_lower_priority, monkeypatch, tmp_path, caplog):
+    """Test user specified path for config file."""
+    path = os.path.join(str(tmp_path), "convert2rhel.ini")
+    with open(path, "w") as file:  # pylint: disable=unspecified-encoding
+        file.write(content)
+    os.chmod(path, 0o600)
+
+    path_lower_priority = os.path.join(str(tmp_path), "convert2rhel_lower.ini")
+    content_lower_priority = "[subscription_manager]\npassword = low_prior_pass"
+    with open(path_lower_priority, "w") as file:  # pylint: disable=unspecified-encoding
+        file.write(content_lower_priority)
+    os.chmod(path_lower_priority, 0o600)
+
+    paths = [path_lower_priority]
+    monkeypatch.setattr(convert2rhel.toolopts, "CONFIG_PATHS", value=paths)
+    # user specified path
+    opts = convert2rhel.toolopts.options_from_config_files(path)
+
+    assert opts["password"] == output["password"]
+    assert opts["activation_key"] == output["activation_key"]
+    if "INVALID_HEADER" in content or "INVALID_HEADER" in content_lower_priority:
+        assert "Unsupported header" in caplog.text
+    if "incorrect_option" in content or "incorrect_option" in content_lower_priority:
+        assert "Unsupported option" in caplog.text
+
+
+@pytest.mark.parametrize(
+    "supported_opts",
+    (
+        {"password": "correct_password", "activation_key": "correct_key"},
+        {"password": "correct_password", "activation_key": "correct_key", "invalid_key": "invalid_key"},
+    ),
+)
+def test_set_opts(supported_opts):
+    tool_opts.__init__()
+    convert2rhel.toolopts.ToolOpts.set_opts(tool_opts, supported_opts)
+
+    assert tool_opts.password == supported_opts["password"]
+    assert tool_opts.activation_key == supported_opts["activation_key"]
+    assert not hasattr(tool_opts, "invalid_key")
--- a/plans/tier1.fmf
+++ b/plans/tier1.fmf
@@ -283,3 +283,18 @@ discover+:
   - name: reboot after conversion
     how: ansible
     playbook: tests/ansible_collections/roles/reboot/main.yml
+
+/config_file:
+  adjust:
+      enabled: false
+      when: >
+        distro != centos-8
+  discover+:
+    test: checks-after-conversion
+  prepare+:
+  - name: main conversion preparation
+    how: shell
+    script: pytest -svv tests/integration/tier1/config-file/test_config_file.py
+  - name: reboot after conversion
+    how: ansible
+    playbook: tests/ansible_collections/roles/reboot/main.yml
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,4 @@ recursive-include convert2rhel/data *
 include man/*
 include LICENSE
 include README.md
+include config/convert2rhel.ini
--- a/config/convert2rhel.ini
+++ b/config/convert2rhel.ini
@@ -1,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# This file should be in mode 0600
+# Example of configuration file convert2rhel.ini for secrets.
+# Possible locations of this file:
+# 1) user specified and passed by -c, --config-file option; highest priority
+# 2) ~/.convert2rhel.ini; lower priority
+# 3) /etc/convert2rhel.ini; the lowest priority
+
+[subscription_manager]
+# password = <insert_password>
+# activation_key = <insert_activation_key>
--- a/packaging/convert2rhel.spec
+++ b/packaging/convert2rhel.spec
@@ -96,6 +96,10 @@ install -d %{buildroot}%{_sharedstatedir}/%{name}/rhsm/
 install -d -m 755 %{buildroot}%{_mandir}/man8
 install -p man/%{name}.8 %{buildroot}%{_mandir}/man8/
 
+# Put the config file into place
+install -d %{buildroot}%{_sysconfdir}/
+install -m 0600 config/convert2rhel.ini %{buildroot}%{_sysconfdir}/convert2rhel.ini
+
 %files
 
 %if 0%{?rhel} && 0%{?rhel} <= 6
@@ -107,6 +111,7 @@ install -p man/%{name}.8 %{buildroot}%{_mandir}/man8/
 %{_datadir}/%{name}/
 %{_sharedstatedir}/%{name}/
 %{python_sitelib}/%{name}*
+%config(noreplace) %attr(0600, root, root) %{_sysconfdir}/convert2rhel.ini
 
 %{!?_licensedir:%global license %%doc}
 %license LICENSE
--- a/tests/integration/tier1/checks-after-conversion/test_release_version.py
+++ b/tests/integration/tier1/checks-after-conversion/test_release_version.py
@@ -12,7 +12,7 @@ def test_correct_distro():
     """
     with open("/etc/migration-results") as json_file:
         json_data = json.load(json_file)
-        source_distro = json_data["activities"][0]["source_os"]
+        source_distro = json_data["activities"][-1]["source_os"]
 
     with open("/etc/system-release", "r") as sys_release:
         destination_distro = sys_release.read()
--- a/tests/integration/tier1/config-file/main.fmf
+++ b/tests/integration/tier1/config-file/main.fmf
@@ -1,0 +1,5 @@
+summary: config file
+
+tier: 0
+
+test: pytest -svv
--- a/tests/integration/tier1/config-file/test_config_file.py
+++ b/tests/integration/tier1/config-file/test_config_file.py
@@ -1,0 +1,125 @@
+import os
+
+from collections import namedtuple
+
+from envparse import env
+
+
+Config = namedtuple("Config", "path content")
+
+
+def create_files(config):
+    for cfg in config:
+        f = open(os.path.expanduser(cfg.path), "w")
+        f.write(cfg.content)
+        f.close()
+        os.chmod(os.path.expanduser(cfg.path), 0o600)
+
+
+def remove_files(config):
+    for cfg in config:
+        os.remove(os.path.expanduser(cfg.path))
+
+
+def test_user_path_custom_filename(convert2rhel):
+    config = [Config("~/.convert2rhel_custom.ini", "[subscription_manager]\nactivation_key = config_activationkey")]
+    create_files(config)
+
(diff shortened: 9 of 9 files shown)


convert2rhel.ini.example
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+# Example of configuration file convert2rhel.ini for secrets.
+# Possible locations of this file:
+# 1) user specified and passed by -c, --config-file option; highest priority
+# 2) ~/convert2rhel.ini; lower priority
+# 3) /etc/convert2rhel.ini; the lowest priority
+
+# [SECRET]
```
hosekadam, line 8: Using a header. Due [1] I decide, that using it is a good practice and can provide better readability if the config file would be used for something else. If the header isn't needed there is possibility to do it without it - adding some dummy section after reading the file so the configparser accepts the structure.

[1] [https://docs.python.org/3/library/configparser.html#supported-ini-file-structure](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure)

convert2rhel/toolopts.py
```diff
@@ -279,9 +288,27 @@ def _process_cli_options(self):
             tool_opts.password = parsed_opts.password
 
         if parsed_opts.password_from_file:
+            loggerinst.warning("Deprecated. Use -c | --config-file instead")
             tool_opts.password_file = parsed_opts.password_from_file
             tool_opts.password = utils.get_file_content(parsed_opts.password_from_file)
 
+        # Processing the configuration file
+        if parsed_opts.config_file or (not parsed_opts.password or not parsed_opts.activationkey):
```
hosekadam, line 296: If the second part of condition would be `(not parsed_opts.password and not parsed_opts.activationkey)`  the difference will be in [this variant of test](https://github.com/oamg/convert2rhel/pull/451/files#diff-5537d9a47132079b0eff4fb09b1247db109b6322ee99aad12628df6772136346R166) with output value equal to `{"password": None, "activation-key": "activation-key"}`. 

I think it's fine, the activation key has precedence. And maybe sometime in the future the convert2rhel could try to use it if the activation key fails?

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,50 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
         convert2rhel.toolopts.CLI()
 
     assert convert2rhel.toolopts.tool_opts.no_rhsm == no_rhsm_value
+
+
+@pytest.mark.parametrize(
```
hosekadam, line 151: Maybe too much tests, but I want to show what happens in different cases.  Believe it will be useful for reviewers and makes your work easier :)

convert2rhel/utils.py
```diff
@@ -618,4 +625,44 @@ def set_locale():
...
+        else:
+            continue
+
+        for secret_name in secret_names:
+            try:
+                secrets[secret_name] = config.get(header, secret_name)
+                loggerinst.debug("Found %s in %s" % (secret_name, path))
+
+            except (configparser.NoOptionError, KeyError, configparser.NoSectionError):
+                continue
```
r0x0d, line 660: I feel this for-loop is a bit unnecessary here. We can access the fields in the config file manually since we are using only 2 fields now, and we don't need to optimize our way in right now (Maybe when we start porting more options to the config file it could be beneficial) 
```suggestion
          try:
              secrets["password"] = config.get("SECRETS",  "password")
              secrets["activation-key"] = config.get("SECRETS",  "activation-key")
          except (configparser.NoOptionError, KeyError, configparser.NoSectionError):
              # Maybe throwing an error here is a better option?
              loggerinst.critical("Failed to load config file options.")
              
          loggerinst.debug("Loaded the config file successfully!")
```

convert2rhel/utils.py
```diff
@@ -618,4 +625,44 @@ def set_locale():
...
+    The order is:
+    1. if filename is set then load from it
+    2. ~/.convert2rhel.ini
+    3. /etc/convert2rhel.ini
+
+    Return {"password": <value>, "activation-key": <value>}.
+    If the secret isn't set, default value is None.
+    """
+    header = "SECRET"  # header in config file
+    secret_names = ("password", "activation-key")
```
r0x0d, line 640: We can remove that 2 variables here if we opt-in to not use the loop there and go with a simpler version.

convert2rhel/utils.py
```diff
@@ -618,4 +625,44 @@ def set_locale():
...
+    config = configparser.ConfigParser()
+    paths = CONFIG_PATHS
```
r0x0d, line 642: I think this global constant is not very useful right here, what do you think to placing it inside the function instead? 

```suggestion
    paths = ["~/convert2rhel.ini", "/etc/convert2rhel.ini"]
```

r0x0d (changes requested):
Awesome work, @hosekadam! Having a config file (specific for secrets now, but) that can evolve to a general config is cool. 

Anyway, left a few comments there for you, if you're in doubt about them, reach out to me and we can have a chat so I can explain a bit further my thoughts about the comments!

r0x0d:
I left out 2 questions I had when reviewing this, so if anyone has a reason/answer, I would appreciate it! 

* Do we really need to have different paths to check for config files? I mean, if the user is automating their conversion, they will have the same effort to place the file in one of the locations or pass them through the command line... I don't mind having the support for different paths, but it seems easier if the file was located in just one place. 

* Do we need a command to copy the example config file to the user? Or this is something that will be available during the package installation through the RPM? For example, this is how `pre-commit` does to generate a sample config file: 
```bash
pre-commit sample-config > .pre-commit-config.yaml  
```

convert2rhel.ini.example
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
```
bocekm, line 1: I don't think the `.example` suffix of the file is necessary. It just gives the user another barrier that does not add much value. It can be actually counter-productive in case we want to introduce some default values that might be possible to change by the user.

convert2rhel.ini.example
bocekm, line 8: Since the config file can later be used to not set only secrets but also to set various other things, I'd keep using the header there.

And because there might be more secrets later on, for instance network proxy settings, we should name the header more specifically so that it's clear what exactly the keys relate to.
```suggestion
[subscription-manager]
```
Also, headers don't need to be commented out I'd say.

convert2rhel.ini.example
```diff
@@ -0,0 +1,10 @@
...
+# Example of configuration file convert2rhel.ini for secrets.
+# Possible locations of this file:
+# 1) user specified and passed by -c, --config-file option; highest priority
+# 2) ~/convert2rhel.ini; lower priority
+# 3) /etc/convert2rhel.ini; the lowest priority
+
+# [SECRET]
+# password = <insert_password>
+# activation-key = <insert_activation_key>
```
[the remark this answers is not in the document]
bocekm: I agree. We can put it to the existing `convert2rhel/data/version-independent/`. I don't think another directory is needed.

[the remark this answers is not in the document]
bocekm: I'd say we can have a mix of keys that are:
- commented out (advantage of being able to provide an example of the value),
- uncommented with a default value

There's another option I'd not vote for as it seems to add no advantage over the above option:
- keys uncommented and without a value (easy to handle when parsing the config: no value = ignoring)

bocekm (changes requested):

convert2rhel.ini.example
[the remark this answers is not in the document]
abadger: I wonder in general if we should move `convert2rhel/data` somewhere else.  setup.py installs that directory into `site-packages/convert2rhel` and I see that we then delete it from there inside of the rpm spec file.

convert2rhel.ini.example
abadger, line 1: Note: We should have a comment that this file should be installed with mode 0600 since it contains secrets and be sure that we do that in the rpm spec file.

convert2rhel.ini.example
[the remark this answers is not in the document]
abadger: This file is also different from the other files in convert2rhel/data since it:
* Is meant to be user edited to become useful in the common case
* Should be installed with mode 0600 for security
* Installs to /etc/ rather than /usr/share/

I wonder if all of those reasons make it better to place it in a different directory than convert2rhel/data ?

convert2rhel/toolopts.py
abadger: I think it would be better to always set `tool_opts.config_file` here since we might have other, non-secret, values in the config file at a later date.

In other projects, I parse the config file before I look into any of the other command line arguments and then, when I merge that with the command line arguments.

Our password and activation parameters are special, though, as they have an interrelation.  I would set the tool_opts for those in one place instead of setting the tool_opts from the cli first, and then setting it from the config file in a later section.  So maybe something like this:

``` python
parsed_opts, _ = self._parser.parse_args()
tool_opts.config_file = parsed_opts.config_file

# Notes:
# options_from_config_files reads from all of the config files
# It skips the passed in config file if it is None (no config file specified on the CLI)
# It raises an exception if the passed in config file is specified but does not exist/is unreadable
# It ignores if one of the hardcoded config file locations does not exist.  It may be wise to warn if it is unreadable.
# It merges the values from the hardcoded and passed in locations.  The passed in location takes precedence followed by ~/.convert2rhel.ini, and then /etc/convert2rhel.ini if neither of the other two have a value.

config_opts = options_from_config_files(parsed_opts.config_file)

# General config options (which don't have cornercases like password + activation key)
# Note, we don't have any of these in the config file yet.
tool_opts.server_url = config_opts.copy()
if parsed_opts.debugl:
    tool_opts.debug = parsed_opts.debug
# [..]

if parsed_opts.password or parsed_opts.activationkey:
    if parsed_opts.password and parsed_opts.activationkey:
        loggerinst.warning("Set only one of password or activation key")
    tool_opts.password = tool_opts.activationkey = None
    if parsed_opts.activationkey:
        tool_opts.activationkey = parsed_opts.activationkey
    if parsed_opts.password:
        tool_opts.password = parsed_opts.password
```

convert2rhel.ini.example
r0x0d: That's a good catch! I think that's probably would be a good idea to implement a check in the code as well, so we can be sure that it's correct even after the installation (or manual creation by the user)

convert2rhel/toolopts.py
SpyTec: @abadger by this suggestion, do you mean to say we hard-code the reading of all entries in the ini file? Would it not be better for us to iterate through all entries in the ini and see if it matches our keys in tool_opts?
```py
# General config options (which don't have cornercases like password + activation key)
# Note, we don't have any of these in the config file yet.
tool_opts.server_url = config_opts.copy()
```

But yes I agree that we should 
1. Parse entries from .ini files in terms of priority
2. Overwrite values set by .ini files if CLI options are set

hosekadam: @abadger @SpyTec Maybe the idea with iterating through the entries in file would be better. I'll try it that way - I was focused just on doing it in one way, this variant didn't come to my mind :smile: 

But anyway, parsing the .ini file at first time and then overwrite the values by values from CLI simplified the part which I called 'Processing the configuration file'.

@r0x0d I think you are half right: When are provided both (password and activation-key) it will be caught by another condition which mentioned @abadger and it would say 'loggerinst.warning("Set only one of password or activation key")'. 

You wrote that we are ignoring password or activation-key when we detect config file. That's not right, because the CLI options have precedence over the options from config file.

convert2rhel.ini.example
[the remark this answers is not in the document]
abadger: From the meeting today,
- We'll keep the files in convert2rhel/data/ and the config file separate.
- Moving convert2rhel/data will be a separate issue
- Place the data file in a subdirectory of the toplevel

Changes to make:
- Store the sample config file in `config/convert2rhel.ini`
- Modify the `MANIFEST.in` to put the config file in the tarball:
  ``` patch
  + include config/convert2rhel.ini
  ```
- Modify the rpm spec file (`packaging/convert2rhel.spec`) to include the config file.  A spec file has  three main parts: Some metadata, shell scripts to control building and installing, and a file          manifest.  We'll need to modify the shell script and the file manifest.
        
  At the end of the `%install` section, add a comment explaining what we're doing and the shell command to install the file.  We prepend the path to install to with `%{buildroot}` because rpm looks for the files to install in a parallel directory tree, the buildroot, rather than directly using what is in `/`.  `%{_sysconfdir}` is an rpm macro that expands to `/etc`. 
  ```                                            
  # Put the config file into place
  install -m 0600 config/convret2rhel.ini %{buildroot}%{_sysconfdir}/convert2rhel.ini          
  ```                                                                                                 
           
  In the `%files` section, add the config file:
  ``` patch
    %{_sharedstatedir}/%{name}
  + %config(noreplace) %attr(0600, root, root) %{_sysconfdir}/convert2rhel.ini
  ```

convert2rhel/toolopts.py
```diff
@@ -179,6 +181,13 @@ def _register_options(self):
             " 'subscription-manager orgs'. From the listed pairs"
             " Name:Key, use the Key here.",
         )
+        group.add_option(
+            "-c",
+            "--config-file",
+            help="Configuration file containing"
+            " password or activation key for subscription-manager."
+            " Example of this file in /etc/convert2rhel.ini.example",
```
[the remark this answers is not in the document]
hosekadam: Yes, this is default path for example.

convert2rhel/utils.py
hosekadam: That was a variant. But I choose this one, because I think that's improve readability of the code - you see the default paths in the beginning of the file without digging to it. And it's easy to add a new one in case somebody needs that.

convert2rhel/utils.py
```diff
@@ -620,4 +627,52 @@ def set_locale():
...
+    3. /etc/convert2rhel.ini
+
+    Return config_opts filled with loaded values.
+    """
+    headers = ["subscription_manager"]  # supported sections in config file
+    config_file = configparser.ConfigParser()
+    paths = CONFIG_PATHS
+    # create dict with all supproted options, all of them set to None
+    # needed for avoiding problems with files priority
+    supproted_opts = {"password": None, "activation_key": None}
```
hosekadam, line 645: Adding support for new option will be just adding the name there. And if needed the new header upper.

hosekadam:
Example file created during the installation process, so ready for re-review

convert2rhel/unit_tests/utils_test.py
```diff
@@ -347,3 +348,56 @@ def test_prompt_user(question, is_password, response, monkeypatch):
...
+        (
+            "[subscription_manager]\nincorrect_option = incorrect_content",
+            {"password": None, "activation_key": None},
+            False,
+        ),
+        ("[INVALID_HEADER]\nactivation_key = correct_key", {"password": None, "activation_key": None}, False),
+        ("nonexisting", {"password": None, "activation_key": None}, False),
+    ),
+)
+def test_options_from_config_files(monkeypatch, content, output, tmp_path, specified, caplog):
```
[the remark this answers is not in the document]
hosekadam: Makes sense, I'll split them :+1:

convert2rhel/utils.py
```diff
@@ -620,4 +627,52 @@ def set_locale():
...
+
+    if cfg_path:
+        if not os.path.exists(cfg_path):
+            raise FileNotFoundError
```
abadger, line 649: Good thinking that this particular case (the cfg_path passed from the user) should raise an exception whereas the other config file location should not!  A couple changes to make

* The exception should have a message.
* FileNotFoundError doesn't exist in python2.x.  You could just use OSError here (which is what FileNotFoundError was in Python-2.  But if you do that you'll want to make sure that you populate all the fields that OSError has.  Maybe the easiest way to see what it should look like is to make an OSError onpurpose and then fill in your raise appropriately (Catch the exception from `open('/nonexistant')` and then use that as a model for what you need to raise.)

convert2rhel/utils.py
```diff
@@ -620,4 +627,52 @@ def set_locale():
...
+        paths.insert(0, cfg_path)  # highest priority
+
+    for path in paths:
+        if os.path.exists(path):
+            if oct(os.stat(path).st_mode)[-4:] != "0600":
+                loggerinst.critical("The %s file need to be in mode 0600" % path)
```
abadger, line 655: * On Python2, `oct(384)` => "0600" but on Python3, it yields "0o600".
* We probably are okay with being only readable by the owner, readable and executable by the owner, or readable, writable, and executable by the owner.  So maybe it is better to check that the bits we don't want are not present:

```suggestion
            if not oct(os.stat(path).st_mode)[-4:].endswith("00"):
                loggerinst.critical("The %s file must only be accessible by the owner (0600)" % path)
```

convert2rhel/utils.py
```diff
@@ -620,4 +627,52 @@ def set_locale():
...
+            config_file.read(path)
+
+            for header in config_file.sections():
+                if header in headers:
+                    for option in config_file.options(header):
+                        if option.lower() in supproted_opts:
```
abadger, line 661: * Spelling: `s/supproted_opts/supported_opts`

convert2rhel/utils.py
```diff
@@ -620,4 +627,52 @@ def set_locale():
...
+                        else:
+                            loggerinst.warning("Unsupported option %s in %s" % (option, path))
+                elif header not in headers and header != "DEFAULT":
+                    loggerinst.warning("Unsupported header %s in %s." % (header, path))
+
+    for key, value in supproted_opts.items():
+        if value is not None:
+            setattr(config_opts, key, value)
+
+    return config_opts
```
abadger, line 675: A colleague once told me that good functions should either return the data they've found or set it as a side effect but not both.  I personally favor returning the data and doing the setting in the caller most of the time (That way is easier to unittest, for instance).

So, seeing how the rest of this function operates, I'd change it to not have to pass in config_opts and instead `return supported_opts`.   The caller can then merge `supported_opts` into `config_opts` (Originally I was thinking we'd construct a ToolOpts in here and pass that back but I see from how you've designed this that returning the dict and doing the merging in the caller makes more sense).

convert2rhel/unit_tests/utils_test.py
```diff
@@ -347,3 +348,79 @@ def test_prompt_user(question, is_password, response, monkeypatch):
...
+@pytest.mark.parametrize(
+    ("content", "output"),
+    (
+        (
+            "[subscription_manager]\nactivation_key = correct_key\nPassword = correct_password",
+            {"password": "correct_password", "activation_key": "correct_key"},
+        ),
+    ),
+)
+def test_options_from_config_files_specified(monkeypatch, content, output, tmp_path, caplog):
```
abadger, line 402: Minor: I would organize this so that the fixtures (moneypatch, tmp_path, and caplog) are grouped separately from the parameters handled by parametrize (content and output).

convert2rhel/unit_tests/utils_test.py
```diff
@@ -347,3 +348,79 @@ def test_prompt_user(question, is_password, response, monkeypatch):
...
+    monkeypatch.setattr(utils, "CONFIG_PATHS", value=paths)
+    # user specified path
+    tool_opts = utils.options_from_config_files(tool_opts, path)
+
+    assert tool_opts.password == output["password"]
+    assert tool_opts.activation_key == output["activation_key"]
+    if "INVALID_HEADER" in content:
+        assert "Unsupported header" in caplog.text
+    if "incorrect_option" in content:
+        assert "Unsupported option" in caplog.text
```
abadger, line 426: Note: I don't see either of these tested in this function.  Perhaps you want to put some of these in the lower priority config sources to test that they won't cause a failure of `utils.options_from_config_files()` in the future.

convert2rhel/unit_tests/utils_test.py
```diff
@@ -347,3 +348,79 @@ def test_prompt_user(question, is_password, response, monkeypatch):
...
+        ),
+        ("[INVALID_HEADER]\nactivation_key = correct_key", {"password": None, "activation_key": None}),
+        ("nonexisting", {"password": None, "activation_key": None}),
+    ),
+)
+def test_options_from_config_files_default(monkeypatch, content, output, tmp_path, caplog):
+    """Test config files in default path."""
+    path = os.path.join(str(tmp_path), "convert2rhel.ini")
+    tool_opts = ToolOpts()
+    if content != "nonexisting":
```
abadger, line 376: This is fine for a test but note that `None` is the usual sentinel value to use.  (And if None has meaning, then `sentinel = object()` can be used)

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,145 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+            mock_cli_arguments([""]),
+            "[subscription_manager]\npassword=conf_pass\nactivation_key=conf_key",
+            {"password": "conf_pass", "activation_key": "conf_key"},
+        ),
+    ),
+)
+def test_config_file(argv, content, output, monkeypatch, tmp_path):
+    # After each test there were left data from previous
+    # Re-init needed delete the set data
+    convert2rhel.toolopts.tool_opts.__init__()
```
abadger, line 189: No need to change this but I thought I should note that, in the future, I'm going to be introducing a fixture to conftest.py that partially handles this.  It looks like this:

``` python
@pytest.fixture
def global_tool_opts(monkeypatch):
    local_tool_opts = toolopts.ToolOpts()
    monkeypatch.setattr(toolopts, "tool_opts", local_tool_opts)
    return local_tool_opts
```

it's a little weird, though, because all of our modules use the global tool_opts with this sort of import: `from convert2rhel.toolopts import tool_opts` so we also have to monkeypatch `tool_opts` in the specific module.  I do that like this:

``` python
from convert2rhel import subscription

@pytest.fixture
def tool_opts(global_tool_opts, monkeypatch):
    monkeypatch.setattr(subscription, "tool_opts", global_tool_opts)
    return global_tool_opts

def test_subscribe_system(tool_opts):
    tool_opts.username = "user"
    tool_opts.password = "pass"
    assert subscription.subscribe_system()
```

(Like I said, no need to change this now.  Just something to be aware of when you port code to pytest in the future).

convert2rhel/unit_tests/toolopts_test.py
abadger: I like how you have all of these test cases!  Thanks!

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,145 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+    (
+        (
+            mock_cli_arguments([""]),
+            "[subscription_manager]\npassword=conf_pass",
+            {"password": "conf_pass", "activation_key": None},
+        ),
+        (
+            mock_cli_arguments(["-p", "password"]),
+            "[subscription_manager]\nactivation_key=conf_key",
+            {"password": "password", "activation_key": None},
```
abadger, line 162: For the cases where we are emitting a warning in the logs, should we also check that the log message shows up in caplog?

convert2rhel/unit_tests/utils_test.py
hosekadam: @abadger Not sure what do you mean by this. Is it like if in the config file with lower priority would be typo it shouldn't warn user about the mistake?
Or it's just about adding new scenarios which will test this path? Because I've forgot these asserts there since splitting the test.

convert2rhel/unit_tests/toolopts_test.py
hosekadam: Sure, good point :+1:

hosekadam:
@abadger I did the changes. Now the function return just a dict and there is new function which set the values. And some changes in unit tests (added case for invalid config file with lower priority). Also fixed the exception, you said something about message - did you mean the standard message `No such file or directory:` or some other?

convert2rhel/unit_tests/utils_test.py
abadger: The latter.  Just adding more scenarios to the tests.

convert2rhel/utils.py
```diff
@@ -620,4 +627,59 @@ def set_locale():
...
+    1. if filename is set then load from it
+    2. ~/.convert2rhel.ini
+    3. /etc/convert2rhel.ini
+
+    Return dict of supported options with their values.
+    """
+    headers = ["subscription_manager"]  # supported sections in config file
+    config_file = configparser.ConfigParser()
+    paths = CONFIG_PATHS
+    # Create dict with all supproted options, all of them set to None
```
abadger, line 643:
```suggestion
    # Create dict with all supported options, all of them set to None
```

convert2rhel/utils.py
```diff
@@ -620,4 +627,59 @@ def set_locale():
...
+    return supported_opts
+
+
+def set_opts(opts, supported_opts):
+    """Set ToolOpts data using dict with values from config file."""
+    for key, value in supported_opts.items():
+        if value is not None and hasattr(opts, key):
+            setattr(opts, key, value)
+
+    return opts
```
abadger, line 682: **Todo Now**: Seeing this broken out into its own function, I think that it would be good as a method of the `ToolOpts` class because:

* The first argument is an instance of `ToolOpts`.
* It mutates the attributes stored in the `ToolOpts`.

**Possible Future Refactor:** One thing that will be a little odd is that setting the cfg data on the `ToolOpts` will then be a method of the `ToolOpts` class but setting the command line arguments will be handled by the `CLI` class. I don't think that's a problem with your code, though.  Instead, I think it means we should look at changing the interface between the `CLI` class and the `ToolOpts` in the future.  

Maybe that eventual refactor should look more like:
``` python
    class ToolOpts
        def__init__(self,
                    args: List = sys.argv[1:],
                    cfg_files: List[Path] = CONFIG_PATHS):
        """
        Instantiate a toolopts, setting the values from configuration files and the programs command
        line arguments.

        Path in cfg_files which is not in CONFIG_PATHS must exist.
        """
```

`ToolOpts` could then call out to the `CLI` class to parse the arguments or the logic from `CLI` could
move into helper methods on `ToolOpts`.  I'm not sure which would be better without looking more closely.

convert2rhel/utils.py
```diff
@@ -620,4 +627,59 @@ def set_locale():
...
+    # The name of supported option MUST correspond with the name in ToolOpts()
+    # Otherwise it won't be used
+    supported_opts = {"password": None, "activation_key": None}
+
+    if cfg_path:
+        if not os.path.exists(cfg_path):
+            raise OSError(2, "No such file or directory: '" + cfg_path + "'")
+        paths.insert(0, cfg_path)  # highest priority
+
+    for path in paths:
```
abadger, line 654: Something I just noticed: `paths` needs to have `os.path.expanduser()` run on the paths contained in it.  Otherwise when we call `os.path.exist()` later, it will think that it's looking for a literal `~/.convert2rhel.ini`, not for `/home/USERNAME/.convert2rhel.ini`.

There's two places we could call `expanduser()`.. Here, or where we define `CONFIG_PATHS`. I lean towards doing it here for the following reason: If cfg_path contains a path with an initial tilde, then it should probably also be expanded. My justification isn't really strong because, in most cases, the user will be specifying cfg_path via a command line argument and the user's shell will do that expansion but it could occur if `options_from_cfg_files()` is called with a cfg_path that isn't from the command line (for instance, a unit test) or someone needs to run convert2rhel programmatically (like `subprocess.Popen(["convert2rhel", "--cfg_file", "~badger/my-convert2rhel.ini"])` ) .

My suggestion for doing it here would be:
```suggestion
    for path in (os.path.join(p) for p in paths):
```

abadger:
@hosekadam Nice job!  I added two  comments about the organization of the config reading code after your latest refactor and one bug that I spotted.

convert2rhel/toolopts.py
```diff
@@ -49,6 +60,58 @@ def __init__(self):
         # set True when credentials (username & password) are given through CLI
         self.credentials_thru_cli = False
 
+    @staticmethod
+    def options_from_config_files(cfg_path=None):
```
abadger, line 64: In other languages, like Java, it is common to make staticmethods but in Python, it's better to create a toplevel function rather than a `@staticmethod` most of the time.

The difference stems from the fact that Python builds namespaces out of each file on the filesystem whereas other programming languages depend on the class to be part of the namespace.  So I'd move this to just be a regular toplevel function in toolopts.py.

abadger:
Great job Adam!  I noted two things, one was to make `options_from_config_files()` a toplevel function instead of a `@staticmethod`.  The other was a mistake I made about where to apply `os.path.expanduser()`.  See if the new suggestion makes sense to you.

convert2rhel/toolopts.py
```diff
@@ -338,5 +388,51 @@ def warn_on_unsupported_options():
...
+    headers = ["subscription_manager"]  # supported sections in config file
+    config_file = configparser.ConfigParser()
+    paths = [os.path.expanduser(p) for p in CONFIG_PATHS]
+    # Create dict with all supported options, all of them set to None
+    # needed for avoiding problems with files priority
+    # The name of supported option MUST correspond with the name in ToolOpts()
+    # Otherwise it won't be used
+    supported_opts = {"password": None, "activation_key": None}
+
+    if cfg_path:
```
abadger, line 410:
```suggestion
    if cfg_path:
        cfg_path = os.path.expanduser(cfg_path)
```

convert2rhel/toolopts.py
```diff
@@ -263,9 +288,20 @@ def _process_cli_options(self):
         parsed_opts, _ = self._parser.parse_args()
 
         global tool_opts  # pylint: disable=C0103
+
         if parsed_opts.debug:
```
hosekadam, line 292: I had to move parsing the debug option, because without that is impossible to use debug logging in `options_from_config_files`

tests/integration/tier0/config-file/test_config_file.py
```diff
@@ -0,0 +1,65 @@
...
+
+    for path, content in paths:
+        f = open(os.path.expanduser(path), "w")
+        f.write(content)
+        f.close()
+        os.chmod(os.path.expanduser(path), 0o600)
+
+    with convert2rhel('--no-rpm-va --debug -c "~/.convert2rhel_custom.ini"') as c2r:
+        c2r.expect("DEBUG - Found activation_key in /root/.convert2rhel_custom.ini")
+        c2r.sendcontrol("c")
```
hosekadam, line 19: I did small performance change - testing script isn't waiting for the first question and is stopped immediately after the debug messages. So the whole testing is faster.

hosekadam:
So I think the code is OK, the problems with expanduser (@abadger) should be now solved and I can't see any new idea for improvements. 

The integration test is there too, but it might need a look from QE guys, if it's ok. @danmyway

convert2rhel/toolopts.py
```diff
@@ -338,5 +388,52 @@ def warn_on_unsupported_options():
...
+    # Create dict with all supported options, all of them set to None
+    # needed for avoiding problems with files priority
+    # The name of supported option MUST correspond with the name in ToolOpts()
+    # Otherwise it won't be used
+    supported_opts = {"password": None, "activation_key": None}
+
+    if cfg_path:
+        cfg_path = os.path.expanduser(cfg_path)
+        if not os.path.exists(cfg_path):
+            raise OSError(2, "No such file or directory: '" + cfg_path + "'")
```
r0x0d, line 413:
```suggestion
            raise OSError(2, "No such file or directory: '%s'" % cfg_path)
```

convert2rhel/toolopts.py
```diff
@@ -338,5 +388,52 @@ def warn_on_unsupported_options():
...
+        paths.insert(0, cfg_path)  # highest priority
```
r0x0d, line 414: I don't think we need this comment anymore! Since we have it in our docstrings, it should be fairly easy to spot that this is the highest priority when looking for a configuration file. 

```suggestion
        paths.insert(0, cfg_path)
```

convert2rhel/toolopts.py
```diff
@@ -338,5 +388,52 @@ def warn_on_unsupported_options():
...
+
+    for path in paths:
+        if os.path.exists(path):
+            if not oct(os.stat(path).st_mode)[-4:].endswith("00"):
```
r0x0d, line 418: What do you think about changing this and instead of checking against permission that ends with `00`, checking that the `st_mode` is returning a `0600` as expected?

```suggestion
            if int(oct(os.stat(path).st_mode)[-4:]) != 0600:
```
An example to illustrate that: 

```bash
touch b
chmod 0700
```

This file should pass on this statement right now since the permission ends on `00`

```python
>>> oct(os.stat('b').st_mode)[-4:]
'0700'
>>> not oct(os.stat('b').st_mode)[-4:].endswith("00")
False
```

Actually, any other permission that ends on `00` should pass on this, and since the message in the `logging.critical()` specifically tells the user that the permissions should be `0600`, it feels misleading looking at the code.

convert2rhel/toolopts.py
```diff
@@ -338,5 +388,52 @@ def warn_on_unsupported_options():
...
+                    for option in config_file.options(header):
+                        if option.lower() in supported_opts:
+                            # Solving priority
+                            if supported_opts[option.lower()] is None:
+                                supported_opts[option] = config_file.get(header, option)
+                                loggerinst.debug("Found %s in %s" % (option, path))
+                        else:
+                            loggerinst.warning("Unsupported option %s in %s" % (option, path))
+                elif header not in headers and header != "DEFAULT":
+                    loggerinst.warning("Unsupported header %s in %s." % (header, path))
```
r0x0d, line 433: I'm wondering, should we include an `else` statement by the end to log that the configuration file was not found? Something like: 
```suggestion
                elif header not in headers and header != "DEFAULT":
                    loggerinst.warning("Unsupported header %s in %s." % (header, path))
        else:
            loggerinst.debug("Couldn't find configuration file at '%s'." % path)
```

It feels to me that this could help us in debugging customer logs if they ever report that convert2rhel couldn't read the configuration the way it should. 

Let me know what you think about it, @hosekadam and @abadger

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,254 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+        (
+            mock_cli_arguments(["--password", "pass", "--config-file"]),
+            "[subscription_manager]\nactivation_key = key_cnf_file",
+            "Command line authentication method take precedence over method in configuration file.",
+            {"password": "pass", "activation_key": None},
+        ),
+    ),
+)
+def test_multiple_auth_src_combined(argv, content, message, output, caplog, monkeypatch, tmp_path):
+    """Test combination of password file or configuration file and CLI arguments."""
```
r0x0d, line 229: I would say we don't need to have docstrings in the test functions, in this case, it is pretty clear to me what the test is about. 
```suggestion
```

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,254 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+    convert2rhel.toolopts.tool_opts.__init__()
+    path = os.path.join(str(tmp_path), "convert2rhel.file")
```
r0x0d, line 231: The extension `.file` is on purpose here? I thought it should be `.ini`

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,254 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+        (
+            mock_cli_arguments(["--password", "pass", "-f", "file", "--config-file", "file"]),
+            ("pass_file", "[subscription_manager]\nactivation_key = pass_cnf_file"),
+            "Password file take precedence over the config file.",
+            {"password": "pass_file", "activation_key": None},
+        ),
+    ),
+)
+def test_multiple_auth_src_files(argv, content, message, output, caplog, monkeypatch, tmp_path):
+    """Test combination of password file, config file and CLI."""
```
r0x0d, line 259: I would say we don't need to have docstrings in the test functions, in this case, it is pretty clear to me what the test is about.
```suggestion
```

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,254 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+    (
+        (
+            mock_cli_arguments(["--password", "pass", "--activationkey", "key"]),
+            "Set only one of password or activation key.",
+            {"password": "pass", "activation_key": "key"},
+        ),
+    ),
+)
+def test_multiple_auth_src_cli(argv, message, output, caplog, monkeypatch):
+    """Test both auth methods in CLI."""
```
r0x0d, line 291:
```suggestion
```

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,254 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+        (
+            "[subscription_manager]\nincorrect_option = incorrect_content",
+            {"password": None, "activation_key": None},
+        ),
+        ("[INVALID_HEADER]\nactivation_key = correct_key", {"password": None, "activation_key": None}),
+        ("nonexisting", {"password": None, "activation_key": None}),
+    ),
+)
+def test_options_from_config_files_default(content, output, monkeypatch, tmp_path, caplog):
+    """Test config files in default path."""
```
r0x0d, line 321:
```suggestion

```

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,254 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+        (
+            "[subscription_manager]\nactivation_key = correct_key\nPassword = correct_password",
+            {"password": "correct_password", "activation_key": "correct_key"},
+        ),
+        (
+            "[subscription_manager]\nincorrect_option = incorrect_content",
+            {"password": None, "activation_key": None},
+        ),
+        ("[INVALID_HEADER]\nactivation_key = correct_key", {"password": None, "activation_key": None}),
+        ("nonexisting", {"password": None, "activation_key": None}),
```
r0x0d, line 317: You could change this "nonexisting" to `None`, it would make it simpler to compare later on in the test: 
```suggestion
        (None, {"password": None, "activation_key": None}),
```

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,254 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+    ),
+)
+def test_options_from_config_files_default(content, output, monkeypatch, tmp_path, caplog):
+    """Test config files in default path."""
+    path = os.path.join(str(tmp_path), "convert2rhel.ini")
+    if content != "nonexisting":
```
r0x0d, line 323: Following my previous comment about changing the `nonexisting`
```suggestion
    if content:
```

convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -145,3 +146,254 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
...
+        ),
+        (
+            "[subscription_manager]\nactivation_key = correct_key\nPassword = correct_password",
+            {"password": "correct_password", "activation_key": "correct_key"},
+            "[subscription_manager]\nincorrect_option = incorrect_option",
+        ),
+    ),
+)
+def test_options_from_config_files_specified(content, output, content_lower_priority, monkeypatch, tmp_path, caplog):
+    """Test user specified path for config file."""
```
r0x0d, line 361:
```suggestion

```

convert2rhel/unit_tests/utils_test.py
```diff
@@ -37,6 +37,7 @@
 from convert2rhel import unit_tests  # Imports unit_tests/__init__.py
 from convert2rhel import utils
 from convert2rhel.systeminfo import system_info
+from convert2rhel.toolopts import ToolOpts
```
r0x0d, line 40: This import doesn't seem to be used anywhere in the `utils_test.py` file. Is it needed by anything else? 
```suggestion

```

tests/integration/tier0/config-file/test_config_file.py
```diff
@@ -0,0 +1,65 @@
...
+    f.close()
+
+    with convert2rhel("--no-rpm-va --debug") as c2r:
+        c2r.expect("DEBUG - Found password in /root/.convert2rhel.ini")
+        c2r.sendcontrol("c")
+    assert c2r.exitstatus != 0
+
+    for path, content in paths:
+        os.remove(os.path.expanduser(path))
+    os.remove(os.path.expanduser("/etc/convert2rhel.ini"))
```
r0x0d, line 65: What do you think about splitting the integration tests into more tests in the same file? I feel that this can help in the readability and debugging process when needed. For starters, you could take all the occurrences of `with convert2rhel(...) as c2r` and split them into different tests. 

What do you think, @danmyway  and @kokesak?

r0x0d (commented):
Pretty good code, @hosekadam! I have spotted only a few things that could benefit from a change. 

Some of them are just stylish suggestions and you can decide if it makes sense to you or not. 

But anyway, let me know if you want to discuss any of the comments from me!

convert2rhel/toolopts.py
```diff
@@ -325,6 +362,19 @@ def _process_cli_options(self):
         if parsed_opts.org:
             tool_opts.org = parsed_opts.org
 
+        # Checks of multiple authentication sources
+        if tool_opts.password and tool_opts.activation_key:
```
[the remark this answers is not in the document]
hosekadam: @r0x0d If the `tool_opts` would be changed to `parsed_opts`, there will be need for more checks, because `password` and `activation_key` can came from more sources (like config file, password file). But in the `tool_opts` is everything merged and is possible to easily check, if there are more than one verification method.

convert2rhel/toolopts.py
```diff
@@ -338,5 +388,52 @@ def warn_on_unsupported_options():
...
 
+def options_from_config_files(cfg_path=None):
+    """Get the options from config file.
+
+    The order is:
+    1. if filename is set then load from it
+    2. ~/.convert2rhel.ini
+    3. /etc/convert2rhel.ini
+
+    Return dict of supported options with their values.
```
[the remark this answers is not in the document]
hosekadam: Yeah, this style looks better. I haven't noted, that something like this is used in our code, but now I see it. I will know for the future, thanks :)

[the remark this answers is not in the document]
hosekadam: @r0x0d Do you know how is this style called, so I can find available options? I can't see it in the PEP's guides mentioned in our `CONTRIBUTING.md`. And from what I've found on internet, the epytext was the closest one, but still not the same

convert2rhel/toolopts.py
hosekadam: @r0x0d that was already one of the ideas. But according to [this](https://github.com/oamg/convert2rhel/pull/451#discussion_r833005319) discussion it's done like this. In short it's because the compatibility between Python 2 & 3.

convert2rhel/unit_tests/toolopts_test.py
hosekadam: @r0x0d Yeah, they are maybe a bit unnecessary, but they helped me a lot while I was creating the test to keep me organized. If they should be removed, then it should be the last step :smile: Or they could help somehow in the future.

hosekadam:
I've updated the PR, based on @r0x0d review (thanks for that :)). Some renaming, improved docstring and splitted integration test.

hosekadam:
rebased

hosekadam:
I slightly remember that @r0x0d and @bocekm were talking about the format of the config file, because there might be need of options with more values (or something like that, not sure). Is it still the thing we should consider? Maybe it would be better to change it now, even it's almost done, but it's not merged. What do you think @abadger?

r0x0d:
> I slightly remember that @r0x0d and @bocekm were talking about the format of the config file, because there might be need of options with more values (or something like that, not sure). Is it still the thing we should consider? Maybe it would be better to change it now, even it's almost done, but it's not merged. What do you think @abadger?

Yes! We were talking about another RFE that would introduce a few more fields to the system config files (the ones under the data/8/x86_64/configs/{centos,oracle,...}-x86_64.cfg), that a new config file format would be good to have instead of the current one.  

As we mentioned in that meeting, probably using YAML/TOML/JSON would be a good choice to allow us more flexibility when adding fields that needs more options than just a string or a number and so on… I remember specifically saying that I would like to try TOML as a default config format since it will be present in python 3.11 in the stdlib, although, we would need to include `tomli` as a dependency for older versions, but the same is true for YAML, and TOML and YAML seems easier to read than JSON is, but that's just my opinion.

This was only an idea, and I don't really know the effort to change from a `*.cfg` format to `*.{yml,json,toml,...}` format. We wanted some inputs from @abadger before going into more details, as I remember he had some opinions about it! Probably in a cabal session we can revisit this topic and progress in the discussion. 

About your question, @hosekadam, IMO I think something like this would give too much trouble to change in the code, testing and so on… We can revisit this in a later PR to introduce these changes consistently across our configurations files, maybe, we figure out a good way of shaping a nice config file that can escalate well with our needs.

IMO, I wouldn't change the way your PR is right now, since it's look excellent and meets the requirements.

convert2rhel/toolopts.py
```diff
@@ -338,5 +385,60 @@ def warn_on_unsupported_options():
...
+       This function will parse the configuration file following a specific
+       order, which is:
+       1) Path provided by the user in cfg_path (Highest priority).
+       2) ~/.convert2rhel.ini (The 2nd highest priority).
+       3) /etc/convert2rhel.ini (The lowest priority).
+    :param cfg_path: Path of a custom configuration file
+    :type cfg_path: str
+
+    :return: Dict with the supported options alongside their values.
+    :rtype: dict[str | None, str | None]
```
r0x0d, line 404: Aesthetic suggestion for this docstring, nothing major. 

```suggestion
       3) /etc/convert2rhel.ini (The lowest priority).
       
    :param cfg_path: Path of a custom configuration file
    :type cfg_path: str
    :return: Dict with the supported options alongside their values.
    :rtype: dict[str | None, str | None]
```

r0x0d (approved):
Looks nice to me. Only two small suggestion regarding docstrings, but no need to apply right now. We can always make another PR to improve things like this, IMHO.

r0x0d (approved):
Looks excellent to me.

abadger (dismissed):
Approved.  And talked with @r0x0d who is also happy with this.  This is ready for QE to review the integration test.

tests/integration/tier0/config-file/test_config_file.py
```diff
@@ -0,0 +1,100 @@
...
+        os.chmod(os.path.expanduser(path), 0o600)
+
+
+def remove_files(paths):
+    for path, _ in paths:
+        os.remove(os.path.expanduser(path))
+
+
+def test_user_path_custom_filename(convert2rhel):
+    paths = (("~/.convert2rhel_custom.ini", "[subscription_manager]\nactivation_key = config_activationkey"),)
```
kokesak, line 18: I don't like the fact that `paths` contains the path to the file and the content as well. I would probably named it something like `config`.  Also using the `namedtuple` could be nice here and would make the code more readable I believe.

I would create a global level object:
`Config = namedtuple('Config', 'path content')`

And then in each test function I would do something like this:
```
config = Config(path="~/.convert2rhel_custom.ini", content="[subscription_manager]\nactivation_key = config_activationkey")
```
You can then easily pass the object to the `create_files` function and accress the variables like this:
`config.path` and `config.content` which is much nicer.

More on that: https://realpython.com/python-namedtuple/

tests/integration/tier0/config-file/test_config_file.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def test_user_path_pswd_file_priority(convert2rhel):
+    paths = (
+        ("~/.convert2rhel.ini", "[subscription_manager]\npassword = config_password"),
+        ("~/password_file", "file_password"),
+    )
+    create_files(paths)
+
+    with convert2rhel(('--no-rpm-va -f "~/password_file" --debug')) as c2r:
```
kokesak, line 61: brackets :D

tests/integration/tier0/config-file/test_config_file.py
```diff
@@ -0,0 +1,100 @@
...
+        c2r.expect("WARNING - Command line authentication method take precedence over method in configuration file.")
+        c2r.sendcontrol("c")
+    assert c2r.exitstatus != 0
+
+    remove_files(paths)
+
+
+def test_user_path_pswd_file_priority(convert2rhel):
+    paths = (
+        ("~/.convert2rhel.ini", "[subscription_manager]\npassword = config_password"),
```
kokesak, line 56: You can maybe create a list of Configs?

kokesak (dismissed):
The test cases are nice, however I would also add a last test case with full conversion. Eg, put all the necessary information inside the config file (load the env. variables with secrets) and do the conversion (whichever method you want). With that move the integration test to the tier1.

Otherwise good job Adam :D

danmyway:
Hey @hosekadam , do you need some help with addressing those change requests?

plans/tier1.fmf
```diff
@@ -283,3 +283,18 @@ discover+:
   - name: reboot after conversion
     how: ansible
     playbook: tests/ansible_collections/roles/reboot/main.yml
+
+/config_file:
+  adjust:
+      enabled: false
+      when: >
+        distro != centos-7
```
danmyway, line 291: Are we running this just for CentOS 7.9, @hosekadam ?

hosekadam: @danmyway Yeah, just one system is fine. I choose the centos7, but I think it doesn't matter which one we would run. So we can just wait for the fix you mentioned.

danmyway: Okay then, I would maybe suggest switching centos7 for centos8 at least, as the test suite might run faster there
Other than that, I am fine with that
Great job, thank you!

hosekadam: Switched :+1:

SpyTec (approved):

convert2rhel/toolopts.py
```diff
@@ -63,6 +80,7 @@ def _get_argparser():
             "  convert2rhel [-h]\n"
             "  convert2rhel [--version]\n"
             "  convert2rhel [-u username] [-p password | -f pswd_file] [--pool pool_id | -a] [--disablerepo repoid]"
+            "  convert2rhel [-c | --config-file conf_file]"
             " [--enablerepo repoid] [--serverurl url] [--keep-rhsm] [--no-rpm-va] [--debug] [--restart]"
             " [--disable-colors] [-y]\n"
```
bocekm, line 85: You changed:
`convert2rhel [-u username] [-p password | -f pswd_file] [--pool pool_id | -a] [--disablerepo repoid] [--enablerepo repoid] [--serverurl url] [--keep-rhsm] [--no-rpm-va] [--debug] [--restart] [--disable-colors] [-y]`
to
`convert2rhel [-u username] [-p password | -f pswd_file] [--pool pool_id | -a] [--disablerepo repoid]  convert2rhel [-c | --config-file conf_file] [--enablerepo repoid] [--serverurl url] [--keep-rhsm] [--no-rpm-va] [--debug] [--restart] [--disable-colors] [-y]`

If somebody runs the command this way, convert2rhel would fail.

As I understand the meaning of pipe it's that you should use either one option or another, not both. In this case, for example, either password or a config file.

It should be (including removing the deprecated `-f pswd_file`):
```python
            "  convert2rhel [-u username] [-p password | -c conf_file_path] [--pool pool_id | -a] [--disablerepo repoid]"
            " [--enablerepo repoid] [--serverurl url] [--keep-rhsm] [--no-rpm-va] [--debug] [--restart]"
            " [--disable-colors] [-y]\n"
            "  convert2rhel [--no-rhsm] [--disablerepo repoid]"
            " [--enablerepo repoid] [--no-rpm-va] [--debug] [--restart] [--disable-colors] [-y]\n"
            "  convert2rhel [-k activation_key | -c conf_file_path] [-o organization] [--pool pool_id | -a] [--disablerepo repoid] [--enablerepo"
            " repoid] [--serverurl url] [--keep-rhsm] [--no-rpm-va] [--debug] [--restart] [--disable-colors] [-y]"
```

convert2rhel/toolopts.py
```diff
@@ -179,6 +197,13 @@ def _register_options(self):
             " 'subscription-manager orgs'. From the listed pairs"
             " Name:Key, use the Key here.",
         )
+        group.add_option(
+            "-c",
+            "--config-file",
+            help="Configuration file containing"
+            " password or activation key for subscription-manager."
```
bocekm, line 204:
```suggestion
            help="A configuration file to safely provide either a user password or an activation key for registering"
                 " the system through subscription-manager. Alternatively, passing these values through the"
                 " --activationkey or --password option would leak them through a list of running processes."
```

To the description of the password and activation key, add something like: 
_"We recommend using the --config-file option instead to prevent leaking the password/activation key through a list of running processes."_

Also remove the mention of `--password-from-file` from the password option description.

convert2rhel/toolopts.py
```diff
@@ -325,6 +362,19 @@ def _process_cli_options(self):
         if parsed_opts.org:
             tool_opts.org = parsed_opts.org
 
+        # Checks of multiple authentication sources
+        if tool_opts.password and tool_opts.activation_key:
+            loggerinst.warning("Set only one of password or activation key. Activation key take precedence.")
```
bocekm, line 367:
```suggestion
loggerinst.warning("Passing the RHSM password or activation key through the --activationkey or --password options is insecure as it leaks the values through the list of running processes. We recommend using the safer --config-file option instead.")
loggerinst.warning("Either a password or an activation key can be used for system registration. We're going to use the activation key.")
```

convert2rhel/toolopts.py
```diff
@@ -325,6 +362,19 @@ def _process_cli_options(self):
...
+
+        if parsed_opts.password and parsed_opts.password_from_file:
+            loggerinst.warning("Password file argument take precedence over the password argument.")
```
bocekm, line 370:
```suggestion
loggerinst.warning("You've passed the RHSM password through both the --password-from-file and the --password option. We're going to use the password from file.")
```

convert2rhel/toolopts.py
```diff
@@ -325,6 +362,19 @@ def _process_cli_options(self):
...
+
+        if (config_opts.activation_key or config_opts.password) and (parsed_opts.activationkey or parsed_opts.password):
+            loggerinst.warning("Command line authentication method take precedence over method in configuration file.")
```
bocekm, line 373:
```suggestion
loggerinst.warning("You've passed either the RHSM password or activation key through both the command line and the configuration file. We're going to use the command line values.")
```

convert2rhel/toolopts.py
```diff
@@ -325,6 +362,19 @@ def _process_cli_options(self):
...
+
+        if (config_opts.activation_key or config_opts.password) and parsed_opts.password_from_file:
+            loggerinst.warning("Password file take precedence over the config file.")
```
bocekm, line 376: I'd change this precedence for two reasons:
- The password file option is deprecated
- We prefer customers to use the activation key rather than the username/password combination as it's more user friendly. So if the user specifies both the AK and the password, I'd take the AK.

bocekm: Ok, what I've proposed above might be too big of a change considering the PR is already merged. So perhaps just changing the wording is enough:
```suggestion
loggerinst.warning("You have passed the RHSM credentials both through a config file and through a password file. We're going to use the password file.")
```

convert2rhel/toolopts.py
hosekadam: Thanks for clarification what the pipe does :+1:

PR #455, opened 2022-03-02
Improve integration test for submanager packages

kokesak:
We are adding a second test case where no "subscription manager"
package is installed. The conversion has to get to the last
point of the rollback without any issues.


r0x0d:
@kokesak does this PR have an issue associated with?

kokesak:
> @kokesak does this PR have an issue associated with?

No it doesn't. I can create a JIRA ticket for it and link it.

tests/integration/tier1/remove-all-submgr-pkgs/test_no_submgr_pkg_installed.py
```diff
@@ -0,0 +1,34 @@
...
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        c2r.expect("Continue with the system conversion?")
+        c2r.sendline("y")
+        # On OracleLinux8 there is one question less than on other distros
+        if "oracle-8" not in platform.platform():
```
danmyway, line 25: If I recall correctly, centos-7 does have one less question as well.

kokesak: I checked the logs and it's ok on CentOS-7.
http://artifacts.osci.redhat.com/testing-farm/e9737651-1226-4697-a677-0c39dd0991ed/work-remove_all_submgr_pkgsnwN0V6/log.txt
It may be unreliable though. As I think it depends wherter we have installed some packages that c2r wants to remove during the pre-conversion checks.

danmyway: Ah, I see, my bad then

danmyway (approved):

danmyway (approved):
LGTM!

PR #458, opened 2022-03-07
Allow converting to RHEL 8 EUS minor versions

r0x0d:
This PR introduces the ability to make conversions from older systems (EUS) like CentOS/Oracle Linux 8.4. 

# What are we introducing with this? 

We are introducing a way for the user to run convert2rhel from an older CentOS/Oracle Linux machine without the need to update their system to the latest release version available. 

Right now, we are supporting these versions here: 

## CentOS Linux

Supporting versions `8.4` and `8.5`.

## Oracle Linux

Supporting version `8.4`.

# What we did in this PR?

There is a lot of changes committed, each of them has a separate commit to help the reviewers to track the changes better. 

* https://github.com/oamg/convert2rhel/pull/458/commits/d8033e26eda90ab4a4ee074c26aa72ca665dbe4e: Introduces the hardcoded repositories for CentOS/Oracle Linux 8.4.
* https://github.com/oamg/convert2rhel/pull/458/commits/db71c9f8281d0454f3a06920afc4fb7a9663991b: Introduces a refactor from some classes that handle the `restore`/`backup` functionality in convert2rhel from `utils.py` to `backup.py`. At first, we've bumped into a circular dependency import problem trying to get our code working for this, so we decided to split a bit and do this refactoring.
* https://github.com/oamg/convert2rhel/pull/458/commits/8124c96dfb4f44095bec662b2b276175e9b8d715: Introduces the fixing of the unit tests that broke because of the refactor, mostly fixing import names, since they changed from `utils.*` to `backup.*`.
* https://github.com/oamg/convert2rhel/pull/458/commits/3d666535609960ab18b033e8a23fe1743ba9215d: Moves around the tests from the refactored classes from `utils_tests.py` to `backup_tests.py`.
* https://github.com/oamg/convert2rhel/pull/458/commits/810dc418d6f5601b1bd508d242122a353b141a6e: Adding new tests for the new functionalities that we are still implementing. 

# ~Blocked by #356~

~Right now, we are depending on the PR#356 to be merged before this one if we want to keep changes to where they belong, the reason is that in that PR we have a function that should be changed to include the usage of the hardcoded repos, more specifically, https://github.com/oamg/convert2rhel/pull/356/files#diff-cf19bc90f65d9e26e1b6fa7dcd84a18dfba097f65706e211a4f4ac71b1b467f6R60.~

~As stated by @bocekm at a previous meeting, we could merge both PRs #356 and #458 individually and later on open a new PR#??? that introduces the desired-blocked change that it's still missing.~

# Notes

Since this is a big change being introduced to something that should not the this big, I think that we need to take closer attention, especially to the refactoring thing, as I tried to fix every single import from the changed files, but who knows, something could be missing and we didn't even notice.

---- 

Jira reference (If any): https://issues.redhat.com/browse/RHELC-291
Bugzilla reference (If any): 

### DIFF
--- a/plans/tier1.fmf
+++ b/plans/tier1.fmf
@@ -191,7 +191,7 @@ discover+:
   - name: allow access to Satellite only
     how: shell
     script: pytest -svv tests/integration/tier1/convert-offline-systems/prepare_system.py
-  - name: run offline conversion
+  - name: run offline conversion using the satellite
     how: shell
     script: pytest -svv tests/integration/tier1/convert-offline-systems/run_conversion.py
   - name: reboot after conversion
@@ -199,13 +199,13 @@ discover+:
     playbook: tests/ansible_collections/roles/reboot/main.yml
 
 /system_up_to_date:
-  # On Oracle Linux 8 we do not have minor version repositories so we skip the check for up-to-date system
+  # On Oracle Linux 8.4 we do not have minor version repositories so we skip the check for up-to-date system.
+  # The other distributions (including latest Oracle Linux 8 ) have all the necessary repositories
+  # to make the check possible.
   adjust:
     enabled: false
     when: >
-      distro != centos-8 and
-      distro != centos-7 and
-      distro != oraclelinux-7
+      distro == oraclelinux-8.4
   discover+:
     test: checks-after-conversion
   prepare+:
@@ -245,3 +245,19 @@ discover+:
     - name: reboot after conversion
       how: ansible
       playbook: tests/ansible_collections/roles/reboot/main.yml
+
+/satellite_non_eus_repos:
+  adjust:
+    enabled: false
+    when: >
+      distro != centos-8.4 and
+      distro != oraclelinux-8.4
+  discover+:
+    test: checks-after-conversion
+  prepare+:
+  - name: main conversion preparation
+    how: shell
+    script: pytest -svv tests/integration/tier1/satellite_non_eus_repos/satellite_non_eus_repos.py
+  - name: reboot after conversion
+    how: ansible
+    playbook: tests/ansible_collections/roles/reboot/main.yml
--- a/tests/ansible_collections/roles/add-custom-repos/main.yml
+++ b/tests/ansible_collections/roles/add-custom-repos/main.yml
@@ -4,6 +4,8 @@
   become: false
 - include: rhel7-repos.yml
   when: ansible_facts['distribution_major_version'] == "7"
+  # When Oracle Linux 8.7 is released, the "8.6" needs to change to "8.7" and the
+  # "8.6" is to be moved to the condition below related to enabling RHEL 8 EUS repos
 - include: rhel8-repos.yml
   when: ansible_facts['distribution_version'] == "8.5" or ansible_facts['distribution_version'] == "8.6"
 - include: rhel8-eus-repos.yml
--- a/tests/ansible_collections/roles/update-system/tasks/main.yml
+++ b/tests/ansible_collections/roles/update-system/tasks/main.yml
@@ -1,5 +1,7 @@
 ---
-# We do not update Oracle Linux 8.4
+# We can't update Oracle Linux 8.4 to the latest 8.4 packages because there are no public Oracle Linux 8.4 repositories available.
+# The CentOS Linux 8.4 can be updated using this role because the repositories on the system image has been
+# tweaked to point to the CentOS Linux 8.4 repositories in the Vault.
 - include: update-system.yml
   when:
-    - ansible_facts['distribution_version'] != "8.4" or ansible_facts['distribution'] != "OracleLinux"
+    - not (ansible_facts['distribution_version'] == "8.4" and ansible_facts['distribution'] == "OracleLinux")
--- a/tests/integration/tier0/backup-release/test_backup_release.py
+++ b/tests/integration/tier0/backup-release/test_backup_release.py
@@ -25,12 +25,12 @@ def test_backup_os_release_no_envar(shell, convert2rhel):
     assert shell("rpm -i {}".format(pkg_dst)).returncode == 0
 
     # Move all repos to other location, so it is not being used
-    assert shell("mkdir /tmp/s_backup")
+    assert shell("mkdir /tmp/s_backup").returncode == 0
     assert shell("mv /etc/yum.repos.d/* /tmp/s_backup/").returncode == 0
 
     # EUS version use hardoced repos from c2r as well
     if "centos-8" in system or "oracle-8.4" in system:
-        assert shell("mkdir /tmp/s_backup_eus")
+        assert shell("mkdir /tmp/s_backup_eus").returncode == 0
         assert shell("mv /usr/share/convert2rhel/repos/* /tmp/s_backup_eus/").returncode == 0
 
     # Since we are moving all repos away, we need to bypass kernel check
--- a/tests/integration/tier1/convert-offline-systems/run_conversion.py
+++ b/tests/integration/tier1/convert-offline-systems/run_conversion.py
@@ -4,15 +4,27 @@ import platform
 from envparse import env
 
 
-def test_convert_offline_systems(shell, convert2rhel):
+def test_convert_offline_systems(convert2rhel):
     """Test converting systems not connected to the Internet but requiring sub-mgr (e.g. managed by Satellite)."""
 
     os.environ["CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK"] = "1"
-    with convert2rhel(
-        ("-y --no-rpm-va -k {} -o {} --keep-rhsm --debug").format(
-            env.str("SATELLITE_KEY"),
-            env.str("SATELLITE_ORG"),
-        )
-    ) as c2r:
-        c2r.expect("Conversion successful!")
-    assert c2r.exitstatus == 0
+    source_distro = platform.platform()
+
+    if "centos-8.4" in source_distro or "oracle-8.4" in source_distro:
+        with convert2rhel(
+            ("-y --no-rpm-va -k {} -o {} --keep-rhsm --debug").format(
+                env.str("SATELLITE_KEY_EUS"),
+                env.str("SATELLITE_ORG"),
+            )
+        ) as c2r:
+            pass
+        assert c2r.exitstatus == 0
+    else:
+        with convert2rhel(
+            ("-y --no-rpm-va -k {} -o {} --keep-rhsm --debug").format(
+                env.str("SATELLITE_KEY"),
+                env.str("SATELLITE_ORG"),
+            )
+        ) as c2r:
+            pass
+        assert c2r.exitstatus == 0
--- a/tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
+++ b/tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
@@ -29,6 +29,7 @@ def test_non_latest_kernel(shell, convert2rhel):
     System has non latest kernel installed, thus the conversion
     has to be inhibited.
     """
+    system_version = platform.platform()
 
     with convert2rhel(
         ("-y --no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug").format(
@@ -38,7 +39,11 @@ def test_non_latest_kernel(shell, convert2rhel):
             env.str("RHSM_POOL"),
         )
     ) as c2r:
-        if "centos-7" in platform.platform() or "oracle-7" in platform.platform():
+        if "centos-7" in system_version or "oracle-7" in system_version:
+            c2r.expect(
+                "The version of the loaded kernel is different from the latest version on the enabled system repositories."
+            )
+        elif "oracle-8" in system_version:
             c2r.expect(
                 "The version of the loaded kernel is different from the latest version on the enabled system repositories."
             )
--- a/tests/ansible_collections/roles/update-system/tasks/update-system.yml
+++ b/tests/ansible_collections/roles/update-system/tasks/update-system.yml
@@ -7,5 +7,4 @@
     # and the version of the package is not the latest in copr repo it would be
     # updated at this point. The package version would then differ from the intended one.
     exclude: convert2rhel
-    skip_broken: yes
   register: update
--- a/tests/integration/tier1/checks-after-conversion/test_enabled_repositories.py
+++ b/tests/integration/tier1/checks-after-conversion/test_enabled_repositories.py
@@ -1,5 +1,7 @@
 import platform
 
+from os.path import exists
+
 
 def test_enabled_repositories(shell):
     """Testing, if the EUS repostitories are enabled after conversion"""
@@ -7,14 +9,20 @@ def test_enabled_repositories(shell):
     enabled_repos = shell("yum repolist").output
 
     # Once we will decide to use EUS 8.6 we have to add them here as well
-    if "redhat-8.4" in system_version:
-        assert "rhel-8-for-x86_64-appstream-eus-rpms" in enabled_repos
-        assert "rhel-8-for-x86_64-baseos-eus-rpms" in enabled_repos
-    elif "redhat-8.5" in system_version or "redhat-8.6" in system_version:
-        assert "rhel-8-for-x86_64-baseos-rpms" in enabled_repos
-        assert "rhel-8-for-x86_64-appstream-rpms" in enabled_repos
-    elif "redhat-7.9" in system_version:
-        assert "rhel-7-server-rpms/7Server/x86_64" in enabled_repos
-
-    # We need to unregister the system after the conversion
-    shell("subscription-manager unregister")
+    try:
+        if "redhat-8.4" in system_version:
+            # Handle the special test case scenario where we do not use the premium account with EUS repositories
+            if exists("/non_eus_repos_used"):
+                assert "rhel-8-for-x86_64-baseos-rpms" in enabled_repos
+                assert "rhel-8-for-x86_64-appstream-rpms" in enabled_repos
+            else:
+                assert "rhel-8-for-x86_64-appstream-eus-rpms" in enabled_repos
+                assert "rhel-8-for-x86_64-baseos-eus-rpms" in enabled_repos
+        elif "redhat-8.5" in system_version or "redhat-8.6" in system_version:
+            assert "rhel-8-for-x86_64-baseos-rpms" in enabled_repos
+            assert "rhel-8-for-x86_64-appstream-rpms" in enabled_repos
+        elif "redhat-7.9" in system_version:
+            assert "rhel-7-server-rpms/7Server/x86_64" in enabled_repos
+    finally:
+        # We need to unregister the system after the conversion
+        shell("subscription-manager unregister")
--- a/tests/integration/tier1/checks-after-conversion/test_release_version.py
+++ b/tests/integration/tier1/checks-after-conversion/test_release_version.py
@@ -6,9 +6,10 @@ def test_basic_conversion(shell):
     assert "Red Hat Enterprise Linux" in os_release
 
 
-def test_correct_distro(shell):
-    shell("cat /etc/migration-results")
-
+def test_correct_distro():
+    """ "
+    Check that we landed on the correct version
+    """
     with open("/etc/migration-results") as json_file:
         json_data = json.load(json_file)
         source_distro = json_data["activities"][0]["source_os"]
--- a/tests/integration/tier1/convert-offline-systems/prepare_system.py
+++ b/tests/integration/tier1/convert-offline-systems/prepare_system.py
@@ -1,3 +1,4 @@
+import os
 import re
 import socket
 
@@ -51,3 +52,5 @@ def test_prepare_system(shell):
     assert shell("iptables -I OUTPUT -d 8.8.8.8 -j DROP").returncode == 0
 
     assert shell("systemctl enable dnsmasq && systemctl restart dnsmasq").returncode == 0
+
+    os.environ["CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK"] = "1"
--- a/tests/integration/tier1/method/custom_repos.py
+++ b/tests/integration/tier1/method/custom_repos.py
@@ -40,6 +40,7 @@ def test_run_conversion_using_custom_repos(shell, convert2rhel):
 
     with convert2rhel("-y --no-rpm-va --disable-submgr {} --debug".format(enable_repo_opt)) as c2r:
         c2r.expect("Conversion successful!")
+    assert c2r.exitstatus == 0
 
     # after the conversion using custom repostitories it is expected to enable repos by yourself
     if system_version.major == 7:
@@ -52,5 +53,3 @@ def test_run_conversion_using_custom_repos(shell, convert2rhel):
         elif system_version.minor == 4:
             enable_repo_opt = "--enable rhel-8-for-x86_64-baseos-eus-rpms --enable rhel-8-for-x86_64-appstream-eus-rpms"
     shell("yum-config-manager {}".format(enable_repo_opt))
-
-    assert c2r.exitstatus == 0
--- a/tests/integration/tier1/method/satellite.py
+++ b/tests/integration/tier1/method/satellite.py
@@ -1,3 +1,5 @@
+import platform
+
 from envparse import env
 
 
@@ -12,11 +14,23 @@ def test_satellite_conversion(shell, convert2rhel):
     pkg_dst = "/usr/share/convert2rhel/subscription-manager/katello-ca-consumer-latest.noarch.rpm"
     assert shell("wget --no-check-certificate --output-document {} {}".format(pkg_dst, pkg_url)).returncode == 0
 
-    with convert2rhel(
-        ("-y --no-rpm-va -k {} -o {} --debug").format(
-            env.str("SATELLITE_KEY"),
-            env.str("SATELLITE_ORG"),
-        )
-    ) as c2r:
-        pass
-    assert c2r.exitstatus == 0
+    source_distro = platform.platform()
+
+    if "centos-8.4" in source_distro or "oracle-8.4" in source_distro:
+        with convert2rhel(
+            ("-y --no-rpm-va -k {} -o {} --debug").format(
+                env.str("SATELLITE_KEY_EUS"),
+                env.str("SATELLITE_ORG"),
+            )
+        ) as c2r:
+            pass
+        assert c2r.exitstatus == 0
+    else:
+        with convert2rhel(
+            ("-y --no-rpm-va -k {} -o {} --debug").format(
+                env.str("SATELLITE_KEY"),
+                env.str("SATELLITE_ORG"),
+            )
+        ) as c2r:
+            pass
+        assert c2r.exitstatus == 0
--- a/tests/integration/tier1/satellite_non_eus_repos/main.fmf
+++ b/tests/integration/tier1/satellite_non_eus_repos/main.fmf
@@ -1,0 +1,6 @@
+summary: test using satellite with non eus account for the EUS distro
+
+tier: 1
+
+test: |
+  pytest -svv
--- a/tests/integration/tier1/satellite_non_eus_repos/satellite_non_eus_repos.py
+++ b/tests/integration/tier1/satellite_non_eus_repos/satellite_non_eus_repos.py
@@ -1,0 +1,31 @@
+from envparse import env
+
+
+def test_missing_os_release(shell, convert2rhel):
+    """
+    This test case verify that it's possible to do full conversion when /etc/os-release
+    file is not present on the system.
+    """
+
+    # Mark the system so the check for the enabled repos after the conversion handles this special case
+    assert shell("touch /non_eus_repos_used").returncode == 0
+
+    # Remove subscription manager if installed
+    assert shell("yum remove subscription-manager -y").returncode == 0
+
+    assert shell("yum install wget -y").returncode == 0
+
+    # Install katello package
+    pkg_url = "https://dogfood.sat.engineering.redhat.com/pub/katello-ca-consumer-latest.noarch.rpm"
+    pkg_dst = "/usr/share/convert2rhel/subscription-manager/katello-ca-consumer-latest.noarch.rpm"
+    assert shell("wget --no-check-certificate --output-document {} {}".format(pkg_dst, pkg_url)).returncode == 0
+
+    with convert2rhel(
+        ("-y --no-rpm-va -k {} -o {} --debug").format(
+            env.str("SATELLITE_KEY"),
+            env.str("SATELLITE_ORG"),
+        )
+    ) as c2r:
+        c2r.expect("WARNING - rhel-8-for-x86_64-baseos-eus-rpms repository is not available")
+        c2r.expect("WARNING - rhel-8-for-x86_64-appstream-eus-rpms repository is not available")
+    assert c2r.exitstatus == 0
--- a/tests/integration/tier1/system-up-to-date/install_non_latest_kernel.py
+++ b/tests/integration/tier1/system-up-to-date/install_non_latest_kernel.py
@@ -33,3 +33,7 @@ def test_install_one_kernel(shell):
         shell("grub2-set-default 'Oracle Linux Server 7.9, with Linux 3.10.0-1160.el7.x86_64'")
     elif "centos-8" in system_version:
         handle_centos8(shell)
+    # Test is being run only for the latest released oracle-linux
+    elif "oracle-8" in system_version:
+        assert shell("yum install kernel-4.18.0-80.el8.x86_64 -y").returncode == 0
+        shell("grub2-set-default 'Oracle Linux Server (4.18.0-80.el8.x86_64) 8.0'")
--- a/tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
+++ b/tests/integration/tier1/system-up-to-date/test_system_up_to_date.py
@@ -14,13 +14,12 @@ def test_skip_kernel_check(shell, convert2rhel):
     any kernel packages present in repos.
     """
 
+    # Move all repos to other location, so it is not being used
     assert shell("mkdir /tmp/s_backup")
     assert shell("mv /etc/yum.repos.d/* /tmp/s_backup/").returncode == 0
-    # TODO this has to be fixed
-    # Move all repos to other location, so it is not being used
-    # EUS version use hardoced repos from c2r
 
-    if "centos-8" in system_version in system_version:
+    # EUS version use hardoced repos from c2r as well
+    if "centos-8" in system_version:
         assert shell("mkdir /tmp/s_backup_eus")
         assert shell("mv /usr/share/convert2rhel/repos/* /tmp/s_backup_eus/").returncode == 0
 
@@ -34,7 +33,7 @@ def test_skip_kernel_check(shell, convert2rhel):
     ) as c2r:
         if "centos-7" in system_version or "oracle-7" in system_version:
             c2r.expect("Could not find any kernel from repositories to compare against the loaded kernel.")
-        elif "centos-8" in system_version:
+        else:
             c2r.expect("Could not find any kernel-core from repositories to compare against the loaded kernel.")
     assert c2r.exitstatus != 0
 


r0x0d:
This PR is waiting for PR#356 to be merged because there has a function that we need to update.

r0x0d:
Converting a CentOS 8.4 system worked just fine. 

```bash
[root@c2r-20220314145824 ~]# cat /etc/system-release
Red Hat Enterprise Linux release 8.4 (Ootpa)
```

r0x0d:
Converting Oracle Linux 8.4 worked as well: 

```bash
[root@c2r-20220314153922 ~]# cat /etc/system-release
Red Hat Enterprise Linux release 8.4 (Ootpa)
```

convert2rhel/data/8/x86_64/repos/centos-8.5/CentOS-Linux-BaseOS.repo
```diff
@@ -0,0 +1,17 @@
...
+# close to the client.  You should use this for CentOS updates unless you are
+# manually picking other mirrors.
+#
+# If the #mirrorlist does not work for you, you can try the commented out
+# baseurl line instead.
+
+[baseos]
+name=CentOS Linux 8.5.2111 - BaseOS
+#mirrorlist=http://#mirrorlist.centos.org/?release=8.5.2111&arch=$basearch&repo=BaseOS&infra=$infra
+baseurl=http://vault.centos.org/$contentdir/8.5.2111/BaseOS/$basearch/os/
```
SpyTec, line 14: All vault links have to be https

convert2rhel/pkghandler.py
```diff
@@ -1043,25 +1055,29 @@ def _get_packages_to_update_yum():
...
-    # when the DNF API gets called, those variables are not populated by default.
-    if system_info.id == "oracle":
-        base.conf.substitutions["ocidomain"] = "oracle.com"
-        base.conf.substitutions["ociregion"] = ""
-
-    # Same thing for CentOS, but this time, the URL for the vault has changed
-    # and instead of just ussing $releasever (8.5, 8.4...), we need either
-    # using the correct $releasever (8.5.2111) or setting up the $contentdir
-    elif system_info.id == "centos":
-        base.conf.substitutions["contentdir"] = system_info.id
```
r0x0d, line 1063: This was only a temporary solution to fix #356. We replaced it with a better way of expanding the variables on the system. 

Basically, on lines 1079-1080, we expand the variables that are located under `/etc/dnf/vars`.

convert2rhel/backup.py
```diff
@@ -0,0 +1,180 @@
...
+    def __init__(self):
+        self.installed_pkgs = []
+        self.removed_pkgs = []
+
+    def track_installed_pkg(self, pkg):
+        """Add a installed RPM pkg to the list of installed pkgs."""
+        self.installed_pkgs.append(pkg)
+
+    def track_installed_pkgs(self, pkgs):
+        """Track packages installed  before the PONR to be able to remove them later (roll them back) if needed."""
```
SpyTec, line 24:
```suggestion
        """Track packages installed before the PONR to be able to remove them later (roll them back) if needed."""
```

convert2rhel/repo.py
```diff
@@ -68,3 +74,31 @@ def restore_yum_repos():
...
+
+
+def is_eus_repos_available():
+    """"""
+    hardcoded_reposdir = _get_hardcoded_repofiles_dir()
+    reposdir = None
+    if os.path.exists(hardcoded_reposdir) and system_info.has_internet_access:
+        reposdir = hardcoded_reposdir
+
+    return reposdir
```
SpyTec: Method implies boolean return but returns a `reposdir` or `None`. I'd rename to `get_eus_repos_available`

convert2rhel/subscription.py
```diff
@@ -441,6 +455,27 @@ def enable_repos(rhel_repoids):
...
+            loggerinst.info("Falling back to using default rhsm repositories.")
+            # Fallback to the default_rhsm_repoids
+            _submgr_enable_repos(system_info.default_rhsm_repoids)
+            repos_to_enable = system_info.default_rhsm_repoids
+    else:
+        # This could be either the default_rhsm repos or any user specific
+        # repoids
+        _submgr_enable_repos(repos_to_enable)
+
+    system_info.submgr_enabled_repos = repos_to_enable
```
SpyTec, line 474: None of this is not covered by tests

convert2rhel/subscription.py
```diff
@@ -441,6 +455,27 @@ def enable_repos(rhel_repoids):
...
+
+
+def _submgr_enable_repos(repos_to_enable):
+    """ """
```
SpyTec, line 478:
```suggestion
    """Go through subscription manager repos and try to enable them through subscription-manager."""
```

convert2rhel/systeminfo.py
```diff
@@ -98,6 +101,7 @@ def __init__(self):
         self.kmods_to_ignore = []
         # Booted kernel VRA (version, release, architecture), e.g. "4.18.0-240.22.1.el8_3.x86_64"
         self.booted_kernel = ""
+        #
```
SpyTec, line 104:
```suggestion
```

convert2rhel/systeminfo.py
```diff
@@ -306,6 +320,37 @@ def get_enabled_rhel_repos(self):
...
+            * Comparision of different methods to check internet connections: https://stackoverflow.com/a/33117579
+            * Redirecting ip addresses: https://stackoverflow.com/a/33117579https://superuser.com/questions/954665/how-to-redirect-route-ip-address-to-another-ip-address
+
+        :param host: The host to stablish to connection. Defatuls to "8.8.8.8".
+        :type host: str
+        :param port: The port to use in the connection. Defaults to 53.
+        :type port: int
+        :param timeout: The maximum number of timeouts for the connection. Defaults to 3.
+        :type port: int
+        """
```
SpyTec, line 344:
```suggestion
        """Check whether or not the machine is connect to the internet.

        This method will try to establish a socket connection through the
        default host in the method signature (8.8.8.8), if it's connected
        successfully, then we know that internet is accessibly from the host.

        .. warnings::
            We might have some problems with this if the host machine is using
            a NAT gateway to route the outbound requests

        .. seealso::
            * Comparison of different methods to check internet connections: https://stackoverflow.com/a/33117579
            * Redirecting IP addresses: https://superuser.com/questions/954665/how-to-redirect-route-ip-address-to-another-ip-address

        :param host: The host to establish a connection to. Defaults to "8.8.8.8".
        :type host: str
        :param port: The port for the connection. Defaults to 53.
        :type port: int
        :param timeout: The timeout in seconds for the connection. Defaults to 3.
        :type port: int
        """
```

tests/ansible_collections/roles/hardcode-minor-repos/tasks/centos-repos.yml
```diff
@@ -0,0 +1,21 @@
...
+    gpgcheck: yes
+    enabled: yes
+    gpgkey: file:///etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
+    file: CentOS-Linux-AppStream
+
+- name: Overwrite BaseOS repository (CentOS8.4)
+  yum_repository:
+    name: baseos
+    description: BaseOS repo
+    baseurl: http://vault.centos.org/$contentdir/8.4.2105/BaseOS/$basearch/os/
```
SpyTec, line 16: Vault should be https here too

SpyTec (changes requested):

convert2rhel/backup.py
r0x0d: Done.

convert2rhel/systeminfo.py
r0x0d: Done.

convert2rhel/utils.py
```diff
@@ -641,4 +468,19 @@ def string_to_version(verstring):
     return (epoch, version, release)
 
 
-changed_pkgs_control = ChangedRPMPackagesController()  # pylint: disable=C0103
+def remove_orphan_folders():
+    """Even after removing redhat-release-* package, some of its folders are
+    still present, are empty, and that blocks us from installing centos-release
+    pkg back. So, by now, we are removing them manually.
+    """
```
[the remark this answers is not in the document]
r0x0d: Done.

convert2rhel/data/8/x86_64/repos/centos-8.4/CentOS-Linux-Extras.repo
```diff
@@ -0,0 +1,17 @@
+# CentOS-Linux-Extras.repo
+#
+# The #mirrorlist system uses the connecting IP address of the client and the
+# update status of each mirror to pick current mirrors that are geographically
+# close to the client.  You should use this for CentOS updates unless you are
+# manually picking other mirrors.
+#
+# If the #mirrorlist does not work for you, you can try the commented out
+# baseurl line instead.
```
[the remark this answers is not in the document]
SpyTec: Not worth it for this PR I don't think

convert2rhel/subscription.py
r0x0d: Done.

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -163,8 +162,8 @@ def test_modified_rpm_files_diff_with_differences_after_conversion(self):
     @unit_tests.mock(utils, "run_subprocess", RunSubprocessMocked(("rpmva\n", 0)))
     def test_generate_rpm_va(self):
         # Check that rpm -Va is executed (default) and stored into the specific file.
+        tool_opts.no_rpm_va = False
```
r0x0d, line 165: I had to include this line in this test, otherwise, it would still be failing after adding the tests in the `subscription_test.py`. Probably we have some leaks when using the `pretend_os` fixture.

SpyTec (approved):
Code itself looks good. Not tested it in VMs. @bocekm should take a look over as well

convert2rhel/data/8/x86_64/repos/oracle-8.4/oracle-linux-ol8.repo
```diff
@@ -0,0 +1,70 @@
...
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-oracle
+gpgcheck=1
+enabled=0
+
+[ol8_addons]
+name=Oracle Linux 8 Addons ($basearch)
+baseurl=https://yum$ociregion.$ocidomain/repo/OracleLinux/OL8/addons/$basearch/
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-oracle
+gpgcheck=1
+enabled=0
```
bocekm, line 70: Only the `ol8_u4_baseos_base` contains OL 8.4 content. The rest is for the OL 8.latest.
```suggestion
[ol8_u4_baseos_base]
name=Oracle Linux 8.4 BaseOS ($basearch)
baseurl=https://yum$ociregion.$ocidomain/repo/OracleLinux/OL8/4/baseos/base/$basearch/
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-oracle
gpgcheck=1
enabled=0
```

convert2rhel/data/8/x86_64/repos/oracle-8.4/uek-ol8.repo
```diff
@@ -0,0 +1,13 @@
+[ol8_UEKR6]
```
bocekm, line 1: Remove this whole file for these reasons:
1) We require users to install and boot into the RHEL-compatible kernel, not UEK, so the UEK doesn't need to be updated
2) the UEK repo is not OL 8.4-specific. It might or might not require dependencies from OL 8.5+ BaseOS/Appstream

convert2rhel/data/8/x86_64/repos/oracle-8.4/virt-ol8.repo
```diff
@@ -0,0 +1,6 @@
+[ol8_kvm_appstream]
```
bocekm, line 1: Remove this whole file for the reason that it's not OL 8.4-specific. It might or might not require dependencies from OL 8.5+ BaseOS/Appstream.

convert2rhel/repo.py
```diff
@@ -20,15 +20,21 @@
...
 loggerinst = logging.getLogger(__name__)
 
 
 def get_rhel_repoids():
     """Get IDs of the Red Hat CDN repositories that correspond to the current system."""
-    repos_needed = system_info.default_rhsm_repoids
+
+    reposdir = get_eus_repos_available()
+
+    if not reposdir:
```
bocekm, line 34:
```suggestion
    EUS_MINOR_VERSIONS = [8.2, 8.4 ,8.6, 8.8, 8.10]  # this can be moved to systeminfo.py

    os_version = ".".join(map(str, system_info.version))
    if os_version not in EUS_MINOR_VERSIONS:
```

convert2rhel/repo.py
```diff
@@ -68,3 +74,35 @@ def restore_yum_repos():
...
+
+    :return: The return can be either the path to the eus repos, or None, meaning it is not a EUS system.
+    :rtype: str | None
+    """
+    hardcoded_reposdir = _get_hardcoded_repofiles_dir()
+    reposdir = None
+    if os.path.exists(hardcoded_reposdir) and system_info.has_internet_access:
+        reposdir = hardcoded_reposdir
+
+    return reposdir
```
bocekm, line 90:
```suggestion
```
I totally don't understand the purpose of this function. Decision whether the RHEL minor version corresponding to the to-be-converted-OS minor version shouldn't rely on what directories we store/hardcode. Because we also hardcode CentOS Linux 8.5 repositories, but corresponding RHEL 8.5 is not an EUS release,.

bocekm (changes requested):

convert2rhel/backup.py
```diff
@@ -0,0 +1,180 @@
...
+class RestorablePackage(object):
+    def __init__(self, pkgname):
+        self.name = pkgname
+        self.path = None
+
+    def backup(self):
+        """Save version of RPM package"""
+        loggerinst.info("Backing up %s" % self.name)
+        if os.path.isdir(BACKUP_DIR):
+            reposdir = get_eus_repos_available()
```
bocekm, line 134:
```suggestion
            reposdir = get_hardcoded_repofiles_dir()
```
This is a bit tricky because we want to use the hardcoded repos for the package backup purposes only in case of non-latest RHEL release. Because imagine you have an air-gapped system with set up to receive updates from an internal Satellite server. With this change the package backup would fail. Without it it wouldn't.

Right of the bat I can't think of a good solution. We need to do some brainstorming here.

convert2rhel/pkghandler.py
```diff
@@ -1011,14 +1020,16 @@ def get_total_packages_to_update():
...
     """
     packages = []
 
+    reposdir = get_eus_repos_available()
+
     if pkgmanager.TYPE == "yum":
         packages = _get_packages_to_update_yum()
     elif pkgmanager.TYPE == "dnf":
-        packages = _get_packages_to_update_dnf()
+        packages = _get_packages_to_update_dnf(reposdir=reposdir)
```
bocekm, line 1032:
```suggestion
        # We're using the reposdir with dnf only because we currently hardcode the repofiles for RHEL 8 derivatives only.
        packages = _get_packages_to_update_dnf(reposdir=reposdir)
```

bocekm (changes requested):

convert2rhel/data/8/x86_64/repos/oracle-8.4/virt-ol8.repo
r0x0d: Done.

convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
```diff
@@ -31,6 +31,10 @@ repofile_pkgs =
 # Delimited by any whitespace(s).
 default_rhsm_repoids = rhel-7-server-rpms
 
+# List of eus repoids to enable through subscription-manager when the --enablerepo option is not used.
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/systeminfo.py
```diff
@@ -90,6 +91,8 @@ def __init__(self):
         self.logger = None
         # IDs of the default Red Hat CDN repositories that correspond to the current system
         self.default_rhsm_repoids = None
+        # IDS of the eus Red Hat CDN repositories that correspond to the current system
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/pkghandler.py
```diff
@@ -1011,14 +1020,16 @@ def get_total_packages_to_update():
     if it's Oracle/CentOS Linux 7 or 6, it will use `yum`, otherwise it should use `dnf`.
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/repo.py
```diff
@@ -68,3 +74,35 @@ def restore_yum_repos():
...
+
+
+def _get_hardcoded_repofiles_dir():
+    """Get the path to the hardcoded repofiles for CentOS/Oracle Linux.
```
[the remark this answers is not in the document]
r0x0d: done with some modifications.

convert2rhel/subscription.py
```diff
@@ -441,6 +455,27 @@ def enable_repos(rhel_repoids):
     else:
         repos_to_enable = rhel_repoids
 
+    # Check if the rhel_repoids is the eus ones
+    if repos_to_enable == system_info.eus_rhsm_repoids:
+        try:
+            # Try a first time and see if it's possible to enable the EUS repositories
+            # Otherwise, if it raiess an exception, try to enable the default rhsm-repos
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/subscription.py
```diff
@@ -441,6 +455,27 @@ def enable_repos(rhel_repoids):
...
+            _submgr_enable_repos(repos_to_enable)
+        except SystemExit:
+            loggerinst.info("Falling back to using default rhsm repositories.")
```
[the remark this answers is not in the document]
r0x0d: done with some modifications.

convert2rhel/pkghandler.py
r0x0d: done.

convert2rhel/backup.py
```diff
@@ -0,0 +1,201 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright(C) 2016 Red Hat, Inc.
```
bocekm, line 3:
```suggestion
# Copyright(C) 2022 Red Hat, Inc.
```

convert2rhel/systeminfo.py
```diff
@@ -306,6 +319,52 @@ def get_enabled_rhel_repos(self):
         #     )
         return self.submgr_enabled_repos if not tool_opts.no_rhsm else tool_opts.enablerepo
 
+    def _check_internet_access(self, host="8.8.8.8", port=53, timeout=3):
+        """Check wether or not the machine is connect to the internet.
```
bocekm, line 323:
```suggestion
        """Check whether or not the machine is connected to the internet.
```

convert2rhel/pkghandler.py
```diff
@@ -1004,21 +1013,28 @@ def compare_package_versions(version1, version2):
     return rpm.labelCompare(evr1, evr2)
 
 
-def get_total_packages_to_update():
+def get_total_packages_to_update(reposdir):
     """Return the total number of packages to update in the system
 
     It uses both yum/dnf depending on weather they are installed on the system,
```
bocekm, line 1019:
```suggestion
    It uses both yum/dnf depending on whether they are installed on the system,
```

convert2rhel/checks.py
```diff
@@ -466,7 +467,13 @@ def check_package_updates():
     """Ensure that the system packages installed are up-to-date."""
     logger.task("Prepare: Checking if the packages are up-to-date")
 
-    packages_to_update = get_total_packages_to_update()
+    reposdir = get_hardcoded_repofiles_dir()
+
+    if (reposdir is None) and (not system_info.has_internet_access):
```
bocekm, line 472: This has a different meaning than suggested in https://github.com/oamg/convert2rhel/pull/458#discussion_r866281472.

The way you have it the check would be skipped only on offline systems we don't have hardcoded repofiles for, e.g. CentOS Linux 7, Alma Linux 8, etc.

However the purpose of the condition should be to skip the check in case we **have** the hardcoded repofiles for and the system is offline. The reason is that our hardcoded repofiles require an internet connection.

convert2rhel/backup.py
```diff
@@ -0,0 +1,201 @@
...
+
+            if reposdir and system_info.corresponds_to_rhel_eus_release():
+                # When backing up the packages, the original system repofiles are still available and for them we can't
+                # use the releasever for RHEL repositories
+                self.path = download_pkg(
+                    self.name,
+                    dest=BACKUP_DIR,
+                    set_releasever=False,
+                    reposdir=reposdir,
+                )
```
bocekm, line 162:
```suggestion
            if reposdir and system_info.corresponds_to_rhel_eus_release():
                # One of the reasons we hardcode repofiles pointing to archived repositories of older system minor versions is that we need to be able to download an older package version as a backup. Because for example the default repofiles on CentOS Linux 8.4 point only to 8.latest repositories that already don't contain 8.4 packages.
                # The reason for the EUS condition is that convert2rhel allows conversions of just those older system minor versions that correspond to a RHEL EUS release.
                self.path = download_pkg(
                    self.name,
                    dest=BACKUP_DIR,
                    # When backing up the packages, the original system repofiles are still available and for them we can't
                    # use the releasever for RHEL repositories
                    set_releasever=False,
                    reposdir=reposdir,
                )
```

bocekm (changes requested):

convert2rhel/checks.py
r0x0d: Oh, my bad. I must have included the `reposdir is None` by mistake while trying to do something different. 

I will update it on Monday.

convert2rhel/pkghandler.py
```diff
@@ -24,6 +24,8 @@
 import rpm
 
 from convert2rhel import pkgmanager, utils
+from convert2rhel.backup import RestorableFile, remove_pkgs
+from convert2rhel.repo import get_hardcoded_repofiles_dir
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/subscription.py
```diff
@@ -22,7 +22,7 @@
 from collections import namedtuple
 from time import sleep
 
-from convert2rhel import pkghandler, utils
+from convert2rhel import backup, pkghandler, systeminfo, utils
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/backup.py
r0x0d: done.

convert2rhel/systeminfo.py
r0x0d: done.

convert2rhel/pkghandler.py
r0x0d: done.

convert2rhel/backup.py
r0x0d: done.

convert2rhel/checks.py
r0x0d: done.

tests/integration/tier1/convert-offline-systems/run_conversion.py
```diff
@@ -10,7 +10,7 @@ def test_convert_offline_systems(shell, convert2rhel):
     # The CentOS8 Extras repo url is unreachable due to offline system setup.
     # The repoquery returns an error, thus we need to disable this repository.
     if "centos-8" in platform.platform():
-        shell("yum-config-manager --disable extras --disable epel-modular")
+        shell("yum-config-manager --disable extras --disable epel-modular --disable appstream")
```
bocekm, line 13: @r0x0d, @kokesak, @danmyway, why is appstream being disabled and not some other repos like baseos?

bocekm (approved):

tests/integration/tier1/convert-offline-systems/run_conversion.py
kokesak: @bocekm this was just me playing with the test to see what happens. 
The issue I'm hitting here is this:
```
20:29:42             out: [05/12/2022 16:29:42] TASK - [Prepare: Checking if --enablerepo repositories are accessible] ****
20:29:42             out: Skipping the check of repositories due to the use of RHSM for the conversion.
20:29:42             out: 
20:29:42             out: [05/12/2022 16:29:42] TASK - [Prepare: Checking if the packages are up-to-date] *****************
20:29:42             out: Errors during downloading metadata for repository 'appstream':
20:29:42             out:   - Curl error (7): Couldn't connect to server for https://vault.centos.org/centos/8.5.2111/AppStream/x86_64/os/repodata/repomd.xml [Failed to connect to vault.centos.org port 443: Connection refused]
20:29:42             out: [05/12/2022 16:29:42] DEBUG - Traceback (most recent call last):
20:29:42             out:   File "/usr/lib/python3.6/site-packages/dnf/repo.py", line 574, in load
20:29:42             out:     ret = self._repo.load()
20:29:42             out:   File "/usr/lib64/python3.6/site-packages/libdnf/repo.py", line 397, in load
20:29:42             out:     return _repo.Repo_load(self)
20:29:42             out: libdnf._error.Error: Failed to download metadata for repo 'appstream': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
20:29:42             out: 
20:29:42             out: During handling of the above exception, another exception occurred:
20:29:42             out: 
20:29:42             out: Traceback (most recent call last):
20:29:42             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 91, in main
20:29:42             out:     checks.perform_pre_checks()
20:29:42             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/checks.py", line 61, in perform_pre_checks
20:29:42             out:     check_package_updates()
20:29:42             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/checks.py", line 476, in check_package_updates
20:29:42             out:     packages_to_update = get_total_packages_to_update(reposdir=reposdir)
20:29:42             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/pkghandler.py", line 1036, in get_total_packages_to_update
20:29:42             out:     packages = _get_packages_to_update_dnf(reposdir=reposdir)
20:29:42             out:   File "/usr/lib/python3.6/site-packages/convert2rhel/pkghandler.py", line 1091, in _get_packages_to_update_dnf
20:29:42             out:     base.fill_sack()
20:29:42             out:   File "/usr/lib/python3.6/site-packages/dnf/base.py", line 399, in fill_sack
20:29:42             out:     self._add_repo_to_sack(r)
20:29:42             out:   File "/usr/lib/python3.6/site-packages/dnf/base.py", line 139, in _add_repo_to_sack
20:29:42             out:     repo.load()
20:29:42             out:   File "/usr/lib/python3.6/site-packages/dnf/repo.py", line 581, in load
20:29:42             out:     raise dnf.exceptions.RepoError(str(e))
20:29:42             out: dnf.exceptions.RepoError: Failed to download metadata for repo 'appstream': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

This happens on C8.5. On C8.4 it has problem with "extras" repo. The system is "offline" so it cannot reach those repos. However it's strange that even though I disabled those repos it still tries to download from them

kokesak: full log here:
C8.5: http://artifacts.osci.redhat.com/testing-farm/801e69bd-b1af-440a-9328-644ba6d4024e/work-convert_offline_systems2BBlv3/log.txt
C8.4: http://artifacts.osci.redhat.com/testing-farm/0ef7b1f9-f411-4856-a320-7936105b8d24/work-convert_offline_systemsRcB199/log.txt

bocekm:
```
Errors during downloading metadata for repository 'appstream':
 - Curl error (7): Couldn't connect to server for https://vault.centos.org/centos/8.5.2111/AppStream/x86_64/os/repodata/repomd.xml [Failed to connect to vault.centos.org port 443: Connection refused]
```
The URL is accessible now so I bet it was some kind of intermittent vault outage.

kokesak: It is accessible just not on the testing machine. This is the "offline conversion" test. The machine has access only to our internal satellite server. This problem is only on CentoOS 8.5/8.4 as far as I looked at the tests, should the repositories be available on the satellite or how should I approach this? Remove the unaccessible repos from the system before the conversion? This seems a bit hackisch to me :D

bocekm: Ah, right, this is the offline test. Then I think disabling all the repositories that require the internet connection is perfectly fine. Or removing the repofiles that define such repositories. That's what expect customer would do. Why would they keep repositories they don't have access to enabled..

bocekm (changes requested):

convert2rhel/systeminfo.py
```diff
@@ -105,12 +108,14 @@ def resolve_system_info(self):
         self.id = self.name.split()[0].lower()
         self.version = self._get_system_version()
         self.arch = self._get_architecture()
+        self.has_internet_access = self._check_internet_access()
```
bocekm, line 111: Please move the function call down, e.g. under `self.booted_kernel = self._get_booted_kernel()` to not break the output this way:
```
[05/17/2022 17:28:38] TASK - [Prepare: Gather system information] *******************************
Name:                Oracle Linux Server
OS version:          8.4
Architecture:        x86_64
Checking internet connectivity using host '8.8.8.8' and port '53'.
Config filename:     oracle-8-x86_64.cfg
```

convert2rhel/checks.py
```diff
@@ -495,15 +506,33 @@ def is_loaded_kernel_latest():
     """Check if the loaded kernel is behind or of the same version as in yum repos."""
     logger.task("Prepare: Checking if the loaded kernel version is the most recent")
 
+    reposdir = get_hardcoded_repofiles_dir()
```
bocekm, line 509:
```suggestion
    LATEST_RHEL_8_MINOR_VERSION = 6
    if system_info.id == "oracle" and system_info.version.major == 8 and system_info.version.minor < LATEST_RHEL_8_MINOR_VERSION:
        logger.info("Skipping the check because there are no publicly available %s %d.%d repositories available."
        % (system_info.name, system_info.version.major, system_info.version.minor))
        return
    reposdir = get_hardcoded_repofiles_dir()
```
@r0x0d, @abadger, we may improve hardcoding the latest minor version by employing scheduled RHEL release dates. But we might do that in a separate PR.

Or shall we perhaps simply skip this check on OL 8, no matter the minor version, to save ourselves from the hassle with hardcoding the minor versions?

bocekm (changes requested):

convert2rhel/systeminfo.py
r0x0d: done.

convert2rhel/backup.py
```diff
@@ -0,0 +1,205 @@
...
+        loggerinst.info("Backing up %s" % self.name)
+        if os.path.isdir(BACKUP_DIR):
+            reposdir = get_hardcoded_repofiles_dir()
+
+            # One of the reasons we hardcode repofiles pointing to archived repositories of older system
+            # minor versions is that we need to be able to download an older package version as a backup.
+            # Because for example the default repofiles on CentOS Linux 8.4 point only to 8.latest repositories
+            # that already don't contain 8.4 packages.
+            # The reason for the EUS condition is that convert2rhel allows conversions of just those older system
+            # minor versions that correspond to a RHEL EUS release.
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/backup.py
```diff
@@ -0,0 +1,205 @@
...
+            if reposdir and system_info.corresponds_to_rhel_eus_release():
+                self.path = download_pkg(
+                    self.name,
+                    dest=BACKUP_DIR,
+                    set_releasever=False,
+                    reposdir=reposdir,
+                )
+            else:
+                self.path = download_pkg(self.name, dest=BACKUP_DIR, set_releasever=False)
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/systeminfo.py
```diff
@@ -305,6 +318,52 @@ def get_enabled_rhel_repos(self):
...
+        """Return whether the current minor version corresponds to a RHEL Extended Update Support (EUS) release.
+
+        For example if we detect CentOS Linux 8.4, the corresponding RHEL 8.4 is an EUS release, but if we detect
+        CentOS Linux 8.5, the corresponding RHEL 8.5 is not an EUS release.
+
+        :return: Whether or not the current system has a EUS correspondent in RHEL.
+        :rtype: bool
+        """
+        system_version = "%s.%s" % (self.version.major, self.version.minor)
+        EUS_MINOR_VERSIONS = ["8.4", "8.6", "8.8", "8.10"]
```
bocekm, line 364:
```suggestion
        EUS_MINOR_VERSIONS = ["8.4", "8.6", "8.8"]
```
I've now realized that the last minor version won't be an EUS minor version: https://access.redhat.com/articles/rhel-eus#c5.

bocekm (changes requested):

convert2rhel/systeminfo.py
r0x0d: done.

r0x0d:
@bocekm I have updated the message that we discussed earlier today, but I don't remember if the phrasing is the correct one. Let me know if it's right, otherwise I can change it right awya.

convert2rhel/checks.py
```diff
@@ -532,13 +576,14 @@ def is_loaded_kernel_latest():
         if match == 0:
             logger.info("Kernel currently loaded is at the latest version.")
         else:
+            repos_message = "in your system repos" if not reposdir else "in repositories defined at: %s " % reposdir
             logger.critical(
-                "The current kernel version loaded is different from the latest version in your repos.\n"
+                "The current kernel version loaded is different from the latest version %s.\n"
                 " Latest kernel version: %s\n"
                 " Current loaded kernel: %s\n\n"
```
bocekm, line 583:
```suggestion
                "The version of the loaded kernel is different from the latest version %s.\n"
                " Latest kernel version available in %s: %s\n"
                " Loaded kernel version: %s\n\n"
```

bocekm (changes requested):

convert2rhel/checks.py
```diff
@@ -532,13 +576,14 @@ def is_loaded_kernel_latest():
         if match == 0:
             logger.info("Kernel currently loaded is at the latest version.")
         else:
+            repos_message = "in your system repos" if not reposdir else "in repositories defined at: %s " % reposdir
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/checks.py
r0x0d: done.

r0x0d: done.

convert2rhel/checks.py
```diff
@@ -532,13 +576,14 @@ def is_loaded_kernel_latest():
...
             logger.critical(
-                "The current kernel version loaded is different from the latest version in your repos.\n"
+                "The current kernel version loaded is different from the latest version %s.\n"
                 " Latest kernel version: %s\n"
                 " Current loaded kernel: %s\n\n"
                 "To proceed with the conversion, update the kernel version by executing the following step:\n\n"
                 "1. yum install %s-%s -y\n"
-                "2. reboot" % (latest_kernel, loaded_kernel, package_to_check, latest_kernel)
+                "2. reboot" % (repos_message, latest_kernel, loaded_kernel, package_to_check, latest_kernel)
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/checks.py
```diff
@@ -495,15 +514,40 @@ def is_loaded_kernel_latest():
...
+        )
+        return
+
+    reposdir = get_hardcoded_repofiles_dir()
+
+    if reposdir and not system_info.has_internet_access:
+        logger.warning("Skipping the check as no internet connection has been detected.")
+        return
+
+    cmd = ["repoquery", "--quiet", "--qf", '"%{BUILDTIME}\\t%{VERSION}-%{RELEASE}"']
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/subscription.py
```diff
@@ -542,3 +584,34 @@ def exit_on_failed_download(paths):
...
+
+        output, ret_code = utils.run_subprocess(cmd, print_output=False)
+        if ret_code != 0:
+            loggerinst.warning(
+                "Locking RHEL repositories failed with return code %d and message:\n%s",
+                ret_code,
+                output,
+            )
+        else:
+            loggerinst.info("RHEL repositories locked on %s" % system_info.releasever)
```
bocekm, line 615:
```suggestion
            loggerinst.info("RHEL repositories locked to the %s minor version." % system_info.releasever)
```

convert2rhel/subscription.py
```diff
@@ -542,3 +584,34 @@ def exit_on_failed_download(paths):
...
+    .. note::
+        This function should only run if the detected system is an EUS correspondent, otherwise, there is no need
+        to lock the releasever in the redhat.repo file as the repository structure when using EUS is slightly
+        different from a non-EUS repository.
+    """
+
+    # We only lock the releasever on rhel repos if we detect that the running system is an EUS correspondent and if te
+    # rhsm is used, otherwise, there's no need to lock the releasever as the subscription-manager won't be available.
+    if system_info.corresponds_to_rhel_eus_release() and not tool_opts.no_rhsm:
+        loggerinst.info("Locking the releasever to %s in RHEL repositories." % system_info.releasever)
```
bocekm, line 604:
```suggestion
        loggerinst.info("Updating /etc/yum.repos.d/rehat.repo to point to RHEL %s instead of the default latest minor version." % system_info.releasever)
```
People might not necessarily know what releasever is so it's better to explain in layman terms what we're doing.

convert2rhel/subscription.py
```diff
@@ -542,3 +584,34 @@ def exit_on_failed_download(paths):
...
+        if ret_code != 0:
+            loggerinst.warning(
+                "Locking RHEL repositories failed with return code %d and message:\n%s",
+                ret_code,
+                output,
+            )
+        else:
+            loggerinst.info("RHEL repositories locked on %s" % system_info.releasever)
+    else:
+        loggerinst.info("Not an EUS system, skipping the RHEL lock releasever.")
```
bocekm, line 617:
```suggestion
        loggerinst.info("Skipping locking RHEL repositories to a specific EUS minor version.")
```
The original message wouldn't be correct when the system is converted to RHEL 8.6 which is an EUS minor version.

bocekm (changes requested):

convert2rhel/checks.py
```diff
@@ -479,12 +497,13 @@ def check_package_updates():
         return
 
     if len(packages_to_update) > 0:
+        repos_message = "in your system repos" if not reposdir else "in repositories defined at: %s " % reposdir
```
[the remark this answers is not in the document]
r0x0d: done.

tests/ansible_collections/roles/update-system/tasks/main.yml
```diff
@@ -1,2 +1,5 @@
 ---
+# We do not update Oracle Linux 8.4
 - include: update-system.yml
+  when:
+    - ansible_facts['distribution_version'] != "8.4" or ansible_facts['distribution'] != "OracleLinux"
```
mkluson, line 5: !(ansible_facts['distribution_version'] == "8.4" and ansible_facts['distribution'] == "OracleLinux") seems more readable

tests/integration/tier0/backup-release/test_backup_release.py
```diff
@@ -22,7 +25,13 @@ def test_backup_os_release_no_envar(shell, convert2rhel):
     assert shell("rpm -i {}".format(pkg_dst)).returncode == 0
 
     # Move all repos to other location, so it is not being used
-    assert shell("mkdir /tmp/s_backup && mv /etc/yum.repos.d/* /tmp/s_backup/").returncode == 0
+    assert shell("mkdir /tmp/s_backup")
```
mkluson, line 28: check for return code?

plans/tier1.fmf
```diff
@@ -191,14 +191,21 @@ discover+:
...
     playbook: tests/ansible_collections/roles/reboot/main.yml
 
 /system_up_to_date:
+  # On Oracle Linux 8 we do not have minor version repositories so we skip the check for up-to-date system
+  adjust:
+    enabled: false
+    when: >
+      distro != centos-8 and
+      distro != centos-7 and
+      distro != oraclelinux-7
```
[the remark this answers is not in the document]
bocekm: @mkluson, actually the adjustment @kokesak means skipping the test on every distro **except** centos-7/centos-8/oraclelinux-7.

However, the comment `# On Oracle Linux 8 we do not have minor version repositories so we skip the check for up-to-date system` is not true anymore. We run the system-up-to-date check on every system except Oracle Linux 8.4. Meaning we run the check on OL 8.6 - for that we use the default repos enabled on the system pointing to OL 8.latest.

That means the adjustment should look like:
```
/system_up_to_date:
  # On Oracle Linux 8.4 we do not have minor version repositories so we skip the check for up-to-date system
  adjust:
    enabled: false
    when: >
      distro == oraclelinux-8.4
```

tests/ansible_collections/roles/add-custom-repos/main.yml
```diff
@@ -5,4 +5,6 @@
 - include: rhel7-repos.yml
   when: ansible_facts['distribution_major_version'] == "7"
 - include: rhel8-repos.yml
-  when: ansible_facts['distribution_major_version'] == "8"
+  when: ansible_facts['distribution_version'] == "8.5" or ansible_facts['distribution_version'] == "8.6"
```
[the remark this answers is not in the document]
bocekm: When Oracle Linux 8.7 is released, we'll need to have in this condition only 8.5 and 8.7 (not 8.6 anymore). OL 8.7 will be supported (because it's going to be the latest minor version), and 8.6 will be supported as well however that will need RHEL 8.6 EUS custom repos. That means 8.6 will move two lines below.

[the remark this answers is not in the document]
bocekm: You can add a comment describing what's to be done half a year later.
```suggestion
  # When Oracle Linux 8.7 is released, the "8.6" needs to change to "8.7" and the
  # "8.6" is to be moved to the condition below related to enabling RHEL 8 EUS repos
  when: ansible_facts['distribution_version'] == "8.5" or ansible_facts['distribution_version'] == "8.6"
```

tests/ansible_collections/roles/update-system/tasks/main.yml
bocekm: @kokesak, help me understand how does the system update using this ansible role work on CentOS Linux 8.4. Is the distro image tweaked before this role is executed so that its repos point to CentOS Linux 8.4 Vault meaning that the system gets updated to the latest 8.4 packages? If so, please update the comment three lines above to:
```
# We can't update Oracle Linux 8.4 to the latest 8.4 packages because there're no public Oracle Linux 8.4 repositories available.
# The CentOS Linux 8.4 can be updated using this role because the repositories on the system image has been tweaked to point to the CentOS Linux 8.4 repositories in the Vault.
```

plans/tier1.fmf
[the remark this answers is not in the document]
kokesak: @bocekm thanks for the clarification, I'll update the test so it run on oracle8.6 as well.

tests/ansible_collections/roles/update-system/tasks/main.yml
kokesak: @bocekm yes the distro is tweaked that only baseos, appstream and extras repos are enabled and poiting to the 8.4 url
(https://github.com/oamg/convert2rhel/pull/458/files#diff-34df60c049b010fc09dfd65d350825b2eafce0b02bd8fa487dddb730d0185193)

I'll update the comment accordingly

tests/ansible_collections/roles/add-custom-repos/main.yml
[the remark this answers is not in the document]
kokesak: Alright I'll update the comment

tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
```diff
@@ -38,7 +39,11 @@ def test_non_latest_kernel(shell, convert2rhel):
...
     ) as c2r:
-        if "centos-7" in platform.platform() or "oracle-7" in platform.platform():
+        if "centos-7" in system_version or "oracle-7" in system_version:
+            c2r.expect(
+                "The version of the loaded kernel is different from the latest version on the enabled system repositories."
+            )
+        elif "oracle-8" in system_version:
             c2r.expect(
                 "The version of the loaded kernel is different from the latest version on the enabled system repositories."
             )
```
bocekm, line 49: Can't these two conditions be merged into one?

convert2rhel/checks.py
```diff
@@ -532,13 +579,18 @@ def is_loaded_kernel_latest():
         if match == 0:
             logger.info("Kernel currently loaded is at the latest version.")
         else:
+            repos_message = (
+                "on the enabled system repositories"
+                if not reposdir
+                else "on repositories defined in the %s folder" % reposdir
```
bocekm, line 585:
```suggestion
                "in the enabled system repositories"
                if not reposdir
                else "in repositories defined in the %s folder" % reposdir
```
Sorry if I made it confusing in the comment https://github.com/oamg/convert2rhel/pull/458#discussion_r883345313 but the `in` preposition was correct here (there isn't `based` before it here).
tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py

bocekm (changes requested):

bocekm (approved):

PR #460, opened 2022-03-08
packit: build SRPM in Copr

TomasTomecek:
and be able to specify the precise list of deps needed to create the
SRPM and since you don't need any extra, we "have to" enforce
this process by defining the `srpm_build_deps` key.


SpyTec (approved):
Thanks Tomas

PR #444, opened 2022-02-22
Remove unused local integration testing

bocekm:
We don't run integration tests locally anymore - we run them just in the Testing Farm nowadays.

PR #464, opened 2022-03-15
Update Makefile

r0x0d:
Since we removed the `local.centos8.requirements.txt` file in #444, the install command in the `Makefile` was wrong and needed to be patched.


abadger (approved):
Simple and straightforward change

PR #465, opened 2022-03-18
Add CentOS 7 to coverage

r0x0d:
Enabling the CentOS 7 docker build to run and send coverage do codecov.io

Jira reference (If any): https://issues.redhat.com/browse/OAMG-6611
Bugzilla reference (If any): 

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -11,7 +11,7 @@ jobs:
     name: coverage-centos${{ matrix.centos_ver }}
     strategy:
       matrix:
-        centos_ver: [8]
+        centos_ver: [7, 8]
 
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


SpyTec:
@r0x0d could you test to see if either overwrites the other report? When I look at docs it seems to imply it merges files when you upload them together but I can't see anything if it's separate

To test that it might be possible to create a follow-up PR with some changes to tests so that some have `@centos7` and some have `@centos8` decorator

r0x0d:
> @r0x0d could you test to see if either overwrites the other report? When I look at docs it seems to imply it merges files when you upload them together but I can't see anything if it's separate
>
> To test that it might be possible to create a follow-up PR with some changes to tests so that some have @centos7 and some have @centos8 decorator

We have a test right now that it's meant to be running only on Centos7 because it mocks the yum API. I believe we could check whether it overrides or not with that case... 

This is the test I mentioned. https://github.com/oamg/convert2rhel/pull/356/files#diff-c3ce6b0e1e7ddae0678f9eb7c71f9c5f8aecbfd5b7db1deb7d8947ccf1492a08R1448

PR #467, opened 2022-03-21
Fix for unavailable repos on CentOS8

kubealex:
This fix adds a check if on CentOS8 and performs the needed steps to switch repos to vault.centos.org.


bocekm (approved):

bocekm:
Thank you, @kubelax!

PR #468, opened 2022-03-21
Update the Secure Boot message

r0x0d:
Update the secure boot message to point out the user on how to disable secure
boot as it is not currently available in convert2rhel.

Jira reference (If any): https://issues.redhat.com/browse/OAMG-6665
Bugzilla reference (If any):


abadger (approved):
The updated message looks good.  It implements the message asked for in the jira ticket.

PR #469, opened 2022-03-23
[RHELC-297] Clean up kmods functionality

r0x0d:
These changes are cleanup of most of the things that were left behind from
PR #193.

Some of the functions were changed to make it easier to read and debug if
necessary, some of them were moved around to the files that made sense to place
they and some of them just needed a docstring to tell what the function was all
about.

This PR it is introduced a unit test for the function `get_enabled_rhel_repos`
since that was introduced in the past and never had a unit test tied to it.

Jira reference (If any): https://issues.redhat.com/browse/OAMG-4668
Bugzilla reference (If any):

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-
+import functools
 import itertools
 import logging
 import os
@@ -28,6 +28,7 @@ import rpm
 from convert2rhel import __version__ as convert2rhel_version
 from convert2rhel import grub, pkgmanager, utils
 from convert2rhel.pkghandler import (
+    _package_version_cmp,
     call_yum_cmd,
     compare_package_versions,
     get_installed_pkg_objects,
@@ -37,13 +38,7 @@ from convert2rhel.pkghandler import (
 from convert2rhel.repo import get_hardcoded_repofiles_dir
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
-from convert2rhel.utils import (
-    ask_to_continue,
-    get_file_content,
-    run_subprocess,
-    store_content_to_file,
-    convert_to_int_or_zero,
-)
+from convert2rhel.utils import ask_to_continue, get_file_content, run_subprocess, store_content_to_file
 
 
 logger = logging.getLogger(__name__)
@@ -284,8 +279,7 @@ def check_custom_repos_are_valid():
 def ensure_compatibility_of_kmods():
     """Ensure that the host kernel modules are compatible with RHEL.
 
-    :raises:
-        SystemExit: Interrupts the conversion because some kernel modules are not supported in RHEL.
+    :raises SystemExit: Interrupts the conversion because some kernel modules are not supported in RHEL.
     """
     host_kmods = get_loaded_kmods()
     rhel_supported_kmods = get_rhel_supported_kmods()
@@ -306,7 +300,7 @@ def ensure_compatibility_of_kmods():
             (
                 "The following kernel modules are not supported in RHEL:\n{kmods}\n"
                 "Make sure you have updated the kernel to the latest available version and rebooted the system.\n"
-                "Remove the unsupported modules and run convert2rhel again to continue with the conversion."
+                "Prevent the unsupported modules from loading by following https://access.redhat.com/solutions/41278 and run convert2rhel again to continue with the conversion."
             ).format(kmods=not_supported_kmods, system=system_info.name)
         )
 
@@ -438,49 +432,14 @@ def get_most_recent_unique_kernel_pkgs(pkgs):
     list_of_sorted_pkgs = []
     for distinct_kernel_pkgs in pkgs_groups:
         if distinct_kernel_pkgs[0].startswith(("kernel", "kmod")):
-            list_of_sorted_pkgs.append(max(distinct_kernel_pkgs[1], key=_repos_version_key))
+            list_of_sorted_pkgs.append(
+                max(
+                    distinct_kernel_pkgs[1],
+                    key=functools.cmp_to_key(_package_version_cmp),
+                )
+            )
 
     return tuple(list_of_sorted_pkgs)
-
-
-def _repos_version_key(pkg_name):
-    """Identify the version key in a given package name.
-
-    Consider the following pkg_name that will be passed to this function::
-
-        pkg_name = 'kernel-core-0:4.18.0-240.10.1.el8_3.x86_64'
-
-    The output of this will be a tuple containing the package version in a tuple::
-
-        result = _repos_version_key(pkg_name=pkg_name)
-        print(result)
-        # (4, 15, 0, 240, 10, 1)
-
-    The function will ignore the package name as it is not an important information here
-    and will only care about the version that is tied to it's name.
-
-    :param pkg_name: The package to extract the version
-    :type pkg_name: str
-    :return: A tuple containing the package version.
-    :rtype: tuple[int]
-    :raises:
-        SystemExit: Raises SytemExit if it can't find the version in the pkg_name.
-    """
-    # TODO(r0x0d): This should be moved to repo.py or utils.py?
-    try:
-        rpm_version = KERNEL_REPO_RE.search(pkg_name).group("version")
-    except AttributeError:
-        logger.critical(
-            "Unexpected package: %s\n Couldn't find the version of the given package.",
-            pkg_name,
-        )
-    else:
-        return tuple(
-            map(
-                convert_to_int_or_zero,
-                KERNEL_REPO_VER_SPLIT_RE.split(rpm_version),
-            )
-        )
 
 
 def get_rhel_kmods_keys(rhel_kmods_str):
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -15,7 +15,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-import functools
 import itertools
 import logging
 import os
@@ -55,6 +54,44 @@ COMPATIBLE_KERNELS_VERS = {
     7: "3.10.0",
     8: "4.18.0",
 }
+
+# Python 2.6 compatibility.
+# This code is copied from Pthon-3.10's functools module,
+# licensed under the Python Software Foundation License, version 2
+try:
+    from functools import cmp_to_key
+except ImportError:
+
+    def cmp_to_key(mycmp):
+        """Convert a cmp= function into a key= function"""
+
+        class K(object):
+            __slots__ = ["obj"]
+
+            def __init__(self, obj):
+                self.obj = obj
+
+            def __lt__(self, other):
+                return mycmp(self.obj, other.obj) < 0
+
+            def __gt__(self, other):
+                return mycmp(self.obj, other.obj) > 0
+
+            def __eq__(self, other):
+                return mycmp(self.obj, other.obj) == 0
+
+            def __le__(self, other):
+                return mycmp(self.obj, other.obj) <= 0
+
+            def __ge__(self, other):
+                return mycmp(self.obj, other.obj) >= 0
+
+            __hash__ = None
+
+        return K
+
+
+# End of PSF Licensed code
 
 
 def perform_system_checks():
@@ -438,7 +475,7 @@ def get_most_recent_unique_kernel_pkgs(pkgs):
             list_of_sorted_pkgs.append(
                 max(
                     distinct_kernel_pkgs[1],
-                    key=functools.cmp_to_key(_package_version_cmp),
+                    key=cmp_to_key(_package_version_cmp),
                 )
             )
 
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -338,7 +338,8 @@ def ensure_compatibility_of_kmods():
         )
         logger.critical(
             "The following loaded kernel modules are not available in RHEL:\n{0}\n"
-            "Prevent the modules from loading by following {1}"
+            "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
+            "If this message appears again after doing the above, prevent the modules from loading by following {1}"
             " and run convert2rhel again to continue with the conversion.".format(
                 "\n".join(not_supported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
             )
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -37,7 +37,13 @@ from convert2rhel.pkghandler import (
 from convert2rhel.repo import get_hardcoded_repofiles_dir
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
-from convert2rhel.utils import ask_to_continue, get_file_content, run_subprocess, store_content_to_file
+from convert2rhel.utils import (
+    ask_to_continue,
+    get_file_content,
+    run_subprocess,
+    store_content_to_file,
+    convert_to_int_or_zero,
+)
 
 
 logger = logging.getLogger(__name__)
@@ -55,7 +61,7 @@ COMPATIBLE_KERNELS_VERS = {
 }
 
 
-def perform_pre_checks():
+def perform_system_checks():
     """Early checks after system facts should be added here."""
 
     check_custom_repos_are_valid()
@@ -202,7 +208,7 @@ def check_tainted_kmods():
         multipath 20480 0 - Live 0x0000000000000000
         linear 20480 0 - Live 0x0000000000000000
         system76_io 16384 0 - Live 0x0000000000000000 (OE)  <<<<<< Tainted
-        system76_acpi 16384 0 - Live 0x0000000000000000 (OE) <<<<< Tainted
+        system76_acpi 16384 0 - Live 0x0000000000000000 (OE) <<<<<< Tainted
     """
     logger.task("Prepare: Checking if loaded kernel modules are not tainted")
     unsigned_modules, _ = run_subprocess(["grep", "(", "/proc/modules"])
@@ -261,7 +267,9 @@ def check_custom_repos_are_valid():
         return
 
     output, ret_code = call_yum_cmd(
-        command="makecache", args=["-v", "--setopt=*.skip_if_unavailable=False"], print_output=False
+        command="makecache",
+        args=["-v", "--setopt=*.skip_if_unavailable=False"],
+        print_output=False,
     )
     if ret_code != 0:
         logger.critical(
@@ -274,11 +282,20 @@ def check_custom_repos_are_valid():
 
 
 def ensure_compatibility_of_kmods():
-    """Ensure if the host kernel modules are compatible with RHEL."""
+    """Ensure that the host kernel modules are compatible with RHEL.
+
+    :raises:
+        SystemExit: Interrupts the conversion because some kernel modules are not supported in RHEL.
+    """
     host_kmods = get_loaded_kmods()
     rhel_supported_kmods = get_rhel_supported_kmods()
     unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods)
-    if unsupported_kmods:
+
+    # Validate the best case first. If we don't have any unsupported_kmods, this means
+    # that everything is compatible and good to go.
+    if not unsupported_kmods:
+        logger.debug("Kernel modules are compatible.")
+    else:
         not_supported_kmods = "\n".join(
             map(
                 lambda kmod: "/lib/modules/{kver}/{kmod}".format(kver=system_info.booted_kernel, kmod=kmod),
@@ -288,12 +305,10 @@ def ensure_compatibility_of_kmods():
         logger.critical(
             (
                 "The following kernel modules are not supported in RHEL:\n{kmods}\n"
-                "Make sure you have updated the kernel to the latest available version and rebooted the system. "
+                "Make sure you have updated the kernel to the latest available version and rebooted the system.\n"
                 "Remove the unsupported modules and run convert2rhel again to continue with the conversion."
             ).format(kmods=not_supported_kmods, system=system_info.name)
         )
-    else:
-        logger.info("Kernel modules are compatible.")
 
 
 def validate_package_manager_transaction():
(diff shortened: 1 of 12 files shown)


convert2rhel/checks.py
```diff
@@ -287,66 +296,83 @@ def get_most_recent_unique_kernel_pkgs(pkgs):
...
             pkg_name,
         )
     else:
         return tuple(
             map(
-                _convert_to_int_or_zero,
+                convert_to_int_or_zero,
                 KERNEL_REPO_VER_SPLIT_RE.split(rpm_version),
             )
         )
```
r0x0d, line 373: I was unsure about this... Maybe we should move this function to `convert2rhel/repos.py` file? Since it's dealing with "repos" even it is only extracting the version from a string.

Not sure if the `convert2rhel/repos.py` or `convert2rhel/utils.py` is a better location for this function.

convert2rhel/systeminfo.py
```diff
@@ -291,12 +296,18 @@ def is_rpm_installed(name):
...
+        `tool_opts.enablerepo` option.
+
+        .. note::
+            The repositories passed through the CLI have more priority than the ones get get from RHSM.
+
+        :return: A list of enabled repos to use during the conversion
+        :rtype: list[str]
         """
         # TODO:
         # if not self.submgr_enabled_repos:
```
r0x0d, line 313: This `TODO` here needs any new information? It's not telling much. 

If it doesn't make sense, can we remove it?

convert2rhel/checks.py
abadger: It feels like maybe a function for pkghandler?  (Unless we want pkghandler to only deal with packages-in-repositories).  It's parsing a version from a single package's NEVR.  One possible reason not to move this (or to change it when we do move it), though: right now, this function is very specifically targeted at the NEVR of kernel package's that we think we are going to encounter.  All of the following NEVRs would output the same value from this function despite being different packages:

* kernel-2.4.0-1.el8_1
* kernel-2.4-0.1.el8_1
* kernel-2.4.0-1.el9_0

convert2rhel/checks.py
```diff
@@ -287,66 +295,82 @@ def get_most_recent_unique_kernel_pkgs(pkgs):
...
```
abadger, line 371: Thinking about the issues I raised earlier with different package versions being evaluated wrongly, I begin to think we should be using `rpm.labelCompare()` to do our sorting rather than trying to do it ourselves.  This is in part because we're not just scanning for kernel packages here, but also kmod packages (whose authors might be doing non-standard things with their naming or versioning).

The code for the two functions would look like this once ported to `rpm.labelCompare`:
``` python
import functools
import rpm

def get_most_recent_unique_kernel_pkgs(pkgs):
    pkgs_groups = itertools.groupby(sorted(pkgs), lambda pkg_name: pkg_name.split(":")[0])
    list_of_sorted_pkgs = []
    for distinct_kernel_pkgs in pkgs_groups:
        if distinct_kernel_pkgs[0].startswith(("kernel", "kmod")):
            list_of_sorted_pkgs.append(max(distinct_kernel_pkgs[1], key=functools.cmp_to_key(_package_version_cmp)))

    return tuple(list_of_sorted_pkgs)

def _package_version_cmp(pkg_1, pkg_2):
    pkg_ver_components = []
    for pkg in pkg_1, pkg_2:
        # Remove the package name and split the rest between epoch + version and release + arch
        epoch_version, release_arch =  pkg.rsplit('-', 2)[-2:]
        # Separate the (optional) epoch from the version
        epoch_version = epoch_version.split(':', 1)
        if len(epoch_version) == 1:
            epoch = '0',
            verson = epoch_version[0]
        else:
            epoch, version = epoch_version
        # Discard the arch
        release = release_arch.rsplit('.', 1)[0]
        pkg_ver_components.append((epoch, version, release))

    return rpm.labelCompare(pkg_ver_components[0], pkg_ver_components[1])
```
The splitting could also be done via a regex.  I believe we have such a regex in pkghandler.py although we would want to audit before using it (for two reasons: (1) yum and repoquery commands output epoch in different ways depending on the command [0:foo-1.0-2.el8.x86_64 vs foo-0:1.0-2.el8.x86_64].  Sometimes an epoch of 0 is omitted and sometimes it is placed in the string, and (2) some of the code we've written cares about the arch while others do not so the rege may grab arch or it might not.

@r0x0d Do you think that switching to `rpm.labelCompare` should be in this PR or a different one?

abadger:
I've reviewed this.  Other than the questions about `get_most_recent_unique_kernel_pkgs()` and its helper functions, this looks good.

convert2rhel/systeminfo.py
abadger: I can see the potential problem with repos not being enabled yet but I'm not sure whether that is ever a problem in actuality (especially since customers aren't reporting failures because of it).  @zhukovgreen Do you happen to know about this TODO and whether we can either enhance what it says or remove it?

zhukovgreen: I don't have any input on this. Seems like some artifact which I forgot to remove

convert2rhel/checks.py
r0x0d: @abadger, I think we can include this switch in this PR without a problem. Although this is a code cleanup and not a refactor, I feel that cleaning up code sometimes leads to a better way of doing things. 

I will take a closer look at the code you provided and see if I can implement it!

r0x0d: @abadger I have included your piece of code in the last commit.

convert2rhel/pkghandler.py
```diff
@@ -1075,3 +1082,45 @@ def _get_packages_to_update_dnf():
         packages.append(package.name)
 
     return packages
+
+
+def package_version_cmp(pkg_1, pkg_2):
```
abadger, line 1087: Hmmm... A  few higher level thoughts on this function:

* Sometimes yum and dnf return NEVR like this: `NetworkManager-1:1.18.8-2.0.1.el7_9.x86_64`.  Other times they return it like this (ENVR form): `1:NetworkManager-1.18.8-2.0.1.el7_9.x86_64`.  Should we handle both of these cases?  In one function or in two?

* There's already a `pkghandler.compare_package_versions` function.  That one takes its input as EVR strings: `1:1.18.8-2.0.1.el7_9.x86_64` which removes the ambiguity around whether the epoch is before or after name because there is no name.  However, we can't use that directly as a key function (since the key will be a NEVR or an ENVR string).

* Should we verify that the name is the same for both strings passed into this function?  If I call the function with `package_version_cmp("test-1.0-1", "foobar-2.0-1") I would expect it to error rather than say foobar is later than test.

* Should we parse arch off the end of the NEVR (arch is `x86_64`, `i686`, `s390x`, etc) and error if arches do not match?

Here's my strawman to address all of those:

* Separate version parsing and version comparison
* The parsing function will take a version string in either ENVR or NEVR format. 
  * If epoch or arch does not exist in the input string, it will be set to `None` in the tuple.
  * Parsing arch will be slightly tricky as I don't think there's a fool-proof way to separate an optional arch from release.  Maybe looking for a list of allowed arch patterns is the best we can do. Something like: `.*(x86_64|s390x|i.86)?$`
* The version comparison function will take two version strings in either ENVR or NEVR format.
* The strings will be passed to the parsing functions to return the information in a normalized form.
  * The names and arches will be compared. 
    * An error will be emitted if names do not match.
    * An error will be emitted if arches do not match and both arches are not None.
* rpm.labelCompare() will then be used to compare the EVR for both packages and the results returned.
* Port existing use of `pkghandler.compare_package_versions()` to use the new comparison function.
* Remove `pkghandler.compare_package_versions()` and `utils.string_to_version()`.

Discuss?

r0x0d: resolving this one since it's a duplicate.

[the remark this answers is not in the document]
r0x0d: @abadger, I think we can follow up on this in our 1:1 sessions. Probably next week, we can discuss it a bit more on which direction we want to move with this function.  

Definitely, there is a lot of room to improve the current function.

kokesak:
@r0x0d once you will have a free time, please rebase the PR

convert2rhel/pkghandler.py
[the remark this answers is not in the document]
r0x0d: I will create a Jira issue as well for specifically this one change.

r0x0d:
@kokesak rebased!

r0x0d:
I will rebase this to resolve the conflict with the checks.py

convert2rhel/checks.py
```diff
@@ -286,12 +297,10 @@ def ensure_compatibility_of_kmods():
         logger.critical(
             (
                 "The following kernel modules are not supported in RHEL:\n{kmods}\n"
-                "Make sure you have updated the kernel to the latest available version and rebooted the system. "
+                "Make sure you have updated the kernel to the latest available version and rebooted the system.\n"
                 "Remove the unsupported modules and run convert2rhel again to continue with the conversion."
```
bocekm, line 301:
```suggestion
                "Prevent the unsupported modules from loading by following https://access.redhat.com/solutions/41278 and run convert2rhel again to continue with the conversion."
```

bocekm (changes requested):

convert2rhel/checks.py
```diff
@@ -272,11 +275,19 @@ def check_custom_repos_are_valid():
...
     host_kmods = get_loaded_kmods()
     rhel_supported_kmods = get_rhel_supported_kmods()
     unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods)
-    if unsupported_kmods:
+
+    # Validate the best case first. If we don't have any unsupported_kmods, this means
+    # that everything is compatible and good to go.
+    if not unsupported_kmods:
+        logger.debug("Kernel modules are compatible.")
+    else:
```
bocekm, line 290: There's still one part of the RHELC-297 missing:

> Improve readability/maintainability of the `ensure_compatibility_of_kmods` function in checks.py as per https://github.com/oamg/convert2rhel/pull/193#discussion_r599881876

```python
def get_unsupported_kmods(host_kmods, rhel_supported_kmods):
    """Return a set of full paths to those installed kernel modules that are not available in RHEL repositories.
    ...
    """
    unsupported_kmods_subpaths = host_kmods - rhel_supported_kmods - set(system_info.kmods_to_ignore)
    unsupported_kmods_full_paths = ["/lib/modules/{kver}/{kmod}".format(kver=kernel_version, kmod=kmod)
                                    for kmod in unsupported_kmods_subpaths]
    return unsupported_kmods_full_paths
```

```python
def ensure_compatibility_of_kmods():
    ...
    if not unsupported_kmods:
        logger.debug("Kernel modules are compatible.")
    else:
        logger.critical(
            (
                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
                "Make sure you have updated the kernel to the latest available version and rebooted the system.\n"
                "Prevent the unsupported modules from loading by following https://access.redhat.com/solutions/41278"
                " and run convert2rhel again to continue with the conversion."
            ).format(kmods="\n".join(unsupported_kmods), system=system_info.name)
        )
```

bocekm (changes requested):

convert2rhel/checks.py
bocekm: Resolved by https://github.com/oamg/convert2rhel/pull/469/commits/5f52da81b3e23fb6a03cb4e5c2f4d251cb6b700f.

convert2rhel/checks.py
bocekm: Resolved by https://github.com/oamg/convert2rhel/commit/5f52da81b3e23fb6a03cb4e5c2f4d251cb6b700f.

bocekm (approved):

convert2rhel/utils.py
```diff
@@ -506,6 +506,23 @@ def string_to_version(verstring):
     return (epoch, version, release)
 
 
+def convert_to_int_or_zero(s):
```
bocekm, line 509: @r0x0d, you can remove this function and the associated unit test. It's not used anywhere anymore.

bocekm (changes requested):

kokesak (changes requested):
@danmyway since we want to improve the tests description, could you please update the summary in `tests/integration/tier0/inhibit-if-kmods-is-not-supported/main.fmf` and maybe even in the pytest file as well if you think it's not good enough

danmyway:
> @danmyway since we want to improve the tests description, could you please update the summary in `tests/integration/tier0/inhibit-if-kmods-is-not-supported/main.fmf` and maybe even in the pytest file as well if you think it's not good enough

Oh, the test got overwritten by rebasing again, I will push it back.

danmyway:
The test is back in @kokesak

r0x0d:
whoops, I think we have a rebase problem in this PR. I will try to fix it

convert2rhel/utils.py
r0x0d: done.

r0x0d:
Okay, I rebased that and applied the two latest commits from @danmyway here. Hope everything works as expected.

bocekm (dismissed):

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -61,16 +75,43 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
...
         c2r.expect("Continue with the system conversion?")
         c2r.sendline("n")
         assert c2r.exitstatus != 0
 
 
-def test_tainted_kernel_inhibitor(shell, convert2rhel):
-    # This test marks the kernel as tainted which is not supported by convert2rhel.
+def test_inhibit_if_module_is_force_loaded(shell, convert2rhel):
+    """
+    Test force loads kmod and checks for Convert2RHEL not being inhibited.
```
kokesak, line 88: I'm confused. The conversion should be inhibited no?

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -61,16 +75,43 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
...
+    With the check for tainted kernel modules being at the beginning of the script, abort the conversion ASAP
```
kokesak, line 89: In the beginning of the file is the check for custom kernel module no? Anyway I would probably removed this section completely

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -61,16 +75,43 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
...
+    for the sake of test speed.
+    """
+    # Force load the kernel module
+    assert shell("modprobe -f -v bonding").returncode == 0
+    # Check for force loaded modules being flagged FE in /proc/modules
+    assert "(FE)" in shell("cat /proc/modules").output
+
+    with convert2rhel("--no-rpm-va --debug") as c2r:
+        assert c2r.expect("Tainted kernel module\(s\) detected") == 0
```
kokesak, line 98:
```suggestion
        assert c2r.expect("Tainted kernel modules detected") == 0
```
So it's the same as below or whatever of these two is correct?

kokesak (changes requested):
Just a few small tweaks in the description in comments, other than that it looks good @danmyway

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
bocekm: A force loaded kernel module is considered tainted as well and the conversion is supposed to be inhibited as well.

danmyway: Oh yeah, I have changed the function name and did not apply the changes to the docstring

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
bocekm: I changed the log message in https://github.com/oamg/convert2rhel/commit/5f52da81b3e23fb6a03cb4e5c2f4d251cb6b700f. @r0x0d, is there a reason why you dropped the commit?

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
danmyway: This was added to check for handling scenarios when module is force loaded. 
The custom one is just loaded from a different custom path, thus considered custom.

danmyway: My bad, the kmod being force loaded should inhibit the conversion.

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
r0x0d: This PR had a rebase problem where we had a lot of commits from main here... I just pushed my local changes and applied Daniel's commits on top of it.

r0x0d: @bocekm do you mean that the commit is not available here anymore or the change proposed is missing?

bocekm: Both. The commit is not among the PR commits and the changes from the commit are not part of any other PR commit.

r0x0d: I see. Could you please comment again the change so I can apply it?

bocekm: All the changes are already in https://github.com/oamg/convert2rhel/commit/5f52da81b3e23fb6a03cb4e5c2f4d251cb6b700f.

You can use the approach @abadger suggested on Monday when you dropped @danmyway's commits from https://github.com/oamg/convert2rhel/pull/554:
```
$ git pull --rebase
$ wget https://github.com/oamg/convert2rhel/commit/5f52da81b3e23fb6a03cb4e5c2f4d251cb6b700f.patch
$ git am 5f52da81b3e23fb6a03cb4e5c2f4d251cb6b700f.patch
```

r0x0d: Okay, I applied the changes from 5f52da81b3e23fb6a03cb4e5c2f4d251cb6b700f.

bocekm: Not completely, for some reason. The map/lambda block is still present.

kokesak (approved):
LGTM

convert2rhel/utils.py
```diff
@@ -32,6 +32,39 @@
...
+
+            def __le__(self, other):
+                return mycmp(self.obj, other.obj) <= 0
+
+            def __ge__(self, other):
+                return mycmp(self.obj, other.obj) >= 0
+
+            __hash__ = None
+
+        return K
```
r0x0d, line 65: @abadger, I took this code from python 3.10 functools module, should I do the same as you did here? https://github.com/oamg/convert2rhel/pull/540/files#diff-f82efabf7bb126100afd0472e3d8a5902da6576c6ddf75e3184cce9eb647c52aR58-R59

convert2rhel/checks.py
```diff
@@ -272,26 +278,32 @@ def check_custom_repos_are_valid():
...
     rhel_supported_kmods = get_rhel_supported_kmods()
     unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods)
-    if unsupported_kmods:
+
+    # Validate the best case first. If we don't have any unsupported_kmods, this means
+    # that everything is compatible and good to go.
+    if not unsupported_kmods:
+        logger.debug("All loaded kernel modules are available in RHEL.")
+    else:
         not_supported_kmods = "\n".join(
```
bocekm, line 294: The map/lambda code block is something I removed under the commit https://github.com/oamg/convert2rhel/commit/5f52da81b3e23fb6a03cb4e5c2f4d251cb6b700f. The commit has been dropped from the PR twice. Please apply the patch or let me know why is it better to keep the map/lambda then the list comprehension.

bocekm (changes requested):

convert2rhel/checks.py
r0x0d: You're correct, actually. The list comprehension is more readable. 

We were having some problems during the rebase/pull of this branch. I will make sure to update it on Monday.

r0x0d: done.

r0x0d:
@danmyway I have rebased this PR to just one commit, hopefully, it will cause less confusion now.

bocekm (approved):
This looks good already from my POV! The tests fail though.

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -22,25 +30,31 @@ def factory():
 
 
 def test_inhibit_if_custom_module_loaded(insert_custom_kmod, convert2rhel):
-    # This test checks that rpmquery works correctly.
-    # If custom module is loaded the conversion has to be inhibited.
+    """
+    This test checks that rpmquery for detecting supported kernel modules in RHEL works correctly.
```
kokesak, line 34:
```suggestion
    This test verifies that rpmquery for detecting supported kernel modules in RHEL works correctly.
```

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -61,16 +75,41 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
...
         c2r.expect("Continue with the system conversion?")
         c2r.sendline("n")
         assert c2r.exitstatus != 0
 
 
-def test_tainted_kernel_inhibitor(shell, convert2rhel):
-    # This test marks the kernel as tainted which is not supported by convert2rhel.
+def test_do_not_inhibit_if_module_is_force_loaded(shell, convert2rhel):
+    """
+    Test force loads kmod and checks for Convert2RHEL being inhibited.
```
kokesak, line 88:
```suggestion
    Test force loads kmod and verifies that Convert2RHEL run is being inhibited.
```

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -61,16 +75,41 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
...
+    """
+    # Force load the kernel module
+    assert shell("modprobe -f -v bonding").returncode == 0
```
kokesak, line 91: @danmyway, If this is failing only on CentOS7/OL7 you can add a simple condition and skip this test case scenario on those el7 distros.

Unless you find a solution to make it work properly on EL7 distros.

r0x0d:
@danmyway, since the conflict is in an integration tests, do you want to try to resolve yourself, or do you want me to take care of it?

convert2rhel/utils.py
abadger: @r0x0d Yes, that would be good. Keeping track of licenses and copyright when copying code is important.

convert2rhel/checks.py
```diff
@@ -273,26 +279,26 @@ def check_custom_repos_are_valid():
...
+        logger.debug("All loaded kernel modules are available in RHEL.")
+    else:
         logger.critical(
-            (
-                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
-                "Make sure you have updated the kernel to the latest available version and rebooted the system. "
-                "Remove the unsupported modules and run convert2rhel again to continue with the conversion."
-            ).format(kmods=not_supported_kmods, system=system_info.name)
+            "The following loaded kernel modules are not available in RHEL:\n{0}\n"
+            "Prevent the modules from loading by following {1}"
```
bocekm, line 297:
```suggestion
            "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
            "If this message appears again after doing the above, prevent the modules from loading by following {1}"
```
@abadger, @SpyTec, @r0x0d, as we discussed on the Cabal session, without having the kernel currency check available on OL 8, a message notifying the user that the kernel needs to be up-to-date is useful.
I was thinking whether to go into more detail, saying that the loaded kernel should be the latest available for the minor version of the system they are converting but:
1. the message would not be concise
2. it would be difficult to phrase in a way that works also for a distro without minor versions (Stream)

bocekm (changes requested):

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
danmyway: Yep, that workaround would be needed for now.

bocekm (approved):

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -61,16 +75,44 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
             c2r.sendline("y")
             prompt_amount -= 1
 
-        assert c2r.expect("Kernel modules are compatible.", timeout=300) == 0
+        # Stop conversion before the point of no return as we do not need to run the full conversion
+        # with the check for kompatible kernel modules being right before the PONR.
```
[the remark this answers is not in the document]
danmyway: Good catch, thank you :+1:

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -61,16 +75,44 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
...
         c2r.sendline("n")
         assert c2r.exitstatus != 0
 
 
-def test_tainted_kernel_inhibitor(shell, convert2rhel):
-    # This test marks the kernel as tainted which is not supported by convert2rhel.
+def test_do_not_inhibit_if_module_is_force_loaded(shell, convert2rhel):
+    """
+    Test force loads kmod and verifies that Convert2RHEL run is being inhibited.
+    Force loaded kmods are denoted (FE) for F = module was force loaded E = unsigned module was loaded.
```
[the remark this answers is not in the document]
danmyway: I'll fix that, thank you for spotting that!

convert2rhel/checks.py
```diff
@@ -381,64 +393,46 @@ def get_most_recent_unique_kernel_pkgs(pkgs):
...
+    """
 
+    pkgs_groups = itertools.groupby(sorted(pkgs), lambda pkg_name: pkg_name.split(":")[0])
+    list_of_sorted_pkgs = []
+    for distinct_kernel_pkgs in pkgs_groups:
+        if distinct_kernel_pkgs[0].startswith(("kernel", "kmod")):
+            list_of_sorted_pkgs.append(
+                max(
+                    distinct_kernel_pkgs[1],
+                    key=functools.cmp_to_key(_package_version_cmp),
```
SpyTec, line 431: Unit tests for centos6 says this doesn't exist, which is confirmed in 2.6 docs https://docs.python.org/2.6/library/functools.html

SpyTec (changes requested):

convert2rhel/checks.py
r0x0d: done.

kokesak (dismissed):
The integration test looks nice.

convert2rhel/checks.py
r0x0d: done.

abadger (approved):
Code looks fine but there were two failing integration tests.  It looks like those are due to a problem with the yum repository.  I've kicked off new runs for both of those.

I also see that Eric had a crequest to change something but I wasn't able to find it in the converaation.  Can you check if that was addressed before merging?

SpyTec (approved):
> I also see that Eric had a crequest to change something but I wasn't able to find it in the converaation. Can you check if that was addressed before merging?

It was related to py2.6 and been addressed, LGTM as well

r0x0d:
Thanks, @abadger for re-running the tests. Now the only thing that failed is the codecov patch/project checks, but they are fine. 

@kokesak could you approve it once again?

kokesak (approved):
Looks nice, merge it whenever you want.

PR #475, opened 2022-04-08
[RHELC-570] Cleanup pexpect terminal size

abadger:
For some of the programs we run, we have to set the terminal size before the process starts.  In the past we've done this by subclassing pexpect.spawn on all platforms.  Newer pexpect gives us the dimensions parameter to spawn() to do this in a better way.  This change will use dimensions.spawn() when available and our hack when it is not.

Fixes [RHELC-570](https://issues.redhat.com/browse/RHELC-570)

### DIFF
--- a/convert2rhel/unit_tests/conftest.py
+++ b/convert2rhel/unit_tests/conftest.py
@@ -1,3 +1,4 @@
+import logging
 import sys
 
 import pytest
@@ -23,6 +24,39 @@ def is_py26():
 @pytest.fixture(scope="session")
 def is_py2():
     return sys.version_info[:2] <= (2, 7)
+
+
+@pytest.fixture
+def clear_loggers():
+    """
+    Remove handlers from the loggers.
+
+    If logger handlers are created while we are using capsys or capfd, then
+    the logger handlers get created with pytest's capture.  If pytest's
+    capture feature is disabled later, the logger handlers will get errors
+    trying to write to a closed file.  This is this issue:
+
+        https://github.com/pytest-dev/pytest/issues/5502
+
+    In our unittests, there are some tests that use pexpect that have to
+    disable pytest's capture otherwise pexpect-2.3 will malfunction.  The
+    combination of these two things causes logging calls during the utils
+    tests to complain that they are trying to log to a closed file handle.
+
+    Adding this fixture to any unit test which both uses capsys/capfd and
+    creates the logger handlers (by calling
+    `convert2rhel.logger.setup_logger_handler()`) fixes the issue.  The
+    fixture works by removing the handlers that `setup_logger_handler()`
+    (actually, all of the handlers) created when the test completes.
+    """
+    # Nothing to do in setup
+    yield
+    # In teardown, we clear all the logger handlers.
+    loggers = [logging.getLogger()] + list(logging.Logger.manager.loggerDict.values())
+    for logger in loggers:
+        handlers = getattr(logger, "handlers", [])
+        for handler in handlers:
+            logger.removeHandler(handler)
 
 
 @pytest.fixture()
--- a/convert2rhel/unit_tests/conftest.py
+++ b/convert2rhel/unit_tests/conftest.py
@@ -52,10 +52,11 @@ def clear_loggers():
     # Nothing to do in setup
     yield
     # In teardown, we clear all the logger handlers.
-    loggers = [logging.getLogger()] + list(logging.Logger.manager.loggerDict.values())
+    loggers = logging.Logger.manager.loggerDict.values()
     for logger in loggers:
-        handlers = getattr(logger, "handlers", [])
-        for handler in handlers:
+        if not hasattr(logger, "handlers"):
+            continue
+        for handler in logger.handlers[:]:
             logger.removeHandler(handler)
 
 
--- a/convert2rhel/unit_tests/logger_test.py
+++ b/convert2rhel/unit_tests/logger_test.py
@@ -21,14 +21,15 @@ import os
 import pytest
 
 from convert2rhel import logger as logger_module
-from convert2rhel.toolopts import tool_opts
 
 
-def test_logger_handlers(tmpdir, caplog, read_std, is_py2, capsys):
+def test_logger_handlers(monkeypatch, tmpdir, caplog, read_std, is_py2, global_tool_opts, clear_loggers):
     """Test if the logger handlers emmits the events to the file and stdout."""
+    monkeypatch.setattr(logger_module, "tool_opts", global_tool_opts)
+
     # initializing the logger first
     log_fname = "convert2rhel.log"
-    tool_opts.debug = True  # debug entries > stdout if True
+    global_tool_opts.debug = True  # debug entries > stdout if True
     logger_module.setup_logger_handler(log_name=log_fname, log_dir=str(tmpdir))
     logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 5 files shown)


abadger:
Marking as a draft for now.  Once #492 is merged, I'll rebase this on top of the changes there and it will be ready for review again.

abadger:
Latest push has coverage for the easier of the two issues (re-raising TypeError when the problem was not the dimensions argument) but the other uncovered line is actually leading me to realize there's an unexpected interaction between pytest's stdout capturing and pexpect.  I'm going to look into that a little more and then summarize here if I can't find a suitable solution.

abadger:
You are correct that the code in the PR doesn't show that setwinsize is executed. But it is happening inside of the pexpect-2.3 (RHEL7) code.   When we call [`super(PexpectSpawnWithDimensions, self).__init__(*args, **kwargs)`](https://github.com/oamg/convert2rhel/pull/475/files#diff-ed1ca2e4a8b27f1bee13d608562d4e94eceb2972790eccf4d84f366b750f221dR280) it calls  `pexpect.spawn.__init__()`   [which calls](https://github.com/pexpect/pexpect/blob/c9b0b6ac40b5dea4084302d6dbcc4c4e28cf414f/pexpect/__init__.py#L493) `self._spawn()` [which calls](https://github.com/pexpect/pexpect/blob/c9b0b6ac40b5dea4084302d6dbcc4c4e28cf414f/pexpect/__init__.py#L622) `self.setwinsize()`

abadger:
I've just rebased.  Michal Bocek is willing to try removing the explicit setwinsize() call in `run_command_in_pty()` so I'm going to let the integration tests run on this version to establish a baseline (hopefully, the baseline is "everything passes") and then push the change to remove https://github.com/oamg/convert2rhel/pull/475/files#diff-ed1ca2e4a8b27f1bee13d608562d4e94eceb2972790eccf4d84f366b750f221dR199 and check that the integration tests continue to work with that.

kokesak:
@abadger please rebase once you will have time to have the latest fix of the integration tests

bocekm (dismissed):

bocekm:
@abadger, I've removed the additional setting of the window size.

When running `make tests`, with or without this change I'm getting bunch of errors like:
```
  File "/usr/lib64/python2.7/logging/__init__.py", line 862, in emit
    stream.write(ufs % msg)
  File "/usr/lib/python2.7/site-packages/py/_io/capture.py", line 16, in write
    return StringIO.write(self, data)
ValueError: I/O operation on closed file
Logged from file utils.py, line 218
```
The unit tests pass but something fishy is going on. Would you please take a look at it?

abadger:
> @abadger, I've removed the additional setting of the window size.
> 
> When running `make tests`, with or without this change I'm getting bunch of errors like:
> 
> ```
>   File "/usr/lib64/python2.7/logging/__init__.py", line 862, in emit
>     stream.write(ufs % msg)
>   File "/usr/lib/python2.7/site-packages/py/_io/capture.py", line 16, in write
>     return StringIO.write(self, data)
> ValueError: I/O operation on closed file
> Logged from file utils.py, line 218
> ```
> 
> The unit tests pass but something fishy is going on. Would you please take a look at it?

I think this is the issue: https://github.com/pytest-dev/pytest/issues/5502  although I'm not quite sure, yet what it is about our code that triggers that behaviour and how to fix it (and why we've only seen it once in a great while before this PR)

abadger:
Okay, based on this: https://github.com/pytest-dev/pytest/issues/5502#issuecomment-526348052  I think what's happening is that:

1. Pytest starts and sbstitutes its capture buffers for stdin and stdout
2. convert2rhel code is imported and the logger is setup which makes use of the capture buffers.
3. In order to make many of the run_in_pty unittests work when run under pytest, they disable pytest's capture via: `with capfd.disabled():`
4. When the code inside of there runs, the pexpect portion of the code works but any log messages get emitted to the closed capture buffers.  Which means we get that error.

abadger:
Further information: the bad interaction appears to go away if we run logger_test.py after all of the other tests have been run.

abadger:
Alright.  So `logger_test.py:test_logger_handlers()` and `logger_test:test_tool_opts_debug()` are using two fixtures: `capsys` and `read_std`.  These two fixtures enable pytest's capturing behaviour.  Then these functions are calling `logger_module.setup_logger_handler()`.  Since capturing is on at that point, the logger ends up setup to use the capture buffers instead of the real stdout and stderr.  Later on, when utils_tests are run which disable capture, logger calls are trying to write to the buffers which are no longer available and so we end up with the IOErrors.

I'm looking into whether we can re-setup the logger afterwards to use the real stdout and stderr.

convert2rhel/unit_tests/conftest.py
```diff
@@ -19,6 +20,39 @@ def is_py2():
...
+    (actually, all of the handlers) created when the test completes.
+    """
+    # Nothing to do in setup
+    yield
+    # In teardown, we clear all the logger handlers.
+    loggers = [logging.getLogger()] + list(logging.Logger.manager.loggerDict.values())
+    for logger in loggers:
+        handlers = getattr(logger, "handlers", [])
+        for handler in handlers:
+            logger.removeHandler(handler)
```
bocekm, line 53:
```suggestion
        if not hasattr(logger, "handlers"):
            continue
        for handler in logger.handlers[:]:
            logger.removeHandler(handler)
```
This for loop does not work correctly because by calling the `removeHandler()` you effectively mutate the list you're iterating on, returned by `getattr(logger, "handlers", [])`. The end result is that not all handlers are removed from the logger. In my testing, just two out of four handlers are removed. Is this the reason for duplicating the code from this fixture in the tests themselves? Because iterating over the handlers the second time allowed to finally remove the rest of them.

What fixes the fixture is making a copy of the handlers list, for example by using a slice notation (inspiration: https://youtu.be/xeLecww65Zg?t=805).

By the way, the credit goes to this person who figured this out: https://stackoverflow.com/a/7484605/3499937).

convert2rhel/unit_tests/logger_test.py
```diff
@@ -64,10 +71,16 @@ def test_tools_opts_debug(tmpdir, read_std, is_py2):
...
     stdouterr_out, stdouterr_err = read_std()
     assert "debug entry 2: data" not in stdouterr_out
+    loggers = [logging.getLogger()] + list(logging.Logger.manager.loggerDict.values())
+    for logger in loggers:
+        handlers = getattr(logger, "handlers", [])
+        for handler in handlers:
+            logger.removeHandler(handler)
 
 
 def test_logger_custom_logger(tmpdir, caplog):
```
bocekm, line 86:
```suggestion
def test_logger_custom_logger(tmpdir, caplog, clear_loggers):
```
This test also call the `convert2rhel.logger.setup_logger_handler()`.

bocekm (changes requested):

bocekm (approved):
Let's wait for the packit to fix the copr build failures. Then it's ready to get merged.

bocekm (approved):

bocekm (changes requested):
The suggestions from my previous review have not been resolved.

abadger:
> The suggestions from my previous review have not been resolved.

As Michal theorized to me on internal chat. this was caused by my rebasing to pick up a CI workaround without pulling the fixes that had been commited via the web ui first.  I managed to recover that and added it back in.

bocekm (approved):

PR #481, opened 2022-04-20
Fix old logs archival

r0x0d:
This PR introduces a fix for the log archival introduced in the
[PR#359](https://github.com/oamg/convert2rhel/pull/359).

The feature introduced there did a `copyfile` instead of a `move` to the
`archive` directory, wich was not the expected way of archiving a log.

Using a `copyfile` instead of a `move` causes the log file to be copied to the
`archive` directory as expected, but we still keep appending the log to the
`convert2rhel.log` file.

Jira reference (If any): https://issues.redhat.com/browse/RHELC-566
Bugzilla reference (If any):

### DIFF
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -126,7 +126,7 @@ def archive_old_logger_files(log_name, log_dir):
 
     file_name, suffix = tuple(log_name.rsplit(".", 1))
     archive_log_file = "%s/%s-%s.%s" % (archive_log_dir, file_name, formatted_time, suffix)
-    shutil.copyfile(current_log_file, archive_log_file)
+    shutil.move(current_log_file, archive_log_file)
 
 
 def _task(self, msg, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


r0x0d:
Output for this change using `move` instead of `copy` 

```bash
[root@c2r-20220420152817 convert2rhel]# wc -l convert2rhel.log archive/*.log
   45 convert2rhel.log
   45 archive/convert2rhel-20220420T183115Z.log
   45 archive/convert2rhel-20220420T183128Z.log
   45 archive/convert2rhel-20220420T183129Z.log
  180 total
```

abadger:
Code looks good.  Should we also modify `unit_tests/logger_test.py:test_archive_old_logger_files()` to assert that `convert2rhel.log` does not exist after `archive_old_logger_files()` is called?

abadger (approved):
Thanks @r0x0d !  Changes look good.  Approved.

PR #492, opened 2022-05-10
Do not put the subscription-manager password onto the command line.

abadger:
Passing values on the command line is insecure.  With this change,
the rhsm password is passed interactively to subscription-manager
instead of being passed on the commandline when we shell out to it.

The structure of this change deserves a bit of description.  Previously,
we called one function to assemble all of the information needed to
invoke subscription-manager and then returned that as a string that
could be run as a command line.  We called a second function with that
string to actually run the command.

To send the password interactively, we need to stop adding the password
to the string of command line arguments but it still makes sense to keep
the code that figures out the password together with the code which
finds the other command line args.  So it makes sense to keep a single
function to do that but return the password and other args separately.

We could use a dict, a class, or a tuple as the return value from the
function.  That doesn't feel too ugly.  But then we need to pass that
structure into the function which takes care of invoking
subscription-manager on the command line and that *does* feel ugly.
That function would have to care about the structure of the data we pass
in (If a tuple, what is the order?  If a dict, what are the field
names?, etc).  To take care of this, we can make the data structure that
we return from assembling the data a class and the function which calls
subscription-manager a method of that class because it's quite natural
for a method to have knowledge of what attributes the class contains.

Hmm... but now that we have a class with behaviours (methods), it starts
to feel like we could do some more things.  A function that fills in the
values of a class, validates that the data is proper, and then returns
an instance of that class is really a constructor, right?  So it makes
sense to move the function which assembles the data and returns the
class a constructor.  But that particular function isn't very generic:
it uses knowledge of our global toolopts.tool_opts to populate the
class.  So let's write a simple __init__() that takes all of the values
needed as separate parameters and then create an alternative constructor
(an @classmethod which returns an instance of the class) which gets the
data from a ToolOpt, and then calls __init__() with those values and
returns the resulting class.

Okay, things are much cleaner now, but there's one more thing that we
can do.  We now have a class that has a constructor to read in data and
a single method that we can call to return some results.  What do we
call an object that can be called to return a result?  A function or
more generically, in python, a Callable.  We can turn this class into a
callable by renaming the method which actually invokes
subscription-manager __call__().

What we have at the end of all this is a way to create a function which
knows about the settings in tool_opts which we can then call to perform
our subscription-manager needs::

    registration_command = RegistrationCommand.from_tool_opts()
    return_code = registration_command()

OAMG-6551 #done  convert2rhel now passes the rhsm password to subscription-manager securely.

* Modify the hiding of secret to hide both --password SECRET and
  --password=SECRET.  Currently we only use it with passwords that we
  are passing in the second form (to subscription-manager) but catching
  both will future proof us in case we use this function for more things
  in the future. (Eric Gustavsson)
  * Note: using generator expressions was tried here but found that they
    only bind the variable being iterated over at definition time, the
    rest of the variables are bound when the generator is used.  This
    means that constructing a set of generators in a loop doesn't really
    work as the loop variables that you use in the generator will have a
    different value by the time you're done.

    So a nested for loop and if-else is the way to implement this.

* Add a comment about activation_key being insecure for now.  We can fix
  this once subscription-manager implements https://issues.redhat.com/browse/ENT-4724
  (Eric Gustavsson)

Unittest enhancements for the subscription RegisterCommand change:

* Add global_tool_opts fixture to conftest.py which monkeypatches a
  fresh ToolOpts into convert2rhel.toolopts.tool_opts.  That way tests
  can modify that without worrying about polluting other tests.

  * How toolopts is imported in the code makes a difference whether this
    fixture is sufficient or if you need to do a little more work. If
    the import is::

      from convert2rhel import toolopts
      do_something(toolopts.tool_opts.username)

   then your test can just do::

      def test_thing_that_uses_toolopts(global_tool_opts):
          global_tool_opts.username = 'badger'

   Most of our code, though, imports like this::

      # In subscription.py, for instance
      do_something(tool_opts.username)

   so the tests should do something like this::

      def test_toolopts_differently(global_test_opts, monkeypatch):
          monkeypatch.setattr(subscription, 'tool_opts', global_tool_opts)

* Add unittests for utils.run_cmd_in_pty()
* Add unittests for the subscription.RegistrationCommand object.
* subscription_test.py::test_hide_secrets was expanded to test without
  equals sign
* Flaky test appeared within systeminfo_test.py::test_generate_rpm_va
  due to changes to tool_opts in other tests. A temporary fix is added
  until the whole test class is changed from unittest to pytest
* Fix failing unittests for RegistrationCommand

* Sometimes a process will close stdout before it is done processing.
  When that happens, we need to wait() for the process to end before
  closing the pty.  If we don't wait, the process will receive a HUP
  signal which may end it before it is done.
  * But on RHEL7, pexpect.wait() will throw an exception if you wait on
    an already finished process.  So we need to ignore that exception
    there.

* Add integration test Check for cve fixes, add 'psutil' to
  install-testing-depends playbook

* Fix Pexpect.spawn truncating lines on RHEL7.
  Along with a comment on why the change fixes the bug.

Fixes: https://issues.redhat.com/browse/RHELC-432


abadger:
I believe the lgtm errors are false positives but I would appreciate it if someone else would take a look as lgtm is flagging them as security issues (revealing of private data).

convert2rhel/utils.py
```diff
@@ -181,28 +188,59 @@ def run_cmd_in_pty(cmd, print_cmd=True, print_output=True, columns=120):
...
+# when the subprocess starts but dnf download checks the terminal's size
+# just before it prints the statusline we care about.  So setting the
+# terminal size via setwinsize() after the process is created works (note:
+# there is a race condition there but it's unlikely to ever trigger as it
+# would require downloading a package to happen quicker than the time
+# between calling spawn.__init__() and spawn.setwinsize())
+class PexpectSizedWindowSpawn(pexpect.spawn):
+    # https://github.com/pexpect/pexpect/issues/134
+    def setwinsize(self, rows, cols):
+        super(PexpectSizedWindowSpawn, self).setwinsize(rows, 120)
```
SpyTec, line 241: We're removing this workaround in #475 for 0.27 and won't fix it for 0.26

SpyTec:
@abadger was there any difference between this and the patch?

convert2rhel/subscription.py
```diff
@@ -135,65 +135,184 @@ def register_system():
...
+        """
+        Run the subscription-manager command.
+
+        Wrapper for running the subscription-manager command that keeps
+        secrets secure.
+        """
+        if self.password:
+            loggerinst.debug("Calling command '%s %s'" % (self.cmd, " ".join(hide_secrets(self.args))))
+            output, ret_code = utils.run_cmd_in_pty(
+                [self.cmd] + self.args, expect_script=(("assword: ", self.password + "\n"),), print_cmd=False
```
SpyTec, line 273:
```suggestion
                [self.cmd] + self.args, expect_script=(("password: ", self.password + "\n"),), print_cmd=False
```

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -463,6 +347,499 @@ def __call__(self, pkg, dest, reposdir=None):
...
+        monkeypatch.setattr(utils, "run_subprocess", mock.Mock(return_value=("", 0)))
+        monkeypatch.setattr(utils, "run_cmd_in_pty", mock.Mock(return_value=("", 0)))
+
+        reg_cmd = subscription.RegistrationCommand(username="me_myself_and_i", password="a password")
+        reg_cmd()
+
+        assert utils.run_cmd_in_pty.called_once_with(
+            "subscription-manager",
+            ["register", "--force", "--username=me_myself_and_i"],
+            expect_script=(("assword: *", "a password\n"),),
```
SpyTec, line 761:
```suggestion
            expect_script=(("password: *", "a password\n"),),
```

SpyTec (commented):
LGMT warnings seem like false positives

convert2rhel/subscription.py
abadger: I'm purposefully leaving off the initial letter to handle both `password` and `Password`.  I think that pexpect supports pattern matching so I can look up the syntax for patterns that it uses and change that appropriately.

abadger:
> @abadger was there any difference between this and the patch?

There were quite a few conflicts (there have been changes from Rodolfo and Andrew which were merged after the patch was created) which I had to resolve.  (I have asked both of them to take a look at this PR specifically to see whether their changes are still intact.)

convert2rhel/subscription.py
SpyTec: Makes sense. Since it wasn't documented in the code I didn't know. Either a comment about it or changing to a pattern lookup would work

r0x0d:
> > @abadger was there any difference between this and the patch?
> 
> There were quite a few conflicts (there have been changes from Rodolfo and Andrew which were merged after the patch was created) which I had to resolve. (I have asked both of them to take a look at this PR specifically to see whether their changes are still intact.)

Looking at the changes right now.

r0x0d:
@abadger for the false positives of leaking passwords, you can add a comment on the line flagged by LGTM as `# lgtm[py/clear-text-logging-sensitive-data]`, this would mute LGTM on the next runs about this.

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -463,6 +347,497 @@ def __call__(self, pkg, dest, reposdir=None):
...
+        )
+        monkeypatch.setattr(subscription.RegistrationCommand, "from_tool_opts", fake_from_tool_opts)
+
+        class FakeProcess(mock.Mock):
+            called_count = 0
+
+            @property
+            def exitstatus(self):
+                self.called_count += 1
+                return self.called_count % 3
```
r0x0d, line 454: Just a minor suggestion for this: 

This could be moved to the outside test method, like the others class mocks to follow the same “pattern” and then, the test method itself would look more clean. 

No need to apply this if you don't think it makes sense.

convert2rhel/utils.py
```diff
@@ -154,15 +154,22 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
...
 
     The pseudo-terminal can be useful when a command prints out a different output with or without an active terminal
     session. E.g. yumdownloader does not print the name of the downloaded rpm if not executed from a terminal.
     Switching off printing the command can be useful in case it contains a password in plain text.
 
-    :param cmd: The command to execute, including the options, e.g. "ls -al"
-    :type cmd: string
+    :param cmd: The command to execute, including the options as a list, e.g. ["ls", "-al"]
+    :type cmd: list
+    :param expect_script: An iterable of pairs of expected strings and response strings.  By giving
```
r0x0d, line 166:
```suggestion
    :param expect_script: An iterable of pairs of expected strings and response strings. By giving
```

convert2rhel/utils.py
```diff
@@ -154,15 +154,22 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
...
+    these pairs, interactive programs can be scripted.  Example:
+        run_cmd_in_pty(['sudo', 'whoami'], [('password: ', 'sudo_password\n')])
+        Note1: The caller is responsible for adding newlines to the response strings where
+        needed.  Note2: This function will await pexpect.EOF after all of the pairs in expect_script
```
r0x0d, line 170:
```suggestion
        needed. Note2: This function will await pexpect.EOF after all of the pairs in expect_script
```

convert2rhel/utils.py
```diff
@@ -181,28 +188,59 @@ def run_cmd_in_pty(cmd, print_cmd=True, print_output=True, columns=120):
...
+# subclass is a kludge for the issue.  On pexpect-2.3, it fixes the issue
+# because of the setwinsize call in __init__() at the cost of never being
+# able to change the column size later.  On later pexpect (RHEL-8 has
+# pexpect-4.3), this doesn't fix the issue of the terminal size being small
+# when the subprocess starts but dnf download checks the terminal's size
+# just before it prints the statusline we care about.  So setting the
+# terminal size via setwinsize() after the process is created works (note:
+# there is a race condition there but it's unlikely to ever trigger as it
+# would require downloading a package to happen quicker than the time
+# between calling spawn.__init__() and spawn.setwinsize())
```
r0x0d, line 237:
```suggestion
# For pexpect released prior to 2015 (RHEL7's pexpect-2.3),
# spawn.__init__() hardcodes a call to setwinsize(24, 80) to set the
# initial terminal size. There is no official way to set the terminal size
# to a custom value before the process starts. This can cause an issue with
# truncated lines for processes which read the terminal size when they
# start and never refresh that value (like yumdownloader)
#
# overriding setwinsize to set the columns to the size we want in this
# subclass is a kludge for the issue. On pexpect-2.3, it fixes the issue
# because of the setwinsize call in __init__() at the cost of never being
# able to change the column size later.  On later pexpect (RHEL-8 has
# pexpect-4.3), this doesn't fix the issue of the terminal size being small
# when the subprocess starts but dnf download checks the terminal's size
# just before it prints the statusline we care about. So setting the
# terminal size via setwinsize() after the process is created works (note:
# there is a race condition there but it's unlikely to ever trigger as it
# would require downloading a package to happen quicker than the time
# between calling spawn.__init__() and spawn.setwinsize())
```

plans/tier0.fmf
```diff
@@ -1,6 +1,10 @@
 discover+:
     tier: 0
 
+/check_cve:
```
r0x0d, line 4: We should be able to pass the CVE number here too, right? 
```suggestion
/check_cve_2022_1662:
```

plans/tier0.fmf
```diff
@@ -1,6 +1,10 @@
...
+  discover+:
+    test: check-cve
```
r0x0d, line 6: Same case here
```suggestion
    test: check-cve-2022-1662
```

tests/integration/tier0/check-cve-2022-1662/main.fmf
```diff
@@ -0,0 +1,6 @@
+summary: check cve fixes
```
r0x0d, line 1:
```suggestion
summary: check cve-2022-1662 is fixed
```

r0x0d (dismissed):
@abadger most of the changes I proposed here are more aesthetic ones, thus, feel free to accept or deny them! 

In an overral, from what I remember from my code in `subscription.py`, everything looks fine with the modifications and merges you did, don't think you missed anything!

plans/tier0.fmf
SpyTec: It's more of a general check for all CVEs we'll ever have I think. Since this is the first one it is just a bit empty atm

r0x0d: Ah, I see! Makes sense from this point of view, then!

convert2rhel/subscription.py
```diff
@@ -135,65 +135,187 @@ def register_system():
...
+        A callable that can register a system with subscription-manager.
+
+        :kwarg server_url: Optional url to the subscription-manager backend.
+            Useful when the customer has an on-prem subscription-manager instance.
+        :kwarg activation_key: subscription-manager activation_key that can be
+            used to register the system.  Org must be specified if this was given.
+        :kwarg org: The organization that the activation_key is associated with.
+            Required if activation_key is specified.
+        :kwarg username: Username to authenticate with subscription-manager.
+            Required if password is specified.
```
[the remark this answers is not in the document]
abadger: I would leave this one without `the` since the argument is named `password` so it follows the rules for proper nouns in English.  (In English we say Alexander instead of The Alexander).

convert2rhel/subscription.py
```diff
@@ -135,65 +135,187 @@ def register_system():
...
+
+        if activation_key and not org:
+            raise ValueError("org must be specified if activation_key is used")
+        self.activation_key = activation_key
+        self.org = org
+
+        self.password = password
+        self.username = username
+        if (password and not username) or (username and not password):
+            raise ValueError("username and password must be used together")
```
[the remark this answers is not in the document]
abadger: I think I was trying to use empty lines to to group attributes being set with the code that was validating or otherwise manipulating them but I see where you're coming from.  This block is a bit longer than the other ones because of all the special interactions between password, username, and activation_key....

/me applies the change.

convert2rhel/subscription.py
```diff
@@ -135,65 +135,187 @@ def register_system():
...
+            registration_attributes["password"] = password
+
+        if tool_opts.serverurl:
+            loggerinst.debug("    ... using custom RHSM URL")
+            registration_attributes["server_url"] = tool_opts.serverurl
+
+        return cls(**registration_attributes)
+
+    @property
+    def args(self):
```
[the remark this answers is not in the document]
abadger: Yeah, that's a great idea.

convert2rhel/subscription.py
```diff
@@ -135,65 +135,187 @@ def register_system():
...
-            loggerinst.info("    ... password detected")
+        return output, ret_code
 
-        password = tool_opts.password
-        while not password:
-            password = utils.prompt_user("Password: ", password=True)
 
-        registration_cmd.extend(("--username=%s" % username, "--password=%s" % password))
+def hide_secrets(args):
+    """Replace secret values with asterisks."""
```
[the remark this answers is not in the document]
abadger: +1.  I'll do that.

convert2rhel/unit_tests/subscription_test.py
abadger: I think in this case, it's specific to the function being tested (It returns the non-zero exit codes 1 and 2 the first two times it's called and then a success (0) exit code the third time).  It could be made more generic and then moved to the toplevel, though.  But if so, I think it should take these things into consideration:

* How many iterations before success?
* Spawn makes available exitstatus and signalstatus, should this allow setting both?
* Should it also set stdout output?
* Should the specific exitstatus per invocation be settable or is using ` count % success_attempt` good enough?

I think those are all good things to consider but maybe we should save moving it to a subsquent PR.

convert2rhel/subscription.py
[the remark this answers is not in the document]
abadger: Okay, docstring added.

convert2rhel/subscription.py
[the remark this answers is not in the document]
abadger: And docstring updated for this function too.

bocekm (dismissed):

convert2rhel/subscription.py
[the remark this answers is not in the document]
r0x0d: Ah, I see. This was a change proposed by my language checker, I don't remember what rule it followed to apply the `the` in the phrase, but if it's not required, that's okay! 

Learned something new just now!

convert2rhel/unit_tests/subscription_test.py
r0x0d: Hmm, yeah, agreed! I think we can include a comment then somewhere saying that if there is any need to reuse this class in another test, then those points you mentioned should be considered as well. 

Maybe in the future there will be more tests regarding pexpect that we need to do, so this might come in hand.

abadger:
Rebased to pick up the fixes for ubi8

danmyway:
Rebased onto main.
Small change in the test_user_response, which was causing the test to fail, though.

danmyway:
All tests have passed, @bocekm

bocekm (approved):

r0x0d:
@abadger +1 for that toolopts fixture.

r0x0d (approved):

PR #494, opened 2022-05-12
Install json-c from UBI 8

bocekm:
To be able to install the subscription-manager packages, we download
them from the Universal Base Image (UBI) 8 BaseOS publicly available
repository. This repository has been recently updated with packages
corresponding to RHEL 8.6 packages.

When trying to install these downloaded RHEL 8.6 subscription-manager
packages on CentOS/Oracle Linux 8.5 and older, the installation fails
because dnf-plugin-subscription-manager depends on
libjson-c.so.4(JSONC_0.14)(64bit). This symbol is newly available in
RHEL 8.6 json-c package and is not available in the json-c package
distributed with CentOS Linux/Oracle Linux/RHEL 8.5 and older.

For this reason we need to download and install the newer json-c package
version from the UBI 8 repository.

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -509,6 +509,7 @@ def download_rhsm_pkgs():
             "dnf-plugin-subscription-manager",
             "python3-syspurpose",
             "python3-cloud-what",
+            "json-c.x86_64",  # there's also an i686 version which we don't need
         ]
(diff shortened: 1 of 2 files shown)


r0x0d (approved):

abadger (approved):
Looks good.

PR #495, opened 2022-05-13
Use YumBase instead of YumBaseCli to get pkg updates

bocekm:
Using YumBaseCli caused doubling of every log message on the output.

The reason is that instantiating of the YumBaseCli class includes calling logging.basicConfig() which sets up handlers on the root logger.

It is possible to use the YumBase class, which does not affect logging, to get the same data - list of packages having updates available.

The YumBaseCli logging handlers interfered with the convert2rhel logging as we use propage=True to be able to use caplog in our unit tests:
https://github.com/oamg/convert2rhel/pull/179

### DIFF
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -1502,8 +1502,6 @@ def test_get_total_packages_to_update(package_manager_type, packages, monkeypatc
 )
 @pytest.mark.parametrize(("packages"), ((["package-1", "package-2", "package-3"],)))
 def test_get_packages_to_update_yum(packages, monkeypatch):
-    sys.path.insert(0, "/usr/share/yum-cli")
-    from cli import YumBaseCli  # pylint: disable=E0401
 
     PkgName = namedtuple("PkgNames", ["name"])
     PkgUpdates = namedtuple("PkgUpdates", ["updates"])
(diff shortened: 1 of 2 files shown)


convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1502,8 +1502,7 @@ def test_get_total_packages_to_update(package_manager_type, packages, monkeypatc
 )
 @pytest.mark.parametrize(("packages"), ((["package-1", "package-2", "package-3"],)))
 def test_get_packages_to_update_yum(packages, monkeypatch):
-    sys.path.insert(0, "/usr/share/yum-cli")
-    from cli import YumBaseCli  # pylint: disable=E0401
+    from yum import YumBase
```
r0x0d, line 1505:
```suggestion
```

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1513,11 +1512,8 @@ def test_get_packages_to_update_yum(packages, monkeypatch):
 
     pkg_lists_mock = mock.Mock(return_value=PkgUpdates(transaction_pkgs))
 
-    monkeypatch.setattr(YumBaseCli, "returnPkgLists", value=pkg_lists_mock)
+    monkeypatch.setattr(YumBase, "doPackageLists", value=pkg_lists_mock)
```
r0x0d, line 1515:
```suggestion
    monkeypatch.setattr(pkgmanager.YumBase, "doPackageLists", value=pkg_lists_mock)
```

r0x0d (changes requested):
With this change, we fix the pylint errors message and make the tests look cleaner without that import dependency.

r0x0d (approved):

abadger (approved):

PR #496, opened 2022-05-13
Migrate from unittest to pytest

abadger:
example_test.py, other_test.py, and redhatrelease_test.py currently have no open
PRs so migrate them to pytest.

We decided that we would write new tests in pytest.  To move some of the existing tests over a bit quicker, I ported test files which currently have no open PRs (to avoid making those PR owners rebase).


convert2rhel/unit_tests/example_test.py
```diff
@@ -19,29 +19,28 @@
...
-        self.assertEqual(utils.run_subprocess.prefix, "this ain't ")
+class RunSubprocessMocked(unit_tests.MockFunction):
+    def __call__(self, *args, **kwargs):
+        """
+        Implementation of the simplest behavior of a mocked function -
+        ignore the input parameters and just return fake value.
+        """
+        self.prefix = "this ain't "
+        self.ret = (self.prefix + self.ret[0], self.ret[1])
+        return self.ret
```
r0x0d, line 34: -- Not something to fix, just a note --

Just a quick note: We have an Issue in our Jira board to transform classes/functions likes this into fixtures in the conftest.py file. 

Maybe it could be a good idea for us, in the future, to move this class to be a fixture and add a couple of examples here on how to use it.

r0x0d (approved):
The changes look good to me, the only thing that I'm not getting is why codecov reports that we lost 5% of coverage :thinking:

SpyTec (approved):
Seems like Codecov has some issues with something. Rebasing off of main might work. But we can also just check the test action and make assumptions. The unit tests give off the right code coverage value and nothing looks off to me with the test conversion. Approving and merging

PR #497, opened 2022-05-13
Improve of Python 2 & 3 compatibility

hosekadam:
Mainly in unit test there was used a condition for importing the corresponding variant of module for python 2 or 3, if they have different names. The better way is to use the `six` library.

Changed the conditions to imports through `six`, removed partial compatibility with Python 2.6 (just in some of unit test, not in all of them - module `unittest2`), removed unused imports in tests and small improvement of `utils.py`: changed deprecated `warn` to `warning` which caused deprecated error in unit tests.

RHELC-588

### DIFF
--- a/Dockerfiles/centos6.Dockerfile
+++ b/Dockerfiles/centos6.Dockerfile
@@ -28,6 +28,9 @@ COPY $PREP_PIP_DEPS $PREP_PIP_DEPS
 COPY $APP_DEV_DEPS $APP_DEV_DEPS
 COPY $APP_PRE_DEV_DEPS $APP_PRE_DEV_DEPS
 RUN chmod +x $PREP_PIP_DEPS
+# The SSL implementation that python-2.6 uses is insecure and pip refuses to install
+# packages using it. This script will use curl to download the packages securely and
+# then pip can be asked to install from the local files
 RUN $PREP_PIP_DEPS
 RUN $PIP install --no-index --find-links /data -r $APP_PRE_DEV_DEPS
 RUN $PIP install --no-index --find-links /data -r $APP_DEV_DEPS
--- a/Dockerfiles/centos6.Dockerfile
+++ b/Dockerfiles/centos6.Dockerfile
@@ -1,0 +1,40 @@
+FROM centos:6 as base
+
+ENV PYTHON python2
+ENV PIP pip2
+ENV PYTHONDONTWRITEBYTECODE 1
+
+ENV URL_GET_PIP "https://fedora-archive.ip-connect.vn.ua/epel/6/x86_64/Packages/p/python-pip-7.1.0-2.el6.noarch.rpm"
+ENV APP_DEV_DEPS "requirements/centos6.requirements.txt"
+ENV APP_PRE_DEV_DEPS "requirements/centos6_pre.requirements.txt"
+ENV PREP_PIP_DEPS "scripts/centos6_pip_prep.sh"
+ENV APP_MAIN_DEPS \
+    python-six \
+    pexpect \
+    gcc \
+    python-devel
+
+WORKDIR /data
+
+FROM base as install_main_deps
+RUN sed -i 's/^mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Base.repo &&\
+    sed -i 's/^#baseurl.*$/baseurl=https:\/\/vault.centos.org\/6.10\/os\/x86_64/g' \
+    /etc/yum.repos.d/CentOS-Base.repo
+RUN yum update -y && yum install -y $APP_MAIN_DEPS && yum clean all
+
+FROM install_main_deps as install_dev_deps
+RUN yum install -y $URL_GET_PIP
+COPY $PREP_PIP_DEPS $PREP_PIP_DEPS
+COPY $APP_DEV_DEPS $APP_DEV_DEPS
+COPY $APP_PRE_DEV_DEPS $APP_PRE_DEV_DEPS
+RUN chmod +x $PREP_PIP_DEPS
+RUN $PREP_PIP_DEPS
+RUN $PIP install --no-index --find-links /data -r $APP_PRE_DEV_DEPS
+RUN $PIP install --no-index --find-links /data -r $APP_DEV_DEPS
+
+FROM install_dev_deps as install_application
+RUN groupadd --gid=1000 -r app && \
+    useradd -r --uid=1000 --gid=1000 app
+RUN chown -R app:app .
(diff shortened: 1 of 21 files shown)


hosekadam:
I see that there is a new PR496 with converting the tests from unittest to pytest, so maybe this PR can wait until the PR496 would be merged and then rebase this one and apply possible changes.

convert2rhel/systeminfo.py
```diff
@@ -15,20 +15,20 @@
...
-try:
-    import ConfigParser as configparser
-except ImportError:
-    import configparser  # pylint: disable=import-error
 
-import difflib
-import logging
-import os
-import re
+six.add_move(six.MovedModule("configparser", "ConfigParser", "configparser"))
```
abadger, line 30: I believe that the configparser move is built into six.  So this line is probably not necessary?

hosekadam:
@abadger I've tried to run the tests on CentOS 6 (which is the system with Python 2.6.6) and I got 4 (like unsupported comma etc.) syntax errors, so I think the tests aren't compatible anymore... 

I wanted to try it once again, but unfortunately my terminal crashed, so I (and there is the problem) rebuilt the images what always crashes due `No matching distribution found for pip<10`. Now I think, that I had some leftover container with Centos6, where I was able to run the first attempt of unit tests. Because of that I don't have any output to attach :(

hosekadam:
I'll investigate how to run the unittest under centos 6 again.

hosekadam:
There is running Centos 6 container with all required dependencies. With `make tests6` you can test it. Now I need to fix the unit tests for Python 2.6.6

abadger (approved):
This looks good to me.  Let's do the python-2.6 fixes in a separate PR. that way we have a checkpoint with all of the improvements that you have here.

abadger:
The coverage.io report that test coverage has decreased with this change is a little strange to me.  From what I can see, the lines it has marked as newly uncovered should not have changed and I do see unittests that should be testing those lines.

abadger:
Running `pytest` with `coverage` On a local centos7 vm with package versions as close to what I see from the coverage github logs I've found a discrepancy in what is being tested:
Local output from pytest --cov convert2rhel
> Name                            Stmts   Miss Branch BrPart  Cover   Missing
> convert2rhel/subscription.py      340     34    148      7    89%   369-371, 391->400, 474-475, 479-480, 508, 518-521, 527-533, 542, 557-558, 565-566, 589-604, 657-658, 706->exit

Output in the github action coverage-centos7 logs:
> Name                            Stmts   Miss Branch BrPart  Cover   Missing
> convert2rhel/subscription.py      335     34    146      7    89%   361-363, 383->392, 466-467, 471-472, 500, 510-513, 519-525, 534, 549-550, 557-558, 581-596, 649-650, 698->exit

For subscription.py (and only subscription.py), the total number of statements found in subscription.py is 5 less than when I run pytest locally.  The listed statements missing coverage all seem to be the same except offset by five lines.  I think that the missing statements account for the .06% drop in coverage but I'm not sure why pytest thinks there are 5 less lines in that file whn run as a github action.

Makefile
```diff
@@ -86,6 +87,10 @@ tests8: images
 	@echo 'CentOS Linux 8 tests'
 	@$(DOCKER) run --rm --user=$(id -ur):$(id -gr) -v $(shell pwd):/data:Z $(IMAGE)/centos8 pytest --show-capture=$(SHOW_CAPTURE) $(PYTEST_ARGS)
 
+tests6: images
+	@echo 'CentOS Linux 6 tests'
+	@$(DOCKER) run --user=$(id -ur):$(id -gr) --rm -v $(shell pwd):/data:Z $(IMAGE)/centos6 pytest
```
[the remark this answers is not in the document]
r0x0d: We don't need to do this right now. Just left thing here, so we can fix in another PR, as this one is good to merge!

r0x0d (approved):

Dockerfiles/centos6.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    /etc/yum.repos.d/CentOS-Base.repo
+RUN yum update -y && yum install -y $APP_MAIN_DEPS && yum clean all
+
+FROM install_main_deps as install_dev_deps
+RUN yum install -y $URL_GET_PIP
+COPY $PREP_PIP_DEPS $PREP_PIP_DEPS
+COPY $APP_DEV_DEPS $APP_DEV_DEPS
+COPY $APP_PRE_DEV_DEPS $APP_PRE_DEV_DEPS
+RUN chmod +x $PREP_PIP_DEPS
+RUN $PREP_PIP_DEPS
```
abadger, line 31: Just add a comment here and in the script about what this is for:

```suggestion
# The SSL implementation that python-2.6 uses is insecure and pip refuses to install
# packages using it. This script will use curl to download the packages securely and
# then pip can be asked to install from the local files
RUN $PREP_PIP_DEPS
```

scripts/centos6_pip_prep.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
```
abadger, line 2:
```suggestion

# The SSL implementation that python-2.6 uses is insecure and pip refuses to install
# packages using it. This script will use curl to download the packages securely and
# then pip can be asked to install from the local files

```

abadger:
@hosekadam Michal wanted to try out the coverage fix so I moved that to another PR.  I'm going to add the suggestions and rebase this PR to latest.  Hope that's okay with you!

scripts/centos6_pip_prep.sh
abadger: Add a comment on why the PREP_PIP_DEPS script is needed

hosekadam:
@abadger Yeah, it's fine :) And thanks for adding the suggestions and rebasing!

Makefile
```diff
@@ -78,6 +79,10 @@ lint-errors: images
 
 tests: tests7 tests8
```
r0x0d, line 80: We should add the `tests6` here as well, so when we execute the whole `make tests` command, we get everything executed. 

But this is not a blocker for me. We can merge this and add just this piece of code in another PR.

r0x0d (approved):
Looks super nice to me. 
I just made one small comment that we can handle later in another PR, no need to add any extra work into this for my comment. 

Also, loved the simplicity of using `six` for both python 2 & 3. Good work!

Makefile
r0x0d: Thanks for the clarification!

abadger:
Thanks Adam for all your work on this!

PR #501, opened 2022-05-18
Improve various log messages

bocekm:
- be consistent (punctuation, words used, etc.)
- change logging level - move low important messages to debug level and bump more important messages up from the debug level
- provide a bit more details where possible
- remove message bringing no value
- add messages where missing (e.g. to untangle the mystery of `DEBUG - Calling command 'grep ( /proc/modules'` under the task `Checking the firmware interface type (BIOS/UEFI)`)

### DIFF
--- a/convert2rhel/backup.py
+++ b/convert2rhel/backup.py
@@ -69,7 +69,7 @@ class ChangedRPMPackagesController(object):
         """Install packages locally available."""
 
         if not pkgs_to_install:
-            loggerinst.info("No package to install")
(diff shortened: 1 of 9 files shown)


r0x0d (approved):

r0x0d:
Trying to re-run the integration tests again to see if the failures are related to the changes (probably not)

abadger (approved):
I agree that the messages read better with these changes.  Approved.

PR #504, opened 2022-05-24
Add basic sanity checks

danmyway:
As a part of morf tool decommission add some basic sanity checks to the
integration tests using tmt.
- check for manpage
- basic smoke test

Rename the test scenario previously named as check-privileges and add
mentioned above.

### DIFF
--- a/plans/tier0.fmf
+++ b/plans/tier0.fmf
@@ -1,6 +1,10 @@
 discover+:
     tier: 0
 
+/basic_sanity_checks:
+  discover+:
+    test: basic-sanity-checks
+
 /check_user_response:
   discover+:
     test: check-user-response
--- a/tests/integration/tier0/basic-sanity-checks/main.fmf
+++ b/tests/integration/tier0/basic-sanity-checks/main.fmf
@@ -1,0 +1,6 @@
+summary: user privileges, manpage exists, basic smoke
+
+tier: 0
+
+test: |
+  pytest -svv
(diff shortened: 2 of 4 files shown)


mkluson (dismissed):

r0x0d:
@danmyway the file `tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks` has no `.py` extension, is this on purpose?

r0x0d (dismissed):

mkluson (approved):

kokesak (dismissed):
Looks good, @danmyway just try if using the `shell` function works also for adding the new user. If not than you can leave it as it is.

danmyway:
> Looks good, @danmyway just try if using the `shell` function works also for adding the new user. If not than you can leave it as it is.

Yep, I'll try if calling the `shell()` will work as expected, if so, I'll change it accordingly.
IIRC we used the `os.system()` to do it in a 'more pythonic way', although as resolved [here](https://github.com/oamg/convert2rhel/pull/356#discussion_r853218047), calling the `shell()` is okay too.

danmyway:
No issues with calling the `shell()` function.
All the `os.system()` calls were changed.

kokesak (approved):
LGTM, the failing OL8 tests are due to satellite content that is missing for 8.6

r0x0d (approved):

danmyway:
The failing tests were the one with the satellite issue now resolved, it should pass now, @r0x0d

kokesak:
All passed. Good job @danmyway!

PR #505, opened 2022-05-31
[RHELC-600] Set gpgcheck to one for ubi repos.

abadger:
We setup UBI and centos6 package repositories for downloading the
subscription-manager packages.  It is good practice to set gpgcheck=1
for any yum repositories.  gpgcheck=0 shouldn't be exploitable since
we're hardcoding the https:// urls to the ubi repos on our CDN but it is
part of a good defense in depth.

Fixes: https://issues.redhat.com/browse/RHELC-600

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -35,7 +35,7 @@ _CENTOS_6_REPO_CONTENT = (
     "[centos-6-contrib-convert2rhel]\n"
     "name=CentOS Linux 6 - Contrib added by Convert2RHEL\n"
     "baseurl=https://vault.centos.org/centos/6/contrib/$basearch/\n"
-    "gpgcheck=0\n"
+    "gpgcheck=1\n"
     "enabled=1\n"
 )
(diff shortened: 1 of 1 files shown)


r0x0d:
@abadger, do you think it's worth to provide some tests to guarantee that we always have these repos with the correct information? Maybe starting with checking the `gpgcheck` field in the repofile.

abadger:
@r0x0d I'm not sure.  I think there's a point where adding a test doesn't do much good.  In this case, there's no logic that we're verifying, just data.  And if someone takes the trouble to change the data, they probably would change the test case too?  It also wouldn't stop us from adding a new  file which had `gpgcheck=0`

In the past, when I had things like this, I would write a "sanity test"... Something that would grep the whole repository looking for patterns that we wanted to avoid.  We don't have anything like that right now (Running pylint and flake8 on the repository is probably the closest analogue).  Do we want to start doing that?

abadger:
@r0x0d I've opened this issue: https://issues.redhat.com/browse/RHELC-614 to look into adding sanity checks for convert2rhel.

PR #514, opened 2022-06-09
Improve python 2 3 compatibility

abadger:
Last thing we're testing


abadger:
Okay, this does not work.  It looks like codecov.io is not honoring  OVERRIDE_COMMIT and COMMIT_PARENT .  We'll just go ahead and configure the checkout action to not do a merge for the coverage tests.

PR #516, opened 2022-06-14
[RHELC-560, RHELC-166] Create a BackupController Framework and a rpm gpg key plugin

abadger:
The BackupController framework is a pluggable way to make changes to the
system that we can revert at a later date.  It consists of
a BackupController object which we instantiate once per convert2rhel
run.  The API of the BackupController consists of:

* a push() method.  When a RestorableChange is pushed onto the
  BackupController's stack, the BackupController will call the
  RestorableChange's enable() method and then store the change on the
  BackupController's stack.
* a pop() method.  This will pop the last RestorableChange off of the
  stack, call the change's restore() method, and then return the change
  to the caller.
* a pop_all() method. This will call the restore() method on all of the
  RestorableChanges() and then return the list of all the changes to the
  caller.

RestorableChanges are objects which implement the following API:
* An enable() method.  When this is called, it should backup any data
  that would be needed to rollback the change and then will make the
  requested change.
* A restore() method.  When this is called, it should undo the changes
  made by the enable() method.

This commit also include a RestorableRpmKey class which implements
RestorableChange for rpm gpg keys.  Upon enable(), the class will record
whether the key is already present in the rpm database.  If it isn't, it
will then be added.  On restore(), it will check whether the key was
installed before enable() was called.  If it wasn't, it will uninstall
that key.

pkghandler.install_gpg_keys() has been enhanced to use the
RestorableRpmKey() and the step that installs gpg keys has been moved
into the Pre-ponr section of the code.


abadger:
This implementation is not quite complete yet:
* Need a function which checks whether a keyid is already present in the rpm database.
* Need unittests for BackupController and RestorableChange
* Need unittests for RestorableRpmKey
* Check whether unittests for install_gpg_key() need adjusting
* Write unittest for find_keyid()
* Decide where to catch the exceptions that from_keyid() can raise.  To be like the rest of the code, probably catch this in install_gpg_key() and use logger.critical() to bubble it up to the user.

abadger:
Okay, if this passes tests, it is ready for review.

kokesak:
Please rebase to upstream as well.

abadger:
Rebased and ready for review.

convert2rhel/backup.py
```diff
@@ -104,6 +108,150 @@ def restore_pkgs(self):
...
+
+    def restore(self):
+        if not self.enabled or self.previous_state in (None, "installed"):
+            self.enabled = False
+            return
+
+        # The self.previous_state == "not_installed" case
+        utils.run_subprocess(["rpm", "-e", "gpg-pubkey-%s" % self.keyid])
+
+        super(RestorableRpmKey, self).restore()
```
r0x0d, line 252: @abadger I see that this whole class is missing the docstrings here. I can see why `restore` and `enabled` might not have one, it makes sense. But for the class itself and the `installed`, I would say that it should be good to at least have a small docstring to indicate what it is.

convert2rhel/backup.py
```diff
@@ -104,6 +108,150 @@ def restore_pkgs(self):
...
+        """
+        Restore the state of the system.
+        """
+        self.enabled = False
+
+
+class RestorableRpmKey(RestorableChange):
+    def __init__(self, keyfile):
+        super(RestorableRpmKey, self).__init__()
+        self.previous_state = None
```
r0x0d, line 207: So, I'm not a big fan of having states as a normal string in the code and not having a good way to verify/assert that the state is actually correct… I don't know how you feel about it, but to me, it seems that this can be error-prone so easily. 

The reason is that I'm imagining that someone could change these states at will in any other PR and introduce new ones, or just mess up the name of the state and lose a lot of time debugging or figuring out why. Another possibility would be someone using this class as a subclass of something else and introducing other states, and we would need to keep track of that in two separate classes. _I know, I went too far in the imagination section, but…_

Since we have two states for now (`not_installed` | `installed`), it seems to be that we could replace that with a boolean property just fine.  Maybe, something like that could work: 

```suggestion
        self.is_installed = False
```

or

```suggestion
        self.is_rpm_installed = False
```

I would say that if we want to keep the states as strings here, I would come up with something to assert that the values can't be different from what we propose/expect here… 

Let me know what you think about this… If you don't think that it makes sense, feel free to just resolve the comment!

convert2rhel/unit_tests/backup_test.py
```diff
@@ -398,3 +399,253 @@ def test_restorable_package_backup(pretend_os, is_eus_system, has_internet_acces
...
+    gpg_key = os.path.realpath(
+        os.path.join(os.path.dirname(__file__), "../data/version-independent/gpg-keys/RPM-GPG-KEY-redhat-release")
+    )
+
+    @pytest.fixture
+    def rpm_key(self):
+        return backup.RestorableRpmKey(self.gpg_key)
+
+    def test_init(self):
+        rpm_key = backup.RestorableRpmKey(self.gpg_key)
```
r0x0d, line 544:
```suggestion
        rpm_key = self.rpm_key
```

r0x0d, line 544: Changing this to use the fixture would be a suitable change? 

```suggestion
    def test_init(self, rpm_key):
```

convert2rhel/backup.py
abadger: I think that Enums would be preferred in the general case of more than three values but we don't have access to Enums until we can run only on Python3 (python-3.4 specifically).  Until then, for more than three values, I think strings should be preferred.  They do suffer from potential typos as you point out, but the other side of that coin is that, they are easily extensible.  And strings are just better than any of the other choices until Enums were added.

In this case, I think you're right to think we have less values so a boolean might be okay.  However:
* This is actually a tristate, not a True|False.  The third state is "unknown". (represented by None).
* I don't think we need to use some sort of temporal reference in the variable name.  This is not about whether the gpg key is imported now.  it's about whether the gpg  key was imported prior to our performing any work on the system.

So maybe something like `self.previously.installed = (True|False|None)`

convert2rhel/unit_tests/backup_test.py
abadger: I'm not sure.  Here's the pro and con that I'm struggling with:
* Pro: We don't have to repeat ourselves since the rpm_key fixture does exactly what this code is doing.  If we ever change the parameters to `backup.RestorableRpmKey.__init__()`, we would only have one place to change it, in the fixture.
* Con: This function is specifically for testing `RestorableRpmKey.__init__()`.  Relying on the fixture to call that function seems like we're relying on the implementation of the fixture rather than being sure that we're actually calling the function that we're testing by making that call explicit in our test function.

I think I lean towards making the call explicit here.  Does that sound okay to you?

convert2rhel/backup.py
abadger: Docstrings now added.

convert2rhel/backup.py
abadger: Switched to `self.previously_installed` and refactored the logic in `restore()` to be clearer.

bocekm:
Rebased to get the changes from https://github.com/oamg/convert2rhel/pull/591.

convert2rhel/backup.py
r0x0d: Like the modifications! Thanks for the explanation as well.

convert2rhel/unit_tests/backup_test.py
r0x0d: Hey, sorry for the late response! Yes, that sounds totally fine by me.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1921,6 +1899,49 @@ def test_find_pkg_names_no_names(output, message):
     assert pkghandler.find_pkg_names(output, message) == set()
 
 
+class TestInstallGpgKeys(object):
+    data_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "data", "version-independent"))
```
r0x0d, line 1903: Just a minor suggestion here: 

We could change it to be like the other examples in this PR: 

```suggestion
    data_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "../data/version-independent"))
```

No need to change this piece of code, just pointing out that we could make it more identical to the other ones.

r0x0d (commented):
Thanks for this work, @abadger! 

I couldn't find anything that would stop me from merging this PR, except the fact that I couldn't see anywhere mentioning that we need an integration test for this change… (haven't looked the jira tickets).

Anyway, I made two small comments that it is up to you if you think it is worth it to address in this PR.

convert2rhel/unit_tests/pkghandler_test.py
abadger: I'm conflicted on this one.  It's better practice to use `os.path.join()` to join paths since the path separator is different on different platforms.  However, convert2rhel is only going to be run on linux systems so we don't strictly need to do this.  And you're right that many of our current uses of `os.path.join()` have `/` hardcoded into the paths that are passed in.

So maybe we should use hardcoded `/` for now and cleanup all of of the places where `/` occurs in `os.path.joins()` as a later PR.

convert2rhel/backup.py
```diff
@@ -104,6 +108,156 @@ def restore_pkgs(self):
...
+class BackupController(object):
+    """
+    Controls backup and restore for all restorable types.
+
+    This is the second version of a backup controller.  It handles all types of things that
+    convert2rhel will change on the system which it can restore in case of a failure before the
+    Point-of-no-return (PONR).
+
+    The basic interface to this is a stack.  When a Restorable is pushed onto the stack, it is
+    backed up.  When it is popped off of the stack, it is restored.
```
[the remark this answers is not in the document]
abadger: It is a LIFO but I can add this information.

convert2rhel/main.py
```diff
@@ -270,6 +272,7 @@ def rollback_changes():
     pkghandler.versionlock_file.restore()
     system_cert = cert.SystemCert()
     system_cert.remove()
+    backup.backup_control.pop_all()
```
abadger, line 275:
```suggestion
    try:
        backup.backup_control.pop_all()
    except IndexError as e:
        if e.args[0] == "No backups to restore":
            loggerinst.info("During rollback there were no backups to restore")
        else:
            raise
```

convert2rhel/backup.py
```diff
@@ -104,6 +108,156 @@ def restore_pkgs(self):
...
+        self._restorables.append(restorable)
+
+    def pop(self):
+        """
+        Restore and then return the last RestorableChange added to the Controller.
+
+        :returns: RestorableChange object that was last added.
+        :raises IndexError: If there are no RestorableChanges currently known to the Controller.
+        """
+        restorable = self._restorables.pop()
```
abadger, line 146:
```suggestion
        try:
            restorable = self._restorables.pop()
        except IndexError as e:
            # Use a more specific error message
            args = list(e.args)
            args[0] = "No backups to restore"
            e.args = tuple(args)
            raise e

```

convert2rhel/unit_tests/backup_test.py
```diff
@@ -398,3 +399,253 @@ def test_restorable_package_backup(pretend_os, is_eus_system, has_internet_acces
...
+        assert popped_restorable1 is restorable1
+        assert popped_restorable2 is restorable2
+        assert popped_restorable3 is restorable3
+
+        assert restorable1.called["restore"] == 1
+        assert restorable2.called["restore"] == 1
+        assert restorable3.called["restore"] == 1
+
+    def test_pop_when_empty(self, backup_controller):
+        with pytest.raises(IndexError, match="pop from empty list"):
```
abadger, line 468:
```suggestion
        with pytest.raises(IndexError, match="No backups to restore"):
```

convert2rhel/unit_tests/backup_test.py
```diff
@@ -398,3 +399,253 @@ def test_restorable_package_backup(pretend_os, is_eus_system, has_internet_acces
...
+        popped_restorables = backup_controller.pop_all()
+        backup_controller.push(restorable2)
+
+        assert len(popped_restorables) == 1
+        assert popped_restorables[0] == restorable1
+        assert len(backup_controller._restorables) == 1
+        assert backup_controller._restorables[0] is restorable2
+
+    def test_pop_all_when_empty(self, backup_controller):
+        with pytest.raises(IndexError, match="pop from empty list"):
```
abadger, line 505:
```suggestion
        with pytest.raises(IndexError, match="No backups to restore"):
```

convert2rhel/backup.py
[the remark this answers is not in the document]
abadger: Information added.

convert2rhel/backup.py
```diff
@@ -104,6 +108,156 @@ def restore_pkgs(self):
...
+
+        :returns: List of RestorableChange objects that were known to the Controller.
+        :raises IndexError: If there are no RestorableChanges currently known to the Controller.
+
+        After running, the Controller object will not know about any RestorableChanges.
+        """
+        restorables = self._restorables
+
+        if not restorables:
+            raise IndexError("pop from empty list")
```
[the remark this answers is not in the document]
abadger: I've changed the error message both here and in `BackupController.pop()`.  I also added an exception handler where this is called in `main.py` to log this as `info` rather than letting it propogate as an error.  If rollback is somehow triggered so early that no changes have been registered with `BackupController`, then it shouldn't be an error.

convert2rhel/backup.py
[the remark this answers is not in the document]
SpyTec: Woops, you're right. Thanks!

SpyTec (approved):

SpyTec:
@r0x0d if it looks good to you and it's tested that it works in a VM. Feel free to move to Dev Complete

r0x0d:
> @r0x0d if it looks good to you and it's tested that it works in a VM. Feel free to move to Dev Complete

Right. Will test it right now.

r0x0d:
Tested manually in a Vagrant VM. Everything seems to be working just fine.

r0x0d (approved):
A couple of codecov misses, but nothing that should stop this PR to get merged.

PR #525, opened 2022-06-20
The new version of ansible-core has removed support for toplevel include

abadger:
import_playook is almost the same thing.  Replace usage of toplevel
include with import_playbook.


kokesak (approved):
LGTM

kokesak:
I reviewed both failing tests. They are failing due to network problems when registering the system -> infrastructure issue. No need to re-run tests. 
Thanks @abadger for the quick fix.

PR #529, opened 2022-06-23
[RHELC-642] Add integration test that install custom kernel

kokesak:
Install custom kernel on the system that is not signed by the running OS and verify that the system inhibits/reports warning.

Tickets that are little bit related to this: https://github.com/oamg/convert2rhel/pull/389, https://github.com/oamg/convert2rhel/pull/424

Bug found: https://issues.redhat.com/browse/RHELC-642

### DIFF
--- a/tests/integration/tier0/custom-kernel/install_custom_kernel.py
+++ b/tests/integration/tier0/custom-kernel/install_custom_kernel.py
@@ -1,0 +1,41 @@
+import platform
+
+from envparse import env
+
+
+def test_install_custom_kernel(shell):
+    # We install CentOS kernel on Oracle Linux and vice versa to mimic the custom kernel
+    # that is not signed by the running OS official vendor
+    system_version = platform.platform()
+
+    if "centos-7" in system_version:
+        # Install dependency for kernel-uek
+        assert (
+            shell(
+                "yum install https://yum.oracle.com/repo/OracleLinux/OL7/latest/x86_64/getPackage/linux-firmware-20200124-999.4.git1eb2408c.el7.noarch.rpm -y"
+            ).returncode
+            == 0
+        )
+        assert (
+            shell(
+                "yum install https://yum.oracle.com/repo/OracleLinux/OL7/UEKR6/x86_64/getPackage/kernel-uek-5.4.17-2011.0.7.el7uek.x86_64.rpm -y"
+            ).returncode
+            == 0
+        )
+        shell("grub2-set-default 'CentOS Linux (5.4.17-2011.0.7.el7uek.x86_64) 7 (Core)'")
+    elif "oracle-7" in system_version:
+        assert (
+            shell(
+                "yum install http://mirror.centos.org/centos/7/os/x86_64/Packages/kernel-3.10.0-1160.el7.x86_64.rpm -y"
+            ).returncode
+            == 0
+        )
+        shell("grub2-set-default 'Oracle Linux Server 7.9, with Linux 3.10.0-1160.el7.x86_64'")
+    elif "centos-8.4" in system_version:
+        assert (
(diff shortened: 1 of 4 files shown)


kokesak:
Once the https://issues.redhat.com/browse/RHELC-642 is resolved we can re-run the tests

tests/integration/tier0/custom-kernel/test_cutsom_kernel.py
```diff
@@ -0,0 +1,18 @@
+import platform
+
+from envparse import env
+
+
+def test_custom_kernel(convert2rhel):
+    # Run c2r with --variant option
```
bocekm, line 7:
```suggestion
    # Run c2r with --variant option
```
Isn't this a copy paste error?

tests/integration/tier0/custom-kernel/install_custom_kernel.py
```diff
@@ -0,0 +1,41 @@
...
+            shell(
+                "yum install https://yum.oracle.com/repo/OracleLinux/OL7/latest/x86_64/getPackage/linux-firmware-20200124-999.4.git1eb2408c.el7.noarch.rpm -y"
+            ).returncode
+            == 0
+        )
+        assert (
+            shell(
+                "yum install https://yum.oracle.com/repo/OracleLinux/OL7/UEKR6/x86_64/getPackage/kernel-uek-5.4.17-2011.0.7.el7uek.x86_64.rpm -y"
+            ).returncode
+            == 0
```
r0x0d, line 23: @kokesak does this still work? Because we have a check in Convert2RHEL to prevent UEK kernels to pass

kokesak: @r0x0d if the check works, then the conversion should inhibit. We can keep the UEK kernel install here as well as add a scenario where the non-uek kernel is installed on centos to test that as well.

r0x0d: Right... I'm having some troubles to reproduce this locally because of those checks. Do you have a step-by-step on how to do it? 

Tried with the UEK and then the conversion halts on that check. Tried OL7 kernel, but it is the same version of CentOS 7 (no downgrades available for both of them)

kokesak: Hmm, maybe I had similar problem when installing OL7 kernel on CentOS so I used the UEK. I'll try it and let you know

r0x0d: Okay, I tried to follow the steps in this test, but didn't work because of the checks we have in place for UEK

r0x0d: @kokesak tried with this one and it worked: https://yum.oracle.com/repo/OracleLinux/OL7/latest/x86_64/getPackage/kernel-3.10.0-1160.76.1.0.1.el7.x86_64.rpm

Maybe we can adapt the test to include this kernel instead of the UEK

r0x0d: Also, @kokesak, do you want to merge this PR with mine, so we get everything in one place? 

I can cherry-pick your changes here and merge into mine if you think that is worth it.

kokesak: @r0x0d nice that it works. It will be nice to have it under the same PR. I will add the commit to your PR myself and close this one, you dont have to do it.

r0x0d:
@kokesak since we merged this in the PR#623, can we close this one?

kokesak:
@r0x0d yeah sure, closing this one

PR #534, opened 2022-07-01
Playbook fixes

### LINKED ISSUE #532: Incorrect logic in setup-convert2rhel.yml when skip_os_version_check is true
In the OS Validation section  it isn't possible to skip the check due to the current logic
```
    - name: Validate Release Version (6.10, 7.9, 8.5)
      fail:
        msg: "Distribution version must be 6.10, 7.9, or 8.5"
      when: (ansible_distribution_version != '6.10' and ansible_distribution_version != '7.9'
             and ansible_distribution_version != '8.5') or skip_os_version_check|bool
```

Should be

```
    - name: Validate Release Version (6.10, 7.9, 8.5)
      fail:
        msg: "Distribution version must be 6.10, 7.9, or 8.5"
      when: (ansible_distribution_version != '6.10' and ansible_distribution_version != '7.9'
             and ansible_distribution_version != '8.5') and not skip_os_version_check

```

### LINKED ISSUE #533: Invalid data passed to loop in setup-convert2rhel
Error message

```
TASK [Comment mirrorlist in all repos.] ***************************************************************************
task path: ./setup-convert2rhel.yml.new:37
fatal: [leapp2rhel]: FAILED! => {
    "msg": "Invalid data passed to 'loop', it requires a list, got this instead: centos8_repos. Hint: If you passed a list/dict of just one element, try adding wantlist=True to your lookup invocation or use q/query instead of lookup."
}

Using Ansible 2.9.27 on Fedora 35

```

abadger:
Fixes two playbook issues.  These issues prevented the ansible playbook from converting centos machines.  People could still run convert2rhel from the command line.

*   Fix syntax of looping over centos8_repos. As reported in #533, the loop syntax we have in the ansible playbook is
    not quite right.  Instead of simply writing the variable name as we have it now, we are responsible for asking for it to be expanded in the loop construct (using "{{ }}").

    Fixes #533

*  Fix setup playbook feature allowing user to skip os version check. As reported in #532, the logic in the ansible playbook was to fail when either the version check showed an unhandled version OR the skip_os_version_check was set to true. This meant that the toggle would either check the version or the playbook will fail.

    By using AND in that logic, we get the behaviour we want.  Either the toggle will check the version or the playbook's check will succeed

    It is confusing because it is a `fail:` task.  So a true value in the  conditional WILL fail the playbook while a false value in the     conditional means that the playbook WILL NOT fail here.

    Fixes #532

https://issues.redhat.com/browse/RHELC-650

### DIFF
--- a/ansible/setup-convert2rhel.yml
+++ b/ansible/setup-convert2rhel.yml
@@ -30,7 +30,7 @@
       fail:
         msg: "Distribution version must be 6.10, 7.9, or 8.5"
       when: (ansible_distribution_version != '6.10' and ansible_distribution_version != '7.9'
-             and ansible_distribution_version != '8.5') or skip_os_version_check|bool
+             and ansible_distribution_version != '8.5') and skip_os_version_check|bool
 
     - name: Patch repositories if dealing with CentOS8.
       block:
--- a/ansible/setup-convert2rhel.yml
+++ b/ansible/setup-convert2rhel.yml
@@ -39,14 +39,14 @@
           path: /etc/yum.repos.d/{{ item }}
           regexp: ^mirrorlist
           replace: "#mirrorlist"
-        loop: centos8_repos
+        loop: "{{ centos8_repos }}"
 
       - name: Switch repo to vault.centos.org.
         replace:
           path: /etc/yum.repos.d/{{ item }}
           regexp: "#baseurl=http://mirror.centos.org"
           replace: baseurl=https://vault.centos.org
-        loop: centos8_repos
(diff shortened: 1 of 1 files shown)


abadger:
note:  The ansible playbooks are not shipped in the rpms to customers so this change does not need an integration test and does not appear in the release notes.

r0x0d (approved):
Looks good to me. 

Clean and clear changes.

r0x0d:
Thanks, @abadger, for the quick fix on this.

PR #539, opened 2022-07-14
Update 6.x config files to have the eus_rhsm_repoids entry.

abadger:
Like RHEL 7 clones, we need an empty entry for this since there are no
eus repos but the code looks for it.


bocekm (approved):

r0x0d (approved):
Thanks for spotting that out, @abadger!

PR #540, opened 2022-07-15
[RHELC-597] Register with subscription-manager via dbus

abadger:
Part of [RHELC-597](https://issues.redhat.com/browse/RHELC-597)

---

Using the dbus API to register, we're able to stop passing the
activation key over the commandline.

This also lets us avoid pexpect for invoking subscription-manager when
passing the password.

Note that this requires that dbus is installed and running and that the
dbus-python/python-dbus packages are installed.  This package is
available on RHEL6-RHEL9 clones (all that we care about) but is a new
dependency.  The subscription-manager rpm already requires dbus-python so when we install that, we should be installing sbud as well.  RHEL7 and above use systemd which requires that dbus be running.  So RHEL6 is the only place where we might have to install and start up DBus.  Since the supported migration would include upgrading RHEL6 to RHEL7 soon after conversion, we might consider it not worthwhile to stop and remove dbus in this case.  Still have to consider rollback.

This code has been successfully tested on centos8 with activation key.  There are quite a few questions that I'm ready to take some feedback on before going further:

## TODOs

- [ ] DBus Daemon
  - [x] For this release and PR we will 
    - [x] Add an rpm requires for the dbus daemon package
    - [x] Check whether dbus is running:
      - [x] systeminfo.py will record whether dbus is running when convert2rhel starts
      - [x] checks.py will have a new check that fails if the dbus daemon is not running. 
    - [x] If it's not running error and ask the user to start it
  - [ ] I will open a jira ticket to make this behaviour better for the next release:
    - [ ] If the daemon is not running, start it.
    - [ ] During rollback, if the daemon was started by us, stop it.
- [ ] Remove RegistrationCommand.args
  - [x] Leave this in for this release as it isa handy way to log what we are using to talk to subscription-manager 
  - [ ] Will open a jira ticket to remove the function in a later PR.
    - [ ] Will change the logging to enumerate what we call subscription-manager's DBus API with
- [x] Workaround python-dbus broken logging (it initializes the root logger with a Handler that outputs to stderr. This causes our messages to be displayed twice and worse, if the user gives us data which can provoke an error when marshalling data to send to subscription-manager over DBus, then DBus will log the credentials in clear to stderr.)
  - [x] Added a NullHandler to the root logger before importing DBus which should stop DBus from adding the stderr handler.
  - [ ] Will open a jira ticket to move the NullHandler addition to another location next release.  @r0x0d is working on an early-initialization step for the yum merge transaction PR which we can use for this as well.
- [ ] Open jira ticket to determine org even for username + password auth
  - Resulted from speaking with Pino.
  - One customer bz: https://bugzilla.redhat.com/show_bug.cgi?id=1946309
  - Michal said: "We still can't handle this use case. For that reason we state in the product documentation that Satellite users should use the activation key/organization combination."
  - This code should be easy to update to get the organization: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/subscription.py#L208
- [ ] Open a jira ticket to enter rollback if a user hits Ctrl-C during registration.  As coded right now, it will retry three times before entering rollback.
- [ ] Open a jira ticket to look into why using DBus to edit the configuration is not working at the end of `RegistrationCommand.__call__()`.  Looks to be dbus policy-related.
- [ ] Open a jira ticket to figure out:
  - Bug in subscription-manager that prevents force from working when we register when the system is already registered.  WIll this be fixed and can we drop the code that works around it?
  - Why we need to stop the rhsm service. Can we limit the rhsm service stop to only some platforms (RHEL7 and older)? 
  - Could we use reload instead of stop?  Do we need to stop even if unregister succeeds?
  - Some experimentation shows that setting register_opts force: True is having an effect.  Maybe the bug was that we weren't setting the DbusDictionary signature to `sv` and the sub-man server really is doing the right thing?
- [ ] Open a jira ticket to make sub-man config rollback aware. 
- [x] What should the return value of RegistrationCommand be?
  - [x] With Pino and Michal's feedback I lean towards RegistrationCommand returning nothing.
  - [x] On error, raise an exception.
  - [x] Be sure to catch the exception in our registration retry code.
- [x] If we fallback to cli, do we still allow activation_key even though it is insecure?
  - Decided not to fallback to cli since that would mean two code paths to test. 
- [x] Similar questions for python-dbus and dbus-python being installed.
  * The bindings are available on pypi as `dbus-python`.
  - [x] This is now required by the rpm.
- [x] Do we want to use DBUS for registering via passwords or only for activation key?
  - Decided to use DBus for registering via both passwords and activation key.
- [x] In _call_via_dbus(), we need to add more logging.  Call_via_cli lets run_subprocess()/run_in_pty() do most of its logging but we don't have that in _call_via_dbus()
- [x] integration tests to check that activation_key is not visible to the process list (because we're no longer calling subscription-manager with activation_key on the cli.
- [x] A few questions about schema, username, and password for pino.
  - [x] Open a jira ticket to look into what subscription-manager cli does with username:password in the --server-url and change our code to do what it does (our code currently ignores )
  - [x] Pino let us know that any username and password supplied to the subscription-manager cli via --serverurl is ignored.  So we can safely ignore those too.
- [x] unittests
  - [x] Fix current tests that fail after moving to dbus
  - [x] Write new tests to cover new code paths. 
- [x] We're going to keep dbus-python as a Requires in the rpm package so we don't need to conditionalize import
  - [ ]  Do we need dbus-python it in some requirements files?  we don't have pexpect in any requirements.

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -66,6 +66,7 @@ def perform_pre_checks():
     check_rhel_compatible_kernel_is_used()
     check_package_updates()
     is_loaded_kernel_latest()
+    check_dbus_is_running()
 
 
 def perform_pre_ponr_checks():
@@ -698,3 +699,21 @@ def is_loaded_kernel_latest():
                 "If you wish to ignore this message, set the environment variable "
                 "'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' to 1." % package_to_check
             )
+
+
+def check_dbus_is_running():
+    """Error out if we need to register with rhsm and the dbus daemon is not running."""
+    logger.task("Prepare: Check that DBus Daemon is running")
+
+    if tool_opts.no_rhsm:
+        logger.info("Skipping the check because we have been asked not to subscribe this system to RHSM.")
+        return
+
+    if system_info.dbus_running:
+        logger.info("DBus Daemon is running")
+        return
+
+    logger.critical(
+        "Could not find a running DBus Daemon which is needed to register with subscription manager.\n"
+        "Please start dbus using `systemctl start dbus` or (on CentOS Linux 6), `service messagebus start`"
+    )
--- a/convert2rhel/i18n.py
+++ b/convert2rhel/i18n.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright(C) 2021 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+"""
+One location to keep all locale related information.
+
+convert2rhel is not currently localized but it does interact with other systems which are. this
+module provides a single place to keep that information.  If we decide to localize convert2rhel in
+the future, this file should point us towards all the locations that we may need to change.
+"""
+
+#
+# Display locales
+#
+
+# These locales are about what we want to display to the user. If we decide to localize, these
+# should be updated to retrieve the locale setting (Using the locale module from the python stdlib).
+CONVERT2RHEL_LOCALE = "C"
+# Subscription-manager's DBus API has its own method to set the locale. I'm not sure whether we
+# would want these to obey the Display locale of the programming locale. We're likely to forward
+# the error messages on to the user so display locale feels like the best fit but sometimes error
+# messages are best left un-localized so that they are easier to perform a web-search for.
+# We can make a decision if/when we're ready to localize convert2rhel.
+SUBSCRIPTION_MANAGER_LOCALE = CONVERT2RHEL_LOCALE
+
+#
+# Programming Locale
+#
+
+# This is the locale to use for programs that we run in a subprocess and then have to interpret what
+# they have done by screenscraping their output.
+SCREENSCRAPED_LOCALE = "C"
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -22,7 +22,13 @@ import re
 from collections import namedtuple
 from time import sleep
 
-from convert2rhel import backup, pkghandler, utils
+import dbus
+import dbus.connection
+import dbus.exceptions
+
+from six.moves import urllib
+
+from convert2rhel import backup, i18n, pkghandler, utils
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
 
@@ -57,13 +63,32 @@ _UBI_8_REPO_CONTENT = (
     "enabled=1\n"
 )
 _UBI_8_REPO_PATH = os.path.join(_RHSM_TMP_DIR, "ubi_8.repo")
+
+# We need to translate config settings between names used for the subscription-manager DBus API and
+# names used for the RHSM config file.  This is the mapping for the settings we care about.
+CONNECT_OPT_NAME_TO_CONFIG_KEY = {
+    "host": "server.hostname",
+    "port": "server.port",
+    "handler": "server.prefix",
+}
+
 MAX_NUM_OF_ATTEMPTS_TO_SUBSCRIBE = 3
 # Using a delay that could help when the RHSM/Satellite server is overloaded.
 # The delay in seconds is a prime number that roughly doubles with each attempt.
 REGISTRATION_ATTEMPT_DELAYS = [5, 11, 23]
+# Seconds to wait for Registration to complete over DBus. If this timeout is exceeded, we retry.
+REGISTRATION_TIMEOUT = 180
 
 
 _SUBMGR_PKG_REMOVED_IN_CL_85 = "subscription-manager-initial-setup-addon"
+
+
+class UnregisterError(Exception):
+    """Raised with problems unregistering a system."""
+
+
+class StopRhsmError(Exception):
+    """Raised with problems stopping the rhsm daemon."""
 
 
 def subscribe_system():
@@ -80,10 +105,6 @@ def subscribe_system():
 def unregister_system():
     """Unregister the system from RHSM."""
     loggerinst.info("Unregistering the system.")
-    if tool_opts.keep_rhsm:
-        loggerinst.info("Skipping due to the use of --keep-rhsm.")
-        return
-
     # We are calling run_subprocess with rpm here because of a bug in
     # Oracle/CentOS Linux 7 in which the process always exits with 1 in case of a
     # rollback when KeyboardInterrupt is raised.  To avoid many changes and
@@ -96,14 +117,11 @@ def unregister_system():
     if ret_code != 0:
         loggerinst.info("The subscription-manager package is not installed.")
         return
+
     unregistration_cmd = ["subscription-manager", "unregister"]
     output, ret_code = utils.run_subprocess(unregistration_cmd, print_output=False)
     if ret_code != 0:
-        loggerinst.warning(
-            "System unregistration failed with return code %d and message:\n%s",
-            ret_code,
-            output,
-        )
+        raise UnregisterError("System unregistration result:\n%s" % output)
     else:
         loggerinst.info("System unregistered successfully.")
 
@@ -120,33 +138,98 @@ def register_system():
                 attempt + 1,
                 MAX_NUM_OF_ATTEMPTS_TO_SUBSCRIBE,
             )
+
         loggerinst.info(
             "%sRegistering the system using subscription-manager ...",
             attempt_msg,
         )
 
-        registration_cmd = RegistrationCommand.from_tool_opts(tool_opts)
-        output, ret_code = registration_cmd()
-        if ret_code == 0:
-            # Handling a signal interrupt that was previously handled by
-            # subscription-manager.
-            if "user interrupted process" in output.lower():
-                raise KeyboardInterrupt
-            return
+        # Force the system to be unregistered
+        # The subscription-manager DBus API has a force parameter but there's
+        # a bug in susbcription-manager where that doesn't take effect.
+        # Explicitly unregister here to workaround that.
+        # Sub-man bug: https://bugzilla.redhat.com/show_bug.cgi?id=2118486
+        loggerinst.info("Unregistering the system to clear the server's state for our registration.")
+        try:
+            unregister_system()
+        except UnregisterError as e:
+            loggerinst.warning(str(e))
 
-        loggerinst.info("System registration failed with return code = %s" % str(ret_code))
-        sleep(REGISTRATION_ATTEMPT_DELAYS[attempt])
-        attempt += 1
-    loggerinst.critical("Unable to register the system through subscription-manager.")
+        # Hack: currently, on RHEL7, subscription-manager unregister is
+        # reporting that the system is not registered but then calling the
+        # subscription-manager Registration API reports that the system is
+        # already registered. We suspect that the test host is being used for
+        # a previous test.  In that test, the system is registered, then
+        # rollback occurs which unregisters the server in some places but it
+        # remains registered in other data.  Then registration fails.
+        #
+        # For the short term, we are going to stop the rhsm service to work
+        # around this issue.  It should restart on its own when needed:
+        loggerinst.info(
+            "Stopping the RHSM service so that registration does not think that the host is already registered."
+        )
+        try:
+            _stop_rhsm()
+        except StopRhsmError as e:
+            # The system really might not be registered yet and also not running rhsm
+            # so ignore the error and try to register the system.
+            loggerinst.info(str(e))
+
+        # Register the system
+        registration_cmd = RegistrationCommand.from_tool_opts(tool_opts)
+
+        try:
+            registration_cmd()
+            loggerinst.info("System registration succeeded.")
+        except KeyboardInterrupt:
+            # When the user hits Control-C to exit, we shouldn't retry
+            raise
+        except Exception as e:
+            loggerinst.info("System registration failed with error: %s" % str(e))
+            sleep(REGISTRATION_ATTEMPT_DELAYS[attempt])
+            attempt += 1
+            continue
+
+        break
+
+    else:  # While-else
+        # We made the maximum number of subscription-manager retries and still failed
+        loggerinst.critical("Unable to register the system through subscription-manager.")
+
+    return None
+
+
+def _stop_rhsm():
+    """Stop the rhsm service."""
+    cmd = ["/bin/systemctl", "stop", "rhsm"]
+    if system_info.version.major <= 6:
+        # On RHEL6, there isn't a service-oriented way to stop rhsm.  It is started on demand so
+        # there isn't an init script to stop it.  If we find we need to stop it, because we're
+        # etting "machine is already registered" errors there, then we'll need to look for
+        # rhsm-service in the process list and send it the TERM signal.
+        loggerinst.info("Skipping RHSM service shutdown on {0} {1}.".format(system_info.name, system_info.version.major))
+        return
+
+    output, ret_code = utils.run_subprocess(cmd, print_output=False)
+    if ret_code != 0:
+        raise StopRhsmError("Stopping RHSM failed with code: %s; output: %s" % (ret_code, output))
+    loggerinst.info("RHSM service stopped.")
 
 
 class RegistrationCommand(object):
-    def __init__(self, activation_key=None, org=None, username=None, password=None, server_url=None):
+    def __init__(
+        self,
+        activation_key=None,
+        org=None,
+        username=None,
+        password=None,
+        rhsm_hostname=None,
+        rhsm_port=None,
+        rhsm_prefix=None,
+    ):
         """
         A callable that can register a system with subscription-manager.
 
-        :kwarg server_url: Optional URL to the subscription-manager backend.
-            Useful when the customer has an on-prem subscription-manager instance.
         :kwarg activation_key: subscription-manager activation_key that can be
             used to register the system. Org must be specified if this was given.
         :kwarg org: The organization that the activation_key is associated with.
@@ -155,12 +238,15 @@ class RegistrationCommand(object):
             Required if password is specified.
         :kwarg password: Password to authenticate with subscription-manager.
             It is required if username is specified.
+        :kwarg rhsm_hostname: Optional hostname of a subscription-manager backend.
+            Useful when the customer has an on-prem subscription-manager instance.
+        :kwarg rhsm_port: Optional port for a subscription-manager backend.
+        :kwarg rhsm_prefix: Optional path element of a subscription-manager backend.
 
         .. note:: Either activation_key and org or username and password must
             be specified.
         """
         self.cmd = "subscription-manager"
-        self.server_url = server_url
 
         if activation_key and not org:
             raise ValueError("org must be specified if activation_key is used")
@@ -178,6 +264,10 @@ class RegistrationCommand(object):
             # No password set
             if not self.activation_key:
                 raise ValueError("activation_key and org or username and password must be specified")
+
+        self.rhsm_hostname = rhsm_hostname
+        self.rhsm_port = rhsm_port
+        self.rhsm_prefix = rhsm_prefix
 
     @classmethod
     def from_tool_opts(cls, tool_opts):
@@ -245,9 +335,15 @@ class RegistrationCommand(object):
 
             registration_attributes["password"] = password
 
-        if tool_opts.serverurl:
-            loggerinst.debug("    ... using custom RHSM URL")
-            registration_attributes["server_url"] = tool_opts.serverurl
+        if tool_opts.rhsm_hostname:
+            loggerinst.debug("    ... using custom RHSM hostname")
+            registration_attributes["rhsm_hostname"] = tool_opts.rhsm_hostname
+        if tool_opts.rhsm_port:
+            loggerinst.debug("    ... using custom RHSM port")
+            registration_attributes["rhsm_port"] = tool_opts.rhsm_port
+        if tool_opts.rhsm_prefix:
+            loggerinst.debug("    ... using custom RHSM prefix")
+            registration_attributes["rhsm_prefix"] = tool_opts.rhsm_prefix
 
         return cls(**registration_attributes)
 
@@ -263,8 +359,18 @@ class RegistrationCommand(object):
         """
         args = ["register", "--force"]
 
-        if self.server_url:
-            args.append("--serverurl=%s" % self.server_url)
+        if self.connection_opts:
+            if self.rhsm_port:
+                netloc = "%s:%s" % (self.rhsm_hostname, self.rhsm_port)
+            else:
+                netloc = self.rhsm_hostname
+
+            prefix = self.rhsm_prefix if self.rhsm_prefix else ""
+            if prefix.startswith("/"):
+                prefix = prefix[1:]
+
+            server_url = urllib.parse.urlunsplit(("https", netloc, prefix, "", ""))
+            args.append("--serverurl=%s" % server_url)
 
         if self.activation_key:
             args.append("--activationkey=%s" % self.activation_key)
@@ -277,28 +383,185 @@ class RegistrationCommand(object):
 
         return args
 
+    @property
+    def connection_opts(self):
+        """
+        This property is a dbus.Dictionary that contains the connection options for RHSM
+        dbus calls.
+
+        Set :attr:`server_url` to affect this value.
+        """
+        connection_opts = {}
+
+        if self.rhsm_hostname:
+            connection_opts["host"] = self.rhsm_hostname
+
+        if self.rhsm_port is not None:
+            connection_opts["port"] = self.rhsm_port
+
+        if self.rhsm_prefix:
+            connection_opts["handler"] = self.rhsm_prefix
+
+        connection_opts = dbus.Dictionary(connection_opts, signature="sv", variant_level=1)
+        return connection_opts
+
     def __call__(self):
         """
-        Run the subscription-manager command.
+        Use dbus to register the system with subscription-manager.
 
-        Wrapper for running the subscription-manager command that keeps
-        secrets secure.
+        Status of dbus on various platforms:
+            * RHEL6:
+                * DBUS-1.2.24 is present but may not be installed.
+                * Install the dbus package and run /etc/rc.d/init.d/messagebus start
+                * dbus-python-0.83.0 is available
+            * RHEL7:
+                * dbus-1.10.24 is installed and run by default
+                * dbus-python-1.1.9 is available
+            * RHEL8 & RHEL9
+                * dbus-1.12.x is installed and run by default
+                * python3-dbus-1.2.x is available
+
+        .. seealso::
+            Documentation for the subscription-manager dbus API:
+            https://www.candlepinproject.org/docs/subscription-manager/dbus_objects.html
         """
-        if self.password:
-            loggerinst.debug(
-                "Calling command '%s %s'" % (self.cmd, " ".join(hide_secrets(self.args)))
-            )  # lgtm[py/clear-text-logging-sensitive-data]
-            output, ret_code = utils.run_cmd_in_pty(
-                [self.cmd] + self.args, expect_script=(("[Pp]assword: ", self.password + "\n"),), print_cmd=False
-            )
-        else:
-            # Warning: Currently activation_key can only be specified on the CLI. This is insecure
-            # but there's nothing we can do about it for now. Once subscription-manager issue:
-            # https://issues.redhat.com/browse/ENT-4724 is implemented, we can change both password
-            # and activation_key to use a file-based approach to passing the secrets.
-            output, ret_code = utils.run_subprocess([self.cmd] + self.args, print_cmd=False)
+        # Note: dbus doesn't understand empty python dicts. Use dbus.Dictionary({}, signature="ss")
+        # if we need one in the future.
+        REGISTER_OPTS_DICT = dbus.Dictionary({"force": True}, signature="sv", variant_level=1)
 
-        return output, ret_code
+        loggerinst.debug("Getting a handle to the system dbus")
+        system_bus = dbus.SystemBus()
+
+        # Create a new bus so we can talk to rhsm privately (For security:
+        # talking on the system bus might be eavesdropped in certain scenarios)
+        loggerinst.debug("Getting a subscription-manager RegisterServer object from dbus")
+        register_server = system_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/RegisterServer")
+        loggerinst.debug("Starting a private DBus to talk to subscription-manager")
+        address = register_server.Start(
+            i18n.SUBSCRIPTION_MANAGER_LOCALE, dbus_interface="com.redhat.RHSM1.RegisterServer"
+        )
+
+        try:
+            # Use the private bus to register the machine
+            loggerinst.debug("Connecting to the private DBus")
+            private_bus = dbus.connection.Connection(address)
+
+            try:
+                if self.password:
+                    loggerinst.info("Registering via username/password: %s" % " ".join(hide_secrets(self.args)))
+                    args = (
+                        self.org or "",
+                        self.username,
+                        self.password,
+                        REGISTER_OPTS_DICT,
+                        self.connection_opts,
+                        i18n.SUBSCRIPTION_MANAGER_LOCALE,
+                    )
+                    private_bus.call_blocking(
+                        "com.redhat.RHSM1",
+                        "/com/redhat/RHSM1/Register",
+                        "com.redhat.RHSM1.Register",
+                        "Register",
+                        "sssa{sv}a{sv}s",
+                        args,
+                        timeout=REGISTRATION_TIMEOUT,
+                    )
+
+                else:
+                    loggerinst.info("Registering via org/activation_key: %s" % " ".join(hide_secrets(self.args)))
+                    args = (
+                        self.org,
+                        [self.activation_key],
+                        REGISTER_OPTS_DICT,
+                        self.connection_opts,
+                        i18n.SUBSCRIPTION_MANAGER_LOCALE,
+                    )
+                    private_bus.call_blocking(
+                        "com.redhat.RHSM1",
+                        "/com/redhat/RHSM1/Register",
+                        "com.redhat.RHSM1.Register",
+                        "RegisterWithActivationKeys",
+                        "sasa{sv}a{sv}s",
+                        args,
+                        timeout=REGISTRATION_TIMEOUT,
+                    )
+
+            except dbus.exceptions.DBusException as e:
+                # Sometimes we get NoReply but the registration has succeeded.
+                # Check the registration status before deciding if this is an error.
+                if e.get_dbus_name() == "org.freedesktop.DBus.Error.NoReply":
+                    # We need to set the connection opts in config before
+                    # checking for registration otherwise we might ask the
+                    # wrong server if the host is registered.
+                    self._set_connection_opts_in_config()
+
+                    if not _is_registered():
+                        # Host is not registered so re-raise the error
+                        raise
+                else:
+                    raise
+                # Host was registered so continue
+            else:
+                # On success, we need to set the connection opts as well
+                self._set_connection_opts_in_config()
+
+        finally:
+            # Always shut down the private bus
+            loggerinst.debug("Shutting down private DBus instance")
+            register_server.Stop(i18n.SUBSCRIPTION_MANAGER_LOCALE, dbus_interface="com.redhat.RHSM1.RegisterServer")
+
+    def _set_connection_opts_in_config(self):
+        """
+        Set the connection opts in the rhsm config.
+
+        The command line subscriptioVn-manager register command sets the
+        config but the DBus API does not.  We need to set it so that
+        subsequent subscription-manager cli calls will use the same connection
+        settings.
+        """
+        # DBus policies are preventing the following from working.
+        # Implement this as calls to the subscription-manager CLI for now.
+        #
+        # config_object = system_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/Config")
+
+        # for option, value in self.connection_opts.items():
+        #     config_object.Set(
+        #         CONNECT_OPT_NAME_TO_CONFIG_KEY[option],
+        #         value,
+        #         i18n.SUBSCRIPTION_MANAGER_LOCALE,
+        #         dbus_interface="com.redhat.RHSM1.ConfigServer",
+        #     )
+        if self.connection_opts:
+            loggerinst.info("Setting RHSM connection configuration.")
+            sub_man_config_command = ["subscription-manager", "config"]
+            for option, value in self.connection_opts.items():
+                sub_man_config_command.append("--%s=%s" % (CONNECT_OPT_NAME_TO_CONFIG_KEY[option], value))
+
+            output, ret_code = utils.run_subprocess(sub_man_config_command, print_cmd=True)
+            if ret_code != 0:
+                raise ValueError("Error setting the subscription-manager connection configuration: %s" % output)
+
+            loggerinst.info("Successfully set RHSM connection configuration.")
+
+
+def _is_registered():
+    """Check if the machine we're running on is registered with subscription-manager."""
+    loggerinst.debug("Checking whether the host was registered.")
+    output, ret_code = utils.run_subprocess(["subscription-manager", "identity"])
+
+    # Registered: ret_code 0 and output like:
+    # system identity: 36dad222-5002-45ba-8840-f41351294213
+    # name: c2r-20220816124728
+    # org name: 13460994
+    # org ID: 13460994
+    if ret_code == 0:
+        loggerinst.debug("Host was registered.")
+        return True
+
+    # Unregistered: ret_code 1 and output like:
+    # This system is not yet registered. Try 'subscription-manager register --help' for more information.
+    loggerinst.debug("Host was not registered.")
+    return False
 
 
 def hide_secrets(args):
@@ -358,7 +621,11 @@ def replace_subscription_manager():
     if not os.path.isdir(SUBMGR_RPMS_DIR) or not os.listdir(SUBMGR_RPMS_DIR):
         loggerinst.critical("The %s directory does not exist or is empty." % SUBMGR_RPMS_DIR)
 
-    unregister_system()
+    try:
+        unregister_system()
+    except UnregisterError as e:
+        loggerinst.warning(str(e))
+
     remove_original_subscription_manager()
     install_rhel_subscription_manager()
 
@@ -643,11 +910,19 @@ def _submgr_enable_repos(repos_to_enable):
 
 def rollback():
     """Rollback subscription related changes"""
+    # Systems using Satellite 6.10 need to stay registered otherwise admins
+    # will lose remote access from the Satellite server.
+    if tool_opts.keep_rhsm:
+        loggerinst.info("Skipping due to the use of --keep-rhsm.")
+        return
+
     try:
         loggerinst.task("Rollback: RHSM-related actions")
         unregister_system()
+    except UnregisterError as e:
+        loggerinst.warning(str(e))
     except OSError:
-        loggerinst.warn("subscription-manager not installed, skipping")
+        loggerinst.warning("subscription-manager not installed, skipping")
 
 
 def check_needed_repos_availability(repo_ids_needed):
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -207,7 +207,9 @@ def _stop_rhsm():
         # there isn't an init script to stop it.  If we find we need to stop it, because we're
         # etting "machine is already registered" errors there, then we'll need to look for
         # rhsm-service in the process list and send it the TERM signal.
-        loggerinst.info("Skipping RHSM service shutdown on {0} {1}.".format(system_info.name, system_info.version.major))
+        loggerinst.info(
+            "Skipping RHSM service shutdown on {0} {1}.".format(system_info.name, system_info.version.major)
+        )
         return
 
     output, ret_code = utils.run_subprocess(cmd, print_output=False)
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -19,6 +19,7 @@ import logging
 import os
 import re
 import socket
+import time
 
 from collections import namedtuple
 
@@ -28,6 +29,9 @@ from convert2rhel import logger, utils
 from convert2rhel.toolopts import tool_opts
 from convert2rhel.utils import run_subprocess
 
+
+# Number of times to retry checking the status of dbus
+CHECK_DBUS_STATUS_RETRIES = 3
 
 # Allowed conversion paths to RHEL. We want to prevent a conversion and minor
 # version update at the same time.
@@ -118,6 +122,7 @@ class SystemInfo(object):
         self.kmods_to_ignore = self._get_kmods_to_ignore()
         self.booted_kernel = self._get_booted_kernel()
         self.has_internet_access = self._check_internet_access()
+        self.dbus_running = self._is_dbus_running()
 
     @staticmethod
     def _get_system_release_file_content():
@@ -358,6 +363,80 @@ class SystemInfo(object):
         :rtype: bool
         """
         return self.releasever in EUS_MINOR_VERSIONS
+
+    def _is_dbus_running(self):
+        """
+        Check whether dbus is running.
+
+        :returns: True if dbus is running.  Otherwise False
+        """
+        retries = 0
+        status = False
+
+        while retries < CHECK_DBUS_STATUS_RETRIES:
+            if self.version.major <= 6:
+                status = _is_sysv_managed_dbus_running()
+            else:
+                status = _is_systemd_managed_dbus_running()
+
+            if status is not None:
+                # We know that DBus is definitely running or stopped
+                break
+
+            # Wait for 1 second, 2 seconds, and then 4 seconds for dbus to be running
+            # (In case it was started before convert2rhel but it is slow to start)
+            time.sleep(2**retries)
+            retries += 1
+
+        else:  # while-else
+            # If we haven't gotten a definite yes or no but we've exceeded or retries,
+            # report that DBus is not running
+            status = False
+
+        return status
+
+
+def _is_sysv_managed_dbus_running():
+    """Get DBus status from SysVinit compatible systems."""
+    # None means the status should be retried because we weren't sure if it is turned off.
+    running = None
+    output, _ret_code = utils.run_subprocess(["/sbin/service", "messagebus", "status"])
+    for line in output.splitlines():
+        if line.startswith("messagebus"):
+            if "running" in line:
+                running = True
+                break
+
+            # Note: SysV has a stopped status but I don't think that toggles until after
+            # the service is running so we could be caught in the case where the service
+            # is starting if we don't retry.
+
+    return running
+
+
+def _is_systemd_managed_dbus_running():
+    """Get DBus status from systemd."""
+    # Reloading, activating, etc will return None which means to retry
+    running = None
+
+    output, ret_code = utils.run_subprocess(["/usr/bin/systemctl", "show", "-p", "ActiveState", "dbus"])
+    for line in output.splitlines():
+        # Note: systemctl seems to always emit an ActiveState line (ActiveState=inactive if
+        # the service doesn't exist).  So this check is just defensive coding.
+        if line.startswith("ActiveState="):
+            state = line.split("=", 1)[1]
+
+            if state == "active":
+                # DBus is definitely running
+                running = True
+                break
+
+            if state in ("inactive", "deactivating", "failed"):
+                # DBus is definitely not running
+                running = False
+                break
+
+    return running
 
 
 # Code to be executed upon module import
--- a/convert2rhel/toolopts.py
+++ b/convert2rhel/toolopts.py
@@ -19,9 +19,10 @@ import copy
 import logging
 import optparse
 import os
+import re
 import sys
 
-from six.moves import configparser
+from six.moves import configparser, urllib
 
 from convert2rhel import __version__, utils
 
@@ -43,7 +44,9 @@ class ToolOpts(object):
         self.enablerepo = []
         self.disablerepo = []
         self.pool = None
-        self.serverurl = None
+        self.rhsm_hostname = None
+        self.rhsm_port = None
+        self.rhsm_prefix = None
         self.autoaccept = None
         self.auto_attach = None
         self.restart = None
@@ -345,7 +348,36 @@ class CLI(object):
                     "Ignoring the --serverurl option. It has no effect when --disable-submgr or --no-rhsm is used."
                 )
             else:
-                tool_opts.serverurl = parsed_opts.serverurl
+                # Parse the serverurl and save the components.
+                try:
+                    url_parts = _parse_subscription_manager_serverurl(parsed_opts.serverurl)
+                    url_parts = _validate_serverurl_parsing(url_parts)
+                except ValueError as e:
+                    # If we fail to parse, fail the conversion. The reason for
+                    # this harsh treatment is that we will be submitting
+                    # credentials to the server parsed from the serverurl. If
+                    # the user is specifying an internal subscription-manager
+                    # server but typo the url, we would fallback to the
+                    # public red hat subscription-manager server. That would
+                    # mean the user thinks the credentials are being passed
+                    # to their internal subscription-manager server but it
+                    # would really be passed externally.  That's not a good
+                    # security practice.
+                    loggerinst.critical(
+                        "Failed to parse a valid subscription-manager server from the --serverurl option.\n"
+                        "Please check for typos and run convert2rhel again with a corrected --serverurl.\n"
+                        "Supplied serverurl: %s\nError: %s" % (parsed_opts.serverurl, e)
+                    )
+
+                tool_opts.rhsm_hostname = url_parts.hostname
+
+                if url_parts.port:
+                    # urllib.parse.urlsplit() converts this into an int but we
+                    # always use it as a str
+                    tool_opts.rhsm_port = str(url_parts.port)
+
+                if url_parts.path:
+                    tool_opts.rhsm_prefix = url_parts.path
 
         if parsed_opts.keep_rhsm:
             if tool_opts.no_rhsm:
@@ -462,6 +494,44 @@ def options_from_config_files(cfg_path=None):
                     loggerinst.warning("Unsupported header %s in %s." % (header, path))
 
     return supported_opts
+
+
+def _parse_subscription_manager_serverurl(serverurl):
+    """Parse a url string in a manner mostly compatible with subscription-manager --serverurl."""
+    # This is an adaptation of what subscription-manager's cli enforces:
+    # https://github.com/candlepin/subscription-manager/blob/main/src/rhsm/utils.py#L112
+
+    # Don't modify http://<something> and https://<something> as they are fine
+    if not re.match("https?://[^/]+", serverurl):
+        # Anthing that looks like a malformed scheme is immediately discarded
+        if re.match("^[^:]+:/.+", serverurl):
+            raise ValueError("Unable to parse --serverurl. Make sure it starts with http://HOST or https://HOST")
+
+        # If there isn't a scheme, add one now
+        serverurl = "https://%s" % serverurl
+
+    url_parts = urllib.parse.urlsplit(serverurl, allow_fragments=False)
+
+    return url_parts
+
+
+def _validate_serverurl_parsing(url_parts):
+    """
+    Perform some tests that we parsed the subscription-manager serverurl successfully.
+
+    :arg url_parts: The parsed serverurl as returned by urllib.parse.urlsplit()
+    :raises ValueError: If any of the checks fail.
+    :returns: url_parts If the check was successful.
+    """
+    if url_parts.scheme not in ("https", "http"):
+        raise ValueError(
+            "Subscription manager must be accessed over http or https.  %s is not valid" % url_parts.scheme
+        )
+
+    if not url_parts.hostname:
+        raise ValueError("A hostname must be specified in a subscription-manager serverurl")
+
+    return url_parts
 
 
 # Code to be executed upon module import
--- a/convert2rhel/unit_tests/__init__.py
+++ b/convert2rhel/unit_tests/__init__.py
@@ -109,11 +109,17 @@ def safe_repr(obj, short=False):
     return result[:_MAX_LENGTH] + " [truncated]..."
 
 
-def get_pytest_mark(request, mark_name):
+def get_pytest_marker(request, mark_name):
     """
     Get a pytest mark from a request.
+
     The pytest API to retrieve a mark changed between RHEL6 and RHEL7.  This function is
     a compatibility shim to retrieve the value.
+
+    Use this function instead of pytest's `request.node.get_closest_marker(mark_name)` so that it will work on all versions of RHEL that we are targeting.
+    .. seealso::
+        * `pytest's get_closest_marker() function which this function wraps <https://docs.pytest.org/en/stable/reference/reference.html#pytest.nodes.Node.get_closest_marker>`_
+        * `A technique you might use where you would need to use this function to retrieve a mark's value. <https://docs.pytest.org/en/stable/how-to/fixtures.html#using-markers-to-pass-data-to-fixtures>`_
     """
     if pytest.__version__.split(".") <= ["3", "6", "0"]:
         mark = request.node.get_marker(mark_name)
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -116,6 +116,7 @@ def test_perform_pre_checks(monkeypatch):
     check_rhel_compatible_kernel_is_used_mock = mock.Mock()
     check_package_updates_mock = mock.Mock()
     is_loaded_kernel_latest_mock = mock.Mock()
+    check_dbus_is_running_mock = mock.Mock()
 
     monkeypatch.setattr(
         checks,
@@ -154,6 +155,7 @@ def test_perform_pre_checks(monkeypatch):
     )
     monkeypatch.setattr(checks, "check_package_updates", value=check_package_updates_mock)
     monkeypatch.setattr(checks, "is_loaded_kernel_latest", value=is_loaded_kernel_latest_mock)
+    monkeypatch.setattr(checks, "check_dbus_is_running", value=check_dbus_is_running_mock)
 
     checks.perform_pre_checks()
 
@@ -164,6 +166,7 @@ def test_perform_pre_checks(monkeypatch):
     check_rhel_compatible_kernel_is_used_mock.assert_called_once()
     check_package_updates_mock.assert_called_once()
     is_loaded_kernel_latest_mock.assert_called_once()
+    check_dbus_is_running_mock.assert_called_once()
 
 
 def test_pre_ponr_checks(monkeypatch):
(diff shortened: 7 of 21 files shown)


convert2rhel/subscription.py
```diff
@@ -278,6 +287,93 @@ def args(self):
...
+        if self.server_url:
+            connection_opts = {"host": self.server_url}
+        else:
+            # DBus seems to want empty string if there are no values even though the type is a dict
+            connection_opts = ""
+
+        # Create a new bus so we can talk to rhsm privately (For security:
+        # talking on the system bus might be eavesdropped in certain scenarios)
+        register_server = system_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/RegisterServer")
+        address = register_server.Start("", dbus_interface="com.redhat.RHSM1.RegisterServer")
```
ptoscano, line 341: the first parameter is the locale for the potential error messages; maybe it'd be better to pass the current locale of convert2rhel; the same applies to the other calls that take a locale as parameter

convert2rhel/subscription.py
```diff
@@ -278,6 +287,93 @@ def args(self):
...
+            # Use the private bus to register the machine
+            private_bus = dbus.connection.Connection(address)
+            register_object = private_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/Register")
+
+            if self.password:
+                register_object.Register(
+                    self.organization or "",
+                    self.username,
+                    self.password,
+                    "",
```
ptoscano, line 353: `options` is a dict, so this must be an empty dict rather than an empty string

convert2rhel/subscription.py
```diff
@@ -278,6 +287,93 @@ def args(self):
...
+                    connection_opts,
+                    "C",
```
ptoscano, line 355: ditto (see earlier notes wrt locale)

convert2rhel/subscription.py
```diff
@@ -278,6 +287,93 @@ def args(self):
...
+            Documentation for the subscription-manager dbus API:
+            https://www.candlepinproject.org/docs/subscription-manager/dbus_objects.html#registerserver
+        """
+        system_bus = dbus.SystemBus()
+
+        if self.server_url:
+            connection_opts = {"host": self.server_url}
+        else:
+            # DBus seems to want empty string if there are no values even though the type is a dict
+            connection_opts = ""
```
ptoscano, line 336: `connection_options` is a dict, so this must be an empty dict (it is bizzarre that it works, however it shouldn't)

convert2rhel/subscription.py
```diff
@@ -278,6 +287,93 @@ def args(self):
...
+                    connection_opts,
+                    "C",
+                    dbus_interface="com.redhat.RHSM1.Register",
+                )
+
+            else:
+                register_object.RegisterWithActivationKeys(
+                    self.org,
+                    [self.activation_key],
+                    "",
```
ptoscano, line 363: `options` is a dict, so this must be an empty dict rather than an empty string

convert2rhel/subscription.py
```diff
@@ -278,6 +287,93 @@ def args(self):
...
+                    connection_opts,
+                    "C",
```
ptoscano, line 365: ditto (see earlier notes wrt locale)

convert2rhel/subscription.py
```diff
@@ -278,6 +287,93 @@ def args(self):
...
+                    dbus_interface="com.redhat.RHSM1.Register",
+                )
+
+        finally:
+            # Always shut down the private bus
+            register_server.Stop("", dbus_interface="com.redhat.RHSM1.RegisterServer")
```
ptoscano, line 371: ditto (see earlier notes wrt locale)

ptoscano (commented):
I left some easy notes on the code; I can try provide some answers to your question, at least from the subscription-manager POV:

> Note that this requires that dbus is installed and running and that the dbus-python/python-dbus packages are installed. This package is available on RHEL6-RHEL9 clones (all that we care about) but is a new dependency.

New for convert2rhel, not for subscription-manager.

>     * [ ]  Should we check whether dbus is running?

As you wrote in the comments: RHEL 7 uses systemd, which mandates D-Bus, while RHEL 6 uses upstart.

Maybe it would be better to try to connect to the system bus, and in case start the D-Bus system service (it should affect only RHEL 6).

>     * [ ]  If it's not running should we install and run it or should we fall back to subscription-managr cli?
>     * [ ]  If we fallback to cli, do we still allow activation_key even though it is insecure?
>     * [ ]  What should the return value of RegistrationCommand be?  Should cli non-zero retcodes raise an exception instead of returning the code?
>     * [ ]  Do we want to use DBUS for registering via passwords or only for activation key?

Personally, I would switch the implementation of registration with subscription-manager to D-Bus exclusively. Otherwise you need to maintain two different code paths (D-Bus and cli), and given that D-Bus should be used when converting RHEL 7 and greater versions, it means the cli version will be tested less.

Also, using D-Bus exclusively means that potentially more interactions with subscription-manager could be done via D-Bus in the future, rather than spawning shell commands (which is a not so good idea, as the console output of subscription-manager is neither stable nor machine parseable).

>     * [ ]  Should checking for DBUS be done in either system_info or checks.py rather than in subscription.py?
> 
Since the D-Bus way also works with subscription-manager in RHEL 6, I'd rather error out in the subscription module if D-Bus is not available (cannot be installer or started). This way it will only matter if the subscription module is used.

Of course this depends on how much effort you want to invest in supporting RHEL 6->7 upgrades.

>       * We could inform the user of the problem and let them install and start dbus like we make them install a compatible kernel and reboot.

Can be a way.

>     * [ ]  If we install it and start the daemon, do we need to revert that after the upgrade/rollback?

I'd say no, as you will get anyway a D-Bus daemon installed in RHEL >= 7.

>     * [ ]  Similar questions for python-dbus and dbus-python being installed.
>     * [ ]  We could have the rpm require dbus and python-dbus which will make sure it is installed.  It won't make sure that dbus is running, though.

I'd recommend to add `python-dbus`/`python3-dbus` explicitly as dependency in the packaging (`convert2rhel.spec`): even if the subscription module is disabled, the Python module is very small in size and dependencies, so it should not be a problem to unconditionally require it even if the subscription module is disabled.

convert2rhel/subscription.py
abadger: Cool.  We always use the C locale for the things that we're calling so I will definitely add that to these function calls.

abadger: Pushed a commit with all locale fields filled in.

convert2rhel/subscription.py
abadger: Pushed a change that changes this to `dbus.Dictionary({}, signature="ss")`

convert2rhel/subscription.py
abadger: Now uses a locale of `C`.

convert2rhel/subscription.py
abadger: Pushed a change that changes this to `dbus.Dictionary({}, signature="ss")`.

Pino and I talked about this on chat.  An empty python dict doesn't work because dbus can't get enough introspection information from that to properly type the keys and values: https://gitlab.freedesktop.org/dbus/dbus-python/-/issues/25 .  `{"": ""}` works for `/bin/dbus-send` but it doesn't work in python-dbus for an unknown reason.

Manually creating the empty DBus Dictionary with signature information should work and better documents what the field means than empty string.

convert2rhel/subscription.py
abadger: Pushed a change that changes this to dbus.Dictionary({}, signature="ss")

convert2rhel/subscription.py
abadger: Now uses the `C` locale.

abadger:
>>    * [ ]  Should checking for DBUS be done in either system_info or checks.py rather than in subscription.py?
> Since the D-Bus way also works with subscription-manager in RHEL 6, I'd rather error out in the subscription module if D-Bus is not available (cannot be installer or started). This way it will only matter if the subscription module is used.

IIRC, subscription always happens.

If we are going to fail if dbus is not running, we would definitely want that to happen in the checks phase so as little as possible has happened to the system (which would mean we don't have any changes we have to rollback).  If we're going to try to install dbus for the user, then we still might want to find that fact out in system_info and then make use of that to decide what to do in subscription.py.

ptoscano (commented):
In addition to the inline comments I left, `RegistrationCommand.args()` is no more needed now, and can be dropped.

convert2rhel/subscription.py
```diff
@@ -279,26 +288,89 @@ def args(self):
...
+        # DBus throws errors for pure empty dict and dbus-python doesn't like
+        # {"": ""} even though the /bin/dbus-send thinks that is okay.  Use
+        # this instead.
+        EMPTY_DBUS_DICT = dbus.Dictionary({}, signature="ss")
+
+        loggerinst.debug("Getting a handle to the system dbus")
+        system_bus = dbus.SystemBus()
+
+        if self.server_url:
+            connection_opts = {"host": self.server_url}
```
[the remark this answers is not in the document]
abadger: Hey @ptoscano I'm back from some PTO.  Should protocol be discarded and user:password be given as parameters in the `Register()` dbus message?  (So only host, port, and path portions of the url will be passed as connection_opts?)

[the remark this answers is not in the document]
ptoscano: > Should protocol be discarded

Yes.

> user:password be given as parameters in the Register() dbus message?

Yes.

> (So only host, port, and path portions of the url will be passed as connection_opts?)

Yes.

convert2rhel/subscription.py
```diff
@@ -279,26 +288,89 @@ def args(self):
...
-            # and activation_key to use a file-based approach to passing the secrets.
-            output, ret_code = utils.run_subprocess([self.cmd] + self.args, print_cmd=False)
+            connection_opts = EMPTY_DBUS_DICT
+
+        # Create a new bus so we can talk to rhsm privately (For security:
+        # talking on the system bus might be eavesdropped in certain scenarios)
+        loggerinst.debug("Getting a subscription-manager RegisterServer object from dbus")
+        register_server = system_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/RegisterServer")
+        loggerinst.debug("Starting a private DBus to talk to subscription-manager")
+        address = register_server.Start("C", dbus_interface="com.redhat.RHSM1.RegisterServer")
```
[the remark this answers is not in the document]
abadger: I like the idea of moving the locale out to a separate file in case we do localize convert2rhel in the future. I'm a little hesitant about changing this to use a non-C locale for several reasons, though:

* convert2rhel isn't localized itself.  Emitting a few localized messages while everything else is unlocalized doesn't seem like a good experience.
* In the past I've had some people make the point that error messages, are better left unlocalized so that people can easily put the errors into google (or bug reports).  I'm not quite convinced either way about that but I think it's appropriate to punt on that question until we decide to localize convert2rhel itself.

convert2rhel/subscription.py
```diff
@@ -279,26 +288,89 @@ def args(self):
...
+
+        if self.server_url:
+            connection_opts = {"host": self.server_url}
         else:
-            # Warning: Currently activation_key can only be specified on the CLI. This is insecure
-            # but there's nothing we can do about it for now. Once subscription-manager issue:
-            # https://issues.redhat.com/browse/ENT-4724 is implemented, we can change both password
-            # and activation_key to use a file-based approach to passing the secrets.
-            output, ret_code = utils.run_subprocess([self.cmd] + self.args, print_cmd=False)
+            connection_opts = EMPTY_DBUS_DICT
```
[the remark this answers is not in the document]
abadger: I like this. For now, though, I've taken a tiny bit of this and combined it with the construction of connection_opts from server_url.  I don't think we'll need to do anything dynamic with connection_opts in the convert2rhel use case so it feels like it's better to move the boilerplate into a separate function.  But if we do need to set connection_opts to different values, I'll keep this in mind as a good strategy.

convert2rhel/subscription.py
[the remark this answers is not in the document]
abadger: Another related question, if the subscription-manager cli is given a username/password pair in both cli arguments and as part of the server-url, which takes precedence?  If they are only in the server-url, does the cli make use of it? If they are in the url but activation-key is given on the cli, are the username and password used?  (I'll probably try to model current behaviour so knowing these things will help me get that right).

convert2rhel/subscription.py
```diff
@@ -279,26 +288,89 @@ def args(self):
...
-            )
+        # Questions:
+        # * Should we check whether dbus is running?
+        # * If it's not running should we install and run it or should we fall back to
+        #   subscription-managr cli?
+        # * If we install it, do we need to revert that after the upgrade/rollback?
+        # * Should checking for DBUS be done in either system_info or checks.py?
+        # * What should the return value of this code be?
+
+        # DBus throws errors for pure empty dict and dbus-python doesn't like
```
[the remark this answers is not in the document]
abadger: Reworded to account for this.

convert2rhel/subscription.py
```diff
@@ -22,6 +22,15 @@
...
 
+
+try:
+    import dbus
+    import dbus.connection
+
+    DBUS = True
+except ImportError:
+    DBUS = False
+
```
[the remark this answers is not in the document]
abadger: <nod>  Added his and removing RegistrationCommaqnd.args to the list of things to do before this code is ready.

convert2rhel/subscription.py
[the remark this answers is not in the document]
ptoscano: > I like the idea of moving the locale out to a separate file

Oh sorry, I did not mean a new file on its own, rather a simple variable declared at the top of the `__call__()` function, eg.
```python
    SUBMAN_LOCALE = "C"
```
and use it in all the calls; this way, whatever you want to do in the future, it's a one place change.

convert2rhel/subscription.py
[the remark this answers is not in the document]
ptoscano: Hmm I need to check myself too, however for now I'd ignore the userinfo part of the server URL.

packaging/convert2rhel.spec
```diff
@@ -41,12 +44,14 @@ Requires:       dnf
 Requires:       dnf-utils
 Requires:       grubby
 Requires:       python3-pexpect
+Requires:       python4-dbus
```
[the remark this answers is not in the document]
abadger: Thanks.  Fixed.

convert2rhel/subscription.py
```diff
@@ -286,6 +288,22 @@ def args(self):
...
         return args
 
+    @property
+    def connection_opts(self):
+        connection_opts = {}
+
+        if self.server_url:
+            url_parts = urllib.parse.urlsplit(self.server_url, allow_fragments=False)
+            connection_opts = {"host": url_parts.hostname}
+            connection_opts = {"port": url_parts.port}
```
ptoscano, line 298: if no port was specified, there is no `port` in `url_part`

convert2rhel/subscription.py
```diff
@@ -286,6 +288,22 @@ def args(self):
...
+
+            path = urllib.parse.urlunsplit(
+                (None, None, url_parts.path, url_parts.query, url_parts.fragment))
+            connection_opts = {"handler": path}
```
ptoscano, line 301: hm why not simply use `url_parts.path`? it is already `/path` as needed for "handler"

convert2rhel/subscription.py
abadger: Fixed

convert2rhel/subscription.py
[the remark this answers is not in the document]
abadger: I moved the locale constants out to their own file in this commit: f90e34f715f955fd4514f6bc07be070c6c16e786

convert2rhel/subscription.py
abadger: So.... I was thinking that if query and fragment were specified in a url, the handler would be the place for those to be passed to the subscription-manager backend.
* Are they just not used in subscription-manager right now?  If so I think this is future-proofing so I think I'd keep it.
* Are they handled by some other connection_opt?  Let me know if so and I'll move them there.

convert2rhel/subscription.py
```diff
@@ -277,28 +288,104 @@ def args(self):
...
-            # https://issues.redhat.com/browse/ENT-4724 is implemented, we can change both password
-            # and activation_key to use a file-based approach to passing the secrets.
-            output, ret_code = utils.run_subprocess([self.cmd] + self.args, print_cmd=False)
+        # Questions:
+        # * Should we check whether dbus is running?
+        # * If it's not running should we install and run it or should we fall back to
+        #   subscription-managr cli?
+        # * If we install it, do we need to revert that after the upgrade/rollback?
+        # * Should checking for DBUS be done in either system_info or checks.py?
+        # * What should the return value of this code be?
```
bocekm, line 336: @abadger, are there still some of the questions unanswered after talking to @ptoscano? I haven't looked at the discussion of you two much so here's how I'd see the answers:

> Should we check whether dbus is running?

Yes.

> If it's not running should we install and run it or should we fall back to subscription-managr cli?

We're already decided to add to spec, so it's installed. If not running, we should start it (no fall back to the sub-man cli).

> If we install it, do we need to revert that after the upgrade/rollback?

If the service was not running and we started it, we should stop the service upon rollback. Let's do that though in a separate PR, after we merge https://github.com/oamg/convert2rhel/pull/516.

> Should checking for DBUS be done in either system_info or checks.py?

I'd say I prefer the checks.py but it's not that important.

> What should the return value of this code be?

I believe you can drop the code in `register_system()` that check the output string (introduced in
https://github.com/oamg/convert2rhel/commit/311b0bd363e3a4b170c9ea981194603e8b4e3858#diff-51e9ff11d39778d3b26778b293fc350430a82e2feed0dab2938da7c0069ffdc6R117). Thus, you don't need to return any string from the `__call__()`.

Regarding the return code - when the registration fails the `Register`/`RegisterWithActivationKeys` services raise an exception, right? We've got the loop for 3 registration attempts and to be able to get to another attempt it seems we should catch the exception. Oh, I now see Pino mentions the same in a comment https://github.com/oamg/convert2rhel/pull/540/files#r923341266.

bocekm (changes requested):

convert2rhel/subscription.py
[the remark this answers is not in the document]
abadger: Good point about args.  I'll think for a bit about whether to remove args and restructure the logging or keep args as a handy way to visualize that.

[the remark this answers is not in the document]
abadger: Conditional import of dbus has now been removed

convert2rhel/subscription.py
abadger: <nod>  I think I've captured all of your (and Pino's) feedback in the checklist in this PR's description.  Let me know if I've missed anything.

I'll remove this comment in the code as it is getting outdated.

convert2rhel/subscription.py
ptoscano: AFAICT, query and fragment are ignored right now. They are ignored after parsing URLs from the command lines, and there are no config knobs for them (sub-man does REST calls, so they would need to be passed to each).

Because of that, I'd simply ignore query and fragment altogether, as (if they would be ever needed) they would require additional work in sub-man.

convert2rhel/subscription.py
```diff
@@ -277,28 +287,103 @@ def args(self):
...
+        Status of dbus on various platforms:
+            * RHEL6:
+                * DBUS is present but may not be installed.
+                * Install the dbus package and run /etc/rc.d/init.d/messagebus start
+                * dbus-python-0.83.0 is available
+            * RHEL7:
+                * dbus looks to be installed and run by default
+                * dbus-python-1.1.9 is available
+            * RHEL8:
+            * RHEL9:
```
r0x0d, line 327: State for RHEL 8 and RHEL 9 is the same? If so, I would group them together like: 

```
RHEL8 & RHEL9:
```

convert2rhel/checks.py
```diff
@@ -698,3 +699,21 @@ def is_loaded_kernel_latest():
...
             )
+
+
+def check_dbus_is_running():
+    """Error out if we need to register with rhsm and the dbus daemon is not running."""
+    logger.task("Prepare: Check that DBus Daemon is running")
+
+    if tool_opts.no_rhsm:
+        logger.info("Skipping the check because we have been asked not to subscribe this system to rhsm.")
+        return None
```
ptoscano, line 710: minor: only `return` is enough

convert2rhel/checks.py
```diff
@@ -698,3 +699,21 @@ def is_loaded_kernel_latest():
...
+
+    if not system_info.dbus_running:
+        logger.critical(
+            "Could not find a running DBus Daemon which is needed to register with subscription manager.\n"
+            "Please start dbus using `systemctl start dbus` or (on CentOS 6), `service messagebus start`"
+        )
+
+    logger.info("DBus Daemon is running")
+    return None
```
ptoscano, line 719: minor: only `return` is enough, or even no `return` is needed here

convert2rhel/subscription.py
abadger: Good point.  I didn't have a RHEL8 vm handy when I was filling this comment out.  I checked things just now and RHEL8 and RHEL9 are almost identical.  RHEL8 just has a slightly older version of python3-dbus so I've merged them as you suggested.

convert2rhel/subscription.py
```diff
@@ -277,28 +287,103 @@ def args(self):
...
+        if self.server_url:
+            url_parts = urllib.parse.urlsplit(self.server_url, allow_fragments=False)
+            connection_opts["host"] = url_parts.hostname
+            if url_parts.port is not None:
+                connection_opts["port"] = url_parts.port
+
+            # Current subscription-manager server-side does not take query and fragments at this
+            # path but include them for future proofing.
+            path = urllib.parse.urlunsplit(("", "", url_parts.path, url_parts.query, url_parts.fragment))
+            connection_opts["handler"] = path
```
ptoscano, line 309: I'd simply leave them out; it would require work in sub-man to support them, if ever, so let's keep things simple here and use only `url_parts.path`

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -35,6 +38,41 @@
...
+    _mocked_rhsm_register_base.Register = mock.Mock(side_effect=rhsm_returns.args[0])
+    return _mocked_rhsm_register_base.Register
+
+
+@pytest.fixture
+def mocked_rhsm_register_with_activation_keys(monkeypatch, request, _mocked_rhsm_register_base):
+    if pytest.__version__.split(".") <= ["3", "6", "0"]:
+        rhsm_returns = request.node.get_marker("rhsm_returns")
+    else:
+        rhsm_returns = request.node.get_closest_marker("rhsm_returns")
```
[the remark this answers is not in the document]
abadger: I hear you, yeah.  What do you think about making a helper function that we can call with the name of the mark and the helper returns the mark to the fixture?

convert2rhel/subscription.py
```diff
@@ -277,28 +287,103 @@ def args(self):
...
+                * DBUS is present but may not be installed.
+                * Install the dbus package and run /etc/rc.d/init.d/messagebus start
+                * dbus-python-0.83.0 is available
+            * RHEL7:
+                * dbus looks to be installed and run by default
+                * dbus-python-1.1.9 is available
+            * RHEL8:
+            * RHEL9:
+                * dbus looks to be installed and run by default
+                * python3-dbus-1.2.18 is available
```
ptoscano, line 329: TBH I'm not sure how the version of dbus-python is useful in this context; I'd simplify thinks a bit as:
- RHEL 6:
  - D-Bus may not be installed or running by default
  - Install the 'dbus' package and run `/etc/rc.d/init.d/messagebus start`
- RHEL 7 and greater:
  - D-Bus is an hard system dependency because of systemd
  - a D-Bus daemon is installed and running

ptoscano (commented):
Thanks for the various updates! At least from my (subscription-manager) POV, the implementation looks sane and matches the sub-man API.

I left some notes for small/minor things; nothing biggie though.

One open question I see in the big checklist:

> A few questions about schema, username, and password for pino. 

username & password specified in `--serverurl URL` are ignored; hence, you can safely ignore them too.

convert2rhel/checks.py
abadger: I was a bit on the fence about this one.  I kept the `None`s in because I am returning in two different levels but the counter argument is that we aren't using the return value from any of the `check_*` functions.  I've restructured this code so that the returns are both "exit early" returns at the top.  Using bare `return`s with that structure feel much better to me.

convert2rhel/subscription.py
```diff
@@ -279,26 +288,89 @@ def args(self):
...
+                    "C",
+                    dbus_interface="com.redhat.RHSM1.Register",
+                )
+
+        finally:
+            # Always shut down the private bus
+            loggerinst.debug("Shutting down private DBus instance")
+            register_server.Stop("C", dbus_interface="com.redhat.RHSM1.RegisterServer")
+
+        return "", 0
```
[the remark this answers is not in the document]
abadger: This has now been changed to return nothing in the success case and raise an exception in the error case.

convert2rhel/unit_tests/subscription_test.py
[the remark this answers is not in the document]
abadger: I've created the helper here: 1d003704b4bb877b800341531be0d0a462fc7a29  I've ported the two test functions that use it in this PR to use it but I did not port the one that you are working on in the RHEL6 PR.

convert2rhel/checks.py
abadger: Changed to use bare return (along with small refactor of the code) in this commit: a049b58f83122b4d83348ca25484db7c4d357d18

convert2rhel/subscription.py
abadger: I decided to leave the versions in because there were some features/bugfixes that I could not make use of since they were added in a later version than some versions of RHEL had.  Having a quick table of relevant versions on all the RHEL versions we're targetting was helpful for quickly verifying that we weren't going to run into a problem.  However, now that I think about it, those might have been versions of the dbus daemon rather than the dbus python bindings.  I'll add the dbus version to the information in this content as well.

convert2rhel/subscription.py
abadger: <nod> I can see your point of view as a subscription-manager dev.  From my point of view, though, I'd rather code something now that might work if subscription-manager should choose to start using query and fragment in the future while the way that subscription-manager and dbus work is fresh in my mind.  If subscription-manager never needs to add query or fragment here, then there is no harm.  If subscription-manager adds them in the way that the code anticipates, then I won't have to try to figure out the mapping of cli->dbus args again.  If subscription-manager adds them in a way that the code doesn't anticipate, at least when someone looks at the code they'll see all the pieces even if they're assembled in the wrong way.

convert2rhel/systeminfo.py
```diff
@@ -359,6 +360,35 @@ def corresponds_to_rhel_eus_release(self):
...
+                    if "run" in line:
+                        running = True
+                        break
+        else:
+            # Systemd
+            output, ret_code = utils.run_subprocess(["/usr/bin/systemctl", "show", "-p", "ActiveState", "dbus"])
+            for line in output.splitlines():
+                if line.startswith("ActiveState="):
+                    state = line.split("=", 1)[1]
+                    if state in ("active", "reloading", "activating"):
```
r0x0d, line 386: Is there any possibility for the `reloading` or `activating` status to turn to be a failure in the end? 

I mean, I don't know if that's even possible, but, let's pretend that after reload something goes wrong, would that affect us?

convert2rhel/unit_tests/subscription_test.py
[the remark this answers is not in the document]
r0x0d: Nice! I will port that out to my RHEL6 PR.

convert2rhel/systeminfo.py
abadger: There is.  I thought about implementing retry on these but got lazy.  I've now pushed a commit which will retry three times if the state is not definitive (for systemd: active, failed, inactive, or deactivating)

convert2rhel/subscription.py
```diff
@@ -385,6 +394,26 @@ def __call__(self):
             register_server.Stop(i18n.SUBSCRIPTION_MANAGER_LOCALE, dbus_interface="com.redhat.RHSM1.RegisterServer")
 
 
+def parse_url(server_url):
+    """Parse a string as a url in a manner mostly compatible with subscription-manager."""
+    # This is an adaptation of what subscription-manager's cli enforces:
+    # https://github.com/candlepin/subscription-manager/blob/main/src/rhsm/utils.py#L112
+
+    # It should look like it has a scheme followed by other information.
+    if not re.match("^.+://.+", server_url):
```
ptoscano, line 403: i'd use `[^:]+` instead of `.+`, to avoid potential greedyness of regexps

ptoscano (commented):
also there's a typo `compatiility` in the commit message

convert2rhel/subscription.py
abadger: Good idea.

abadger: change pushed in a new commit.

abadger:
Okay, I think I've addressed all the comments that have been raised here and in our internal chat.  I have run my checkout against one OEL8 vagrant instance and I don't see any duplicated log messages in the run.

I don't like where it is that I am adding NullHandler to the root logger but I think it will do for now.  @r0x0d  is working on adding a dedicated setup step as part of his work on merging the yum transactions.  When that happens, I think moving the root log handler setup to there would make sense.

In the initial port for this PR, I have a list of checklist items.  All the ones left undone are things that I think are things I think we should change with this implementation but we can defer them to the next release.I'll open jira tickets for all of those.

convert2rhel/subscription.py
ptoscano: I don't think subscription-manager is going to get that functionality any time soon. It is not only about taking query and fragment from an URL passed to `--serverurl`, but also needs new configuration bits (as hostname, port, etc are stored separately there) and all the associated machinery for it, and there is simply nothing that ATM needs them.

So, while I understand your POV, I'd keep things simple and not adding things that could misbehave.

ptoscano:
> I don't like where it is that I am adding NullHandler to the root logger but I think it will do for now.

From what I saw in the dbus-python code, the logged is initialized only when printing "severe problems" that happen when using dbus-python. For example, passing `None` as value for e.g. a string. subscription-manager uses dbus-python way more than this code, and TTBOMK we have not seen any problem related to this, neither as printed directly to terminal nor in the log file.
While I understand your POV on defensive programming, from what I see all the parameters are already checked in one way or another. So IMHO this extra commit could not be needed, now that the code is fixed.

convert2rhel/subscription.py
```diff
@@ -277,28 +287,131 @@ def args(self):
...
-            output, ret_code = utils.run_subprocess([self.cmd] + self.args, print_cmd=False)
+        connection_opts = {}
+
+        url_parts = None
+        if self.server_url:
+            try:
+                url_parts = parse_url(self.server_url)
+            except ValueError:
+                # url_parts will remain None which will mean we return an empty dbus.DIctionary
+                pass
```
SpyTec, line 306: Untested

convert2rhel/checks.py
```diff
@@ -698,3 +699,21 @@ def is_loaded_kernel_latest():
...
                 "'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' to 1." % package_to_check
             )
+
+
+def check_dbus_is_running():
+    """Error out if we need to register with rhsm and the dbus daemon is not running."""
+    logger.task("Prepare: Check that DBus Daemon is running")
+
+    if tool_opts.no_rhsm:
+        logger.info("Skipping the check because we have been asked not to subscribe this system to rhsm.")
```
SpyTec, line 709:
```suggestion
        logger.info("Skipping the check because we have been asked not to subscribe this system to RHSM.")
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1215,3 +1218,43 @@ def test_is_loaded_kernel_latest_unsupported_skip(
         is_loaded_kernel_latest()
 
     assert expected in caplog.records[-1].message
+
+
+@pytest.mark.parametrize(
+    ("no_rhsm", "dbus_running", "log_msg"),
+    (
+        (True, True, "Skipping the check because we have been asked not to subscribe this system to rhsm."),
+        (True, False, "Skipping the check because we have been asked not to subscribe this system to rhsm."),
```
SpyTec, line 1227:
```suggestion
        (True, True, "Skipping the check because we have been asked not to subscribe this system to RHSM."),
        (True, False, "Skipping the check because we have been asked not to subscribe this system to RHSM."),
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1215,3 +1218,43 @@ def test_is_loaded_kernel_latest_unsupported_skip(
...
+    monkeypatch.setattr(checks, "system_info", global_system_info)
+    global_system_info.dbus_running = False
+
+    with pytest.raises(SystemExit):
+        checks.check_dbus_is_running()
+
+    log_msg = (
+        "Could not find a running DBus Daemon which is needed to"
+        " register with subscription manager.\nPlease start dbus using `systemctl"
+        " start dbus` or (on CentOS 6), `service messagebus start`"
```
SpyTec, line 1257:
```suggestion
        " start dbus` or (on CentOS Linux 6), `service messagebus start`"
```

convert2rhel/checks.py
```diff
@@ -698,3 +699,21 @@ def is_loaded_kernel_latest():
...
+        return
+
+    if system_info.dbus_running:
+        logger.info("DBus Daemon is running")
+        return
+
+    logger.critical(
+        "Could not find a running DBus Daemon which is needed to register with subscription manager.\n"
+        "Please start dbus using `systemctl start dbus` or (on CentOS 6), `service messagebus start`"
```
SpyTec, line 718:
```suggestion
        "Please start dbus using `systemctl start dbus` or (on CentOS Linux 6), `service messagebus start`"
```

convert2rhel/subscription.py
```diff
@@ -277,28 +287,131 @@ def args(self):
...
+
+    url_parts = urllib.parse.urlsplit(server_url, allow_fragments=False)
+
+    if url_parts.scheme not in ("https", "http"):
+        raise ValueError(
+            "Subscription manager must be accessed over http or https.  %s is not valid" % url_parts.scheme
+        )
 
-        return output, ret_code
+    return url_parts
```
SpyTec, line 414: Method has no unit tests. If we want to make it private then we are missing a branch for the `ValueError`, can be covered with the exception handling where this method is called

convert2rhel/systeminfo.py
```diff
@@ -359,6 +364,80 @@ def corresponds_to_rhel_eus_release(self):
...
+            # Note: SysV has a stopped status but I don't think that toggles until after
+            # the service is running so we could be caught in the case where the service
+            # is starting if we don't retry.
+
+    return running
+
+
+def _get_dbus_status_from_systemd():
+    """Get DBus status from systemd."""
+    # Reloading, activating, etc will return None which means to retry
```
SpyTec, line 419: We're not retrieving the status, but rather checking whether it's running
```suggestion
def _check_dbus_service_running():
    """
    Check that DBus systemd service is running.
    :returns: true/false if status is determined. None if status is indeterminate
    """
```

convert2rhel/systeminfo.py
```diff
@@ -359,6 +364,80 @@ def corresponds_to_rhel_eus_release(self):
...
+        :returns: True if dbus is running.  Otherwise False
+        """
+        retries = 0
+        status = False
+
+        while retries < CHECK_DBUS_STATUS_RETRIES:
+            if self.version.major <= 6:
+                status = _get_dbus_status_from_sysv()
+            else:
+                status = _get_dbus_status_from_systemd()
```
SpyTec, line 380: See review comment for `_get_dbus_status_from_systemd()`
```suggestion
                status = _check_dbus_service_running()
```

convert2rhel/unit_tests/__init__.py
```diff
@@ -107,6 +109,21 @@ def safe_repr(obj, short=False):
     return result[:_MAX_LENGTH] + " [truncated]..."
 
 
+def get_pytest_mark(request, mark_name):
```
SpyTec, line 112: This name and docstring is confusing to me. I can't tell what it's doing until I read the code

SpyTec (dismissed):

convert2rhel/i18n.py
```diff
@@ -0,0 +1,45 @@
...
+convert2rhel is not currently localized but it does interact with other systems which are. this
+module provides a single place to keep that information.  If we decide to localize convert2rhel in
+the future, this file should point us towards all the locations that we may need to change.
+"""
+
+#
+# Display locales
+#
+
+# These locales are about what we want to display to the user.  If we decide to localize, these
```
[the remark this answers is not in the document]
bocekm: This is a matter of style: https://en.wikipedia.org/wiki/Sentence_spacing. For example:

> The text-editing environment in Emacs uses a double space following a period to identify the end of sentences unambiguously; the double-space convention prevents confusion with periods within sentences that signify abbreviations.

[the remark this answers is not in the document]
r0x0d: Oh. I always thought it was something wrong with vim when doing `gq` for wrapping the text... Good to know it's a convnetion then.

convert2rhel/subscription.py
```diff
@@ -277,28 +287,131 @@ def args(self):
...
+        system_bus = dbus.SystemBus()
+
+        # Create a new bus so we can talk to rhsm privately (For security:
+        # talking on the system bus might be eavesdropped in certain scenarios)
+        loggerinst.debug("Getting a subscription-manager RegisterServer object from dbus")
+        register_server = system_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/RegisterServer")
+        loggerinst.debug("Starting a private DBus to talk to subscription-manager")
+        address = register_server.Start(
+            i18n.SUBSCRIPTION_MANAGER_LOCALE, dbus_interface="com.redhat.RHSM1.RegisterServer"
+        )
```
r0x0d, line 359: One question that strikes me is, this piece of code can't fail when we run it? I'm not sure about the dbus-library, but that could rise an unexpected exception?

bocekm (changes requested):

convert2rhel/subscription.py
```diff
@@ -120,24 +125,29 @@ def register_system():
...
-            if "user interrupted process" in output.lower():
-                raise KeyboardInterrupt
-            return
 
-        loggerinst.info("System registration failed with return code = %s" % str(ret_code))
-        sleep(REGISTRATION_ATTEMPT_DELAYS[attempt])
-        attempt += 1
-    loggerinst.critical("Unable to register the system through subscription-manager.")
+        try:
+            registration_cmd()
```
bocekm, line 137:
```suggestion
            registration_cmd()
            loggerinst.info("System registration succeeded.")
```

bocekm (changes requested):
There's some weird behavior of the registration through dbus. I would expect it to raise exception if credentials are invalid. However it does not raise it and I get an error later on when trying to attach a subscription (at a stage where the system is supposed to be already registered).

On CentOS Linux 7 with `--username, `--password`, and `--pool` options used:

```
[08/10/2022 14:47:46] DEBUG - Getting a subscription-manager Register object from dbus
[08/10/2022 14:47:46] DEBUG - Registering via username/password: register --force --serverurl=subscription.rhsm.stage.redhat.com --username=oamg_acc
[08/10/2022 14:47:55] DEBUG - Shutting down private DBus instance
Attaching provided subscription pool ID to the system ...
[08/10/2022 14:47:55] DEBUG - Calling command 'subscription-manager attach --pool 8a99f9af7c95644d017cbb877eec04f8'
Invalid Credentials
Unable to sync syspurpose with server: HTTP error (401 - Unauthorized): Invalid Credentials
```

convert2rhel/i18n.py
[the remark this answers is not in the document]
abadger: <nod>  Yeah.  I think if you're old enough, you were taught in (US schools at least) to use two spaces between sentences. (Love this headline: "[Nothing Says Over 40 Like Two Spaces after a Period!](https://www.cultofpedagogy.com/two-spaces-after-period" :-)  In addition to vim and emacs, google's android keyboard will insert a period if you type two concurrent spaces.  (Interestingly, I just googled for a way to set vim to change two spaces after period in python comments to single space and found out that PEP8 specifies two spaces after period due to the same tooling criteria, Emacs and vim use two spaces after period in their reflow tools.)

I'm committing all the places where Eric (or anyone has spotted) me using two spaces, but please be patient with me, I'm finding it very hard to retrain my fingers to only type one space after a period.

convert2rhel/subscription.py
abadger: This code can fail.  You are probably asking why it isn't part of the `try: finally:` that starts on the following line?  The reason for that is we only need the `try: finally:` to ensure we tear down the private DBus server that we're creating in this line.  If this line fails, then there won't be a registration server to tear down.

I'm assuming that register_server.Start won't fail after it creates the private DBus server and before it returns to us.  If it does fail there, we could end up with the server still running.  If that's the case, we'd also have to make the `finally:` clause handle both the server having been started and the case where it had not yet been started.

The current subscription-manager starts the server, logs it, and then returns so it seems pretty safe to let subscription-manager's code worry about this cornercase.

convert2rhel/unit_tests/__init__.py
abadger: Would get_pytest_get_marker be better?  The pytest API uses `get_marker()` (old) and `get_closest_marker()` (new) for this functionality.
```suggestion
def get_pytest_marker(request, mark_name):
```

convert2rhel/unit_tests/__init__.py
```diff
@@ -107,6 +109,21 @@ def safe_repr(obj, short=False):
...
+    """
+    Get a pytest mark from a request.
+
+    The pytest API to retrieve a mark changed between RHEL6 and RHEL7.  This function is
+    a compatibility shim to retrieve the value.
```
abadger, line 117:
```suggestion
    a compatibility shim to retrieve the value.
    
    Use this function instead of pytest's `request.node.get_closest_marker(mark_name)` so that it will work on all versions of RHEL that we are targeting.
    .. seealso::
        * `pytest's get_closest_marker() function which this function wraps <https://docs.pytest.org/en/stable/reference/reference.html#pytest.nodes.Node.get_closest_marker>`
        * `A technique you might use where you would need to use this function to retrieve a mark's value. <https://docs.pytest.org/en/stable/how-to/fixtures.html#using-markers-to-pass-data-to-fixtures>`
```

packaging/convert2rhel.spec
```diff
@@ -25,12 +25,17 @@ BuildRequires:  python%{python_pkgversion}-six
 BuildRequires:  python3-pexpect
 # rpm is being imported through utils.py
 BuildRequires:  python3-rpm
+# DBus is used by subscription.py
+BuildRequires:  python3-dbus
 %endif
 %if 0%{?rhel} && 0%{?rhel} <= 7
 BuildRequires:  pexpect
 BuildRequires:  rpm-python
+BuildRequires:  dbus-python
```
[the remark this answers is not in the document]
abadger: Interesting, okay.  At some point I'll have to look into why we need pexpect and rpm-python to build but not dbus.

convert2rhel/subscription.py
```diff
@@ -277,28 +297,144 @@ def args(self):
...
+            loggerinst.debug("Getting a subscription-manager Register object from dbus")
+            register_object = private_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/Register")
+
+            if self.password:
+                loggerinst.debug("Registering via username/password: %s" % " ".join(hide_secrets(self.args)))
+                register_object.Register(
+                    self.org or "",
+                    self.username,
+                    self.password,
+                    EMPTY_DBUS_DICT,
```
bocekm, line 384: Please replace the empty dict with the options dict enabling the `force` option.

bocekm (changes requested):

convert2rhel/systeminfo.py
abadger: The three functions involved in checking whether dbus is running have been renamed here: https://github.com/oamg/convert2rhel/pull/540/commits/f0f7ffd0f5b51b824abc6e0f80f3fa5a46c8352c

convert2rhel/subscription.py
r0x0d: Oh, I see... It totally makes sense now, and you're correct, my question was regarding the `try: finally:` block. In my point of view, I would expect that the server start would fail and start a rollback immediately (or worse, raise an exception and prevent  us to do a rollback)

convert2rhel/i18n.py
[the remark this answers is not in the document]
bocekm: I'd say - feel free to use double-space. There's no harm in doing so :)

convert2rhel/subscription.py
abadger: This has now been fixed in the PR.

convert2rhel/subscription.py
abadger: I've made this private and added unit_tests.  We're debating some more changes and cornercases in the serverurl so there might be a branch or two left to cover in the unit_tests but at least parse_url() (renamed) and the ValueError catch are being tested now.

convert2rhel/subscription.py
abadger: This branch is now tested.  But we're looking into some changes to this behaviour (error instead of pass).

bocekm:
Rebased to resolve a conflict coming from the merged https://github.com/oamg/convert2rhel/pull/552.

convert2rhel/subscription.py
```diff
@@ -245,9 +278,15 @@ def from_tool_opts(cls, tool_opts):
...
-            registration_attributes["server_url"] = tool_opts.serverurl
+        if tool_opts.rhsm_hostname:
+            loggerinst.debug("    ... using custom RHSM hostname")
+            self.rhsm_hostname = tool_opts.rhsm_hostname
+        if tool_opts.rhsm_port:
+            loggerinst.debug("    ... using custom RHSM port")
+            self.rhsm_port = tool_opts.rhsm_port
+        if tool_opts.rhsm_handler:
+            loggerinst.debug("    ... using custom RHSM handler")
+            self.rhsm_handler = tool_opts.rhsm_handler
```
bocekm, line 289:
```suggestion
            registration_attributes["rhsm_hostname"] = tool_opts.rhsm_hostname
        if tool_opts.rhsm_port:
            loggerinst.debug("    ... using custom RHSM port")
            registration_attributes["rhsm_port"] = tool_opts.rhsm_port
        if tool_opts.rhsm_handler:
            loggerinst.debug("    ... using custom RHSM handler")
            registration_attributes["rhsm_handler"] = tool_opts.rhsm_handler
```

convert2rhel/subscription.py
```diff
@@ -120,28 +134,43 @@ def register_system():
...
+    else:  # While-else
+        # We made the maximum number of subscription-manager retries and still failed
+        loggerinst.critical("Unable to register the system through subscription-manager.")
+
+    return None
 
 
 class RegistrationCommand(object):
-    def __init__(self, activation_key=None, org=None, username=None, password=None, server_url=None):
+    def __init__(
```
bocekm, line 164: You removed `server_url`. Because of that you need to remove the line `self.server_url = server_url` from the `__init__()` body.

Also, some unit tests expect the parameter to be present, e.g. `test_calling_registration_command_with_serverurl`.

convert2rhel/toolopts.py
```diff
@@ -19,7 +19,9 @@
 import logging
 import optparse
 import os
+import re
 import sys
+import urllib.parse
```
bocekm, line 24: `No name 'parse' in module 'urllib'` on CentOS Linux 7.

convert2rhel/toolopts.py
```diff
@@ -43,7 +45,9 @@ def __init__(self):
         self.enablerepo = []
         self.disablerepo = []
         self.pool = None
-        self.serverurl = None
```
bocekm, line 46: There are some unit tests working with the toolopts.serverurl, like `test_custom_serverurl`.

convert2rhel/systeminfo.py
```diff
@@ -359,6 +364,80 @@ def corresponds_to_rhel_eus_release(self):
...
+            status = False
+
+        return status
+
+
+def _is_sysv_managed_dbus_running():
+    """Get DBus status from SysVinit compatible systems."""
+    # None means the status should be retried because we weren't sure if it is turned off.
+    running = None
+    output, _ret_code = utils.run_subprocess(["/sbin/service", "messagebus", "status"])
```
bocekm, line 403: `FileNotFoundError: [Errno 2] No such file or directory: '/sbin/service': '/sbin/service'` on CentOS Linux 8.

bocekm (changes requested):
Pylint/unit test errors.

convert2rhel/toolopts.py
r0x0d: The CI is failing right now with the same errror

https://download.copr.fedorainfracloud.org/results/@oamg/convert2rhel/epel-7-x86_64/04726137-convert2rhel/builder-live.log.gz
https://download.copr.fedorainfracloud.org/results/@oamg/convert2rhel/epel-6-x86_64/04726137-convert2rhel/builder-live.log.gz

convert2rhel/subscription.py
```diff
@@ -245,9 +279,15 @@ def from_tool_opts(cls, tool_opts):
...
-            registration_attributes["server_url"] = tool_opts.serverurl
+        if tool_opts.rhsm_hostname:
+            loggerinst.debug("    ... using custom RHSM hostname")
+            self.rhsm_hostname = tool_opts.rhsm_hostname
+        if tool_opts.rhsm_port:
+            loggerinst.debug("    ... using custom RHSM port")
+            self.rhsm_port = tool_opts.rhsm_port
+        if tool_opts.rhsm_prefix:
+            loggerinst.debug("    ... using custom RHSM prefix")
+            self.rhsm_prefix = tool_opts.rhsm_prefix
```
r0x0d, line 290: `self` is not defined in this function

r0x0d, line 290: `self` is not defined here, it's a `classmethod`

```suggestion
            cls.rhsm_hostname = tool_opts.rhsm_hostname
        if tool_opts.rhsm_port:
            loggerinst.debug("    ... using custom RHSM port")
            cls.rhsm_port = tool_opts.rhsm_port
        if tool_opts.rhsm_prefix:
            loggerinst.debug("    ... using custom RHSM prefix")
            cls.rhsm_prefix = tool_opts.rhsm_prefix
```

convert2rhel/subscription.py
abadger: This has been updated to follow ptoscano's suggestion.  Rationale here: https://github.com/oamg/convert2rhel/pull/540#discussion_r943661175

convert2rhel/subscription.py
abadger: Thanks.  Fixed.

convert2rhel/subscription.py
abadger: Yeah, there were a lot of these left to be fixed.  I wanted to get the code up where   it could be looked at asap.  I think I've now fixed all of these places.

convert2rhel/toolopts.py
abadger: Yeah, urllib.parse was introduced in python3.  I switched the imports to use `from six.moves import urllib` which should resolve this.

convert2rhel/toolopts.py
abadger: unit tests should now be updated to stop using serverurl

convert2rhel/subscription.py
abadger: Michal actually had a slightly different solution here: https://github.com/oamg/convert2rhel/pull/540#discussion_r943563943

It's a little tricky, if you set `cls.rhsm_hostname`, then you are setting the attribute on the class.  And then when the instance of the class is initialized in `__init__()`, the instance attribute will mask the class attribute.  (ie: this would end up being None because it is set to None by `__init__()`).  Adding the values to `registration_attributes` which are the parameters which are passed into `__init__()` is the right way to fix my mistake.

convert2rhel/toolopts.py
```diff
@@ -464,5 +496,43 @@ def options_from_config_files(cfg_path=None):
...
+def _parse_subscription_manager_serverurl(serverurl):
+    """Parse a url string in a manner mostly compatible with subscription-manager --serverurl."""
+    # This is an adaptation of what subscription-manager's cli enforces:
+    # https://github.com/candlepin/subscription-manager/blob/main/src/rhsm/utils.py#L112
+
+    # Don't modify http://<something> and https://<something> as they are fine
+    if not re.match("https?://[^/]+", serverurl):
+        # Anthing that looks like a malformed scheme is immediately discarded
+        if re.match("^[^:]+:/.+", serverurl):
+            raise ValueError("serverurl looks unparsable. Make sure it starts with http://HOST or https://HOST")
```
r0x0d, line 508: Does anyone know a way to rephrase this? The sentence starting with a lowercase seems strange. Maybe:

```suggestion
            raise ValueError("Problem while parsing serverurl. Make sure it starts with http://HOST or https://HOST")
```

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -761,37 +766,118 @@ def test_args(self, registration_kwargs):
...
+        monkeypatch.setattr(utils, "run_subprocess", run_subprocess_mocked)
+
         reg_cmd()
 
-        utils.run_cmd_in_pty.assert_called_once_with(
-            ["subscription-manager", "register", "--force", "--username=me_myself_and_i"],
-            expect_script=(("[Pp]assword: ", "a password\n"),),
-            print_cmd=False,
+        assert run_subprocess_mocked.call_count == 1
+        print(run_subprocess_mocked.call_args.args[0])
```
r0x0d, line 854: I think you forgot a print here

convert2rhel/toolopts.py
abadger:
```suggestion
            raise ValueError("Unable to parse --serverurl.  Make sure it starts with http://HOST or https://HOST")
```

convert2rhel/unit_tests/subscription_test.py
abadger: Thanks.  That code was failing on centos8 so I was trying to see what was going on.  I think I have it fixed and all the prints have been removed.

convert2rhel/systeminfo.py
abadger: I think I've tracked this down to the pretend_os fixture which populates system_info.  Inside of a container we might not have an init system so we need to hardcode a value for dbus_running rather than trying to use `service` or `systemctl` to probe for it.  Should be fixed now.

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -494,6 +494,10 @@ def test_register_system_fail_interactive(self, tool_opts, monkeypatch, caplog,
     @pytest.mark.rhsm_returns((KeyboardInterrupt("bang")))
     def test_register_system_keyboard_interrupt(self, tool_opts, monkeypatch, caplog, mocked_rhsm_register):
         """Test that we stop retrying if the user hits Control-C.."""
+
+        print(os.environ)
```
bocekm, line 498: Unfortunately there's nothing more than in a docker image when executed locally:
```
 {'LANG': 'C', 'APP_MAIN_DEPS': 'python-six     dbus-python     pexpect', 'LC_ALL': 'C', 'PYTHON': 'python2', 'HOSTNAME': 'c80479088321', 'PYTHONDONTWRITEBYTECODE': '1', 'PYTEST_CURRENT_TEST': 'convert2rhel/unit_tests/subscription_test.py::TestRegisterSystem::test_register_system_keyboard_interrupt (call)', 'APP_DEV_DEPS': 'requirements/centos7.requirements.txt', 'URL_GET_PIP': 'https://bootstrap.pypa.io/pip/2.7/get-pip.py', 'HOME': '/home/app', 'PIP': 'pip', 'PATH': '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'}
 ```

**Comment by bocekm** [comment]
The integration tests fail on `test_terminate_on_subscription_prompt` (https://github.com/oamg/convert2rhel/blob/main/tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py#L150):
```
System registration failed with error: com.redhat.RHSM1.Error: {"exception": "DBusException", "message": "This system is already registered", "severity": "error"}
```

File: convert2rhel/unit_tests/subscription_test.py (hunk 49 above)
Comment (abadger) [reply convert2rhel/unit_tests/subscription_test.py]: mbocekm later spotted that this test was calling `register_system()` twice and one of those was outside of the `with pytest.raises()`.  I've removed the extra `register_system()` and the unit_tests now pass.

File: convert2rhel/subscription.py [hunk 50]
```diff
@@ -120,33 +134,57 @@ def register_system():
...
         loggerinst.info(
             "%sRegistering the system using subscription-manager ...",
             attempt_msg,
         )
 
+        # Force the system to be unregistered
+        # The subscription-manager DBus API has a force parameter but there's
+        # a bug in susbcription-manager where that doesn't take effect.
+        # Explicitly unregistration here to workaround that.
+        loggerinst.info("Try to unrregister the system to clear the server's state for our registration.")
```
r0x0d, line 147: Typo
```suggestion
        loggerinst.info("Try to unregister the system to clear the server's state for our registration.")
```

convert2rhel/subscription.py
```diff
@@ -277,28 +338,125 @@ def args(self):
...
+
+        try:
+            # Use the private bus to register the machine
+            loggerinst.debug("Connecting to the private DBus")
+            private_bus = dbus.connection.Connection(address)
+            loggerinst.debug("Getting a subscription-manager Register object from dbus")
+            register_object = private_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/Register")
+
+            if self.password:
+                loggerinst.debug("Registering via username/password: %s" % " ".join(hide_secrets(self.args)))
```
r0x0d, line 407: A minor thing here: 

I feel that this should be a `info` instead of a `debug` to actually inform the user that we are registering the system. What you think?

bocekm: I don't think it's necessary. This is what they see now.

``` subscription-manager registration info ... 
09:45:31                 out:     ... activation key not found, using given username and password
09:45:31                 out:     ... username detected
09:45:31                 out:     ... password detected
```

If the credentials are incorrect, they'll get the `"Invalid username or password."` error. Then they can look at the convert2rhel.log file to see what username they used or even better look at how they called convert2rhel so they can see even the password they used.

r0x0d: Makes sense to me! Thanks

bocekm (changes requested):

convert2rhel/subscription.py
```diff
@@ -120,33 +137,86 @@ def register_system():
...
+            unregister_system()
+        except UnregisterError as e:
+            loggerinst.warning(str(e))
+
+            # Hack: currently, on RHEL7, subscription-manager unregister is
+            # reporting that the system is not registered but then calling the
+            # subscription-manager Registration API reports that the system is
+            # already registered. For the short term, we are going to stop the
+            # rhsm service to work around this issue:
+            loggerinst.info("Stopping rhsm so that registration does not think that the host is already registered.")
```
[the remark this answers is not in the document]
abadger: I've changed this in the latest push.

convert2rhel/toolopts.py
```diff
@@ -345,7 +348,36 @@ def _process_cli_options(self):
...
+                        "Please check for typos and run convert2rhel again with a corrected --serverurl.\n"
+                        "Supplied serverurl: %s\nError: %s" % (parsed_opts.serverurl, e)
+                    )
+
+                tool_opts.rhsm_hostname = url_parts.hostname
+
+                if url_parts.port:
+                    # urllib.parse.urlsplit() convert2 this into an int but we
+                    # always use it as a str
+                    tool_opts.rhsm_port = str(url_parts.port) if url_parts.port else ""
```
bocekm, line 377:
```suggestion
                    tool_opts.rhsm_port = str(url_parts.port)
```
This line already is under the condition `if url_parts.port:`.

bocekm (changes requested):

convert2rhel/subscription.py
```diff
@@ -277,28 +377,125 @@ def args(self):
...
+                * dbus-1.12.x is installed and run by default
+                * python3-dbus-1.2.x is available
+
+        .. seealso::
+            Documentation for the subscription-manager dbus API:
+            https://www.candlepinproject.org/docs/subscription-manager/dbus_objects.html
+        """
+        # Note: dbus doesn't understand empty python dicts. Use dbus.Dictionary({}, signature="ss")
+        # if we need one in the future.
+        REGISTER_OPTS_DICT = dbus.Dictionary({"force": True}, signature="ss")
```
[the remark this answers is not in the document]
abadger: `ofrce` in `REGISTER_OPTS` not working seems to be a bug.  So it seems like it would be better to leave it in?

convert2rhel/toolopts.py
abadger: Good point.  I have a vague recollection that the `else ""` might be needed so I changed the code to not do `if url_parts.port` first.

convert2rhel/subscription.py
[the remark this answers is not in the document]
abadger: Comment in the code has been updated to link to the bugzilla.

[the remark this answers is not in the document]
bocekm: Ok, let's leave this in.

convert2rhel/subscription.py
```diff
@@ -277,28 +376,125 @@ def args(self):
...
```
bocekm, line 423: The signature is incorrect. The registration fails on OL 8.6 with:
```
ERROR:dbus.connection:Unable to set arguments ('', 'username', 'password', dbus.Dictionary({'force': True}, signature=dbus.Signature('ss')), dbus.Dictionary({'host': 'subscription.rhsm.redhat.com'}, signature=dbus.Signature('ss')), 'C') according to signature None: <class 'TypeError'>: Expected a string or unicode object
```

convert2rhel/subscription.py
```diff
@@ -22,7 +22,12 @@
 from collections import namedtuple
 from time import sleep
 
-from convert2rhel import backup, pkghandler, utils
+import dbus
+import dbus.connection
```
bocekm, line 26:
```suggestion
```
Unless it's about some kind of optimization, it isn't necessary.

convert2rhel/subscription.py
```diff
@@ -96,14 +114,11 @@ def unregister_system():
...
+
     unregistration_cmd = ["subscription-manager", "unregister"]
     output, ret_code = utils.run_subprocess(unregistration_cmd, print_output=False)
     if ret_code != 0:
-        loggerinst.warning(
-            "System unregistration failed with return code %d and message:\n%s",
-            ret_code,
-            output,
-        )
+        raise UnregisterError("System unregistration failed with return code %d and message:\n%s" % (ret_code, output))
```
bocekm, line 121:
```suggestion
        raise UnregisterError("System unregistration result:\n%s" % output)
```
Because most customers will see this warning (before registering), I'd ease out the wording.

bocekm (changes requested):

convert2rhel/subscription.py
abadger: This should be fixed.  Correct signature is `dbus.Dictionary({}, signature="sv", variant_level=1)`.  If that doesn't work (It's possible I don't understand variant_level) then we can removethe register_opts dict for this PR and revisit adding it back in later.

convert2rhel/subscription.py
abadger: I like to have this because whether this works: `import dbus ; dbus.connection` is an implementation detail of the dbus module.  It will work if `dbus/__init__.py` has something like:
``` python
from dbus import connection
```
If `dbus/__init__.py` does not have something like that line, then `import dbus.connection` would be necessary in our code.

convert2rhel/subscription.py
```diff
@@ -277,28 +377,173 @@ def args(self):
...
+            else:
+                # On success, we need to set the connection opts as well
+                self._set_connection_opts_in_config()
+
+        finally:
+            # Always shut down the private bus
+            loggerinst.debug("Shutting down private DBus instance")
+            register_server.Stop(i18n.SUBSCRIPTION_MANAGER_LOCALE, dbus_interface="com.redhat.RHSM1.RegisterServer")
+
+        loggerinst.info("Successfully registered")
```
bocekm, line 493: I think you added this later after you added `loggerinst.info("System registration succeeded.")`.

Now two similar messages are printed:
```
Successfully registered
System registration succeeded.
```

convert2rhel/subscription.py
```diff
@@ -120,33 +136,94 @@ def register_system():
...
+
+    return None
+
+
+def _stop_rhsm():
+    """Stop the rhsm service."""
+    cmd = ["/bin/systemctl", "stop", "rhsm"]
+    if system_info.version.major <= 6:
+        # RHEL6 used upstart.  After that, we use systemd
+        cmd = ["service", "rhsm", "stop"]
```
bocekm, line 205:
```
[08/17/2022 20:01:59] DEBUG - Calling command 'service rhsm stop'
Stopping RHSM failed with code: 1; output: rhsm: unrecognized service
```

I looked at the services (`service --status-all`) and there's just `rhsmcertd` which I'm not sure makes sense stopping (service description: `Enable periodic update of entitlement certificates.`).
```
rpc.svcgssd is stopped
rpc.mountd is stopped
nfsd is stopped
rpc.statd (pid  1162) is running...
master (pid  1326) is running...
rdisc is stopped
restorecond is stopped
rhsmcertd is stopped
rpcbind (pid  1140) is running...
rpc.gssd is stopped
rpc.idmapd is stopped
rpc.svcgssd is stopped
rsyslogd (pid  1111) is running...
sandbox is stopped
saslauthd is stopped
openssh-daemon (pid  19422) is running...
```

bocekm (changes requested):

bocekm (approved):
Tested manually and works well. Code looks good. My previous review doesn't talk about anything critical.

convert2rhel/subscription.py
abadger: I agree.  I think I added this as I was writing unit_tests and realized that there isn't much to go on for a successful exit.  I'll remove it and depending on returning None as meaning success here.

convert2rhel/subscription.py
abadger: Hmmm.... Okay, so I just ran our dbus test script on CentOS6 and then looked at the process list and found this:
> 3485 ?        Sl     0:00 python /usr/libexec/rhsm-service

I think that rhsm-service is started on-demand when something uses the DBus API.  On systemd hosts (CentOS7 and up), this is handled by systemd which waits for a socket to be opened and then uses that to launch rhsm-service so that it can handle the request.  On centos6, the subscription-manager code must do this itself and they haven't made a SysVinit script to start the service there.  If we need to stop the service, I think we'll need to find the rhsm-service process and send it a `kill -TERM`.

I haven't been able to provoke the system is already registered error with the dbus reg.py script so I'm not certain whether RHEL6 has this problem or not.

Here's what I propose... I will take out the code path that operates on CentOS-6.  If you want to look into whether it's needed on CentOS 6 (or if we want to wait to see if a user runs into a problem) we can implement sending a signal to rhsm-service then.  Does that sound good?

bocekm: I was also thinking that on RHEL6-like systems we may not need to stop the service because it works differently there. So leaving the service-stopping out for OS 6 sounds good to me. Thanks for looking into it.

convert2rhel/subscription.py
```diff
@@ -201,8 +201,12 @@ def _stop_rhsm():
     """Stop the rhsm service."""
     cmd = ["/bin/systemctl", "stop", "rhsm"]
     if system_info.version.major <= 6:
-        # RHEL6 used upstart.  After that, we use systemd
-        cmd = ["service", "rhsm", "stop"]
+        # On RHEL6, there isn't a service-oriented way to stop rhsm.  It is started on demand so
+        # there isn't an init script to stop it.  If we find we need to stop it, because we're
+        # etting "machine is already registered" errors there, then we'll need to look for
+        # rhsm-service in the process list and send it the TERM signal.
+        loggerinst.info("Skipping RHSM service shutdown on CentOS 6.")
```
bocekm, line 208:
```suggestion
loggerinst.info("Skipping RHSM service shutdown on {0} {1}.".format(system_info.name, system_info.version.major))
```
The message would be confusing on Oracle Linux 6. But it's nothing major.

abadger:
With the latest changes, OEL-8.6 has passed integration tests.  I'm going to update the log message that Michal mentioned and then make sure all of the integration tests are restarted successfully.

danmyway:
The `check_cve` test got stuck for some reason.
Yesterday I have sent the test tiers separately to the testing farm, where it passed. 
Results to be seen here: http://artifacts.osci.redhat.com/testing-farm/2f67bf47-dee8-4b73-9671-22fa75563ba5/
Specifically the `check_cve` test here http://artifacts.osci.redhat.com/testing-farm/2f67bf47-dee8-4b73-9671-22fa75563ba5/work-check_cvejwWpdh/log.txt
Also I have run the test locally and it passed

danmyway:
Another successful separate run http://artifacts.osci.redhat.com/testing-farm/1ce38c8c-cd01-4c2c-8c85-c5c71f1b83cf/

bocekm:
After talking with @danmyway about details of the integration test runs on OL 8.6, all of them passed when he scheduled them manually outside the GitHub CI. He considers the PR verified.

bocekm (approved):

PR #546, opened 2022-07-22
[RHELC-640] chore(renovate): Ignore CentOS 6 deps

SpyTec:
Ignoring CentOS Linux 6 dependencies in requirements file as we don't want to update them.

This also only allows 7.0.x updates for pytest with centos8 requirements file

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,5 +2,13 @@
   "extends": [
     "config:base"
   ],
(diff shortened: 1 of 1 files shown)


r0x0d (approved):
Integration tests are not necessary to pass for this PR. Merging it right now.

PR #547, opened 2022-07-22
Improve GPG key import and repository-related messages

bocekm:
Repositories should not be abbreviated to "repos" in user facing messages.

The GPG key message was mentioning "keys" being imported while it referred to a single key.

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -592,7 +592,9 @@ def check_package_updates():
 
     if len(packages_to_update) > 0:
         repos_message = (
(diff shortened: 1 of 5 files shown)


r0x0d (approved):
Looks good to me. The only thing is that codecov is pointing out that some lines are not covered by unit_tests. Would like to make them right now, or tackle this in another PR? 

For me, I would say it's not something critical, and we can merge it.

abadger:
I agree with @r0x0d that since the lines are presently not covered and the code inside is just to log messages when certain conditions are true that we don't have to add new unittests for those cases in this PR.  Merging.

PR #550, opened 2022-07-26
Add type stubs for six library

r0x0d:
Add the type stubs for the six library so we can have better insights when
using static analysis tools, like pyright.

### DIFF
--- a/requirements/centos7.requirements.txt
+++ b/requirements/centos7.requirements.txt
@@ -4,3 +4,4 @@ pytest==4.6.11
 pathlib2==2.3.7.post1
 mock==3.0.5
 pytest-cov==2.12.1
+types-six==1.16.18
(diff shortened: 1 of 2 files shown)


bocekm:
@r0x0d, do we use pyright or something else that would use the types-six?

r0x0d:
For vscode, the default is pylance right now, but I believe any text editor that has a static code analysis plugin would benefit from this library. 

I need to double-check this, because when I did the PR I was testing out on vim... I will check for vscode and pyhcharm.

r0x0d:
Turns out that pycharm has a different mechanism to enable type support for `six`, and this method didn't work as expected for vscode :thinking:.

Maybe I got fooled by another solution. Closing this PR since it seems to not work properly, and it's not required now.

PR #551, opened 2022-07-29
WIP: Enable integration testing for RHEL 6 conversions 

kokesak:
Enable conversion to RHEL-6.

Current stauts:
Image in TF is not working. We need to upload/update working image with Centos6. One option can be to use from 1minutetip

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -40,9 +40,11 @@ jobs:
 - job: tests
   metadata:
     targets:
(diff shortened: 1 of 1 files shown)


bocekm:
Next step: Adding CentOS Linux and Oracle Linux 6 images to Artemis.

SpyTec:
Closing. We won't be doing any more el6 conversions after next version

PR #553, opened 2022-08-03
[RHELC-678] Remove UEFI condition from shim-x64 workaround

r0x0d:
shim-x64 packages can be installed on systems that are running BIOS too,
so our previous workaround included only UEFI systems.

This commit removes that condition by backing up and removing the
shim-x64 package protection on both UEFI and BIOS now.

### DIFF
--- a/tests/integration/tier1/handle-special-pkgs-case/test_handle_shim_x64_pkg.py
+++ b/tests/integration/tier1/handle-special-pkgs-case/test_handle_shim_x64_pkg.py
@@ -26,7 +26,7 @@ def test_handle_shim_x64_pkg(shell, convert2rhel):
             env.str("RHSM_POOL"),
         )
     ) as c2r:
-        c2r.expect_exact("Removing shim-x64 package yum protection.") == 0
+       assert c2r.expect_exact("Removing shim-x64 package yum protection.") == 0
         c2r.expect("removed in accordance with")
     assert c2r.exitstatus == 0
 
--- a/tests/integration/tier1/handle-special-pkgs-case/test_handle_shim_x64_pkg.py
+++ b/tests/integration/tier1/handle-special-pkgs-case/test_handle_shim_x64_pkg.py
@@ -26,7 +26,7 @@ def test_handle_shim_x64_pkg(shell, convert2rhel):
             env.str("RHSM_POOL"),
         )
     ) as c2r:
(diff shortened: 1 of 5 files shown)


tests/integration/tier1/handle-special-pkgs-case/test_handle_shim_x64_pkg.py
```diff
@@ -0,0 +1,36 @@
...
+    # run utility until the reboot
+    with convert2rhel(
+        ("-y --no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug").format(
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        c2r.expect_exact("Removing shim-x64 package yum protection.") == 0
```
danmyway, line 29:
```suggestion
       assert c2r.expect_exact("Removing shim-x64 package yum protection.") == 0
```
I would just assert this to not waste time waiting for child.expect_exact() time out

bocekm (approved):

SpyTec:
@danmyway can you re-review the integration tests? If it looks good to you please merge

tests/integration/tier1/handle-special-pkgs-case/test_handle_shim_x64_pkg.py
danmyway: @r0x0d I believe the suggestion commit got overwritten by rebasing, I am committing that again

danmyway (approved):
Added previously agreed submission otherwise looks good to me.

bocekm:
On a BIOS-based system without the shim package installed we print:
```
09:46:41                 out: Removing shim-x64 package yum protection.
09:46:41                 out: Backing up /etc/yum/protected.d/shim-x64.conf.
09:46:41                 out: Can't find /etc/yum/protected.d/shim-x64.conf.
09:46:41                 out: ERROR - Unable to remove '/etc/yum/protected.d/shim-x64.conf': No such file or directory
09:46:41                 out: Checking if the iwl7260-firmware and iwlax2xx-firmware packages are installed.
```

Before trying to remove the file we should check whether it exists (or whether the package is installed).

r0x0d:
I will do a quick fix for that.

PR #555, opened 2022-08-04
[RHELC-673] convert2rhel rpm to require efibootmgr installed

bocekm:
In https://github.com/oamg/convert2rhel/pull/223 we introduced support for UEFI-based system conversions. With that we call the efibootmgr utility on a couple of places. Yet, we don't require the utility to be installed and that might lead into unwanted behaviour of convert2rhel.

I've verified that the package is named the same across all major versions of RHEL we allow converting.

Jira: https://issues.redhat.com/browse/RHELC-673

### DIFF
--- a/packaging/convert2rhel.spec
+++ b/packaging/convert2rhel.spec
@@ -31,6 +31,7 @@ BuildRequires:  pexpect
 BuildRequires:  rpm-python
 %endif
 
+Requires:       efibootmgr
 Requires:       rpm
 Requires:       python%{python_pkgversion}
 Requires:       python%{python_pkgversion}-setuptools


r0x0d:
Changing the title to include the RHELC key.

r0x0d (approved):

r0x0d (approved):

pirat89:
@bocekm @r0x0d  just thinking about this PR, in this way you will require to install the package (with possible deps) on systems where it's useless (bios boot). the efibootmgr utility should be used only on EFI systems (where it's already required, otherwise the system is basically in unsupported state). wouldn't be better to add a check with err msg that efibootmgr must be installed on systems with EFI boot?

bocekm:
> the efibootmgr utility should be used only on EFI systems (where it's already required, otherwise the system is basically in unsupported state)

@pirat89, I've just removed efibootmgr without problems from a UEFI-based system. Where is it stated that not having this tool installed makes the system unsupported? From the rpm description: _"efibootmgr displays and allows the user to edit the Intel Extensible Firmware Interface (EFI) Boot Manager variables."_ 

We had a similar discussion around adding grub2-tools to the spec file (because not every system needs to be using GRUB2): https://github.com/oamg/convert2rhel/pull/417#issuecomment-1154623887.

I wrote there something that seems to fit this case as well:

> It seems to me that adding grub2-tools as a new convert2rhel dep is the most elegant solution ideal for most people. Yes, the downside is that the dep is unnecessary bloat in case GRUB2 is not being used. What are the other options of installing the dep though? I can imagine installing the dep during the convert2rhel execution only if GRUB2 is detected. But it seems to me like adding unnecessary complexity to the conversion process that can be prone to issues and worsening user experience.

pirat89:
> @pirat89, I've just removed efibootmgr without problems from a UEFI-based system. Where is it stated that not having this tool installed makes the system unsupported? From the rpm description: "efibootmgr displays and allows the user to edit the Intel Extensible Firmware Interface (EFI) Boot Manager variables."

Well, sorry for the possibly misleading information. Well I discussed that stuff in the past with bootloader guys (so not answering the supportability question from the CEE POV)  and I've been told that EFI system without the efibootmgr tool is basically nonsense. It's possible, but it's always expected to have efibootmgr installed when you have UEFI-based system. That's also one of reasons why we just write the report for user to install the efibootmgr  pkg on EFI systems to users.

> We had a similar discussion around adding grub2-tools to the spec file (because not every system needs to be using GRUB2): https://github.com/oamg/convert2rhel/pull/417#issuecomment-1154623887

Interesting. I thought that conversions support only intel architectures where I would expect that GRUB2 is the only supported bootloader (excluding RHEL 6 using Legacy GRUB). I know that you could still find somewhere lilo that is kinda death and there is possibility of windows bootloader on dual boot, but that is unsupported as well.

PR #556, opened 2022-08-05
fix el6 and el7 builds in Packit

TomasTomecek:
workaround for https://github.com/packit/packit-service/issues/1601

before we resolve the root cause - just to unblock you


TomasTomecek:
> TomasTomecek can't run tests (and builds) internally

:( :D

r0x0d (approved):

PR #562, opened 2022-08-09
[RHELC-691] Fix of the cfg file messages 

hosekadam:
Edits of convert2rhel help option and corner cases messages.
Based on [Michal's review.](https://github.com/oamg/convert2rhel/pull/451#discussion_r940112480)

### DIFF
--- a/convert2rhel/toolopts.py
+++ b/convert2rhel/toolopts.py
@@ -79,13 +79,12 @@ class CLI(object):
             "\n"
             "  convert2rhel [-h]\n"
             "  convert2rhel [--version]\n"
-            "  convert2rhel [-u username] [-p password | -f pswd_file] [--pool pool_id | -a] [--disablerepo repoid]"
-            "  convert2rhel [-c | --config-file conf_file]"
+            "  convert2rhel [-u username] [-p password | -c conf_file_path] [--pool pool_id | -a] [--disablerepo repoid]"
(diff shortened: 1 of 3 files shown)


hosekadam:
Maybe editing the unit tests would be good idea... :D My bad

SpyTec (approved):

bocekm (changes requested):

convert2rhel/toolopts.py
```diff
@@ -200,8 +202,9 @@ def _register_options(self):
         group.add_option(
             "-c",
             "--config-file",
-            help="Configuration file containing"
-            " password or activation key for subscription-manager."
+            help="A configuration file to safely provide either a user password or an activation key for registering"
+            " the system through subscription-manager. Alternatively, passing these values through the"
+            " --activationkey or --password option would leak them through a list of running processes."
```
r0x0d, line 207: Just wanted to open this comment for a question, wouldn't be better to use “would leak sensitive information through a list of running processes” instead of specifying that the activation key or password would leak?

r0x0d: I feel that we can cover everything with “sensitive information” in this context, even if we find other types of things that can be leaked, as this kind of text is way general everywhere.

bocekm: Currently the config file is used just for the two specific pieces of information.
I expect that once we add more config options we'll need to update the description here.
Till then IMHO it's better to be specific - this way it's clear. The way you propose it would make it open to questions and assumptions.

bocekm (approved):

PR #563, opened 2022-08-10
[RHELC-693] Check if org and key passed

hosekadam:
The check making sure the organization is specified by the user is currently skipped when you don't specify the organization on the command line AND you pass the activation key through either the CLI OR the config file.

[RHELC-693](https://issues.redhat.com/browse/RHELC-693)

### DIFF
--- a/convert2rhel/toolopts.py
+++ b/convert2rhel/toolopts.py
@@ -420,6 +420,11 @@ class CLI(object):
                 " We're going to use the password file."
             )
 
+        if (tool_opts.org and not tool_opts.activation_key) or (not tool_opts.org and tool_opts.activation_key):
+            loggerinst.critical(
+                "Either the --organization or the --activationkey option is missing. You can't use one without the other."
+            )
+
         if tool_opts.username and tool_opts.password:
             tool_opts.credentials_thru_cli = True
 
--- a/convert2rhel/unit_tests/toolopts_test.py
+++ b/convert2rhel/unit_tests/toolopts_test.py
@@ -213,28 +213,28 @@ def test_both_disable_submgr_and_no_rhsm_options_work(argv, raise_exception, no_
             " the configuration file. We're going to use the command line values.",
         ),
         (
-            mock_cli_arguments(["-k", "activation_key"]),
+            mock_cli_arguments(["-k", "activation_key", "-o", "org"]),
             "[subscription_manager]\nactivation_key=conf_key",
             {"password": None, "activation_key": "activation_key"},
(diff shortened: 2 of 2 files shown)


hosekadam:
Not working now, I've did some "tuning which can't broke anything" and it broke...

convert2rhel/toolopts.py
```diff
@@ -375,6 +375,9 @@ def _process_cli_options(self):
         if (config_opts.activation_key or config_opts.password) and parsed_opts.password_from_file:
             loggerinst.warning("Password file take precedence over the config file.")
 
+        if (tool_opts.org and not tool_opts.activation_key) or (not tool_opts.org and tool_opts.activation_key):
+            loggerinst.critical("You haven't passed organization with activation key. Please pass both.")
```
r0x0d, line 379: So, I did a bit of rephrasing in the message to sound more “generic” and not assume that `organization` is present and `acitivation key` is not (or the other way around).

```suggestion
            loggerinst.critical("Either organization or activation key are missing. You have to use both of them.")
```

If you feel that it doesn't make sense, feel free to resolve this comment.

hosekadam: Changed!

r0x0d:
@hosekadam please, rebase when you can.

hosekadam:
@r0x0d Done!

r0x0d (dismissed):

r0x0d:
@hosekadam, I rebased this PR to trigger the integration tests with the latest changes we had in upstream.

r0x0d:
Oh, it seems that we have a merge conflict here, @hosekadam. Could you take care of it whenever you have time?

convert2rhel/toolopts.py
```diff
@@ -420,6 +420,9 @@ def _process_cli_options(self):
                 " We're going to use the password file."
             )
 
+        if (tool_opts.org and not tool_opts.activation_key) or (not tool_opts.org and tool_opts.activation_key):
+            loggerinst.critical("Either organization or activation key are missing. You have to use both of them.")
```
bocekm, line 424:
```suggestion
            loggerinst.critical("Either the --organization or the --activationkey option is missing. You can't use one without the other.")
```

bocekm (changes requested):

r0x0d (approved):
Looks good from my point of view, @hosekadam! Not sure if we need integration tests here (will let @SpyTec or @danmyway to decide).

bocekm (approved):

PR #566, opened 2022-08-11
[RHELC-670] Clean up new version check

Andrew-ang9:
This PR was made to clean up PR #415, which was merged without taking care of some of the other comments that were left on the PR. Those comments weren't vital enough to stop the PR from being merged. 

All comments have been put in the Jira ticket and addressed for the first commit of this PR.

Jira ticket: [RHELC-670](https://issues.redhat.com/browse/RHELC-670)

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -21,11 +21,12 @@ import logging
 import os
 import os.path
 import re
+import shutil
 import tempfile
 
 import rpm
 
-from convert2rhel import __version__ as convert2rhel_version
+from convert2rhel import __version__ as installed_convert2rhel_version
 from convert2rhel import grub, pkgmanager, utils
 from convert2rhel.pkghandler import (
     call_yum_cmd,
@@ -45,6 +46,25 @@ logger = logging.getLogger(__name__)
 KERNEL_REPO_RE = re.compile(r"^.+:(?P<version>.+).el.+$")
 KERNEL_REPO_VER_SPLIT_RE = re.compile(r"\W+")
 BAD_KERNEL_RELEASE_SUBSTRINGS = ("uek", "rt", "linode")
+
+RPM_GPG_KEY_PATH = os.path.join(utils.DATA_DIR, "gpg-keys", "RPM-GPG-KEY-redhat-release")
+# The SSL certificate of the https://cdn.redhat.com/ server
+SSL_CERT_PATH = os.path.join(utils.DATA_DIR, "redhat-uep.pem")
+CDN_URL = "https://cdn.redhat.com/content/public/convert2rhel/$releasever/$basearch/os/"
+CONVERT2RHEL_REPO_CONTENT = """\
+[convert2rhel]
+name=Convert2RHEL Repository
+baseurl=%s
+gpgcheck=1
+enabled=1
+sslcacert=%s
+gpgkey=file://%s""" % (
+    CDN_URL,
+    SSL_CERT_PATH,
+    RPM_GPG_KEY_PATH,
+)
+
+PKG_NEVR = r"\b(\S+)-(?:([0-9]+):)?(\S+)-(\S+)\b"
 
 LINK_KMODS_RH_POLICY = "https://access.redhat.com/third-party-software-support"
 # The kernel version stays the same throughout a RHEL major version
@@ -76,81 +96,91 @@ def perform_pre_ponr_checks():
 
 
 def check_convert2rhel_latest():
-    """Make sure that we are running the latest version of convert2rhel"""
-    gpg_path = os.path.join(utils.DATA_DIR, "gpg-keys", "RPM-GPG-KEY-redhat-release")
-    ssl_cert_path = os.path.join(utils.DATA_DIR, "redhat-uep.pem")
-    repo_content = (
-        "[convert2rhel]\n"
-        "name=Convert2RHEL Repository\n"
-        "baseurl=https://cdn.redhat.com/content/public/convert2rhel/$releasever/x86_64/os/\n"
-        "gpgcheck=1\n"
-        "enabled=1\n"
-        "sslcacert=%s\n"
-        "gpgkey=file://%s\n" % (ssl_cert_path, gpg_path)
-    )
+    """Make sure that we are running the latest downstream version of convert2rhel"""
+    logger.task("Prepare: Checking if this is the latest version of Convert2RHEL")
+
+    if not system_info.has_internet_access:
+        logger.warning("Skipping the check because no internet connection has been detected.")
+        return
+
+    repo_dir = tempfile.mkdtemp(prefix="convert2rhel_repo.", dir=utils.TMP_DIR)
+    repo_path = os.path.join(repo_dir, "convert2rhel.repo")
+    store_content_to_file(filename=repo_path, content=CONVERT2RHEL_REPO_CONTENT)
 
     cmd = [
         "repoquery",
+        "--disablerepo=*",
+        "--enablerepo=convert2rhel",
         "--releasever=%s" % system_info.version.major,
+        "--setopt=reposdir=%s" % repo_dir,
+        "convert2rhel",
     ]
 
-    repo_dir = tempfile.mkdtemp(prefix="convert2rhel_repo.", dir=utils.TMP_DIR)
+    # Note: This is safe because we're creating in utils.TMP_DIR which is hardcoded to
+    # /var/lib/convert2rhel which does not have any world-writable directory components.
     utils.mkdir_p(repo_dir)
-    repo_path = os.path.join(repo_dir, "convert2rhel.repo")
-    store_content_to_file(filename=repo_path, content=repo_content)
 
-    cmd.append("--setopt=reposdir=%s" % repo_dir)
-    cmd.append("convert2rhel")
-
-    raw_output_convert2rhel_versions, return_code = run_subprocess(cmd, print_output=False)
+    try:
+        raw_output_convert2rhel_versions, return_code = run_subprocess(cmd, print_output=False)
+    finally:
+        shutil.rmtree(repo_dir)
 
     if return_code != 0:
         logger.warning(
-            "Couldn't check if this is the latest version of convert2rhel\n"
-            "repoquery failed %s, %s" % (return_code, raw_output_convert2rhel_versions)
+            "Couldn't check if the current installed Convert2RHEL is the latest version.\n"
+            "repoquery failed with the following output:\n%s" % (raw_output_convert2rhel_versions)
         )
         return
 
-    PKG_NEVR = r"\b(\S+)-(?:([0-9]+):)?(\S+)-(\S+)\b"
     convert2rhel_versions = re.findall(PKG_NEVR, raw_output_convert2rhel_versions, re.MULTILINE)
-    latest_version = ("0", "0.00", "0")
+    logger.debug("Found %s convert2rhel package(s)" % len(convert2rhel_versions))
+    latest_available_version = ("0", "0.00", "0")
 
+    # This loop will determine the latest available convert2rhel version in the yum repo.
+    # It assigns the epoch, version, and release ex: ("0", "0.26", "1.el7") to the latest_available_version variable.
     for package_version in convert2rhel_versions:
-        # rpm.lableCompare(pkg1, pkg2) compare two kernel version strings and return
-        # -1 if str2 is greater then str1, 0 if they are equal, 1 if str1 is greater the str2
-        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
+        # rpm.labelCompare(pkg1, pkg2) compare two package version strings and return
+        # -1 if latest_version is greater than package_version, 0 if they are equal, 1 if package_version is greater than latest_version
+        ver_compare = rpm.labelCompare(package_version[1:], latest_available_version)
         if ver_compare > 0:
-            latest_version = package_version[1:]
+            logger.debug(
+                "...convert2rhel version %s is newer than %s, updating latest_available_version variable"
+                % (package_version[1:][1], latest_available_version[1])
+            )
+            latest_available_version = package_version[1:]
 
-    # After the for loop latest_version will have the epoch ,version, and release ex:("0" "0.26" "1.el7") information from convert2rhel yum repo.
-    # the release for latest_verion will be "1.el7" and the relase for convert2rhel_version will be hard coded as "0" below,
-    # therefore when the versions are the same the latest_version's release field will cause it to evaluate as later
-    ver_compare = rpm.labelCompare(("0", convert2rhel_version, "0"), ("0", latest_version[1], "0"))
+    # After the for loop, the latest_available_version variable will gain the epoch, version, and release
+    # (e.g. ("0" "0.26" "1.el7")) information from the Convert2RHEL yum repo
+    # when the versions are the same the latest_available_version's release field will cause it to evaluate as a later version.
+    # Therefore we need to hardcode "0" for both the epoch and release below for installed_convert2rhel_version
+    # and latest_available_version respectively, to compare **just** the version field.
+    ver_compare = rpm.labelCompare(("0", installed_convert2rhel_version, "0"), ("0", latest_available_version[1], "0"))
     if ver_compare < 0:
-        if "CONVERT2RHEL_UNSUPPORTED_VERSION" in os.environ:
+        if "CONVERT2RHEL_ALLOW_OLDER_VERSION" in os.environ:
             logger.warning(
-                "You are currently running %s and the latest version of convert2rhel is %s.\n"
-                "'CONVERT2RHEL_UNSUPPORTED_VERSION' environment detected, continuing conversion"
-                % (convert2rhel_version, latest_version[1])
+                "You are currently running %s and the latest version of Convert2RHEL is %s.\n"
+                "'CONVERT2RHEL_ALLOW_OLDER_VERSION' environment variable detected, continuing conversion"
+                % (installed_convert2rhel_version, latest_available_version[1])
             )
 
         else:
             if int(system_info.version.major) <= 6:
                 logger.warning(
-                    "You are currently running %s and the latest version of convert2rhel is %s.\n"
-                    "Only the latest version is supported for conversion." % (convert2rhel_version, latest_version[1])
+                    "You are currently running %s and the latest version of Convert2RHEL is %s.\n"
+                    "We encourage you to update to the latest version."
+                    % (installed_convert2rhel_version, latest_available_version[1])
                 )
 
             else:
                 logger.critical(
-                    "You are currently running %s and the latest version of convert2rhel is %s.\n"
+                    "You are currently running %s and the latest version of Convert2RHEL is %s.\n"
                     "Only the latest version is supported for conversion. If you want to ignore"
-                    " this you can set 'CONVERT2RHEL_UNSUPPORTED_VERSION' to continue"
-                    % (convert2rhel_version, latest_version[1])
+                    " this check, then set the environment variable 'CONVERT2RHEL_ALLOW_OLDER_VERSION=1' to continue."
+                    % (installed_convert2rhel_version, latest_available_version[1])
                 )
 
     else:
-        logger.debug("Latest available convert2rhel version is installed.\n" "Continuing conversion.")
+        logger.debug("Latest available Convert2RHEL version is installed.\n" "Continuing conversion.")
 
 
 def check_efi():
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -322,6 +322,50 @@ def test_c2r_up_to_date(caplog, monkeypatch, convert2rhel_latest_version_test):
 
 
 @pytest.mark.parametrize(
+    ("convert2rhel_latest_version_test",),
+    ([{"local_version": "1.10", "package_version": "convert2rhel-0:0.18-1.el7.noarch", "pmajor": "8"}],),
+    indirect=True,
+)
+def test_c2r_up_to_date_repoquery_error(caplog, convert2rhel_latest_version_test, monkeypatch):
+    monkeypatch.setattr(checks, "run_subprocess", mock.Mock(return_value=("Repoquery did not run", 1)))
+
+    checks.check_convert2rhel_latest()
+
+    log_msg = (
+        "Couldn't check if the current installed Convert2RHEL is the latest version.\n"
+        "repoquery failed with the following output:\nRepoquery did not run"
+    )
+    assert log_msg in caplog.text
+
+
+@pytest.mark.parametrize(
+    ("convert2rhel_latest_version_test",),
+    (
+        [
+            {
+                "local_version": "0.19",
+                "package_version": "convert2rhel-0:0.18-1.el7.noarch\nconvert2rhel-0:0.17-1.el7.noarch\nconvert2rhel-0:0.20-1.el7.noarch",
+                "pmajor": "8",
+            }
+        ],
+    ),
+    indirect=True,
+)
+def test_c2r_up_to_date_multiple_packages(caplog, convert2rhel_latest_version_test, monkeypatch):
+
+    with pytest.raises(SystemExit):
+        checks.check_convert2rhel_latest()
+
+    log_msg = (
+        "You are currently running 0.19 and the latest version of Convert2RHEL is 0.20.\n"
+        "Only the latest version is supported for conversion. If you want to ignore"
+        " this check, then set the environment variable 'CONVERT2RHEL_ALLOW_OLDER_VERSION=1' to continue."
+    )
+
+    assert log_msg in caplog.text
+
+
+@pytest.mark.parametrize(
     (
         "host_kmods",
         "exception",
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -94,14 +94,17 @@ REPOQUERY_L_STUB_BAD = (
 
 
 @pytest.fixture
-def convert2rhel_latest_version_test(monkeypatch, tmpdir, request):
+def convert2rhel_latest_version_test(monkeypatch, tmpdir, request, global_system_info):
+    monkeypatch.setattr(checks, "system_info", global_system_info)
+    global_system_info.has_internet_access = True
+
     marker = request.param
-    monkeypatch.setattr(checks, "convert2rhel_version", marker["local_version"])
+    monkeypatch.setattr(checks, "installed_convert2rhel_version", marker["local_version"])
 
     run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(marker["package_version"], 0))
 
     monkeypatch.setattr(checks, "run_subprocess", run_subprocess_mocked)
-    monkeypatch.setattr(system_info, "version", namedtuple("Version", ["major", "minor"])(marker["pmajor"], 0))
+    monkeypatch.setattr(global_system_info, "version", namedtuple("Version", ["major", "minor"])(marker["pmajor"], 0))
     monkeypatch.setattr(utils, "TMP_DIR", str(tmpdir))
 
     return marker["local_version"], marker["package_version"]
@@ -187,36 +190,36 @@ def test_pre_ponr_checks(monkeypatch):
     create_transaction_handler_mock.assert_called_once()
 
 
-def test_repoquery__failure(caplog, monkeypatch, tmpdir, request):
-    repoquery_output = " error message"
-    return_code = 1
-    run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(repoquery_output, return_code))
-
-    monkeypatch.setattr(checks, "run_subprocess", run_subprocess_mocked)
-    monkeypatch.setattr(system_info, "version", namedtuple("Version", ["major", "minor"])(6, 0))
-    monkeypatch.setattr(utils, "TMP_DIR", str(tmpdir))
-
+@pytest.mark.parametrize(
+    ("convert2rhel_latest_version_test",),
+    ([{"local_version": "0.20", "package_version": "convert2rhel-0:0.22-1.el7.noarch", "pmajor": "7"}],),
+    indirect=True,
+)
+def test_convert2rhel_latest_offline(caplog, convert2rhel_latest_version_test, global_system_info):
+    global_system_info.has_internet_access = False
     checks.check_convert2rhel_latest()
 
-    log_msg = "Couldn't check if this is the latest version of convert2rhel\n" "repoquery failed %s, %s" % (
-        return_code,
-        repoquery_output,
-    )
+    convert2rhel_latest_version_test
+    log_msg = "Skipping the check because no internet connection has been detected."
     assert log_msg in caplog.text
 
 
 @pytest.mark.parametrize(
     ("convert2rhel_latest_version_test",),
-    ([{"local_version": "0.20", "package_version": "convert2rhel-0:0.22-1.el7.noarch", "pmajor": "6"}],),
+    (
+        [{"local_version": "0.20", "package_version": "convert2rhel-0:0.22-1.el7.noarch", "pmajor": "6"}],
+        [{"local_version": "0.18", "package_version": "convert2rhel-0:1.10-1.el7.noarch", "pmajor": "6"}],
+    ),
     indirect=True,
 )
 def test_convert2rhel_latest_out_of_date_el6(caplog, convert2rhel_latest_version_test):
     checks.check_convert2rhel_latest()
 
-    local_version, dummy_ = convert2rhel_latest_version_test
+    local_version, package_version = convert2rhel_latest_version_test
+    package_version = package_version[15:19]
     log_msg = (
-        "You are currently running %s and the latest version of convert2rhel is 0.22.\n"
-        "Only the latest version is supported for conversion." % (local_version)
+        "You are currently running %s and the latest version of Convert2RHEL is %s.\n"
+        "We encourage you to update to the latest version." % (local_version, package_version)
     )
     assert log_msg in caplog.text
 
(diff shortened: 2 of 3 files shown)


convert2rhel/checks.py
```diff
@@ -46,6 +46,18 @@
 KERNEL_REPO_VER_SPLIT_RE = re.compile(r"\W+")
 BAD_KERNEL_RELEASE_SUBSTRINGS = ("uek", "rt", "linode")
 
+GPG_PATH = os.path.join(utils.DATA_DIR, "gpg-keys", "RPM-GPG-KEY-redhat-release")
```
r0x0d, line 49: I would rename this to 

```suggestion
RPM_GPG_KEY_PATH = os.path.join(utils.DATA_DIR, "gpg-keys", "RPM-GPG-KEY-redhat-release")
```

So it's more consistent with the actual file you want to use.

Andrew-ang9: ok cool! I'll go ahead and make that change, thank you for pointing that out!

convert2rhel/checks.py
```diff
@@ -95,56 +97,59 @@ def check_convert2rhel_latest():
     repo_dir = tempfile.mkdtemp(prefix="convert2rhel_repo.", dir=utils.TMP_DIR)
     utils.mkdir_p(repo_dir)
     repo_path = os.path.join(repo_dir, "convert2rhel.repo")
-    store_content_to_file(filename=repo_path, content=repo_content)
+    store_content_to_file(filename=repo_path, content=REPO_CONTENT)
 
     cmd.append("--setopt=reposdir=%s" % repo_dir)
```
bocekm, line 102: On CentOS Linux 6:
```
WARNING - Couldn't check if this is the latest version of convert2rhel
repoquery failed 2, Usage: repoquery [options]

repoquery: error: no such option: --setopt
```

We use this option on a couple more places. We weren't testing much on CentOS/OL 6 recently. Please try to find an alternative to the setopt for this old system version.

bocekm (changes requested):

convert2rhel/checks.py
Andrew-ang9: Ok, yeah I can defiantly check this out!

convert2rhel/checks.py
```diff
@@ -46,6 +46,18 @@
...
+SSL_CERT_PATH = os.path.join(utils.DATA_DIR, "redhat-uep.pem")
+REPO_CONTENT = (
+    "[convert2rhel]\n"
+    "name=Convert2RHEL Repository\n"
+    "baseurl=https://cdn.redhat.com/content/public/convert2rhel/$releasever/$basearch/os/\n"
+    "gpgcheck=1\n"
+    "enabled=1\n"
+    "sslcacert=%s\n"
+    "gpgkey=file://%s\n" % (SSL_CERT_PATH, RPM_GPG_KEY_PATH)
+)
```
r0x0d, line 60: I would suggest using the triple quotes (`"""`) for this to make the readability better, for example: 

```suggestion
REPO_CONTENT = """\
[convert2rhel]
name=Convert2RHEL Repository
baseurl=https://cdn.redhat.com/content/public/convert2rhel/$releasever/$basearch/os/
gpgcheck=1
enabled=1
sslcacert=%s
gpgkey=file://%s
""" % (SSL_CERT_PATH, RPM_GPG_KEY_PATH)
```

Not something that you need to add only if you think that it improves the readability of the code,

convert2rhel/checks.py
```diff
@@ -95,56 +97,59 @@ def check_convert2rhel_latest():
...
-        )
-        return
-
     PKG_NEVR = r"\b(\S+)-(?:([0-9]+):)?(\S+)-(\S+)\b"
     convert2rhel_versions = re.findall(PKG_NEVR, raw_output_convert2rhel_versions, re.MULTILINE)
-    latest_version = ("0", "0.00", "0")
+    latest_available_version = ("0", "0.00", "0")
 
+    # This loop will determin the latest_available_version in the convert2rhel yum repo
+    # it assigns the epoch ,version, and release ex:("0" "0.26" "1.el7") to the latest_available_version var.
```
r0x0d, line 112:
```suggestion
    # This loop will determine the latest available convert2rhel version in the yum repo.
    # It assigns the epoch, version, and release ex: ("0", "0.26", "1.el7") to the latest_available_version variable.
```

convert2rhel/checks.py
```diff
@@ -95,56 +97,59 @@ def check_convert2rhel_latest():
...
     for package_version in convert2rhel_versions:
-        # rpm.lableCompare(pkg1, pkg2) compare two kernel version strings and return
-        # -1 if str2 is greater then str1, 0 if they are equal, 1 if str1 is greater the str2
-        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
+        # rpm.lableCompare(pkg1, pkg2) compare two package version strings and return
+        # -1 if lastest_version is greater then package_version, 0 if they are equal, 1 if package_version is greater the lastest_version
```
r0x0d, line 115:
```suggestion
        # -1 if lastest_version is greater than package_version, 0 if they are equal, 1 if package_version is greater than lastest_version
```

convert2rhel/checks.py
```diff
@@ -95,56 +97,59 @@ def check_convert2rhel_latest():
...
+    # we hard code 0 for the epoch and release below because latest_verion will
+    # contain "1.el7" for the release but we don't know what the release for
+    # installed_convert2rhel_version will be, if we did not hard code to 0 latest_available_version
+    # will evaluate as later then installed_convert2rhel_version even when the versions
+    # were the same.
+    #
+    # ver_compare is set to
+    # * -1 if lastest_version is greater then installed_convert2rhel_version,
+    # * 0 if they are equal
+    # * 1 if installed_convert2rhel_version is greater the lastest_version
```
r0x0d, line 129: Do we need this comment since we have a more generic one in line 115? I mean, this might help to identify what is the expected output of the `ver_compare` variable, but I would say it's redundant since we have the same information on line 115.

convert2rhel/checks.py
```diff
@@ -95,56 +97,59 @@ def check_convert2rhel_latest():
...
             logger.warning(
                 "You are currently running %s and the latest version of convert2rhel is %s.\n"
-                "'CONVERT2RHEL_UNSUPPORTED_VERSION' environment detected, continuing conversion"
-                % (convert2rhel_version, latest_version[1])
+                "'CONVERT2RHEL_UNSUPPORTED_VERSION' environment variable detected, continuing conversion"
+                % (installed_convert2rhel_version, latest_available_version[1])
             )
 
         else:
             if int(system_info.version.major) <= 6:
```
r0x0d, line 140: I think that you don't need to cast the `system_info.version.major` here to an integer, as this is intended to be an integer already. 

You can see other examples in the codebase that don't have this implicit cast to `int()`. Let me know if you had to add this because it was failing without the cast (it shouldn't).

convert2rhel/checks.py
```diff
@@ -95,56 +97,59 @@ def check_convert2rhel_latest():
...
                 logger.warning(
                     "You are currently running %s and the latest version of convert2rhel is %s.\n"
-                    "Only the latest version is supported for conversion." % (convert2rhel_version, latest_version[1])
+                    "We encurage you to update to the latest version"
```
r0x0d, line 143:
```suggestion
                    "We encurage you to update to the latest version."
```

convert2rhel/checks.py
```diff
@@ -95,56 +97,59 @@ def check_convert2rhel_latest():
...
+                    % (installed_convert2rhel_version, latest_available_version[1])
                 )
 
             else:
                 logger.critical(
                     "You are currently running %s and the latest version of convert2rhel is %s.\n"
                     "Only the latest version is supported for conversion. If you want to ignore"
-                    " this you can set 'CONVERT2RHEL_UNSUPPORTED_VERSION' to continue"
-                    % (convert2rhel_version, latest_version[1])
+                    " this check set the enviroment variable 'CONVERT2RHEL_UNSUPPORTED_VERSION=1' to continue."
```
r0x0d, line 151: I don't know, but it seems to me that a `, then` is missing in that phrase. It looks strange without it. 

```suggestion
                    " this check, then set the environment variable 'CONVERT2RHEL_UNSUPPORTED_VERSION=1' to continue."
```

I'm not an expert in English, so feel free to deny this comment if it's wrong!

convert2rhel/unit_tests/checks_test.py
```diff
@@ -226,8 +208,8 @@ def test_convert2rhel_latest_log_check_exit(caplog, convert2rhel_latest_version_
 
     log_msg = (
         "You are currently running %s and the latest version of convert2rhel is 0.22.\n"
-        "Only the latest version is supported for conversion. If you want to ignore this you can set 'CONVERT2RHEL_UNSUPPORTED_VERSION' to continue"
-        % (local_version)
+        "Only the latest version is supported for conversion. If you want to ignore"
+        " this check set the enviroment variable 'CONVERT2RHEL_UNSUPPORTED_VERSION=1' to continue." % (local_version)
```
r0x0d, line 212:
```suggestion
        " this check set the environment variable 'CONVERT2RHEL_UNSUPPORTED_VERSION=1' to continue." % (local_version)
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -206,7 +188,7 @@ def test_convert2rhel_latest_out_of_date_el6(caplog, convert2rhel_latest_version
     local_version, dummy_ = convert2rhel_latest_version_test
     log_msg = (
         "You are currently running %s and the latest version of convert2rhel is 0.22.\n"
-        "Only the latest version is supported for conversion." % (local_version)
+        "We encurage you to update to the latest version" % local_version
```
r0x0d, line 191:
```suggestion
        "We encurage you to update to the latest version." % local_version
```

r0x0d (changes requested):
I made a couple of suggestions here, but other than that, it looks super good! 

Good progress with that PR, @Andrew-ang9!

convert2rhel/checks.py
Andrew-ang9: Thanks for pointing these typos out I will get them fixed!

convert2rhel/checks.py
Andrew-ang9: I will definitely get this fixed thank you for spotting this!

convert2rhel/checks.py
Andrew-ang9: you are definitely right about this "then" should be there!

convert2rhel/unit_tests/checks_test.py
Andrew-ang9: Thanks for spotting all of those grammar errors. I must have been off my game or didn't go back and spell/grammar check like I usually do!

convert2rhel/checks.py
Andrew-ang9: I have resolved this in this commit: https://github.com/oamg/convert2rhel/pull/566/commits/41e0a2f2819ad5f3e5d0504f61c1da78bd220a3b

convert2rhel/checks.py
Andrew-ang9: I have implemented this change in the code. thank you for the advice!

convert2rhel/checks.py
Andrew-ang9: I agree that I shouldn't repeat this comment as it is redundant. Thank you!

convert2rhel/checks.py
Andrew-ang9: I'm not sure exactly why you are seeing this var like this. I don't see this in my code, but I do have this in my current code and have committed it!

convert2rhel/checks.py
Andrew-ang9: I have added this to the code, thank you for spotting this out!

convert2rhel/checks.py
```diff
@@ -75,77 +89,63 @@ def perform_pre_ponr_checks():
...
 
-    if return_code != 0:
-        logger.warning(
-            "Couldn't check if this is the latest version of convert2rhel\n"
-            "repoquery failed %s, %s" % (return_code, raw_output_convert2rhel_versions)
-        )
-        return
-
     PKG_NEVR = r"\b(\S+)-(?:([0-9]+):)?(\S+)-(\S+)\b"
     convert2rhel_versions = re.findall(PKG_NEVR, raw_output_convert2rhel_versions, re.MULTILINE)
```
SpyTec, line 108:
```suggestion
    convert2rhel_versions = re.findall(PKG_NEVR, raw_output_convert2rhel_versions, re.MULTILINE)
    logger.debug("Found %s convert2rhel packages" % len(convert2rhel_versions))
```

convert2rhel/checks.py
```diff
@@ -75,77 +89,63 @@ def perform_pre_ponr_checks():
...
     for package_version in convert2rhel_versions:
-        # rpm.lableCompare(pkg1, pkg2) compare two kernel version strings and return
-        # -1 if str2 is greater then str1, 0 if they are equal, 1 if str1 is greater the str2
-        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
+        # rpm.lableCompare(pkg1, pkg2) compare two package version strings and return
+        # -1 if lastest_version is greater than package_version, 0 if they are equal, 1 if package_version is greater than lastest_version
+        ver_compare = rpm.labelCompare(package_version[1:], latest_available_version)
         if ver_compare > 0:
-            latest_version = package_version[1:]
+            latest_available_version = package_version[1:]
```
SpyTec, line 118:
```suggestion
            logger.debug("...convert2rhel version %s is newer than %s, updating latest known version" % (package_version[1:][1], latest_available_version[1]))
            latest_available_version = package_version[1:]
```

convert2rhel/checks.py
```diff
@@ -75,77 +89,63 @@ def perform_pre_ponr_checks():
...
-        "[convert2rhel]\n"
-        "name=Convert2RHEL Repository\n"
-        "baseurl=https://cdn.redhat.com/content/public/convert2rhel/$releasever/x86_64/os/\n"
-        "gpgcheck=1\n"
-        "enabled=1\n"
-        "sslcacert=%s\n"
-        "gpgkey=file://%s\n" % (ssl_cert_path, gpg_path)
-    )
+    """Make sure that we are running the latest downstream version of convert2rhel"""
+    logger.task("Prepare: Checking if this is the latest version of convert2rhel")
```
SpyTec, line 94: In beginning of file, check if repos can be used (the internet check), if not we skip the check with a warning

```suggestion
    """Make sure that we are running the latest downstream version of convert2rhel"""
    logger.task("Prepare: Checking if this is the latest version of convert2rhel")

    if not system_info.has_internet_access:
        logger.warning("Skipping the check as no internet connection has been detected.")
        return
```

SpyTec (changes requested):
A few changes, mostly regarding logging, typos, and rewording. But also one about restoring a deleted code section - I admit the acceptance criteria in the issue was a bit confusing with this but it wasn't about removing the section but about preventing bloat by checking internet access in the beginning of the check

convert2rhel/checks.py
```diff
@@ -75,77 +89,63 @@ def perform_pre_ponr_checks():
...
+                "'CONVERT2RHEL_UNSUPPORTED_VERSION' environment variable detected, continuing conversion"
+                % (installed_convert2rhel_version, latest_available_version[1])
             )
 
         else:
             if int(system_info.version.major) <= 6:
                 logger.warning(
                     "You are currently running %s and the latest version of convert2rhel is %s.\n"
-                    "Only the latest version is supported for conversion." % (convert2rhel_version, latest_version[1])
+                    "We encurage you to update to the latest version."
```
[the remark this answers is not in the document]
Andrew-ang9: Thank you for catching this typo!

convert2rhel/checks.py
```diff
@@ -75,77 +89,63 @@ def perform_pre_ponr_checks():
...
+        ver_compare = rpm.labelCompare(package_version[1:], latest_available_version)
         if ver_compare > 0:
-            latest_version = package_version[1:]
+            latest_available_version = package_version[1:]
+
+    # we hard code 0 for the epoch and release below because latest_verion will
+    # contain "1.el7" for the release but we don't know what the release for
+    # installed_convert2rhel_version will be, if we did not hard code to 0 latest_available_version
+    # will evaluate as later then installed_convert2rhel_version even when the versions
+    # were the same.
```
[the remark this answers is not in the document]
Andrew-ang9: Thanks for catching these typos too, i will go back fix all of them!

convert2rhel/checks.py
```diff
@@ -75,77 +89,79 @@ def perform_pre_ponr_checks():
...
     for package_version in convert2rhel_versions:
-        # rpm.lableCompare(pkg1, pkg2) compare two kernel version strings and return
-        # -1 if str2 is greater then str1, 0 if they are equal, 1 if str1 is greater the str2
-        ver_compare = rpm.labelCompare(package_version[1:], latest_version)
+        # rpm.labelCompare(pkg1, pkg2) compare two package version strings and return
+        # -1 if latest_version is greater than package_version, 0 if they are equal, 1 if package_version is greater than latest_version
+        ver_compare = rpm.labelCompare(package_version[1:], latest_available_version)
         if ver_compare > 0:
-            latest_version = package_version[1:]
+            logger.debuge(
```
SpyTec, line 130:
```suggestion
            logger.debug(
```

SpyTec (changes requested):

convert2rhel/checks.py
Andrew-ang9: Thank you for spotting this, I will go ahead and get this changed then push it with the unit test!

convert2rhel/checks.py
```diff
@@ -75,77 +89,79 @@ def perform_pre_ponr_checks():
...
-        "enabled=1\n"
-        "sslcacert=%s\n"
-        "gpgkey=file://%s\n" % (ssl_cert_path, gpg_path)
-    )
+    """Make sure that we are running the latest downstream version of convert2rhel"""
+    logger.task("Prepare: Checking if this is the latest version of convert2rhel")
+
+    if not system_info.has_internet_accsess:
+        logger.warning("Skipping the check because no internet connection has been detected.")
+        return
```
SpyTec, line 97:
```suggestion
    if not system_info.has_internet_access:
        logger.warning("Skipping the check because no internet connection has been detected.")
        return
```

Andrew-ang9: thank you for pointing this out, I was actually working the unit test and noted changed that!

convert2rhel/checks.py
```diff
@@ -75,77 +89,63 @@ def perform_pre_ponr_checks():
...
 
     raw_output_convert2rhel_versions, return_code = run_subprocess(cmd, print_output=False)
 
-    if return_code != 0:
-        logger.warning(
-            "Couldn't check if this is the latest version of convert2rhel\n"
-            "repoquery failed %s, %s" % (return_code, raw_output_convert2rhel_versions)
-        )
-        return
-
```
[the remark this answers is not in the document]
Andrew-ang9: I have made this change and will be pushing the changes soon!

convert2rhel/checks.py
Andrew-ang9: I have made this change and will be pushing the changes soon!

convert2rhel/checks.py
```diff
@@ -75,77 +89,79 @@ def perform_pre_ponr_checks():
...
-    repo_dir = tempfile.mkdtemp(prefix="convert2rhel_repo.", dir=utils.TMP_DIR)
-    utils.mkdir_p(repo_dir)
-    repo_path = os.path.join(repo_dir, "convert2rhel.repo")
-    store_content_to_file(filename=repo_path, content=repo_content)
-
-    cmd.append("--setopt=reposdir=%s" % repo_dir)
+    cmd.append(
+        "--repofrom='convert2rhel,https://cdn.redhat.com/content/public/convert2rhel/$releasever/$basearch/os/'convert2rhel"
+    )
     cmd.append("convert2rhel")
```
r0x0d, line 107: I know it's a bit late to suggest this, but I would recommend grouping those `cmd` appends you have into just one… The reason to that is that you don't have any conditions to append then dynamically, they are all after each other.

Probably something like this: 

```python
    cmd = [
        "repoquery",
        "--releasever=%s" % system_info.version.major,
        "--repofrom='convert2rhel,https://cdn.redhat.com/content/public/convert2rhel/$releasever/$basearch/os/'convert2rhel",
        "convert2rhel"
    ]
```

convert2rhel/checks.py
```diff
@@ -75,77 +89,79 @@ def perform_pre_ponr_checks():
...
 
     raw_output_convert2rhel_versions, return_code = run_subprocess(cmd, print_output=False)
 
     if return_code != 0:
         logger.warning(
-            "Couldn't check if this is the latest version of convert2rhel\n"
-            "repoquery failed %s, %s" % (return_code, raw_output_convert2rhel_versions)
+            "Couldn't check if currently installed convert2rhel is latest version.\n"
```
r0x0d, line 113:
```suggestion
            "Couldn't check if current installed convert2rhel is latest version.\n"
```

convert2rhel/checks.py
Andrew-ang9: Your not late at all, I appreciate the comments and,I agree with this change. I will make the change and push it along with the rest of them

convert2rhel/checks.py
Andrew-ang9: Thank you for catching this typo!

convert2rhel/checks.py
```diff
@@ -75,77 +89,79 @@ def perform_pre_ponr_checks():
...
+            "repoquery failed with the following output:\n%s" % (raw_output_convert2rhel_versions)
         )
         return
 
     PKG_NEVR = r"\b(\S+)-(?:([0-9]+):)?(\S+)-(\S+)\b"
```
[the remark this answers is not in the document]
Andrew-ang9: yeah good point! I'll take care of both of these comments! Thank you!

convert2rhel/checks.py
```diff
@@ -75,77 +91,75 @@ def perform_pre_ponr_checks():
...
-    cmd.append("--setopt=reposdir=%s" % repo_dir)
-    cmd.append("convert2rhel")
-
     raw_output_convert2rhel_versions, return_code = run_subprocess(cmd, print_output=False)
 
     if return_code != 0:
         logger.warning(
-            "Couldn't check if this is the latest version of convert2rhel\n"
-            "repoquery failed %s, %s" % (return_code, raw_output_convert2rhel_versions)
+            "Couldn't check if the current installed convert2RHEL is the latest version.\n"
```
r0x0d, line 112:
```suggestion
            "Couldn't check if the current installed Convert2RHEL is the latest version.\n"
```

Convert2RHEL should be capitalized

Andrew-ang9: Thank you, I don't know why but recently I've been missing this! I will get this changed and pushed!

convert2rhel/unit_tests/checks_test.py
```diff
@@ -271,8 +271,8 @@ def test_convert2rhel_latest_log_check_env(caplog, monkeypatch, convert2rhel_lat
 
     local_version, dummy_ = convert2rhel_latest_version_test
     log_msg = (
-        "You are currently running %s and the latest version of convert2rhel is 0.22.\n"
-        "'CONVERT2RHEL_UNSUPPORTED_VERSION' environment detected, continuing conversion" % (local_version)
+        "You are currently running %s and the latest version of Convert2RHEL is 0.22.\n"
+        "'CONVERT2RHEL_UNSUPPORTED_VERSION' environment variable detected, continuing conversion" % local_version
```
SpyTec, line 275: I'd like to expand the unit test as well in the parametrize to include versions with different major versions as well.

E.g.
`local_version` of `0.18`, `package_version` of `1.1`
`local_version` of `1.1`, `package_version` of `0.25`

convert2rhel/checks.py
```diff
@@ -75,81 +91,79 @@ def perform_pre_ponr_checks():
...
+
+    if not system_info.has_internet_access:
+        logger.warning("Skipping the check because no internet connection has been detected.")
+        return
 
     cmd = [
         "repoquery",
         "--releasever=%s" % system_info.version.major,
+        "--repofrom='convert2rhel,https://cdn.redhat.com/content/public/convert2rhel/$releasever/$basearch/os/'convert2rhel",
+        "convert2rhel",
```
SpyTec, line 105: We now define the CDN URL both here and in `REPO_CONTENT`. I'd consolidate the two into one variable that's used in both

I see it was added in `3e162a01fdb2ab0246dddefd084bb5989a9c82f7`, but that commit message is hiding the changes in the commit. If you need to be concise with commit messages I'd suggest to mention changes and refactors first, then fixes, and lastly rewordings/fixing typos

SpyTec (changes requested):

convert2rhel/checks.py
Andrew-ang9: Thank you for this comment! I know that's an area where I could get better with! I will consolidate those two together and push that.

convert2rhel/unit_tests/checks_test.py
Andrew-ang9: yeah I can do that! I'll add the changes in and get it pushed on the next commit!

convert2rhel/checks.py
danmyway: I am a bit confused by this.
AD1 Do we need to inform the user about `updating latest known version`? If so, what is the purpose of this? Could the message be a bit clearer?
AD2 The log file displays this DEBUG message `...convert2rhel version 1.0 is newer than 0.00, updating latest known version`, even though the set version is supposed to be `42.0`, am I missing something @SpyTec @Andrew-ang9 ?
Thank you.

convert2rhel/checks.py
```diff
@@ -75,81 +93,79 @@ def perform_pre_ponr_checks():
...
+                "...convert2rhel version %s is newer than %s, updating latest known version"
+                % (package_version[1:][1], latest_available_version[1])
+            )
+            latest_available_version = package_version[1:]
+
+    # We hard code 0 for the epoch and release below because latest_version will
+    # contain somthing like "1.el7" for the release but we don't know what the release for
+    # installed_convert2rhel_version will be, if we did not hard code to 0 latest_available_version
+    # will evaluate as later then installed_convert2rhel_version even when the versions
+    # were the same.
```
danmyway, line 140:
```suggestion
    # After the for loop, the latest_available_version variable will gain the epoch, version, and release 
    # (e.g. ("0" "0.26" "1.el7")) information from the Convert2RHEL yum repo
    # when the versions are the same the latest_available_version's release field will cause it to evaluate as a later version.
    # Therefore we need to hardcode "0" for both the epoch and release below for installed_convert2rhel_version 
    # and latest_available_version respectively, to compare **just** the version field.
```
I liked the previous phrasing better TBH, this one is confusing to me. WDYT @bocekm ?

convert2rhel/checks.py
Andrew-ang9: hmmm I'm not 100% sure why it is showing 1.0 I will look into that and see what I can find. Thank you for catching this!

danmyway: 1.0 is fine @Andrew-ang9, the 0.0 is the part, that confuses me

Andrew-ang9: ahhh my apologies, I believe that I have fixed this issue before I will try and go back and see if I can remember what I did with this!

convert2rhel/checks.py
bocekm: Yeah, the one you suggest is more clear, @danmyway.

Reading it though the functionality should be a bit different. We should compare not only the version but also the release. The reason is that from time to time we need to make a downstream-only release (applying a patch in dist-git only). Doing that results in a bumped up release, for instance `1.1-2.el8`. I'll do a PR review later on to expand more on that.

danmyway: Alright, thank you for the clarification!

convert2rhel/checks.py
Andrew-ang9: Hey @danmyway I think I see whats going on, I had changed the duplicated use of the cdn url and from what I am seeing I messed up the use of the format string to use it as one var. I am working on fixing it now and will push the fix as soon as I get it back working!

Thank you again for finding this!

convert2rhel/checks.py
Andrew-ang9: Yeah thank you both for going over this!

convert2rhel/checks.py
```diff
@@ -75,81 +93,78 @@ def perform_pre_ponr_checks():
...
+        ver_compare = rpm.labelCompare(package_version[1:], latest_available_version)
         if ver_compare > 0:
-            latest_version = package_version[1:]
-
-    # After the for loop latest_version will have the epoch ,version, and release ex:("0" "0.26" "1.el7") information from convert2rhel yum repo.
-    # the release for latest_verion will be "1.el7" and the relase for convert2rhel_version will be hard coded as "0" below,
-    # therefore when the versions are the same the latest_version's release field will cause it to evaluate as later
-    ver_compare = rpm.labelCompare(("0", convert2rhel_version, "0"), ("0", latest_version[1], "0"))
+            logger.debug(
+                "...convert2rhel version %s is newer than %s, updating latest known version"
```
abadger, line 131:
```suggestion
                "...convert2rhel version %s is newer than %s, updating latest_available_version variable"
```

convert2rhel/checks.py
```diff
@@ -75,81 +93,79 @@ def perform_pre_ponr_checks():
...
+                % (package_version[1:][1], latest_available_version[1])
+            )
+            latest_available_version = package_version[1:]
+
+    # After the for loop, the latest_available_version variable will gain the epoch, version, and release
+    # (e.g. ("0" "0.26" "1.el7")) information from the Convert2RHEL yum repo
+    # when the versions are the same the latest_available_version's release field will cause it to evaluate as a later version.
+    # Therefore we need to hardcode "0" for both the epoch and release below for installed_convert2rhel_version
+    # and latest_available_version respectively, to compare **just** the version field.
+    ver_compare = rpm.labelCompare(("0", installed_convert2rhel_version, "0"), ("0", latest_available_version[1], "0"))
```
bocekm, line 142: We need to compare not only the version but also the release. The reason is that from time to time we need to make a downstream-only release (applying a patch in dist-git only). Doing that results in a bumped up release, for instance 1.1-2.el8.

Doing that can be a separate PR though.

convert2rhel/checks.py
```diff
@@ -75,81 +93,79 @@ def perform_pre_ponr_checks():
...
     if ver_compare < 0:
         if "CONVERT2RHEL_UNSUPPORTED_VERSION" in os.environ:
```
bocekm, line 144:
```suggestion
        if "CONVERT2RHEL_ALLOW_OLDER_VERSION" in os.environ:
```
As we've decided to stop using the "UNSUPPORTED" word, @abadger, @SpyTec, @r0x0d, are you ok with this suggestion?

bocekm (changes requested):

convert2rhel/checks.py
```diff
@@ -75,81 +93,79 @@ def perform_pre_ponr_checks():
...
+    if not system_info.has_internet_access:
+        logger.warning("Skipping the check because no internet connection has been detected.")
+        return
 
     cmd = [
         "repoquery",
+        "--disablerepo=*",
         "--releasever=%s" % system_info.version.major,
+        "--repofrom=convert2RHELrepo,%s" % CDN_URL,
+        "convert2rhel",
```
bocekm, line 108: @Andrew-ang9, I'm sure you know that this main part of the whole feature, running the repoquery command, doesn't work. First, you're disabling all repos, including the one you're defining (convert2RHELrepo). Second, you're not using the certificate so the repoquery request is refused. Do you have at least a working repoquery command at hand that you are about to implement here?

bocekm (changes requested):

convert2rhel/checks.py
abadger: @bocekm I worked with Andrew on this.... We tested this on a Centos8 box and found that `--repofrom` ignores the  `--disablerepo`.  That did surprise me.  It occurs to me that we should have tested this on other versions as well, though.  I'm now wondering if `--disablerepo=*` takes precedence on RHEL7, where the backend is yum-utils instead of all being a part of dnf. (Although testing on the matrix of distros that we support is one of the reasons we have CI right?  to catch issues from differences in the various distros?)

I'm trying to reason about the certificate in my head..... If the user has installed convert2rhel using the procedure we publicize here: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/converting_from_an_rpm-based_linux_distribution_to_rhel/index#proc_preparing-for-a-rhel-conversion_converting-from-a-linux-distribution-to-rhel  then they will have installed the certificate with `curl --create-dirs -o /etc/rhsm/ca/redhat-uep.pem https://ftp.redhat.com/redhat/convert2rhel/redhat-uep.pem`.  Ah, but we're creating our own repository so the certificate's location on disk won't be known.  Right... So we need something like `--setopt=sslcacert=PATH/TO/redhat-uep.pem`  (In testing, I lead Andrew astray... I thought that when installed a s documented, the certificate would be available in the system, so I had him add it on the system we were testing on [using update-ca-trust].  But we aren't telling the user to install the certificate into ca-trust, just to drop into a file on the filesystem, and then that specific file is referenced in the convert2rhel.repo file....)

And now I've checked RHEL7.... indeed, on RHEL7, we need `--disablerepo=* --enablerepo=convert2RHELrepo`.  And on RHEL7, `--setopt=sslcacert=redhat-uep.pem` does not work (but it does work on RHEL8)....  Going to have to see if there's a way to fix that on the command line or if we'll have to generate an actual convert2rhel.repo file.

convert2rhel/checks.py
abadger: Note:  release won't be as reliable as version.  We can get version from the code that is actually running.  We can  only get release by querying the rpm database.  (So, for instance, if I yum install convert2rhel-0.17-1 into the system but then I git clone convert2rhel and run from the checkout, we could detect the checkout version, but if we get the release, that would be wrong.)  Maybe we can compare the version from the running code with the version from rpm and if they match, we can also use the rpm release.  If they don't match, then use only the version.

convert2rhel/checks.py
abadger: The new name works for me.

convert2rhel/checks.py
bocekm: Yeah, I was trying it on RHEL 7 and `--disablerepo=* --enablerepo=convert2RHELrepo` did the trick regarding the disabled convert2RHELrepo.

Regarding the cert, I was trying `--setopt=convert2RHELrepo.sslcacert=/usr/share/convert2rhel/redhat-uep.pem`, described in yum man, but that didn't work either. I was thinking that we may need to get back to the original solution - creating the .repo file.

By the way, I was surprised that `--repofrom` works when the man page mentions just `--repofrompath`.

abadger: Since repoquery on RHEL7's documentation implies that `--setopt=sslcacert=FILE` should work but it doesn't, I had to be creative in looking for a suitable alternative.  It looks like we can go back to creating a repo file that points to the convert2rhel repo, setting the path to the ssl certificate in there, and then telling repoquery to use that repofile by pointing it at the directory that holds the file.  For instance:

`repoquery --releasever=8 convert2rhel --setopt=reposdir=convert2rhel/data/version-independent/yum.repos.d`

That's a bit of a hack, but since RHEL7 doesn't pass the `sslcacert` setting from the commandline to curl for `--repofrom` and there doesn't appear to be a way to pass a repo file to repoquery, it might be the best we can do.

convert2rhel/checks.py
r0x0d: Works for me too.

convert2rhel/checks.py
abadger: This should be done now.  repo is back to being set via  a file.

SpyTec (approved):

PR #568, opened 2022-08-15
Revert "fix el6 and el7 builds in Packit"

bocekm:
Reverts oamg/convert2rhel#556.

It was a temporary fix which is no longer needed as the packit issue https://github.com/packit/packit-service/issues/1601 has been resolved.

Ccing @TomasTomecek.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -8,27 +8,27 @@ srpm_build_deps: []
 jobs:
 - job: copr_build
   trigger: pull_request
-  owner: "@oamg"
-  project: convert2rhel
-  targets:
-  - epel-6-x86_64
-  - epel-7-x86_64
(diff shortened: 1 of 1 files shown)


r0x0d:
I would say that the changes @TomasTomecek applied are still valid, except for the 

```yaml
    # workaround for https://github.com/packit/packit-service/issues/1601
    fix-spec-file: []
```

The `metadata` field is deprecated by now, and we don't need to place them back. I would suggest creating a new PR that removes just the two lines of workaround.

bocekm:
Ah, good point. Thanks, @r0x0d.

bocekm:
Separate PR to remove the `fix-spec-file`: https://github.com/oamg/convert2rhel/pull/571.

PR #571, opened 2022-08-15
Remove fix-spec-file from packit configuration

bocekm:
We introduced the `fix-spec-file` only as a workaround for https://github.com/packit/packit-service/issues/1601. The issue has been fixed on the packit side so the workaround is no longer needed.

Ccing @TomasTomecek.


r0x0d (approved):

TomasTomecek:
builds \o/

tests /o\ those messages look scary

bocekm:
We've been seeing those for months so they do not scare us anymore. That's because we're DDOSing GitLab as we've discussed earlier. We're waiting patiently for the Testing Farm to adopt the new tmt versions that implements a few remediations to this issue (caching, retries).

PR #572, opened 2022-08-15
Check if shim-x64 package is present in the system

r0x0d:
In PR #553 we removed a condition from the shim-x64 special case workaround
where it was only supposed to handle UEFI systems, and now we are handling BIOS
as well. The problem is that in the BIOS systems, this package may not be
installed at all.

This commit introduces a quick check to validate if the package is present on
the system before trying to backup it's protection file under
`/etc/yum/protected.d`.

### DIFF
--- a/convert2rhel/special_cases.py
+++ b/convert2rhel/special_cases.py
@@ -101,25 +101,33 @@ def perform_java_openjdk_workaround():
 
 
 def unprotect_shim_x64():
-    """Remove the shim-x64 package yum protection on UEFI-based Oracle Linux 7 systems as it causes a yum traceback.
+    """Remove the shim-x64 package protection on Oracle Linux 7 as it causes yum to traceback.
 
-    The package is protected through the /etc/yum/protected.d/shim-x64.conf file. It is installed with the
-    Oracle Linux 7 shim-x64 package. The same package on RHEL 7 does not install this file - it's OL specific - no
-    need to add it back after a successful conversion to RHEL.
+    The package is protected through the /etc/yum/protected.d/shim-x64.conf
+    file. It is installed with the Oracle Linux 7 shim-x64 package. The same
+    package on RHEL 7 does not install this file - it's OL specific - no need to
+    add it back after a successful conversion to RHEL.
 
     Related: https://bugzilla.redhat.com/show_bug.cgi?id=2009368
     """
-    logger.info("Removing shim-x64 package yum protection.")
-    if system_info.id == "oracle" and system_info.version.major == 7:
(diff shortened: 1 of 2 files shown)


danmyway:
Do we want to have a sanity check for this one @r0x0d, @bocekm ?

r0x0d:
A sanity check in this case would mean (not) having the `shim-x64` package installed prior to the conversion, right? Well, I think that since we run the tests on BIOS only for now, I would say that it could be good thing to have. 

This problem occurs because on BIOS, the package might not be installed at all, and this was a leftover mistake from my previous PR that I did not handle that properly...

danmyway:
That is exactly what I had in mind.

convert2rhel/special_cases.py
```diff
@@ -101,25 +101,34 @@ def perform_java_openjdk_workaround():
 
 
 def unprotect_shim_x64():
-    """Remove the shim-x64 package yum protection on UEFI-based Oracle Linux 7 systems as it causes a yum traceback.
+    """Remove the shim-x64 package yum protection on UEFI-based Oracle Linux 7
+    systems as it causes a yum traceback.
```
abadger, line 105: Minor note that doesn't need to be done at all; just bringing it up since this is a change:

Standard docstring format differs from this in two ways:

* The summary should be on a single line rather than two.
* When a docstring takes up more than one line, it should start on the line after the `"""`.
We don't extract docs from the docstrings so it's okay that these aren't formatted that way; just mentioning it for your information

abadger (approved):
This looks good.  The main change is just adding a single condition to perform the action only when the shim-x64 package is installed.

+1

convert2rhel/special_cases.py
r0x0d: Glad that you picked that up! I can fix it since it's an easy change to the docstring. I will do that tomorrow.

r0x0d: Done!

bocekm (approved):

SpyTec:
@r0x0d @bocekm did we have a Jira issue for this?

bocekm:
@SpyTec, it's a follow up to https://issues.redhat.com/browse/RHELC-708 and its fix https://github.com/oamg/convert2rhel/pull/553. After we merged it I spotted the implementation is incomplete and I mentioned it here: https://github.com/oamg/convert2rhel/pull/553#issuecomment-1215052429.

Technically, there's no issue to this follow up bug fix. Even though I think it can be considered as part of the https://issues.redhat.com/browse/RHELC-708. But feel free to create a separate issue for this.

PR #573, opened 2022-08-16
chore: Cache containers in workflows

SpyTec:
We don't update our image within our workflow that often and use it for every commit.

By caching between workflows and uploading to the registry we should have results much faster


r0x0d:
The build/upload worked as expected: https://github.com/orgs/oamg/packages?repo_name=convert2rhel

r0x0d (approved):
Looks good to me. @SpyTec do you want to merge this changes in order to test it in another PR, or, do you want to test it in your own fork to validate that the build/usage is correct?

SpyTec:
I created a branch on this repo so I will merge now and if it doesn't work I'll just create a revert PR while I figure it out in my own repo

PR #554, opened 2022-08-03
[RHELC-67] Add gnome-documents-libs to excluded packages

r0x0d:
The gnome-documents-libs package is causing file conflicts with gnome-documents
during the package transaction as they both share similar files.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2043724

Signed-off-by: Rodolfo Olivieri <rolivier@redhat.com>

PR #576, opened 2022-08-19
Bump version to 1.0

bocekm:
This version includes the following changes.

Enhancements 

- Checking if a new version of convert2rhel is available (#415)
- Updating grub bootloader images (#417)
- Warning if multiple authentication sources are specified (#451)
- Improved log messages (#501, #499, #458)

Security hardening

- Introduced a user-customizable config file for passing RHSM credentials safely (#451, #562)
- Using dbus API for RHSM registration to safely pass the activation key (#540)
- Verifying GPG key for UBI and CentOS 6 Linux repositories (#505)

Deprecation

- Deprecated `-f` `--password-from-file` parameter option (#451)

Bug fixes

- Fixed logging error with unavailable RHSM certificate (#447)
- Fixed gnome-documents-libs package conflict (#554)
- Fixed handling shim-x64 package protection on non-UEFI systems (#553)
- Fixed unknown tag in an rpm query (#389, #424)
- Clearing yum cache before first yum call (#398)
- Added missing eus_rhsm_repoids key into OS 6 config files (#539)
- Fixed Ansible playbook example for CentOS Linux 8 conversions (#534)


hosekadam (approved):
LGTM

PR #574, opened 2022-08-16
[RHELC-635] Call integration tests through GitHub Actions

r0x0d:
We are replacing the integration tests that run with packit to run with
a custom GitHub Actions workflow to prevent DDoS in Red Hat's internal
giltab.

Signed-off-by: Rodolfo Olivieri <rolivier@redhat.com>

PR #585, opened 2022-08-24
[RHELC-635] Fix integration tests on github actions

r0x0d:
This commit introduces the fix to the github actions that requests the
integration tests through a custom GitHub Actions.

# Related PRs

- https://github.com/oamg/convert2rhel/pull/584
- https://github.com/oamg/convert2rhel/pull/582
- https://github.com/oamg/convert2rhel/pull/577
- https://github.com/oamg/convert2rhel/pull/574

# Changes that are worth to mention

The wait time for feedback is way less than with packit, that's because we are sending lots of requests to the tft API to run the tiers for each operating system we want. This way, we have the following matrix for our tests: 

| Tier 0     | Tier 1     |
|------------|------------|
| OL 7       | OL 7       |
| OL 8.4     | OL 8.4     |
| OL 8.6     | OL 8.6     |
| CentOS 7   | CentOS 7   |
| CentOS 8.4 | CentOS 8.4 |
| CentOS 8.5 | CentOS 8.5 |

You can check the wait time for each of them here: https://github.com/r0x0d/convert2rhel/runs/8000210821?check_suite_focus=true. Keep in mind that some tests in tier1 seems fast enough, but that's because I was using the wrong credentials/secrets, so it may be because it failed fast rather succeeded fast.

# Leftover TODO from #577 

- [ ] Check with QE if it's okay for them the way I'm setting up the tests
- [ ] Check if the link is correct for opening issues about our CI (Maybe use jira for that instead of github issues?)

# Proof of that the action is working as expected

https://github.com/r0x0d/convert2rhel/pull/65

It is good to keep in mind that even if the integration tests fails, the `check` in the list will appear as a green check.

# Suggestions to use this workflow

It would be nice to have a label to place after a reviewer/qe checks that the integration tests pass without problems. Maybe, something like `integration-tests-passes` or `ready-to-merge`.

# List of files that have been added/changed from the PRs

```
.github/workflows/greetings.yml
.github/workflows/reuse-int-tests.yml
.github/workflows/tmt-tests.yml
.github/workflows/reuse-get-copr-id.yml
plans/main.fmf
scripts/get_copr_build_id.py
```
# ❗❗Important notes❗❗

Any reviews should be posted under this PR to simplify the visualization, tracking, and progress. Please, do not post reviews on the closed PRs, as GitHub won't update the changes properly. 

If the reviewer prefer, pull this branch locally, make code changes in the listed files above, generate a diff for me, and I can apply that diff without an issue. 

Steps for that: 

1. Checkout this remote/branch locally
2. Do code changes
3. git diff > review.diff
4. Upload here the diff

### DIFF
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -21,13 +21,11 @@ jobs:
             ## **Thank you for contributing to the Convert2RHEL project!**
             👋 Hello @${{ github.actor }}, thank you for submitting a PR 🚀!
             Please note that every PR needs to comply with the [Convert2RHEL Guidelines](https://github.com/oamg/convert2rhel/blob/main/CONTRIBUTING.md) and must pass all tests in order to be mergable.
-            If you want to rebuild a package in copr, you can use following commands as a comment:
-            - **/packit copr-build** to trigger a copr build using packit
+            If you want to rebuild a package in [Fedora Copr](https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/), you can use the following command as a comment:
+            - **/packit copr-build** to trigger a copr build using [packit](https://packit.dev/)
 
-            To launch regression testing public members of oamg organization don't need any extra step. The automation kicks that off for you!
+            To execute integration tests, members of [the oamg organization](https://github.com/oamg/) don't need any extra step. The automation kicks that off for you!
 
-            If you're an external collaborator (Meaning that you're not part of oamg), ping the `@oamg/convert2rhel-developers` group to assign the testing label to your pr.
-
-            Please [open ticket](https://github.com/oamg/convert2rhel/issues) in case you experience technical problem with the CI.
+            If you're an external collaborator (Meaning that you're not part of oamg organization), wait for somebody from the `@oamg/convert2rhel-developers` group to add the `testing` label to the PR to trigger the tests.
 
             **Note:** In case there are problems with tests not being triggered automatically on new PR/commit or pending for a long time, please tag `@oamg/convert2rhel-developers` in the pr comments.
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -22,7 +22,7 @@ jobs:
             👋 Hello @${{ github.actor }}, thank you for submitting a PR 🚀!
             Please note that every PR needs to comply with the [Convert2RHEL Guidelines](https://github.com/oamg/convert2rhel/blob/main/CONTRIBUTING.md) and must pass all tests in order to be mergable.
             If you want to rebuild a package in copr, you can use following commands as a comment:
-            - **/packit copr-build** to submit a public copr build using packit
+            - **/packit copr-build** to trigger a copr build using packit
 
             To launch regression testing public members of oamg organization don't need any extra step. The automation kicks that off for you!
 
--- a/.github/workflows/reuse-int-tests.yml
+++ b/.github/workflows/reuse-int-tests.yml
@@ -57,7 +57,7 @@ on:
         type: string
         required: false
         default: "epel-8-x86_64"
-      distros:
+      matrix:
         type: string
         required: true
       update_pull_request_status:
@@ -74,8 +74,7 @@ jobs:
     name: Schedule integration tests on Testing Farm
     strategy:
       fail-fast: false
-      matrix:
-        distro: ${{ fromJson(inputs.distros) }}
+      matrix: ${{ fromJson(inputs.matrix) }}
     runs-on: ubuntu-latest
     steps:
       - name: Schedule integration testing for ${{ inputs.pull_request_status_name }}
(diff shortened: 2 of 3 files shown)


.packit.yaml
```diff
@@ -35,11 +35,11 @@ jobs:
...
-  use_internal_tf: True
-  trigger: pull_request
+      #- job: tests
+      #  targets:
+      #    epel-7-x86_64:
+      #      distros: [centos-7, oraclelinux-7]
+      #    epel-8-x86_64:
+      #      distros: [centos-8.4, centos-8, oraclelinux-8.4, oraclelinux-8.6]
+      #  use_internal_tf: True
+      #  trigger: pull_request
```
r0x0d, line 45: Just to be sure, we will still allow integration tests through `packit` for a while?

SpyTec: That's right, let packit remain for now

SpyTec:
> It is good to keep in mind that even if the integration tests fails, the check in the list will appear as a green check.

Is there a way for us to get pass/fail in the checks some other way?

.packit.yaml
r0x0d: Thanks!

r0x0d: Well, if we enable the packit AND use the GitHub actions, we will make a LOT more of DDoS to GitLab, since the packit integration will need to pull the environment files from GitLab defined in `plans/main.fmf`. 

@danmyway or @kokesak any you know of a way to conditionally pull that? I'm asking that because I think there is no difference from the `tmt` perspective in the execution from packit/github actions. Meaning that, if I uncomment the file in `plans/main.fmf` to pull the environment files, all jobs will pull that file.

r0x0d:
> > It is good to keep in mind that even if the integration tests fails, the check in the list will appear as a green check.
> 
> Is there a way for us to get pass/fail in the checks some other way?

I'm looking in the action itself to see if it's possible to do.

r0x0d:
> > It is good to keep in mind that even if the integration tests fails, the check in the list will appear as a green check.
> 
> Is there a way for us to get pass/fail in the checks some other way?

So, it seems that the workflow checks are broken for now. There is this issue opened a long time ago by @fernflower (https://github.com/sclorg/testing-farm-as-github-action/issues/50) that is experiencing the same issue as us (when the tests fails, the checks still remains all green)

r0x0d:
@SpyTec investigated it a bit, it seems that the problem lies in the action we are using. For some reason, when they send the request to the GitHub API to update the status, the `sha` is missing from the request. 

I have posted the logs in their issue https://github.com/sclorg/testing-farm-as-github-action/issues/50. Once that's fixed, we can update it in our end, and it should hopefully give the correct status.

.packit.yaml
bocekm: I don't think it's worth now spending time on investigating what kind of condition would limit using the environment file when tmt is triggered only through GitHub action and not from when triggered through Packit.

The new tmt version is to be deployed to the testing farm these days and then we won't care that much about the number of gitlab requests.

danmyway: Yeah, don't know much about conditioning that.
All I know of is the `trigger` context, but I don't think that it would be of any help.
Hopefully the TF workaround will solve the issue.

r0x0d: Enabled again.

.github/workflows/greetings.yml
```diff
@@ -22,7 +22,7 @@ jobs:
             👋 Hello @${{ github.actor }}, thank you for submitting a PR 🚀!
             Please note that every PR needs to comply with the [Convert2RHEL Guidelines](https://github.com/oamg/convert2rhel/blob/main/CONTRIBUTING.md) and must pass all tests in order to be mergable.
             If you want to rebuild a package in copr, you can use following commands as a comment:
```
bocekm, line 24:
```suggestion
            If you want to rebuild a package in [Fedora Copr](https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/), you can use the following command as a comment:
```

.github/workflows/greetings.yml
```diff
@@ -22,7 +22,7 @@ jobs:
...
-            - **/packit copr-build** to submit a public copr build using packit
+            - **/packit copr-build** to trigger a copr build using packit
```
bocekm, line 25:
```suggestion
            - **/packit copr-build** to trigger a copr build using [packit](https://packit.dev/)
```

.github/workflows/greetings.yml
```diff
@@ -22,7 +22,7 @@ jobs:
...
 
             To launch regression testing public members of oamg organization don't need any extra step. The automation kicks that off for you!
```
bocekm, line 27:
```suggestion
            To execute integration tests, members of [the oamg organization](https://github.com/oamg/) don't need any extra step. The automation kicks that off for you!
```

.github/workflows/greetings.yml
```diff
@@ -22,7 +22,7 @@ jobs:
...
```
bocekm, line 28: Remove the line:
`Please [open ticket](https://github.com/oamg/convert2rhel/issues) in case you experience technical problem with the CI.`

It is kind of duplicate to the _"problems with tests not being triggered automatically"_ note below it.

bocekm (changes requested):

.github/workflows/greetings.yml
r0x0d: done!

.github/workflows/greetings.yml
```diff
@@ -21,13 +21,11 @@ jobs:
...
             Please note that every PR needs to comply with the [Convert2RHEL Guidelines](https://github.com/oamg/convert2rhel/blob/main/CONTRIBUTING.md) and must pass all tests in order to be mergable.
-            If you want to rebuild a package in copr, you can use following commands as a comment:
-            - **/packit copr-build** to submit a public copr build using packit
+            If you want to rebuild a package in [Fedora Copr](https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/), you can use the following command as a comment:
+            - **/packit copr-build** to trigger a copr build using [packit](https://packit.dev/)
 
-            To launch regression testing public members of oamg organization don't need any extra step. The automation kicks that off for you!
+            To execute integration tests, members of [the oamg organization](https://github.com/oamg/) don't need any extra step. The automation kicks that off for you!
 
             If you're an external collaborator (Meaning that you're not part of oamg), ping the `@oamg/convert2rhel-developers` group to assign the testing label to your pr.
```
bocekm, line 29: As I'm thinking about it, maybe we shouldn't be telling them to ping us. We should know what to do. And if somebody creates a PR for us we want the tests to pass so we will add the label even without them pinging us.

So I'd change it to:

```suggestion
            If you're an external collaborator (meaning that you're not part of the oamg organization), wait for somebody from the @oamg/convert2rhel-developers group to add the testing label to the PR to trigger the tests.
```

bocekm (dismissed):

r0x0d:
Rebased to get the latest changes from the docker workflows

bocekm (approved):

PR #587, opened 2022-08-30
[RHELC-637, RHELC-742] Improve internet connection check

r0x0d:
This check was introduced in PR #458 as a quick way to validate if the internet
connectionw as up or not. The way that it was introduced was not very robust,
as it was checking for a public Google DNS that may be blocked on some
machines/servers as not essential.

This commit tries to improve it by, instead of checking Google Public DNS, we
check a static file hosted by Red Hat. That should be a little more robust
since `*.redhat.com` is supposed to be reachable from the customer servers when
running convert2rhel.

### DIFF
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -32,8 +32,11 @@ from convert2rhel.utils import run_subprocess
 # Number of times to retry checking the status of dbus
 CHECK_DBUS_STATUS_RETRIES = 3
 
-# Number of times to retry checking if the internet connection is up
+# Number of times to retry checking if the internet connection is up.
 CHECK_INTERNET_CONNECTION_RETRIES = 3
+
+# The address that will be used to check if there is a internet connection.
+CHECK_INTERNET_CONNECTION_ADDRESS = "http://static.redhat.com/test/rhel-networkmanager.txt"
 
 # Allowed conversion paths to RHEL. We want to prevent a conversion and minor
 # version update at the same time.
@@ -352,29 +355,33 @@ class SystemInfo(object):
     def _check_internet_access(self):
         """Check whether or not the machine is connected to the internet.
 
-        This method will try to estabilish a socket connection through a default
-        address (static.redhat.com). If it's we can successfully send the
-        request, validate that the status_code is 200, and assert that the
-        content of the response is the string 'OK', then it is most likely that
-        we are connected to the internet.
+        This method will try to retrieve a web page on the Red Hat network that
+        we know to exist (http://static.redhat.com/test/rhel-networkmanager.txt).
+        If we can successfully retrieve that page, validate that the status_code
+        is 200, and assert that the content of the response is the string 'OK',
+        then we decide we are connected to the internet.
+
+        We check a web page because we will need working https to retrieve
+        packages from Red Hat infrastructure during the conversion.
 
         .. warnings::
             We might have some problems with this if the host machine is using
             a NAT gateway to route the outbound requests to any other service.
 
-            It's valid to note that there are dnsmasq that can redirect the
-            content of a URL to another address.
+            DNS could also be used to redirect the URL we test to another address.
 
         :return: Return boolean indicating whether or not we have internet
             access.
         :rtype: bool
         """
-        address = "http://static.redhat.com/test/rhel-networkmanager.txt"
         retries = 0
         is_reachable = False
-        self.logger.info("Checking internet connectivity using address '%s'.", address)
+        self.logger.info(
+            "Checking internet connectivity using address '%s'.",
+            CHECK_INTERNET_CONNECTION_ADDRESS,
+        )
         while retries <= CHECK_INTERNET_CONNECTION_RETRIES:
-            response = urllib.request.urlopen(address)
+            response = urllib.request.urlopen(CHECK_INTERNET_CONNECTION_ADDRESS)
             status_code = response.getcode()
             text_response = response.read().decode("utf-8")
             response.close()
@@ -390,9 +397,9 @@ class SystemInfo(object):
             retries += 1
         else:
             self.logger.warning(
-                "Couldn't connect to the address '%s', assuming no internet connection is present.", address
+                "Couldn't connect to the address '%s', assuming no internet connection is present.",
+                CHECK_INTERNET_CONNECTION_ADDRESS,
             )
-            is_reachable = False
 
         return is_reachable
 
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -384,9 +384,9 @@ class SystemInfo(object):
             return True
         except urllib.error.URLError as err:
             self.logger.warning(
-                "There was a problem while trying to connect to '%s' to check internet connectivity, "
-                "it seems that the host appears to be offline or not responding. Convert2RHEL will continue the conversion "
-                "but without support for online resources.",
+                "There was a problem while trying to connect to '%s' to check internet connectivity. "
+                "This could be due to the host being offline, or the network blocking access to the endpoint... "
+                "Some checks and actions will be skipped.",
                 CHECK_INTERNET_CONNECTION_ADDRESS,
             )
             self.logger.debug("Failed to retrieve data from host, reason: %s", err.reason)
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -389,7 +389,7 @@ class SystemInfo(object):
                 "but without support for online resources.",
                 CHECK_INTERNET_CONNECTION_ADDRESS,
             )
-            self.logger.debug("Faild to retrieve data from host, reason: %s", err.reason)
+            self.logger.debug("Failed to retrieve data from host, reason: %s", err.reason)
             return False
 
     def corresponds_to_rhel_eus_release(self):
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -18,7 +18,6 @@ import difflib
 import logging
 import os
 import re
-import socket
 import time
 
 from collections import namedtuple
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -354,9 +354,8 @@ class SystemInfo(object):
 
         This method will try to retrieve a web page on the Red Hat network that
         we know to exist (http://static.redhat.com/test/rhel-networkmanager.txt).
-        If we can successfully retrieve that page, validate that the status_code
-        is 200, and assert that the content of the response is the string 'OK',
-        then we decide we are connected to the internet.
+        If we can successfully access that page, then we decide we are connected
+        to the internet.
 
         We check a web page because we will need working https to retrieve
         packages from Red Hat infrastructure during the conversion.
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -23,7 +23,7 @@ import time
 
 from collections import namedtuple
 
-from six.moves import configparser
+from six.moves import configparser, urllib
 
 from convert2rhel import logger, utils
 from convert2rhel.toolopts import tool_opts
@@ -32,6 +32,9 @@ from convert2rhel.utils import run_subprocess
 
 # Number of times to retry checking the status of dbus
 CHECK_DBUS_STATUS_RETRIES = 3
+
+# Number of times to retry checking if the internet connection is up
+CHECK_INTERNET_CONNECTION_RETRIES = 3
 
 # Allowed conversion paths to RHEL. We want to prevent a conversion and minor
 # version update at the same time.
@@ -347,38 +350,52 @@ class SystemInfo(object):
         #     )
         return self.submgr_enabled_repos if not tool_opts.no_rhsm else tool_opts.enablerepo
 
-    def _check_internet_access(self, host="8.8.8.8", port=53, timeout=3):
+    def _check_internet_access(self):
         """Check whether or not the machine is connected to the internet.
 
-        This method will try to estabilish a socket connection through the
-        default host in the method signature (8.8.8.8). If it's connected
-        successfully, then we know that internet is accessible from the host.
+        This method will try to estabilish a socket connection through a default
+        address (static.redhat.com). If it's we can successfully send the
+        request, validate that the status_code is 200, and assert that the
+        content of the response is the string 'OK', then it is most likely that
+        we are connected to the internet.
 
         .. warnings::
             We might have some problems with this if the host machine is using
-            a NAT gateway to route the outbound requests
+            a NAT gateway to route the outbound requests to any other service.
 
-        .. seealso::
-            * Comparison of different methods to check internet connections: https://stackoverflow.com/a/33117579
-            * Redirecting IP addresses: https://superuser.com/questions/954665/how-to-redirect-route-ip-address-to-another-ip-address
+            It's valid to note that there are dnsmasq that can redirect the
+            content of a URL to another address.
 
-        :param host: The host to establish a connection to. Defaults to "8.8.8.8".
-        :type host: str
-        :param port: The port for the connection. Defaults to 53.
-        :type port: int
-        :param timeout: The timeout in seconds for the connection. Defaults to 3.
-        :type port: int
-        :return: Return boolean indicating whether or not we have internet access.
+        :return: Return boolean indicating whether or not we have internet
+            access.
         :rtype: bool
         """
-        try:
-            self.logger.info("Checking internet connectivity using host '%s' and port '%s'." % (host, port))
-            socket.setdefaulttimeout(timeout)
-            socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect((host, port))
-            return True
-        except (socket.timeout, socket.error):
-            self.logger.warning("Couldn't connect to the host '%s'." % host)
-            return False
+        address = "http://static.redhat.com/test/rhel-networkmanager.txt"
+        retries = 0
+        is_reachable = False
+        self.logger.info("Checking internet connectivity using address '%s'.", address)
+        while retries <= CHECK_INTERNET_CONNECTION_RETRIES:
+            response = urllib.request.urlopen(address)
+            status_code = response.getcode()
+            text_response = response.read().decode("utf-8")
+            response.close()
+
+            if status_code == 200 and text_response.upper() == "OK":
+                self.logger.info("Internet connection available.")
+                is_reachable = True
+                break
+
+            # Wait for 1 second, 2 seconds, and then 4 seconds for another check.
+            # The service might be offline or with instability.
+            time.sleep(2**retries)
+            retries += 1
+        else:
+            self.logger.warning(
+                "Couldn't connect to the address '%s', assuming no internet connection is present.", address
+            )
+            is_reachable = False
+
+        return is_reachable
 
     def corresponds_to_rhel_eus_release(self):
         """Return whether the current minor version corresponds to a RHEL Extended Update Support (EUS) release.
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -377,14 +377,19 @@ class SystemInfo(object):
         try:
             response = urllib.request.urlopen(CHECK_INTERNET_CONNECTION_ADDRESS)
             response.close()
-            self.logger.info("Internet connection available.")
+            self.logger.info(
+                "Successfully connected to address '%s', internet connection seems to be available."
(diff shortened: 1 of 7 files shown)


tests/integration/tier0/check-internet-connection/test_internet_connection.py
```diff
@@ -0,0 +1,61 @@
...
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        c2r.expect("Checking internet connectivity using address")
+        c2r.expect("assuming no internet connection is present.")
+        c2r.send(chr(3))
+
+    _remove_changes_from_dnsmasq()
+    assert c2r.exitstatus == 1
```
r0x0d, line 61: @danmyway and @kokesak since I'm modifying the contents of a file for this test, should I move it to `tier1`? I mean, the way I tried to make this basic integration tests was to check if an internet connection is available or not (breaking earlier if it's (not) available)

tests/integration/tier1/convert-offline-systems/prepare_system.py
```diff
@@ -47,10 +47,6 @@ def test_prepare_system(shell):
 
     configure_connection()
 
-    # c2r checks the internet connectivity against the 8.8.8.8 DNS server
-    # pinging specific IP address is still possible so we need to block this as well
-    assert shell("iptables -I OUTPUT -d 8.8.8.8 -j DROP").returncode == 0
```
r0x0d, line 52: IIRC, this was mainly placed here because of PR#458 where we were checking against google dns, since we are using an address now, I think this is not required anymore.

kokesak: Yes that should not be needed anymore.

tests/integration/tier0/check-internet-connection/test_internet_connection.py
kokesak: Keep it `tier0` as you are not doing the full conversion.

tests/integration/tier0/check-internet-connection/main.fmf
```diff
@@ -0,0 +1,6 @@
+summary: internet connection checks
+
+tier: 1
```
kokesak, line 3:
```suggestion
tier: 0
```

tests/integration/tier0/check-internet-connection/test_internet_connection.py
```diff
@@ -0,0 +1,61 @@
...
+    with open("/etc/dnsmasq.conf", "a") as f:
+        # Everything else is resolved to localhost
+        f.write("address=/#/127.0.0.1")
+
+    with open("/etc/resolv.conf", "w") as f:
+        f.write("nameserver 127.0.0.1")
+
+
+def _remove_changes_from_dnsmasq():
+    """Delete only the last line that was appended to both files."""
```
kokesak, line 14: Correct me if I'm wrong but for me it seems that since you are opening the `resolv.conf` in `w` mode you are overwriting it's content no? So removing the last line in this case will not restore the file?

convert2rhel/systeminfo.py
```diff
@@ -320,38 +322,52 @@ def get_enabled_rhel_repos(self):
...
         """
-        try:
-            self.logger.info("Checking internet connectivity using host '%s' and port '%s'." % (host, port))
-            socket.setdefaulttimeout(timeout)
-            socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect((host, port))
-            return True
-        except (socket.timeout, socket.error):
-            self.logger.warning("Couldn't connect to the host '%s'." % host)
-            return False
+        address = "http://static.redhat.com/test/rhel-networkmanager.txt"
```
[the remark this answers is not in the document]
r0x0d: Liked the idea. Thanks!

convert2rhel/systeminfo.py
```diff
@@ -320,38 +322,52 @@ def get_enabled_rhel_repos(self):
...
+
+            # Wait for 1 second, 2 seconds, and then 4 seconds for another check.
+            # The service might be offline or with instability.
+            time.sleep(2**retries)
+            retries += 1
+        else:
+            self.logger.warning(
+                "Couldn't connect to the address '%s', assuming no internet connection is present.", address
+            )
+            is_reachable = False
```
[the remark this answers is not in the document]
r0x0d: done.

tests/integration/tier0/check-internet-connection/test_internet_connection.py
r0x0d: So, I have changed it a bit after @abadger comments below. Decided to save the original state of the file, modify it, and then restore after the conversion. 

Seems simpler this way, and we don't have to deal with impartial rewrites too.

abadger (dismissed):

abadger:
Looks good, developer approval.

tests/integration/tier0/check-internet-connection/test_internet_connection.py
```diff
@@ -0,0 +1,78 @@
...
+        c2r.expect("Checking internet connectivity using address")
+        assert c2r.expect_exact("Internet connection available.") == 0
+        c2r.send(chr(3))
+
+    assert c2r.exitstatus == 1
+
+
+def test_check_if_internet_connection_is_not_reachable(convert2rhel):
+    """Test a case where the internet connection is not reachable by any means."""
+    _modify_configuration_files()
```
kokesak, line 64: Test failed because the `dnsmasq.conf` is not present on the image by default. You will need to install dnsmasq package manually before calling the function. See https://github.com/oamg/convert2rhel/blob/main/tests/integration/tier1/convert-offline-systems/prepare_system.py#L38

Maybe you will need to even restart the dnsmasq service to apply changes.

tests/integration/tier0/check-internet-connection/test_internet_connection.py
```diff
@@ -0,0 +1,78 @@
...
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        c2r.expect("Checking internet connectivity using address")
+        assert c2r.expect_exact("Internet connection available.") == 0
+        c2r.send(chr(3))
+
+    assert c2r.exitstatus == 1
```
kokesak, line 59: The c2r returned exic code 130.
https://privatebin.corp.redhat.com/?65fd22902d1c97a8#BKzohheYwPpGQ8xkMf4G6vwk975CyKstj9Ne5pexeNhF
I'm not sure why, maybe because it's right in the beginning of the conversion.

Maybe the safest option here would be to use:?
```suggestion
    assert c2r.exitstatus != 0
```

kokesak:
The offline conversion fails tho on the following:
https://privatebin.corp.redhat.com/?2ec94bc2531a19aa#AW5VsNLCeEnxppT6dje2D6TmCxZcSN9UR1Hq999HCss9

tests/integration/tier0/check-internet-connection/test_internet_connection.py
r0x0d: I'm guessing that's because we are sending SIGTERM right when we are executing the systemctl call to assert that dbus is present. Let me take a new look.

r0x0d:
> The offline conversion fails tho on the following: https://privatebin.corp.redhat.com/?2ec94bc2531a19aa#AW5VsNLCeEnxppT6dje2D6TmCxZcSN9UR1Hq999HCss9

I believe my latest push should fix this now! The problem was that I was not catching the exception that raises from `urllib.request.urlopen()`, as I was assuming that the URL would be always accessible. Glad that we have this kind of check in place to catch things like that!

convert2rhel/systeminfo.py
```diff
@@ -320,38 +325,62 @@ def get_enabled_rhel_repos(self):
...
+        )
+        while retries <= CHECK_INTERNET_CONNECTION_RETRIES:
+            try:
+                response = urllib.request.urlopen(CHECK_INTERNET_CONNECTION_ADDRESS)
+            except urllib.error.URLError as err:
+                self.logger.debug("Faild to retrieve data from host, reason: %s", err.reason)
+                break
+
+            status_code = response.getcode()
+            text_response = response.read().decode("utf-8")
```
r0x0d, line 364: @abadger, do you think that this part can fail? I mean, if the connection is successful and the host doesn't return anything to us (probably they have deployed a new version of that server, and it's misbehaving), do you think that we need to extra careful here and do another try-except check? 

I mean, what are the chances of something like that happening? I can see the scenario where this host is offline for some reason, but not having its content changed.

tests/integration/tier0/check-internet-connection/test_internet_connection.py
kokesak: @r0x0d Sadly I don't think there is a faster "way". It's caused by some delay during processing the line during the c2r run and reacting to it.

You can try to run the interactive way (without the `-y` option). Since it's so early in the conversion, you can then send the `n` to the first prompt. This seems like a more clean way and we should always get the `1` exit code.

Also you will not need to pass the credentials to the call, simple 
```
with convert2rhel(
        ("--no-rpm-va --debug")
        ...
```
should be enough.

convert2rhel/systeminfo.py
```diff
@@ -320,38 +325,62 @@ def get_enabled_rhel_repos(self):
...
+            response.close()
+
+            if status_code == 200 and text_response.upper() == "OK":
+                self.logger.info("Internet connection available.")
+                is_reachable = True
+                break
```
SpyTec, line 370: I don't see why we need to check the text response or status code. We're not checking the status of the service, we're checking that the internet connectivity is working. 

If we only test status code, we would have to test all status codes, in which the if-statement becomes pointless. If we don't then we're naive. The server could have moved the endpoint, changed the format, etc.

All we need to do is ensure we can get a response from the server. Content or status does not matter. In the event that the server is completely down we need to be able to handle that scenario as well

SpyTec (changes requested):

tests/integration/tier0/check-internet-connection/test_internet_connection.py
r0x0d: Okay, seems a better approach indeed. I will adapt the tests to that. Thanks!

convert2rhel/systeminfo.py
r0x0d: I implemented that part because it was part of the acceptance criteria defined in the ticket. 

> Download and verify the content of the file returned from the pre-defined URL. In the above case, it's an XML file containing some repo information.

I vaguely remember that we discussed that, in one cabal session, that it would be good to verify the content of the response. Regardless of that, I can change this and remove this condition if you think that's appropriate.

tests/integration/tier0/check-internet-connection/test_internet_connection.py
r0x0d: Done!

convert2rhel/systeminfo.py
SpyTec: If we want to verify the contents of the file the workflow should be as follows:

1. Verify internet connectivity
2. Read contents of XML file
    1. If contents can be read, get repo information
    2. If contents can not be read, continue as normal

r0x0d: > If we want to verify the contents of the file the workflow should be as follows (oversimplified):
> 
>     1. Verify internet connectivity
>        
>        1. If internet connectivity exists, continue function
>        2. If internet connectivity does not exist, return false

Yeah… This is exactly what the code is doing right now. If the connection can be established through the `urllib.request.urlopen()`, then we progress with `1.`, otherwise, if the exception is raised `urllib.error.URLError`, meaning that the connection was refused (no connection, dnsmasq redirects, server offline, nat gateway redirects, firewalls, url not being accessible etc… Anything that can refuse a request) then we go with `2.`. 

>     2. Read contents of XML file
>        
>        1. If contents can be read, get repo information
>        2. If contents can not be read, continue as normal

That's what we are doing. The exception is that we are not using XML files to check repo availability, but a static file that just returns 'OK' in case of being reachable. I don't see a difference between using the example URL that just returns an XML and this one, that just returns a piece of information. To be honest, any internet check can be bypassed by an experienced network admin, but I tend to believe they won't do that on purpose.

Sorry, but that's still not clear to me what is expected with that comment, if you could detail a bit more on why we need to remove, I will be glad… Maybe I'm just missing some point here.

Note: I'm not against removing this condition, I agree that status code here might not be the best information for a check, might as well make sense to remove it. Just asking for a bit more clarification on what you expect to see in this change.

SpyTec: > This is exactly what the code is doing right now

No the code is only asserting that the connection is good if the content is OK and status code is 200. Otherwise it fails. That's what I'm trying to establish is an issue here as we tie the connection to the service rather than if we can connect

The thing I'm trying to say is that we should not establish if we have connectivity based on status code or content

r0x0d: Now it's more clear, I see your point now, thanks :) 

So, simple removing this condition and letting the `try-except` the way it is right now should be sufficient for that.

convert2rhel/systeminfo.py
```diff
@@ -320,37 +322,43 @@ def get_enabled_rhel_repos(self):
...
         """Check whether or not the machine is connected to the internet.
 
-        This method will try to estabilish a socket connection through the
-        default host in the method signature (8.8.8.8). If it's connected
-        successfully, then we know that internet is accessible from the host.
+        This method will try to retrieve a web page on the Red Hat network that
+        we know to exist (http://static.redhat.com/test/rhel-networkmanager.txt).
+        If we can successfully retrieve that page, validate that the status_code
+        is 200, and assert that the content of the response is the string 'OK',
+        then we decide we are connected to the internet.
```
SpyTec, line 332:
```suggestion
        This method will try to retrieve a web page on the Red Hat network that
        we know to exist (http://static.redhat.com/test/rhel-networkmanager.txt).
        If we can successfully access that page, then we decide we are connected
        to the internet.
```

convert2rhel/systeminfo.py
```diff
@@ -320,37 +322,43 @@ def get_enabled_rhel_repos(self):
...
+            self.logger.info("Internet connection available.")
             return True
-        except (socket.timeout, socket.error):
-            self.logger.warning("Couldn't connect to the host '%s'." % host)
+        except urllib.error.URLError as err:
+            self.logger.debug("Faild to retrieve data from host, reason: %s", err.reason)
+            self.logger.warning(
+                "Couldn't connect to the address '%s', assuming no internet connection is present.",
+                CHECK_INTERNET_CONNECTION_ADDRESS,
+            )
```
SpyTec, line 361: This doesn't encompass the QoL change to make the message less confusing as part of [RHELC-742](https://issues.redhat.com/browse/RHELC-742)

SpyTec (changes requested):

convert2rhel/systeminfo.py
r0x0d: Any suggestions for the message?

convert2rhel/systeminfo.py
```diff
@@ -350,14 +350,19 @@ def _check_internet_access(self):
...
+                % CHECK_INTERNET_CONNECTION_ADDRESS
+            )
             return True
         except urllib.error.URLError as err:
-            self.logger.debug("Faild to retrieve data from host, reason: %s", err.reason)
             self.logger.warning(
-                "Couldn't connect to the address '%s', assuming no internet connection is present.",
+                "There was a problem while trying to connect to '%s' to check internet connectivity, "
+                "it seems that the host appears to be offline or not responding. Convert2RHEL will continue the conversion "
+                "but without support for online resources.",
```
SpyTec, line 362: Continuing on [the last convo](https://github.com/oamg/convert2rhel/pull/587#discussion_r986900426). Maybe something like this. `Inaccessible` here can refer to both that it's blocked by the customer or that the server is down.

@bocekm what do you think?

> There was a problem while trying to connect to '%s' to check internet connectivity, it appears that the host is inaccessible. This could be due to the host being offline, or the network blocking access to the endpoint. Some checks and actions will be skipped, which will decrease the chance for a successful conversion.

convert2rhel/systeminfo.py
```diff
@@ -350,14 +350,19 @@ def _check_internet_access(self):
...
                 CHECK_INTERNET_CONNECTION_ADDRESS,
             )
+            self.logger.debug("Faild to retrieve data from host, reason: %s", err.reason)
```
SpyTec, line 365: Typo

```suggestion
            self.logger.debug("Failed to retrieve data from host, reason: %s", err.reason)
```

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -264,8 +264,8 @@ def test_get_release_ver_other(
 @pytest.mark.parametrize(
     ("side_effect", "expected", "message"),
     (
-        (urllib.error.URLError(reason="fail"), False, "Couldn't connect to the address"),
-        (None, True, "Internet connection available."),
+        (urllib.error.URLError(reason="fail"), False, "Faild to retrieve data from host"),
```
SpyTec, line 267:
```suggestion
        (urllib.error.URLError(reason="fail"), False, "Failed to retrieve data from host"),
```

convert2rhel/systeminfo.py
bocekm: I feel that the part "it appears that the host is inaccessible" is not necessary there, it adds no additional information.
Similarly, the part

tests/integration/tier0/check-internet-connection/test_internet_connection.py
```diff
@@ -42,17 +42,19 @@ def _restore_configuration_files():
...
 
 
+@pytest.mark.available_connection
 def test_check_if_internet_connection_is_reachable(convert2rhel):
     """Test if convert2rhel can access the internet."""
-    with convert2rhel(("--no-rpm-va --debug")) as c2r:
+    with convert2rhel("--no-rpm-va --debug") as c2r:
         c2r.expect("Checking internet connectivity using address")
-        assert c2r.expect_exact("Internet connection available.") == 0
+        assert c2r.expect_exact("internet connection seems to be available") == 0
```
danmyway, line 50: I have modified the expected string, the previous could not be read, as the log message is defined differently.

tests/integration/tier0/check-internet-connection/main.fmf
```diff
@@ -1,6 +1,20 @@
...
+    tag+:
+        - available_connection
+    test: |
+        pytest -svv -m available_connection
+
+/unavailable_connection:
+    tag+:
+        - unavailable-connection
+    test: |
+        pytest -svv -m unavailable_connection
```
danmyway, line 20: I have modified the description and the metadata to align with proposed changes for tier0 refactoring.
Hope you don't mind @r0x0d :)

r0x0d: That's fine by me

convert2rhel/systeminfo.py
```diff
@@ -347,37 +349,47 @@ def get_enabled_rhel_repos(self):
...
+        self.logger.info(
+            "Checking internet connectivity using address '%s'.",
+            CHECK_INTERNET_CONNECTION_ADDRESS,
+        )
         try:
-            self.logger.info("Checking internet connectivity using host '%s' and port '%s'." % (host, port))
-            socket.setdefaulttimeout(timeout)
-            socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect((host, port))
+            response = urllib.request.urlopen(CHECK_INTERNET_CONNECTION_ADDRESS)
+            response.close()
```
abadger, line 379: I should have spotted this earlier... I wonder if we should also check the response code and body of this response.  If the user was behind some sort of captive portal, the internet check would not timeout but it would not return the red hat web site either.  So checking that the response body is `OK` feels like it is important.

SpyTec: That becomes tricky. I brought it up before as that was the initial implementation - checking the content and the status code. But at that point you are not checking the internet connectivity, you're checking the website

If the website ever goes down with gateway error etc, we'd run into an issue. We should ignore response code. And as for the content, again if the website goes into a [503](https://http.cat/503) the content wouldn't be returned but the user still has internet - so checking for that means a reliance on it the service working as we expect it to work rather than just checking that the user has connectivity

abadger: <nod> Makes sense.  Resolving this comment.

abadger (approved):
Approved.

SpyTec (approved):

kokesak (approved):
LGTM, merge whenever you want.

danmyway (approved):
I like it, thank you for the alignment with the new tier0 proposal @r0x0d !

PR #590, opened 2022-09-02
chore: remove download artifact

SpyTec:
We don't create an artifact for the container builds so the workflows don't pass

This should still be relatively fast I believe

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -24,12 +24,6 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
-      - name: Download artifact
-        uses: actions/download-artifact@v3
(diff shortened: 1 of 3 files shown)


SpyTec:
Interestingly the workflow runs don't show in here but do in the project actions

@r0x0d can you merge so we can test? I have one more idea to do after this if it still doesn't show up in the checks

r0x0d (approved):

r0x0d:
Merging this right now! If this approach doesn't work, I would suggest you to take a look at https://github.com/r0x0d/convert2rhel/blob/fix-integration-tests-gh-actions/.github/workflows/tmt-tests.yml#L24-L35, this is a custom action I made to wait for other workflows to finish. 

Those custom `workflow_*` events on github are always strange.

PR #592, opened 2022-09-05
chore: build GHCR images on main

SpyTec:
Forks don't have permission to update the image. We also don't need to update the image for every commit in a PR, intead we can simply update the images on push to main and use that image

Caveat: with this update we won't be able to update the images and have the workflows reflect that in the PRs themselves. We can utilize workflow_dispatch together with reusable workflows and environment variables to manually trigger this in the future if we want to. But for that to work we will have to update the images from a branch on this repo instead of a fork due to permissions

### DIFF
--- a/.github/workflows/build_images.yml
+++ b/.github/workflows/build_images.yml
@@ -3,8 +3,6 @@ name: build_and_publish_images
 on:
   push:
     branches: [ main ]
-  pull_request:
-    branches: [ main ]
 
 jobs:
   build_and_publish:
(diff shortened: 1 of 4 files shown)


r0x0d (approved):
Looks reasonable to me. It is kinda rare to see any changes to the dockerfiles that would need to land on `main` first. 

In any way, this suffers from the same thing that my tests in gh actions suffers (forks needing more permissions/info in PR), which is unfortunate that we need to come with different ways to make this work, but I think that will work just fine for our use case (at least, for now).

PR #598, opened 2022-09-06
chore: PR template

SpyTec:
This template will be shown on the website when a new PR is opened. The checklist is primarily for our own sake in making our development better. Will be shown as follows:

---

Jira Issue: [RHELC-]()

Checklist
- [ ] PR meets acceptance criteria
- [ ] The commits squashed to as few commits as possible (without losing data) 


.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+<!-- Write a description of what the PR solves and how -->
+
+<!-- Link to relevant Jira issue, add multiple if necessary -->
+Jira Issue: [RHELC-]()
+
+Checklist
+- [ ] PR meets acceptance criteria
+- [ ] PR has been tested manually in a VM (either author or reviewer)
+- [ ] Jira issue has been made public if possible
```
r0x0d, line 9: I don't think this is a relevant check for someone that is opening a pull request… I think that most of us never made any issues public when we pick up something from the board to work on.

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+<!-- Write a description of what the PR solves and how -->
+
+<!-- Link to relevant Jira issue, add multiple if necessary -->
+Jira Issue: [RHELC-]()
```
r0x0d, line 4: Do we still need to place the Jira issue in the description? I thought that we dropped it in favor of having it in the title because of the integration...

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
...
+<!-- Link to relevant Jira issue, add multiple if necessary -->
+Jira Issue: [RHELC-]()
+
+Checklist
+- [ ] PR meets acceptance criteria
+- [ ] PR has been tested manually in a VM (either author or reviewer)
+- [ ] Jira issue has been made public if possible
+- [ ] Code and tests are documented properly
+- [ ] The commits squashed to as few commits as possible (without losing data) <!-- The commits can be squashed to 1 commit, but then we might lose data regarding moving something to a new file and then refactoring for example. Hence squash without losing data -->
+- [ ] When merged: Jira issue has been updated to `Release Pending`
```
r0x0d, line 12: I would mark this with something to identify that it is only for Convert2RHEL developers, external contributors don't need to care about it.

.github/pull_request_template.md
SpyTec: While true, and it seems like it is the case where we need it in the title (unless the tool just didn't work when I tested) it still helps to have a link somewhere to quickly get to an issue

.github/pull_request_template.md
SpyTec: It is a way to make sure that your task is made public, we want to make as many issues as public as possible

.github/pull_request_template.md
SpyTec: I don't even think an external contributor would have permission to do so anyway. But it's more of a check for the one who merged it

r0x0d (dismissed):
Looks good to me. All my questions were addressed in previous comments.

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+<!-- Write a description of what the PR solves and how -->
+
+<!-- Link to relevant Jira issue, add multiple if necessary -->
+Jira Issue: [RHELC-]()
+
+Checklist
+- [ ] PR meets acceptance criteria
```
bocekm, line 7:
```suggestion
- [ ] PR meets acceptance criteria specified in the Jira issue

```

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
...
+
+<!-- Link to relevant Jira issue, add multiple if necessary -->
+Jira Issue: [RHELC-]()
+
+Checklist
+- [ ] PR meets acceptance criteria
+- [ ] PR has been tested manually in a VM (either author or reviewer)
+- [ ] Jira issue has been made public if possible
+- [ ] Code and tests are documented properly
+- [ ] The commits squashed to as few commits as possible (without losing data) <!-- The commits can be squashed to 1 commit, but then we might lose data regarding moving something to a new file and then refactoring for example. Hence squash without losing data -->
```
bocekm, line 11:
```suggestion
- [ ] The commits are squashed to as few commits as possible (without losing data) <!-- The commits can be squashed to 1 commit, but then we might lose data regarding moving something to a new file and then refactoring for example. Hence squash without losing data -->

```

.github/pull_request_template.md
bocekm: If we need to have the [RHELC-xy] in the PR title then it can be another item in the checklist:
`[ ] [RHELC-] is part of the PR title <!-- For a proper sync with Jira -->`

bocekm (changes requested):

.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+<!-- Write a description of what the PR solves and how -->
+
```
r0x0d, line 2: Not something that should be changed, but I would add (as an example, still) one or two PR examples, so people can see some references on how to write their description. It may be helpful sometimes.

SpyTec: Good point. I'll look through later and find something to link

.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+<!-- Link to relevant Jira issue, add multiple if necessary -->
+Jira Issue: [RHELC-]()
+
+Checklist
+- [ ] PR meets acceptance criteria specified in the Jira issue
+- [ ] PR has been tested manually in a VM (either author or reviewer)
+- [ ] Jira issue has been made public if possible
+- [ ] `[RHELC-]` is part of the PR title <!-- For a proper sync with Jira -->
+- [ ] Code and tests are documented properly
+- [ ] The commits are squashed to as few commits as possible (without losing data) <!-- The commits can be squashed to 1 commit, but then we might lose data regarding moving something to a new file and then refactoring for example. Hence squash without losing data -->
```
r0x0d, line 12: I would place a short link to some page that shows people how to squash commits. This might not be as straight forward to people as we may think it is.

bocekm (approved):

r0x0d (approved):
I made some comments, but I don't think they matter for the state of the PR right now. I would say that this is something we can come back later and include... No worries to adapt them now. 

Looks pretty nice to me.

PR #599, opened 2022-09-06
Debug Testing Farm GH Action permissions

bocekm:
Currently the Testing Farm GitHub action is being skipped. We think it might be related to a PR authorship/permissions.


bocekm:
Ok, we think we know what we need - adding CONTRIBUTOR to the roles allowed to run tests. @r0x0d will prepare a followup fix.

PR #603, opened 2022-09-09
Revert "Fix codecov checks by passing a token"

SpyTec:
Reverts oamg/convert2rhel#593. 

Since PRs from forks can't access the token we're getting errors overall where it tries to upload with an empty token


SpyTec:
Closing in favor #604

PR #608, opened 2022-09-13
 [RHELC-740] Fix permissions check for certificates.

abadger:
We check whether we can access the certificate files shipped with convert2rhel prior to installing them.  The check for the directory they live in was checking whether the directory was readable and writable. We don't write to that directory, though, so the check shuld be whether the directory is readable and executable.

Note that convert2rhel should always be run by root so this bug won't be encountered in normal use.

This issue mentioned in:
Jira Issue: [RHELC-740](https://issues.redhat.com/browse/RHELC-740)

Checklist

### DIFF
--- a/convert2rhel/cert.py
+++ b/convert2rhel/cert.py
@@ -35,7 +35,7 @@ class SystemCert(object):
 
     @staticmethod
     def _get_cert():
-        """Return name of certificate and his directory."""
+        """Return name of certificate and this directory."""
         cert_dir = os.path.join(utils.DATA_DIR, "rhel-certs")
         if not os.access(cert_dir, os.R_OK | os.X_OK):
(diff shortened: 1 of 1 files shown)


convert2rhel/cert.py
```diff
@@ -37,7 +37,7 @@ def __init__(self):
     def _get_cert():
         """Return name of certificate and his directory."""
```
SpyTec, line 38: Seems I had a typo
```suggestion
        """Return name of certificate and this directory."""
```

SpyTec (dismissed):
LGTM! Don't think we need any tests for it

SpyTec (approved):

SpyTec:
Integration tests failing are unrelated, merging

abadger:
THanks @SpyTec !

PR #611, opened 2022-09-14
[RHELC-736] Fix CI builds to use the PR HEAD.

abadger:
CI was only testing the latest head of the main branch (what was built into the container images) instead of the head of the PR branch.  This PR will fix that.

* Change github workflows for testing and coverage to use the Makefile targets for unit testing.
* Modify the unit test targets to get images from the github container repository by default.
* Add instructions in the Makefile for authenticating to the githb container repository or to use make tests with BUILD_IMAGES=1
* Change the make tests targets to remove the test containers when finished (may be disabled with KEEP_TEST_CONTAINER=1)
* When using make tests with podman, output a message that the convert2rhel checkout will be read-only while the test is running (this is due to how podman uses user-namespaces and our need to make the convert2rhel checkout writable inside of the container.

Fixes https://issues.redhat.com/browse/RHELC-736

Jira Issue: [RHELC-]()

Checklist


abadger:
Note: The second commit in this PR is a workaround.  It can be reverted once https://github.com/oamg/convert2rhel/pull/608 is merged.

SpyTec:
> Note: The second commit in this PR is a workaround. It can be reverted once #608 is merged.

We can drop the commit, #608 has been merged

SpyTec (approved):

PR #613, opened 2022-09-20
[RHELC-244] allowing the overriding of kmod inhibitors

Andrew-ang9:
This PR allows the user to convert using kmods that are not signed by Red Hat, though this conversion would not be supported. 

In this PR the code has been changed to add an environment variable, as well as a case for when the user has activated the envar. The original log message has been changed as well to reflect this. The unit test has been added for this as well.

Jira Issue: [RHELC-244](https://issues.redhat.com/browse/RHELC-244)

Checklist

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -410,7 +410,7 @@ def validate_package_manager_transaction():
                 "Make sure you have updated the kernel to the latest available version and rebooted the system. "
                 "Only kernel modules supported in RHEL are preferred for conversion, if you want to ignore this "
                 " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS=1' to continue".format(
-                    kmods=not_supported_kmods, system=system_info.name
+                    kmods=not_supported_kmods
                 )
             )
 
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -379,8 +379,8 @@ def ensure_compatibility_of_kmods():
 
         if "CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS" in os.environ:
             logger.warning(
-                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
-                "'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable detected, continuing conversion.".format(
+                "Detected 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable."
+                " We will continue the conversion, with the following kernel modules that are not supported:\n{kmods}\n".format(
                     kmods=not_supported_kmods
                 )
             )
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -373,9 +373,9 @@ def ensure_compatibility_of_kmods():
 
         unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods)
 
-        if "CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS" in os.environ:
+        if "CONVERT2RHEL_ALLOW_UNCHECKED_KMODS" in os.environ:
             logger.warning(
-                "Detected 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable."
+                "Detected 'CONVERT2RHEL_ALLOW_UNCHECKED_KMODS' environment variable."
                 " We will continue the conversion, with the following kernel modules:\n{kmods}\n".format(
                     kmods="\n".join(unsupported_kmods)
                 )
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -385,14 +385,16 @@ def ensure_compatibility_of_kmods():
                 )
             )
 
-        logger.critical(
-            "The following loaded kernel modules are not available in RHEL:\n{0}\n"
-            "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
-            "If this message appears again after doing the above, prevent the modules from loading by following {1}"
-            " and run convert2rhel again to continue with the conversion.".format(
-                "\n".join(unsupported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
+        else:
+            logger.critical(
+                "The following loaded kernel modules are not available in RHEL:\n{0}\n"
+                "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
+                "If this message appears again after doing the above, prevent the modules from loading by following {1}"
+                " and run convert2rhel again to continue with the conversion.".format(
+                    "\n".join(not_supported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
+                )
             )
-        )
+
 
 def validate_package_manager_transaction():
     """Validate the package manager transaction is passing the tests."""
@@ -401,21 +403,7 @@ def validate_package_manager_transaction():
     transaction_handler.run_transaction(
         validate_transaction=True,
     )
-                "Only kernel modules supported in RHEL are preferd for conversion, if you want to ignor this "
-                " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' to continue"
-            ).format(kmods=not_supported_kmods, system=system_info.name)
-        else:
-            logger.critical(
-                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
-                "Make sure you have updated the kernel to the latest available version and rebooted the system. "
-                "Only kernel modules supported in RHEL are preferred for conversion, if you want to ignore this "
-                " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS=1' to continue".format(
-                    kmods=not_supported_kmods
-                )
-            )
 
-    else:
-        logger.info("Kernel modules are compatible.")
 
 def get_loaded_kmods():
     """Get a set of kernel modules loaded on host.
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -371,27 +371,33 @@ def ensure_compatibility_of_kmods():
         logger.debug("All loaded kernel modules are available in RHEL.")
     else:
         not_supported_kmods = "\n".join(
-            map(
-                lambda kmod: "/lib/modules/{kver}/{kmod}".format(kver=system_info.booted_kernel, kmod=kmod),
-                unsupported_kmods,
+            "/lib/modules/{kver}/{kmod}".format(kver=system_info.booted_kernel, kmod=kmod) for kmod in unsupported_kmods
+        )
+
+        logger.critical(
+            "The following loaded kernel modules are not available in RHEL:\n{0}\n"
+            "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
+            "If this message appears again after doing the above, prevent the modules from loading by following {1}"
+            " and run convert2rhel again to continue with the conversion.".format(
+                not_supported_kmods, LINK_PREVENT_KMODS_FROM_LOADING
             )
         )
 
         if "CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS" in os.environ:
             logger.warning(
-                "Detected 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable."
-                " We will continue the conversion, with the following kernel modules that are not supported:\n{kmods}\n".format(
+                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
+                "'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable detected, continuing conversion.".format(
                     kmods=not_supported_kmods
                 )
             )
 
         else:
             logger.critical(
-                "The following loaded kernel modules are not available in RHEL:\n{0}\n"
-                "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
-                "If this message appears again after doing the above, prevent the modules from loading by following {1}"
-                " and run convert2rhel again to continue with the conversion.".format(
-                    "\n".join(not_supported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
+                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
+                "Make sure you have updated the kernel to the latest available version and rebooted the system. "
+                "Only kernel modules supported in RHEL are preferred for conversion, if you want to ignore this "
+                " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS=1' to continue".format(
+                    kmods=not_supported_kmods, system=system_info.name
                 )
             )
 
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -370,9 +370,6 @@ def ensure_compatibility_of_kmods():
     if not unsupported_kmods:
         logger.debug("All loaded kernel modules are available in RHEL.")
     else:
-
-        unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods)
-
         if "CONVERT2RHEL_ALLOW_UNCHECKED_KMODS" in os.environ:
             logger.warning(
                 "Detected 'CONVERT2RHEL_ALLOW_UNCHECKED_KMODS' environment variable."
@@ -380,7 +377,6 @@ def ensure_compatibility_of_kmods():
                     kmods="\n".join(unsupported_kmods)
                 )
             )
-
         else:
             logger.critical(
                 "The following loaded kernel modules are not available in RHEL:\n{0}\n"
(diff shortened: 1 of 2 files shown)


Andrew-ang9:
I am currently working on the unit test right now so at the moment it will fail if it is ran. I will update this after I get it running.

convert2rhel/checks.py
```diff
@@ -284,13 +284,20 @@ def ensure_compatibility_of_kmods():
...
+                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
+                "'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable detected, continuing conversion."
+            ).format(kmods=not_supported_kmods, system=system_info.name)
+
+        else:
+            logger.critical(
                 "The following kernel modules are not supported in RHEL:\n{kmods}\n"
                 "Make sure you have updated the kernel to the latest available version and rebooted the system. "
-                "Remove the unsupported modules and run convert2rhel again to continue with the conversion."
+                "Only kernel modules supported in RHEL are preferd for conversion, if you want to ignor this "
```
r0x0d, line 297:
```suggestion
                "Only kernel modules supported in RHEL are preferred for conversion, if you want to ignore this "
```

convert2rhel/checks.py
```diff
@@ -284,13 +284,20 @@ def ensure_compatibility_of_kmods():
...
+                " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' to continue"
```
r0x0d, line 298: I would add the `=1` to the message to help the user simply copy and paste our suggestion, this way they don't need to think about what they need to set, but rather, knowing that if they set `=1` will work just fine. 

```suggestion
                " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS=1' to continue"
```

convert2rhel/checks.py
```diff
@@ -284,13 +284,20 @@ def ensure_compatibility_of_kmods():
                 unsupported_kmods,
             )
         )
-        logger.critical(
-            (
+        if "CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS" in os.environ:
+            logger.warning(
+                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
+                "'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable detected, continuing conversion."
+            ).format(kmods=not_supported_kmods, system=system_info.name)
```
r0x0d, line 291: Couple of suggestions here… 

First, the `system` parameter in the `format()` is not being used in the string message, did you forget to add this variable in the `str`, or it is not necessary? I would argue that this is not needed at all, and you can remove it. 

```suggestion
            logger.warning(
                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
                "'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable detected, continuing conversion."
            ).format(kmods=not_supported_kmods)
```

Second, I would rephrase this message to output the detection of the environment variable first and then output the kernel modules, like this: 

```suggestion
            logger.warning(
                "Detected 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable, we will continue the conversion even with the following kernel modules that are not supported: ".
                "Kernel modules that are not supported in RHEL:\n{kmods}\n"
            ).format(kmods=not_supported_kmods, system=system_info.name)
```

Feel free to re-arrange this message if you (or any other reviewer) think that it could benefit from a diferent message.

convert2rhel/checks.py
```diff
@@ -284,13 +284,20 @@ def ensure_compatibility_of_kmods():
...
+
+        else:
+            logger.critical(
                 "The following kernel modules are not supported in RHEL:\n{kmods}\n"
                 "Make sure you have updated the kernel to the latest available version and rebooted the system. "
-                "Remove the unsupported modules and run convert2rhel again to continue with the conversion."
+                "Only kernel modules supported in RHEL are preferd for conversion, if you want to ignor this "
+                " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' to continue"
             ).format(kmods=not_supported_kmods, system=system_info.name)
```
r0x0d, line 299: The `system` variable is not present in the string itself, therefore I think it's okay to remove it. 

```suggestion
            ).format(kmods=not_supported_kmods)
```

convert2rhel/checks.py
r0x0d, line 291: There is one missing part of this string formatting that is missing: 
```suggestion
            logger.warning(
                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
                "'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable detected, continuing conversion.".format(kmods=not_supported_kmods, system=system_info.name)
            )
```

You were trying to call the `format()` method that should come from whatever `warning` returns, but that wouldn't work since `warning()` doesn't return anything.  By moving the `format()` to be at the end of the string, then it will be correct.

convert2rhel/checks.py
r0x0d, line 299:
```suggestion
logger.critical(
                "The following kernel modules are not supported in RHEL:\n{kmods}\n"
                "Make sure you have updated the kernel to the latest available version and rebooted the system. "
                "Only kernel modules supported in RHEL are preferd for conversion, if you want to ignor this "
                " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' to continue".format(kmods=not_supported_kmods, system=system_info.name)
            )
```

The same here is true as the other comment I made about where to place the `.format()` function call.

r0x0d:
@Andrew-ang9 I have seen that you are still development the unit_test for this PR, let me know if you need any assistance with the comments I made or with the unit_test itself.

convert2rhel/checks.py
Andrew-ang9: Thank you for catching this, I will have this changed in the next commit that I push!

convert2rhel/checks.py
Andrew-ang9: I will definitely add that in! Thank you

convert2rhel/checks.py
Andrew-ang9: Thank you for this great suggestion!, I will definitely re-write this changed!

convert2rhel/checks.py
Andrew-ang9: Cool thank you for letting me know I will get rid of this here!

convert2rhel/checks.py
Andrew-ang9: ahhh, now that you explain it, you are totally right! thank you for teaching me about this!

convert2rhel/checks.py
Andrew-ang9: gotcha! thank you again for the help with all of your comments for this PR!

convert2rhel/checks.py
```diff
@@ -284,13 +284,24 @@ def ensure_compatibility_of_kmods():
...
+        else:
+            logger.critical(
                 "The following kernel modules are not supported in RHEL:\n{kmods}\n"
                 "Make sure you have updated the kernel to the latest available version and rebooted the system. "
-                "Remove the unsupported modules and run convert2rhel again to continue with the conversion."
-            ).format(kmods=not_supported_kmods, system=system_info.name)
-        )
+                "Only kernel modules supported in RHEL are preferred for conversion, if you want to ignore this "
+                " check, set the environment variable 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS=1' to continue".format(
+                    kmods=not_supported_kmods, system=system_info.name
```
r0x0d, line 301:
```suggestion
                    kmods=not_supported_kmods,
```

You forgot the `system=system_info.name` here.

Andrew-ang9: oops thank you! I thought I got that one I will make sure I get this here as well!

abadger (dismissed):
Code and unittest reviewed.  Looks good to me.  Integration tests didn't pass but it looked like infrastructural problems.  Re-running the failed tests.

r0x0d:
As much as I would like to merge this as soon as possible, there is a merge conflict now :(

Andrew-ang9:
@r0x0d  I am working on rebasing the code now, it has changed one thing in the code so I am fixing it now I believe I am almost done and should have it pushed up asap! 

Thank you for the help with this!

r0x0d (dismissed):

convert2rhel/unit_tests/checks_test.py
```diff
@@ -438,14 +439,34 @@ def test_ensure_compatibility_of_kmods(
...
+        value=run_subprocess_mock,
     )
 
-    checks.validate_package_manager_transaction()
+    checks.ensure_compatibility_of_kmods()
+    should_be_in_logs = (
+        ".*Detected 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable."
+        " We will continue the conversion, with the following kernel modules that are not supported:.*"
+    )
+    assert re.match(should_be_in_logs in caplog.records[-1].message, re.MULTILINE, re.DOTALL)
```
r0x0d, line 469: Your unit tests are failing right now because of this regex, there are a couple of things that need to be changed. 

First, the `re.match` accepts an `pattern`, one `string`, and some `flags`. The way you wrote that was partially correct, except from a few things that were missing, below I have a suggestion that fixes this, but let me explain what changed as well.

```suggestion
    assert re.match(pattern=should_be_in_logs, string=caplog.records[-1].message, flags=re.MULTILINE | re.DOTALL)
```

Instead of having an `pattern` and `string`, you were using only “one” parameter that acted as the `pattern` (`should_be_in_logs in caplog.records[-1].message`), probably this was a leftover after refactoring the unit tests in a previous commit, so no worries. 

Lastly, you are passing the flags to `re.match` as multiple arguments `re.match(..., re.MULTILINE, re.DOTALL)`, kind of something like `re.match(..., flag1=re.MULTILINE, flag=2re.DOTALL)` (just an example, `flag1` and `flag2` doesn't exist), while the correct way of passing more than one flag to `re.match` is by using the `|` operator, like this `re.match(..., flags=re.MULTILINE | re.DOTALL)`. 

Hope you can apply the changes easily and hopefully all unit tests will pass now!

r0x0d (changes requested):

convert2rhel/checks.py
```diff
@@ -370,15 +370,32 @@ def ensure_compatibility_of_kmods():
...
-            "If this message appears again after doing the above, prevent the modules from loading by following {1}"
-            " and run convert2rhel again to continue with the conversion.".format(
-                "\n".join(unsupported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
+
+        not_supported_kmods = "\n".join(
+            map(
+                lambda kmod: "/lib/modules/{kver}/{kmod}".format(kver=system_info.booted_kernel, kmod=kmod),
+                unsupported_kmods,
             )
         )
```
r0x0d, line 379: Is this really necessary? It's different from upstream, which assumes that anything returned from `get_unsupported_kmods` will be the unsupported kmods, so, this variable seems unnecessary in my point of view. 

Maybe something changed since I last reviewed?

r0x0d (dismissed):

convert2rhel/checks.py
Andrew-ang9: I have removed this code and pushed the changes! Thank you for spotting this!

convert2rhel/checks.py
```diff
@@ -370,14 +370,24 @@ def ensure_compatibility_of_kmods():
...
+            )
+
+        else:
+            logger.critical(
+                "The following loaded kernel modules are not available in RHEL:\n{0}\n"
+                "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
+                "If this message appears again after doing the above, prevent the modules from loading by following {1}"
+                " and run convert2rhel again to continue with the conversion.".format(
+                    "\n".join(not_supported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
+                )
```
r0x0d, line 389:
```suggestion
        unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods)
      
        if "CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS" in os.environ:
            logger.warning(
                "Detected 'CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS' environment variable."
                " We will continue the conversion, with the following kernel modules that are not supported:\n{kmods}\n".format(
                    kmods="\n".join(unsupported_kmods)
                )
            )

        else:
            logger.critical(
                "The following loaded kernel modules are not available in RHEL:\n{0}\n"
                "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
                "If this message appears again after doing the above, prevent the modules from loading by following {1}"
                " and run convert2rhel again to continue with the conversion.".format(
                    "\n".join(unsupported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
                )
```

r0x0d: @Andrew-ang9, you removed too much in your last commit. The proper thing here would be this, just like it is in upstream.

convert2rhel/checks.py
```diff
@@ -371,21 +371,24 @@ def ensure_compatibility_of_kmods():
...
                 "The following loaded kernel modules are not available in RHEL:\n{0}\n"
-                "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
-                "If this message appears again after doing the above, prevent the modules from loading by following {1}"
-                " and run convert2rhel again to continue with the conversion.".format(
-                    "\n".join(not_supported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
+                "Kernel modules need to be up-to-date to minimize the risk for issues occurring related to first-party kernel modules.\n"
+                "Ensure you have updated the kernel to the latest available version and rebooted the system.\n"
+                "If this message persists, you can prevent the modules from loading by following {1} and rerun convert2rhel.\n"
+                "To circumvent this check and allow the risk you can set environment variable 'CONVERT2RHEL_ALLOW_UNCHECKED_KMODS=1'.".format(
+                    "\n".join(unsupported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
```
SpyTec, line 391: Since we're changing `CONVERT2RHEL_UNSUPPORTED_UNCHECKED_KMODS` to `CONVERT2RHEL_ALLOW_UNCHECKED_KMODS` you should rename it elsewhere too

Andrew-ang9: gotcha! I will go ahead and make that change, and push it! yeah I don't know how but I forgot to add that change in when I was editing the message.

danmyway:
The verification tests should not be affected by this change.
I would just do a sanity check for the envar.
WDYT @r0x0d @Andrew-ang9 @kokesak ?

r0x0d:
> The verification tests should not be affected by this change. I would just do a sanity check for the envar. WDYT @r0x0d @Andrew-ang9 @kokesak ?

The sanity check would consist of running the conversion with both the new and old env var name, here? What type of sanity check do you have in mind for this?

kokesak:
> > The verification tests should not be affected by this change. I would just do a sanity check for the envar. WDYT @r0x0d @Andrew-ang9 @kokesak ?
> 
> The sanity check would consist of running the conversion with both the new and old env var name, here? What type of sanity check do you have in mind for this?

There is just one env var for this no? What did you mean by old env var? We already have a test for unsupported kmods so I think we can extend this one and just add another test that would use the `CONVERT2RHEL_ALLOW_UNCHECKED_KMODS` env var and validates that the conversion continues.

@danmyway @r0x0d

r0x0d:
> > > The verification tests should not be affected by this change. I would just do a sanity check for the envar. WDYT @r0x0d @Andrew-ang9 @kokesak ?
> > 
> > 
> > The sanity check would consist of running the conversion with both the new and old env var name, here? What type of sanity check do you have in mind for this?
> 
> There is just one env var for this no? What did you mean by old env var? We already have a test for unsupported kmods so I think we can extend this one and just add another test that would use the `CONVERT2RHEL_ALLOW_UNCHECKED_KMODS` env var and validates that the conversion continues.
> 
> @danmyway @r0x0d

Sorry, I miss-read the PR. I agree with you, @kokesak.

convert2rhel/checks.py
```diff
@@ -370,14 +370,27 @@ def ensure_compatibility_of_kmods():
...
         logger.debug("All loaded kernel modules are available in RHEL.")
     else:
-        logger.critical(
-            "The following loaded kernel modules are not available in RHEL:\n{0}\n"
-            "First, make sure you have updated the kernel to the latest available version and rebooted the system.\n"
-            "If this message appears again after doing the above, prevent the modules from loading by following {1}"
-            " and run convert2rhel again to continue with the conversion.".format(
-                "\n".join(unsupported_kmods), LINK_PREVENT_KMODS_FROM_LOADING
+
+        unsupported_kmods = get_unsupported_kmods(host_kmods, rhel_supported_kmods)
```
r0x0d, line 374: Oh, now I realize that we have this line above this if-else statement... I think you can remove it safely now, as we already have the unsupported kmods available. 

I believe it was me that asked you to place it here, so my bad. 
```suggestion

```

Andrew-ang9: gotcha I can definitely make that change and no worries at all!

Andrew-ang9:
I was trying to do a rebase with main and it didn't go right I am leaving work right now and will return tomorrow(12/8/22) to fix it with help!

r0x0d (approved):
Code looks good to me! Will check the test results later.

SpyTec (approved):

SpyTec:
Failing test unrelated. Merging

PR #615, opened 2022-09-28
[RHELC-745] Fix conversion with "LANGUAGE" envar 

Andrew-ang9:
This PR fixes the conversion when you set the "LANGUAGE" envar, because currently you would run into an traceback.

in this PR:
- utf-8 will beset as the default codec to fix the traceback
- "LANGUAGE" will be added to the set of envar alonge with "LC_ALL" and "LANG" 

Jira Issue: [RHELC-745](https://issues.redhat.com/browse/RHELC-745)

Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,14 @@
 minimum_pre_commit_version: "2.9.0"
 repos:
-  - repo: "https://github.com/psf/black"
-    rev: "21.12b0"
+  - repo: local
     hooks:
-      - id: "black"
-        name: "Format code (black)"
+      - id: black
+        name: black
+        entry: black
+        types: [python]
+        language: python
         language_version: "python3"
-        additional_dependencies: [click==7.1.2]
+        additional_dependencies: [click==7.1.2, black==21.12b0]
         args:
           - "--target-version=py27"
           - "--target-version=py36"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,21 +2,26 @@ minimum_pre_commit_version: "2.9.0"
 repos:
   - repo: local
     hooks:
-      - id: black
-        name: black
-        entry: black
-        types: [python]
-        language: python
-        language_version: "python3"
-        additional_dependencies: [click==7.1.2, black==21.12b0]
-        args:
-          - "--target-version=py27"
-          - "--target-version=py36"
-          - "--target-version=py37"
-          - "--target-version=py38"
-          - "--target-version=py39"
-          - "--target-version=py310"
-
+    - id: black
+      name: black
+      entry: black
+      types: [python]
+      language: python
+      language_version: "python3"
+      additional_dependencies: [click==7.1.2, black==21.12b0]
+  - repo: local
+    hooks:
+    - id: pylint
+      name: pylint
+      entry: pylint
+      types: [python]
+      language: python
+      additional_dependencies: [pylint==2.15.10]
+      args:
+        [
+          "-sn", # Don't display the score
+          "--rcfile=.pylintrc", # Link to your config file
+        ]
   - repo: "https://github.com/timothycrosley/isort"
     rev: 5.11.4
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,27 +1,20 @@
 minimum_pre_commit_version: "2.9.0"
 repos:
-  - repo: local
+  - repo: "https://github.com/psf/black"
+    rev: "21.12b0"
     hooks:
-    - id: black
-      name: black
-      entry: black
-      types: [python]
-      language: python
-      language_version: "python3"
-      additional_dependencies: [click==7.1.2, black==21.12b0]
-  - repo: local
-    hooks:
-    - id: pylint
-      name: pylint
-      entry: pylint
-      types: [python]
-      language: python
-      additional_dependencies: [pylint==2.15.10]
-      args:
-        [
-          "-sn", # Don't display the score
-          "--rcfile=.pylintrc", # Link to your config file
-        ]
+      - id: "black"
+        name: "Format code (black)"
+        language_version: "python3"
+        additional_dependencies: [click==7.1.2]
+        args:
+          - "--target-version=py27"
+          - "--target-version=py36"
+          - "--target-version=py37"
+          - "--target-version=py38"
+          - "--target-version=py39"
+          - "--target-version=py310"
+
   - repo: "https://github.com/timothycrosley/isort"
     rev: 5.11.4
     hooks:
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -868,7 +868,7 @@ class DummyPopenOutput(unit_tests.MockFunction):
             return b""
 
         self.call_count += 1
-        return next_line
+        return next_line + b""
 
     def communicate(self):
         pass
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -868,7 +868,7 @@ class DummyPopenOutput(unit_tests.MockFunction):
             return b""
 
         self.call_count += 1
-        return next_line + b""
+        return next_line
 
     def communicate(self):
         pass
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -847,3 +847,40 @@ def test_write_json_object_to_file(data, expected, tmpdir):
         json.load(handler) == expected
 
     assert oct(os.stat(json_file_path).st_mode)[-4:].endswith("00")
+
+
+class DummyPopenOutput(unit_tests.MockFunction):
+    def __init__(self, output):
+        self.call_count = 0
+        self.output = output
+
+    def __call__(self, args, stdout, stderr, bufsize):
+        return self
+
+    @property
+    def stdout(self):
+        return self
+
+    def readline(self):
+        try:
+            next_line = self.output[self.call_count]
+        except IndexError:
+            return b""
+
+        self.call_count += 1
+        return next_line
+
+    def communicate(self):
+        pass
+
+    def poll(self):
+        return 0
+
+
+def test_run_subprocess_env(monkeypatch):
+    test_output = DummyPopenOutput([u"test of nonascii output: café".encode("utf-8")])
+    monkeypatch.setattr(utils.subprocess, "Popen", test_output)
+
+    output, rc = utils.run_subprocess(["echo", "foobar"])
+    assert u"test of nonascii output: café" == output
+    assert 0 == rc
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -150,9 +150,10 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
     )
     output = ""
     for line in iter(process.stdout.readline, b""):
-        output += line.decode()
+        line = line.decode("utf8")
+        output += line
         if print_output:
-            loggerinst.info(line.decode().rstrip("\n"))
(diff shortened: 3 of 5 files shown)


convert2rhel/utils.py
```diff
@@ -144,9 +144,10 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
     )
     output = ""
     for line in iter(process.stdout.readline, b""):
-        output += line.decode()
+        line.decode("utf8")
+        output += line
         if print_output:
-            loggerinst.info(line.decode().rstrip("\n"))
+            loggerinst.info(line.rstrip("\n"))
```
SpyTec, line 150: Previously the output of `decode()` was added to output and logger. With this we just call decode and then use `line` instead. Does `decode()` modify the string itself?

SpyTec (changes requested):

convert2rhel/utils.py
Andrew-ang9: no decode doesn't modify the string itself, that is actually a bug that I introduced thank you for pointing it out!

r0x0d (dismissed):
Looks nice and pretty straightforward to me, just wondering if we should update the unit_tests we have for run_subprocess in utils_test.py to include this change and validate that other locales doesn't interfere with the change.

r0x0d:
Oops, sorry! I clicked the *approved* instead of *comment* :man_facepalming:

r0x0d:
Regarding the error seen here: https://github.com/oamg/convert2rhel/actions/runs/3154772464/jobs/5140788299#step:5:5598 I think if you just rebase and push again, then it should work just fine. I saw that on my PR the other day, and a new rebase worked just fine.

r0x0d:
Hey, @Andrew-ang9, could you rebase this PR to get the latest changes from main? There are some changes in main to overcome the situation where the GitHub actions tests are skipped, and since your PR passed all of them, I believe this is a good PR to have for comparison. 

Thanks!

SpyTec (approved):

convert2rhel/utils.py
```diff
@@ -144,9 +144,10 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
     )
     output = ""
     for line in iter(process.stdout.readline, b""):
-        output += line.decode()
+        line = line.decode("utf8")
+        output += line
```
SpyTec, line 148: Unit tests are failing here

SpyTec (dismissed):

convert2rhel/unit_tests/utils_test.py
```diff
@@ -460,3 +460,27 @@ def test_remove_orphan_folders(path_exists, list_dir, expected, tmpdir, monkeypa
...
+        return self.output.split()
+
+
+def test_run_subprocess_env(monkeypatch):
+    test_output = DummyPopenOutput("test of nonascii output: cafe".encode("utf-8"))
+    monkeypatch.setattr(utils.subprocess, "Popen", test_output)
+
+    output = utils.run_subprocess(["echo", "foobar"])
+
+    assert "test of nonascii output: cafe" == output
```
bocekm, line 486: The `run_subprocess()` function returns a tuple `output, return_code`, not just `output`.

convert2rhel/unit_tests/utils_test.py
```diff
@@ -460,3 +460,27 @@ def test_remove_orphan_folders(path_exists, list_dir, expected, tmpdir, monkeypa
...
+
+    def __call__(self, args, stdout, stderr, bufsize):
+        return self
+
+    @property
+    def stdout(self):
+        return self
+
+    def readline(self):
+        return self.output.split()
```
bocekm, line 477: `split()` splits a string by a whitespace character. So splitting the string you're passing to it "test of nonascii output: cafe" gives you a list of words. The `readline()` function should return a string though.

Also, the stdout is supposed to be a binary stream. So you should convert the string to a binary stream.

bocekm (changes requested):

convert2rhel/utils.py
Andrew-ang9: I have taken care of this error message in the unit test and will push the changes!

convert2rhel/unit_tests/utils_test.py
Andrew-ang9: I have changed this in my new draft of the unit test.

convert2rhel/unit_tests/utils_test.py
Andrew-ang9: Thank you for pointing this out I have made the adjustments to this and will push it!

convert2rhel/utils.py
```diff
@@ -197,7 +198,11 @@ def run_cmd_in_pty(cmd, expect_script=(), print_cmd=True, print_output=True, col
     process = PexpectSpawnWithDimensions(
         cmd[0],
         cmd[1:],
-        env={"LC_ALL": i18n.SCREENSCRAPED_LOCALE, "LANG": i18n.SCREENSCRAPED_LOCALE},
+        env={
+            "LC_ALL": i18n.SCREENSCRAPED_LOCALE,
+            "LANG": i18n.SCREENSCRAPED_LOCALE,
+            "LANGUAGE": i18n.SCREENSCRAPED_LOCALE,
```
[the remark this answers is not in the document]
Andrew-ang9: Thank you for pointing this out I will look into this and make the proper adjustments to the unit test!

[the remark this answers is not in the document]
Andrew-ang9: Hey @bocekm is this something I would implement in the code and unit test or is this strictly for the integration test?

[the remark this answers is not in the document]
bocekm: This is just about updating the existing or creating a new integration test.

As of now we've got this tier1/destructive test, setting just the LANG env var: https://github.com/oamg/convert2rhel/blob/v1.0/tests/integration/tier1/set-locale/use_non_english_language.py. That tests the issue where LANG unexpectedly has a precedence over LC_ALL in subscription-manager: https://github.com/oamg/convert2rhel/pull/386#issuecomment-987584120.

However, we should expect that other tools follow POSIX where LC_ALL has a precedence over LANG. So I think we should have a test with setting LANG to a non-English locale.

And lastly, there's the LANGUAGE variable. It's not defined in [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_02) but in [GNU](https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html). It has a precedence over any other locale env var. I'd say that we should have a test with setting LANGUAGE to a non-English locale.

@abadger, @danmyway, I'm not sure whether it's "safe" setting LANGUAGE, LC_ALL, and LANG in one test. I.e. whether having one test per env var would not be capable of revealing more. For the sake of saving computational power/time, I'm leaning towards saying set them all in one test. WDYT?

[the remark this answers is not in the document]
r0x0d: @Andrew-ang9 I believe @bocekm is talking about the integration tests that should be added to make it more robust. It's not regarding your code/unit tests.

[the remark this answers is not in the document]
abadger: @bocekm I agree, it is safe to set all of them in one test. The reason just a single test is fine is that the failure case we're looking for is when a program that we call returns translated strings instead of the strings from the `C` locale.  It's enough for us to set all of the environment variables that might cause this in a single test as any of them causing the program to return translated strings will cause the test to fail.  We might also want to set `LC_MESSAGES` to a non-English locale (This variable should not take precedence over `LC_ALL` or `LANGUAGE` but it also affects translations so it is probably good to make sure no programs are getting the precedence order incorrect).

abadger:
Shoot, looks like when we rebased, we lost a lot of work.  I'll help Andrew recover those changes.

r0x0d:
> Shoot, looks like when we rebased, we lost a lot of work. I'll help Andrew recover those changes.

Maybe using reflog you guys can go back in history? Let me know if you guys need helps

abadger:
Note: The rebase issue has been fixed.

.pre-commit-config.yaml
```diff
@@ -1,11 +1,19 @@
 minimum_pre_commit_version: "2.9.0"
 repos:
   - repo: "https://github.com/psf/black"
-    rev: "22.10.0"
+    rev: "21.12b0"
```
r0x0d, line 4: Just to note that this is not enough to keep pre-commit to open a pull request every week, updating the version o black for us. 

One of the temporary workarounds (as far as I know) is to switch this to a `local` repo instead of the black GitHub URL. 

It would be something like this: 

```yaml
-   repo: local
    hooks:
    -   id: black
        name: black
        entry: black
        types: [python]
        language: system
        additional_dependencies: [click==7.1.2, black==21.12b0]
```

Although, I haven't tested this suggestion, but I think something like this would prevent pre-commit from auto-updating the version for black every week. See https://github.com/pre-commit/pre-commit-hooks/issues/332 for more details.

r0x0d: @abadger

r0x0d (dismissed):
Made a comment about the current state for the pre-commit config file, while this is not a blocker, and we can push this to upstream (in my opinion), it is good to take a look at that before we keep receiving PRs from pre-commit to update `black` version.

.pre-commit-config.yaml
SpyTec: One thing to note here. Newer black versions don't support running on v3.6. Meaning it won't work for my dev container PR 

#672

r0x0d: You need the version `22.8.0` in #672, and you also downgraded the version for pre-commit-hooks... 

@abadger do you know if the issue that Andrew was having because of black is happening in `22.8.0`? If not, maybe we can try to pin that version down as it is more recent. 

As far #672, I would suggest you to do the same there, @SpyTec, at least for the pre-commit-hooks repo.

abadger: 21.12b0 is the latest version that won't break python2 code.  It looks like #672 is pinning a downgraded version of black so we should try to pin 21.12b0 there as well?

r0x0d: Honestly, I think that for reducing complexity, we should create another PR that has the change and merge it right away... Since it's a change to the pre-commit config, we don't require any other CI check other than the pre-commit verification.

r0x0d (approved):

r0x0d:
Since @Andrew-ang9 is out for PTO for the next two weeks, I'm rebasing this one and fixing the merge conflict it has in the `.pre-commit-config.yaml` file, as the changes were originated in a thread discussion here.

convert2rhel/utils.py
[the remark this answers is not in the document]
r0x0d: Also, the tests seem to be of tier1, so it's a destructive test. I think it's fine to set them without taking care of restoring it to original.

r0x0d (changes requested):
I think your implementation here is correct, @Andrew-ang9. I haven't checked the result of the integration tests yet, but as far as it seems, you need to rebase this PR to fix the .pre-commit-config.yaml. 

After your rebase, I will double-check the PR and see the results of the integration tests.

.pre-commit-config.yaml
```diff
@@ -17,6 +17,7 @@ repos:
           - "--target-version=py39"
           - "--target-version=py310"
 
+
```
r0x0d, line 20: @Andrew-ang9, you can remove this blank space between the hooks, normally, this wouldn't have any spaces between them.

Andrew-ang9: I have went in and gotten rid of this extra space, thank you for letting me know about this!

SpyTec (commented):
Overall seems good, not tested. @r0x0d I'll leave it up to you to merge

r0x0d:
I will give this a round of tests later, if it works, I will merge it.

r0x0d (changes requested):
@Andrew-ang9, please, do a new rebase to fix the conflict.

r0x0d (approved):

PR #557, opened 2022-08-05
[RHELC-689] Handle non zero output from repoquery

r0x0d:
We were not handling non 0 return code from repoquery call when checking
for latest kernel available on the repositories.

This is intended to fix and not let the tool to raise an unexpected
exception.

<details>
<summary>Click to see an example of the traceback as logged by convert2rhel.</summary>

```
15:58:01         out: [08/05/2022 11:58:01] DEBUG - Calling command 'repoquery --quiet --qf "%{BUILDTIME}\t%{VERSION}-%{RELEASE}\t%{REPOID}" kernel'
15:58:01         out: [08/05/2022 11:58:01] DEBUG - Traceback (most recent call last):
15:58:01         out:   File "/usr/lib/python2.7/site-packages/convert2rhel/main.py", line 94, in main
15:58:01         out:     checks.perform_pre_checks()
15:58:01         out:   File "/usr/lib/python2.7/site-packages/convert2rhel/checks.py", line 68, in perform_pre_checks
15:58:01         out:     is_loaded_kernel_latest()
15:58:01         out:   File "/usr/lib/python2.7/site-packages/convert2rhel/checks.py", line 660, in is_loaded_kernel_latest
15:58:01         out:     _, latest_kernel, repoid = packages[0]
15:58:01         out: ValueError: need more than 1 value to unpack
```
</details>

Signed-off-by: Rodolfo Olivieri <rolivier@redhat.com>

PR #607, opened 2022-09-13
[RHELC-738] Add RHEL 8.6 among the supported EUS minor versions

bocekm:
When RHEL 8.7 is released, RHEL 8.6 will transition into the Extended Update Support. Our next release will happen post RHEL 8.7 GA so this is to make sure we then correctly detect 8.6 as EUS.

Some integration may be failing now as:
- they don't expect 8.6 to be EUS
- different RHEL repos are being enabled
- some checks during the conversion are being skipped if converting to EUS, namely loaded kernel currency check and system up-to-date check on Oracle Linux (because Oracle does not make repos of older minor versions available)
- the convert2rhel will start locking to the minor version through RHSM

Jira Issue: [RHELC-738](https://issues.redhat.com/browse/RHELC-738)

Checklist
- [ ] PR meets acceptance criteria specified in the Jira issue
- [x] PR has been tested manually in a VM (either author or reviewer)
- [x] Jira issue has been made public if possible
- [x] `[RHELC-738]` is part of the PR title 
- [ ] Code and tests are documented properly
- [x] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending`

PR #617, opened 2022-09-29
[RHELC-689] Verification test - Latest kernel check: Handle unsuccessful repoquery call

danmyway:
Add a verification test with intentionally tainted repository to verify handling unsuccessful repoquery.

* add verification test for #557
* tested with old build and build for PR#557

Jira Issue: [RHELC-689](https://issues.redhat.com/browse/RHELC-689)

Checklist

### DIFF
--- a/tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
+++ b/tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
@@ -1,3 +1,5 @@
+import platform
+
 import pytest
 
 
@@ -8,15 +10,16 @@ def test_verify_latest_kernel_check_passes_with_failed_repoquery(shell, convert2
     Failed repoquery subsequently caused the latest kernel check to fail.
     Introduced fixes should get the process past the latest kernel check.
     """
-
-    repopath = "/etc/yum.repos.d/tainted_repo.repo"
-    tainted_repo_setup = "[tainted-repo]\nname=Tainted repository\nbaseurl=http://download.eng.bos.redhat.com/beakerrepos/client/Fedor36/\nenabled=1\ngpgcheck=0"
+    get_system_release = platform.platform()
+    repofile = "broken_repo"
 
     # Add a tainted repository with intentionally bad baseurl so the repoquery fails successfully.
-    with open(repopath, "w") as tainted_repo:
-        tainted_repo.write(tainted_repo_setup)
-
-    assert shell(f"cat {repopath}").output == tainted_repo_setup
+    # For CentOS EUS we are working with hardcoded repos in /usr/share/convert2rhel/repos/
+    # We are skipping (at the test metadata level) the test for Oracle Linux EUS version,
+    # as the repositories are not available at all
+    if "centos-8.4" in get_system_release:
+        shell(f"cp -r files/{repofile}.repo /usr/share/convert2rhel/repos/")
+    shell(f"cp -r files/{repofile}.repo /etc/yum.repos.d/")
 
     # Run the conversion just past the latest kernel check, if successful, end the conversion there.
     with convert2rhel("--debug --no-rpm-va") as c2r:
@@ -34,4 +37,6 @@ def test_verify_latest_kernel_check_passes_with_failed_repoquery(shell, convert2
     assert c2r.exitstatus != 0
 
     # Cleanup the tainted repository.
-    assert shell(f"rm -f {repopath}").returncode == 0
+    if "centos-8.4" in get_system_release:
+        assert shell(f"rm -f /usr/share/convert2rhel/repos/{repofile}.repo") == 0
+    assert shell(f"rm -f /etc/yum.repos.d/{repofile}.repo").returncode == 0
--- a/tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
+++ b/tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
@@ -12,13 +12,19 @@ def test_verify_latest_kernel_check_passes_with_failed_repoquery(shell, convert2
     """
     get_system_release = platform.platform()
     repofile = "broken_repo"
+    centos_custom_reposdir = "/usr/share/convert2rhel/repos/"
 
     # Add a tainted repository with intentionally bad baseurl so the repoquery fails successfully.
-    # For CentOS EUS we are working with hardcoded repos in /usr/share/convert2rhel/repos/
+    # For CentOS we are working with hardcoded repos in /usr/share/convert2rhel/repos/centos-8.{4,5}
     # We are skipping (at the test metadata level) the test for Oracle Linux EUS version,
     # as the repositories are not available at all
+
+    # TODO after the #619 gets merged, squash condition to centos-8 only
+    # TODO and copy to {centos_custom_reposdir}/{get_system_release}/
     if "centos-8.4" in get_system_release:
-        shell(f"cp -r files/{repofile}.repo /usr/share/convert2rhel/repos/")
+        shell(f"cp -r files/{repofile}.repo {centos_custom_reposdir}/centos-8.4/")
+    elif "centos-8.5" in get_system_release:
+        shell(f"cp -r files/{repofile}.repo {centos_custom_reposdir}/centos-8.5/")
     shell(f"cp -r files/{repofile}.repo /etc/yum.repos.d/")
 
     # Run the conversion just past the latest kernel check, if successful, end the conversion there.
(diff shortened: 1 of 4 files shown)


tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
```diff
@@ -0,0 +1,27 @@
...
+
+    repopath = "/etc/yum.repos.d/beaker.repo"
+
+    # Add a tainted repository with intentionally bad baseurl so the repoquery fails successfully.
+    assert shell(f"touch {repopath}").returncode == 0
+    assert (
+        shell(
+            "printf '[beaker-client]\nname=Beaker Client - Fedora36\nbaseurl=http://download.eng.bos.redhat.com/\nenabled=1\ngpgcheck=0' > /etc/yum.repos.d/beaker.repo"
+        )
+    ).returncode == 0
```
r0x0d, line 16: A simple question here... Do we really need this to be a shell call only to assert that it created successfully the file? 

I'm thinking more towards using the `with open(...)` in python to make it more readable, but I'm not sure if using a shell here to assert the file creation is more correct than a python statement.

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
```diff
@@ -0,0 +1,27 @@
...
+
+    # Run the conversion just past the latest kernel check, if successful, end the conversion there.
+    with convert2rhel("--debug --no-rpm-va") as c2r:
+        assert c2r.expect("Continue with the system conversion?", timeout=300) == 0
+        c2r.sendline("y")
+        assert c2r.expect("Continue with the system conversion?", timeout=300) == 0
+        c2r.sendline("n")
+    assert c2r.exitstatus != 0
```
r0x0d, line 24: Should we assert that the message was correctly sent to the user too? 

```python
    with convert2rhel("--debug --no-rpm-va") as c2r:
        assert c2r.expect("Couldn't fetch the list of the most recent kernels available in the repositories. Skipping the loaded kernel check." ,timeout=300) == 0
        assert c2r.expect("Continue with the system conversion?", timeout=300) == 0
        c2r.sendline("y")
        assert c2r.expect("Continue with the system conversion?", timeout=300) == 0
        c2r.sendline("n")
    assert c2r.exitstatus != 0
```

r0x0d (commented):
The tests looks good for me... I only got one question regarding the 8.4 conversions, what should we do in this case? 

I see that you are skipping if the system is ol-8.4, but what in the case of centos-8.4? We have a condition in the code to append the custom reposdir to the repoquery command if we detect the custom reposdir based on the version (currently, only centos-8.4). Should we include a test that add the tainted repository to the custom reposdir as well?

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
danmyway: I'd like to assert, the repofile was created successfully.
But you are right, I may do it more in a python way, and assert a simple `cat` to check for the repofile.

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
danmyway: Good point!
Thanks @r0x0d.

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
```diff
@@ -0,0 +1,37 @@
...
+@pytest.mark.latest_kernel_check_with_failed_repoquery
+def test_verify_latest_kernel_check_passes_with_failed_repoquery(shell, convert2rhel):
+    """
+    This test verifies, that failed repoquery is handled correctly.
+    Failed repoquery subsequently caused the latest kernel check to fail.
+    Introduced fixes should get the process past the latest kernel check.
+    """
+
+    repopath = "/etc/yum.repos.d/tainted_repo.repo"
+    tainted_repo_setup = "[tainted-repo]\nname=Tainted repository\nbaseurl=http://download.eng.bos.redhat.com/beakerrepos/client/Fedor36/\nenabled=1\ngpgcheck=0"
```
kokesak, line 13: I would create a separate file called something like "broken_repo.repo" (maybe not use the word tainted, but that just my personal preference, so you can keep it :D) as a part  of this test folder instead of having the string here in the pytest file, that is hard to read. In the pytest just copy/move the file to the "/etc/yum.repos.d/". When the test will run, tmt should copy all the files to the CWD.

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Add a tainted repository with intentionally bad baseurl so the repoquery fails successfully.
+    with open(repopath, "w") as tainted_repo:
+        tainted_repo.write(tainted_repo_setup)
+
+    assert shell(f"cat {repopath}").output == tainted_repo_setup
```
kokesak, line 19: I don't think this is necessary

kokesak (changes requested):

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
danmyway: You're right, I will fix that, thanks

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
danmyway: Indeed, that is probably some debuging remnant. Thank you!

danmyway:
> The tests looks good for me... I only got one question regarding the 8.4 conversions, what should we do in this case?
> 
> I see that you are skipping if the system is ol-8.4, but what in the case of centos-8.4? We have a condition in the code to append the custom reposdir to the repoquery command if we detect the custom reposdir based on the version (currently, only centos-8.4). Should we include a test that add the tainted repository to the custom reposdir as well?



> The tests looks good for me... I only got one question regarding the 8.4 conversions, what should we do in this case?
> 
> I see that you are skipping if the system is ol-8.4, but what in the case of centos-8.4? We have a condition in the code to append the custom reposdir to the repoquery command if we detect the custom reposdir based on the version (currently, only centos-8.4). Should we include a test that add the tainted repository to the custom reposdir as well?

Good point @r0x0d, I have added the condition

kokesak (approved):
LGTM! Thanks

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
```diff
@@ -0,0 +1,50 @@
...
+    Failed repoquery subsequently caused the latest kernel check to fail.
+    Introduced fixes should get the process past the latest kernel check.
+    """
+    get_system_release = platform.platform()
+    repofile = "broken_repo"
+    centos_custom_reposdir = "/usr/share/convert2rhel/repos/"
+
+    # Add a tainted repository with intentionally bad baseurl so the repoquery fails successfully.
+    # For CentOS we are working with hardcoded repos in /usr/share/convert2rhel/repos/centos-8.{4,5}
+    # We are skipping (at the test metadata level) the test for Oracle Linux EUS version,
```
r0x0d, line 19: Should we skip this test for OL8.6 too? After we merge #607 it will be considered EUS too.

r0x0d (approved):
Looks good to me, only one comment about considering OL 8.6 as EUS too (not sure if in this PR, or in #607)

tests/integration/tier0/verify-latest-kernel-check-with-failed-repoquery/test_verify_kernel_check_passes.py
kokesak: Good point, I don't mind both. Maybe so it's more clear we can keep it as is and make it skip in #607

danmyway: Let's address all the EUS related changes in #607 
Good catch though @r0x0d !

r0x0d: Seems good to me

r0x0d:
@danmyway I think this is good to go... Do you want to merge it or wait a bit more?

danmyway:
Let's merge it @r0x0d, thank you

PR #621, opened 2022-09-30
[RHELC-749] Add tmt-lint to pre-commit hooks

danmyway:
Add tmt-lint to verify correct fmf syntax.

For now introduce tmt-tests-lint, once we move away from GitLab hosted values, add tmt-plans-lint which is now failing due to SSL error whilst trying to reach to internal GitLab.

[RHELC-749](https://issues.redhat.com/browse/RHELC-749)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,3 +23,8 @@ repos:
       - id: "check-toml"
       - id: "check-yaml"
       - id: "check-merge-conflict"
+
+  - repo: "https://github.com/teemtee/tmt"
+    rev: main
+    hooks:
+      - id: tmt-lint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,8 +23,7 @@ repos:
       - id: "check-toml"
       - id: "check-yaml"
       - id: "check-merge-conflict"
-
-  - repo: "https://github.com/teemtee/tmt"
-    rev: main
+  - repo: "https://github.com/teemtee/tmt.git"
+    rev: "1.17.0"
     hooks:
       - id: tmt-lint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,4 +26,4 @@ repos:
   - repo: "https://github.com/teemtee/tmt.git"
     rev: "1.17.0"
     hooks:
-      - id: tmt-lint
+      - id: "tmt-tests-lint"


r0x0d (changes requested):
Lovely idea, @danmyway! I will just recommend using tags instead of branches for pre-commit!

.pre-commit-config.yaml
```diff
@@ -23,3 +23,8 @@ repos:
       - id: "check-toml"
       - id: "check-yaml"
       - id: "check-merge-conflict"
+
+  - repo: "https://github.com/teemtee/tmt"
+    rev: main
+    hooks:
+      - id: tmt-lint
```
r0x0d, line 30:
```suggestion
  - repo: "https://github.com/teemtee/tmt"
    rev: "1.17.0"
    hooks:
      - id: tmt-lint
```

r0x0d:
Also, I noticed that this take extremely long to run and finish... Don't know if that has something to do with cache generation on my side, have you noticed the same?

.pre-commit-config.yaml
danmyway: Oh cool, I was wondering why it was complaining :))
Thank you, @r0x0d!

danmyway:
> Also, I noticed that this take extremely long to run and finish... Don't know if that has something to do with cache generation on my side, have you noticed the same?

Do you mean the pre-commit hook?

danmyway:
There might be an issue with it trying all fmf files.
We might introduce just tmt-tests-lint and tmt-plans-lint as we are not yet using stories

r0x0d:
> > Also, I noticed that this take extremely long to run and finish... Don't know if that has something to do with cache generation on my side, have you noticed the same?
> 
> Do you mean the pre-commit hook @r0x0d ?

Yeah, exactly... Sorry for not clarifying. The hook itself takes too long to finish.

danmyway:
It takes a bit more seconds to finish locally, but it does finish

r0x0d:
> It takes a bit more seconds to finish locally, but it does finish

Cool. I see that the new hook you are using is fast, so no worries. The `tmt-lint` was super slow for some reason...

danmyway:
Yep, going with just tmt-tests-lint now, there is some issue with plans linting.
I will reach out to the tmt folks on Monday.

r0x0d:
@danmyway I'm okay to merge that as it looks okay from my point of view, but it is still in draft, does it need anything else to be added?

danmyway:
> @danmyway I'm okay to merge that as it looks okay from my point of view, but it is still in draft, does it need anything else to be added?

@r0x0d I was looking into the tmt-plans-lint issue, it seems it has problem to reach to the secrets file over at gitlab
So for the time being I would go with just tests-lint and add the plans whenever we move away from the gitlab secrets

r0x0d (approved):

danmyway:
> @danmyway can we get a public Jira issue with story point of 1 created? Once that's done feel free to merge

Issue created @SpyTec, merging.
Thank you all.

PR #619, opened 2022-09-30
[RHELC-757] Make tier0 tests run sequentially

danmyway:
Signed-off-by: Daniel Diblik <ddiblik@redhat.com>

PoC PR to see how running tier0 tests in sequence will affect the CI.

Jira Issue: [RHELC-757](https://issues.redhat.com/browse/RHELC-757)

Checklist
- [x] PR meets acceptance criteria specified in the Jira issue
- [x] PR has been tested manually in a VM (either author or reviewer)
- [x] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [x] Code and tests are documented properly
- [x] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending`

PR #623, opened 2022-10-05
[RHELC-642] Fix incorrect kernel signature check

r0x0d:
Currently, we found a corner case in our kernel verification where if you have a different kernel vendor installed on the system (A CentOS system with a OL kernel), then this verification somestimes may fail, since we were getting only the first occurrence of the kernel installed on the system, and not the one that is booted.

This commit is to fix this behavior and get the correct instellad kernel rather than the first one that it found when querying the installed kernels on the system.

Jira Issue: [RHELC-642](https://issues.redhat.com/browse/RHELC-642)

Checklist

# TODO 

- [x] Merge the commit from https://github.com/oamg/convert2rhel/pull/529 into this one (or vice-versa)
- [ ] Copied over a fixture present in https://github.com/oamg/convert2rhel/pull/619


r0x0d:
The tests locally were done using this command to install & make the kernel default: 

```bash
yum install https://yum.oracle.com/repo/OracleLinux/OL7/latest/x86_64/getPackage/kernel-3.10.0-1160.76.1.0.1.el7.x86_64.rpm -y && grub2-set-default 'Oracle Linux Server (3.10.0-1160.76.1.0.1.el7.x86_64) 7.9' && reboot
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -856,14 +856,14 @@ def test_bad_kernel_substring(kernel_release, exp_return, monkeypatch):
...
-            "kernel-core",
+            "4.18.0&240.22.1.el8_3&x86_64&kernel-core",
             "05b555b38483c65d",
             "yajl.x86_64",
             False,
         ),
         (
             "4.18.0-240.22.1.el8_3.x86_64",
-            "kernel-core",
+            "4.18.0&240.22.1.el8_3&x86_64&kernel-core",
```
SpyTec, line 866: Shouldn't we test with multiple kernel packages in different orders to make sure we meet the acceptance criteria?

r0x0d: good point! let me work on that

r0x0d: So, @SpyTec, I came back to this code and tested it once again, I just validated with the help of the integration test that @kokesak made that we only return 1 package in get_installed_pkg_objects because we are passing the NEVRA of the package in the search. I don't think there is a way to return more than one package to be returned in that function, since we are getting the package NEVRA from a file located in vmlinuz_path = "/boot/vmlinuz-%s" % kernel_release, and, assuming that the user has a custom kernel installed (e.g: an Oracle Linux kernel installed on a CentOS Linux system), and, the file located in vmlinuz_path points to that package, then I'm not sure if we will return anything else.

The test the way it is right now is somewhat correct, as it tries to validate (in)correct package signatures.

Probably I misunderstood your question in the first place, too. Could you provide a simple example here in the comments, so I can try to see if that would be worth it to include?

SpyTec: > we only return 1 package in get_installed_pkg_objects because we are passing the NEVRA of the package in the search

Ohhh gotcha, then yeah you can ignore my suggestion

SpyTec (approved):

danmyway (changes requested):
LGTM, nice!
I have just added a couple of change requests to align with the PR #619 proposals.
* move the install_custom_kernel() to the test_custom_kernel.py
  * add a `tmt-reboot` call, therefore we would be able to get rid of the preparation phase in the test plan
* add a cleanup phase to restore the system
* rename the `system_version` to `get_system_release` so there would be less refactoring after the tier0 PoC gets merged

tests/integration/tier0/custom-kernel/test_cutsom_kernel.py
```diff
@@ -0,0 +1,25 @@
+import platform
+
+
+def test_custom_kernel(convert2rhel):
+    """
```
danmyway, line 5:
```suggestion

@pytest.mark.custom_kernel
def test_custom_kernel(convert2rhel):
    """
```

tests/integration/tier0/custom-kernel/test_cutsom_kernel.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def install_custom_kernel(shell):
+    """
+    Install CentOS kernel on Oracle Linux and vice versa to mimic the custom
+    kernel that is not signed by the running OS official vendor.
+    """
+    get_system_release = platform.platform()
+
+    yum_install_cmd, grub_cmd = DISTRO_KERNEL_MAPPING[get_system_release].values()
```
kokesak, line 43:
```
14:56:03                 out:     def install_custom_kernel(shell):
14:56:03                 out:         """
14:56:03                 out:         Install CentOS kernel on Oracle Linux and vice versa to mimic the custom
14:56:03                 out:         kernel that is not signed by the running OS official vendor.
14:56:03                 out:         """
14:56:03                 out:         get_system_release = platform.platform()
14:56:03                 out:     
14:56:03                 out: >       yum_install_cmd, grub_cmd = DISTRO_KERNEL_MAPPING[get_system_release].values()
14:56:03                 out: E       KeyError: 'Linux-4.18.0-305.25.1.el8_4.x86_64-x86_64-with-centos-8.4.2105'
14:56:03                 out: 
14:56:03                 out: test_cutsom_kernel.py:43: KeyError
```

The problem is that `platform.platform()` returns more than just "centos-8". It's somehow simmilar to `uname -a` command.

kokesak: @r0x0d maybe We can try to use/copy this: https://github.com/oamg/convert2rhel/pull/619/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56R243-R287

Just copy paste it into this file and once @danmyway PR is merged we can remove it and use it just as a pytest fixture?

WDYT? @danmyway @r0x0d ?

r0x0d: Oh :facepalm:. You're correct, @kokesak. 

I don't mind copying this over and later opening a PR to turn into a fixture. But since his code is in a `conftest.py`, I think I can apply the same here and use it as a fixture right now... Probably we can get away with merge conflicts if I place the fixture in the right place. 

Let me try.

kokesak: Yeah, why not. Lte's try it

r0x0d: Also, @kokesak, not sure on what to do with the clean_up_custom_kernel function... Any ideas?

kokesak: @r0x0d, before the installation of different kernel I would try to save to some variable the currently loaded kernel, because we will want to get back to it. Maybe try to call `grubby --default-kernel` and save it. Then procceed with the test installing the new kernels.
When the test finishes, set the default krenel back to the original that it is saved inside the variable and do the reboot using the `tmt-reboot` again. Lastly remove the newly installed kernel (something like `yum undo last` or like that have the systsem in same state as before the test)

r0x0d: Cool, I wll try that. Thanks!

r0x0d:
Hand-picking some changes from https://github.com/oamg/convert2rhel/pull/653

kokesak:
Hmm
```
20:08:19                 out: default_kernel = Result(returncode=0, output='kernel-core-4.18.0-348.7.1.el8_5.x86_64\n')
20:08:19                 out: installed_kernel = Result(returncode=0, output='kernel-core-4.18.0-348.7.1.el8_5.x86_64\n')
20:08:19                 out: shell = <function shell.<locals>.factory at 0x7fc1e8e86c80>
20:08:19                 out: 
20:08:19                 out:     def clean_up_custom_kernel(default_kernel, installed_kernel, shell):
20:08:19                 out:         """
20:08:19                 out:         Remove the current installed kernel and install the machine default kernel.
20:08:19                 out:         """
20:08:19                 out: >       assert shell("rpm -e %s" % installed_kernel).returncode == 0
20:08:19                 out: E       TypeError: not all arguments converted during string formatting
```
I wonder what went wrong here

tests/integration/tier0/custom-kernel/test_cutsom_kernel.py
```diff
@@ -0,0 +1,109 @@
+import os
+
+import pytest
+
+
+SYSTEM_RELEASE = os.environ["SYSTEM_RELEASE"]
```
kokesak, line 6: I believe here it should be:

```suggestion
SYSTEM_RELEASE = os.environ.get("SYSTEM_RELEASE")
```

r0x0d: applied.

SpyTec: Either or works fine, we use normal array access in most cases

tests/integration/tier0/custom-kernel/test_custom_kernel.py
```diff
@@ -0,0 +1,129 @@
...
+
+    shell("tmt-reboot -t 600")
+
+
+def clean_up_custom_kernel(shell, original_kernel_uname):
+    """
+    Remove the current installed kernel and install the machine default kernel.
+    """
+    original_kernel, custom_kernel = DISTRO_KERNEL_MAPPING[SYSTEM_RELEASE].values()
+    original_kernel_release = original_kernel.rsplit("/")[-1].replace(".rpm", "").split("-", 2)[-1]
```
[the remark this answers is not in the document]
SpyTec: This inline notices showing up on the timeline of the PR is great. Indeed, `original_kernel_release` is not being used in the test

r0x0d:
Why we don't make the integration tests here as a tier1 and make it destructible? I think that we are having a lot of trouble trying to figure out a good way of cleaning up the custom kernel

kokesak:
The test finallyw works, I would leave it as is at the moment and merge it. We can fix the clean-up phase later when the tier0 will be merged into one big plan (https://github.com/oamg/convert2rhel/pull/619).

danmyway (approved):
Looks good, thank you.

PR #624, opened 2022-10-07
[RHELC-760] Increase the line length in a pseudo-terminal

bocekm:
The name of the rpm we download during a package backup can be so long that the length of the line of 120 characters we set in the pseudo-terminal may not be sufficient. If the downloaded package file name (NEVRA.rpm) is too long, the yumdownloader truncates it and then our output parsing fails. That leads to not tracking the downloaded package and, more importantly, not installing the package back during a rollback.

Real world example of a truncated rpm name:
`libvirt-daemon-driver-storage-core-8.0.0-5.4.0.1.module+el8.6.0+20743+999ad699.x86_64.r 326 kB/s | 253 kB     00:00`

Jira Issue: [RHELC-760](https://issues.redhat.com/browse/RHELC-760)

Checklist

### DIFF
--- a/convert2rhel/utils.py
+++ b/convert2rhel/utils.py
@@ -156,7 +156,7 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
     return output, return_code
 
 
-def run_cmd_in_pty(cmd, expect_script=(), print_cmd=True, print_output=True, columns=120):
+def run_cmd_in_pty(cmd, expect_script=(), print_cmd=True, print_output=True, columns=150):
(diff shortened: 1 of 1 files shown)


SpyTec:
This seems like a bug fix so I would still create a Jira issue for it. @bocekm could you create one for Q4 with Story point of 1 in that case?

It relates to the codebase in some way so it is always good to document what we've done

r0x0d (approved):
LGTM, approved and ready to be merged after the Jira issue is created, and the PR updated accordingly.

PR #625, opened 2022-10-10
[DRAFT][RHELC-753] Add tier1 sanity test

danmyway:
* add a test for basic conversion just to verify sanity

* Test to verify sanity of a full conversion.
* To be possible to run locally 
  * `tmt run -a provision -h connect -g {vm-machine-ip} plans --name /plans/tier1/sanity`
  * `tmt run -a provision -h connect -g {vm-machine-ip} tests --filter 'tier: 1' --filter 'tag: sanity'`

Jira Issue: [RHELC-753](https://issues.redhat.com/browse/RHELC-753)

Checklist

### DIFF
--- a/plans/tier1.fmf
+++ b/plans/tier1.fmf
@@ -328,3 +328,7 @@
   - name: reboot after conversion
     how: ansible
     playbook: tests/ansible_collections/roles/reboot/main.yml
+
+/sanity:
+    discover+:
+        test: sanity
--- a/tests/integration/tier1/sanity/main.fmf
+++ b/tests/integration/tier1/sanity/main.fmf
@@ -1,0 +1,11 @@
+summary: Conversion sanity test
+description: Verify basic conversion with no preparation phase as a sanity test.
+
+tier: 1
+
+tag+:
+    - sanity
+    - basic-conversion
+
+test: |
+    pytest -svv
--- a/tests/integration/tier1/sanity/test_sanity_conversion.py
+++ b/tests/integration/tier1/sanity/test_sanity_conversion.py
@@ -1,0 +1,20 @@
+from envparse import env
+
+
+def test_sanity_conversion(convert2rhel, shell):
+    """
+    Sanity only.
+    Verify successful full conversion.
+    No preparation or adjustments.
+    """
+
+    with convert2rhel(
+        ("-y --no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug").format(
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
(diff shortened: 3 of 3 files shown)


SpyTec:
Awesome. Can you add a command to the Makefile too so it's even easier?

r0x0d:
Following @SpyTec suggestion, @danmyway if you need to add it to the Makefile, which certainly is a nice idea, you can use this as a inspiration to get the machine ip (using vagrant) 

```
copy-logs:
        $(eval HOST=$(shell vagrant ssh-config | grep HostName | tr -d 'HostName' | tr -d ' '))
        scp -i $(PWD)/.vagrant/machines/default/libvirt/private_key root@$(HOST):/var/log/convert2rhel/convert2rhel.log convert2rhel.log
```

danmyway:
> Awesome. Can you add a command to the Makefile too so it's even easier?

I would gladly do so, I'm not sure about the approach, though.
Any tips @SpyTec @r0x0d ?

SpyTec:
In the [`Makefile` in the project root](https://github.com/oamg/convert2rhel/blob/main/Makefile) I'd make something similar to how lint is setup

I expect it would be something like this, though I haven't tested it:
```
sanity-copy-ip:
	$(eval HOST=$(shell vagrant ssh-config | grep HostName | tr -d 'HostName' | tr -d ' '))
sanity-integration-test: sanity-copy-ip
	tmt run -a provision -h connect -g $(HOST) tests --filter 'tier: 1' --filter 'tag: sanity'
```

r0x0d:
> In the [`Makefile` in the project root](https://github.com/oamg/convert2rhel/blob/main/Makefile) I'd make something similar to how lint is setup
> 
> I expect it would be something like this, though I haven't tested it:
> 
> ```
> sanity-copy-ip:
> 	$(eval HOST=$(shell vagrant ssh-config | grep HostName | tr -d 'HostName' | tr -d ' '))
> sanity-integration-test: sanity-copy-ip
> 	tmt run -a provision -h connect -g $(HOST) tests --filter 'tier: 1' --filter 'tag: sanity'
> ```

That might get a bit tricky, as I think the `vagrant ssh-config` command only works if you're inside the folder that has a `Vagrantfile`. @danmyway I will follow-up with you on gchat about that

r0x0d (approved):
It looks good for a basic sanity check, only wondering if we should define the other methods of conversion here as well, like, the satellite one... Not sure if it's worth it to set a "baseline" with a clean run.

danmyway:
I don't see this being beneficial anymore.
If needed, we can call any of the basic conversion methods locally.
Those are well defined and do not introduce any extensive preparations (apart from updating the system, and complying with the convertability status)
We can possibly revisit this with the use of Packit's labels.
Closing the PR

PR #627, opened 2022-10-10
[RHELC-16] Do not disable RHSM repos before pkg backup

hosekadam:
There is a problem with backing up the packages on systems which use satellite server. During the conversion is system unregistered from previous subscription - in that step we lose access to the CentOS repositories. And then we need to backup packages, which could come from CentOS. 

So the step with backup and remove the package is moved before the unregistration. With that show up another problem - the backed up and then removed package remove `os-release` file too. The file is needed for subscribing the system, so it is restored from backup and then removed (if it won't be removed, the file would stay there after conversion as  `os-release.rpmorig`)

Jira Issue: [RHELC-16](https://issues.redhat.com/browse/RHELC-16)

Checklist

### DIFF
--- a/convert2rhel/backup.py
+++ b/convert2rhel/backup.py
@@ -328,8 +328,8 @@ class RestorableFile(object):
         try:
             os.remove(self.filepath)
             loggerinst.debug("File %s removed." % self.filepath)
-        except (OSError, IOError, TypeError):
-            loggerinst.debug("Could't remove restored file %s" % self.filepath)
+        except (OSError, IOError):
+            loggerinst.debug("Couldn't remove restored file %s" % self.filepath)
 
 
 class RestorablePackage(object):
--- a/convert2rhel/backup.py
+++ b/convert2rhel/backup.py
@@ -298,10 +298,13 @@ class RestorableFile(object):
         else:
             loggerinst.info("Can't find %s.", self.filepath)
 
-    def restore(self):
+    def restore(self, rollback=True):
         """Restore a previously backed up file"""
         backup_filepath = os.path.join(BACKUP_DIR, os.path.basename(self.filepath))
-        loggerinst.task("Rollback: Restoring %s from backup" % self.filepath)
+        if rollback:
+            loggerinst.task("Rollback: Restoring %s from backup" % self.filepath)
+        else:
+            loggerinst.info("Restoring %s from backup" % self.filepath)
 
         if not os.path.isfile(backup_filepath):
             loggerinst.info("%s hasn't been backed up." % self.filepath)
@@ -314,7 +317,19 @@ class RestorableFile(object):
             # IOError for py2 and OSError for py3
             loggerinst.warning("Error(%s): %s" % (err.errno, err.strerror))
             return
-        loggerinst.info("File %s restored." % self.filepath)
+
+        if rollback:
+            loggerinst.info("File %s restored." % self.filepath)
+        else:
+            loggerinst.debug("File %s restored." % self.filepath)
+
+    def remove(self):
+        """Remove restored file from original place, backup isn't removed"""
+        try:
+            os.remove(self.filepath)
+            loggerinst.debug("File %s removed." % self.filepath)
+        except (OSError, IOError, TypeError):
+            loggerinst.debug("Could't remove restored file %s" % self.filepath)
 
 
 class RestorablePackage(object):
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -222,6 +222,14 @@ def pre_ponr_conversion():
         loggerinst.task("Convert: Install RHEL certificates for RHSM")
         system_cert = cert.SystemCert()
         system_cert.install()
+
+    # remove non-RHEL packages containing repofiles or affecting variables in the repofiles
+    # This needs to be before attempt to unregister the system because after unregistration can be lost
+    # access to repositories needed for backuping removed packages
+    loggerinst.task("Convert: Remove packages containing .repo files")
+    pkghandler.remove_repofile_pkgs()
+
+    if not toolopts.tool_opts.no_rhsm:
         loggerinst.task("Convert: Subscription Manager - Subscribe system")
         subscription.subscribe_system()
         loggerinst.task("Convert: Get RHEL repository IDs")
@@ -230,10 +238,6 @@ def pre_ponr_conversion():
         subscription.check_needed_repos_availability(rhel_repoids)
         loggerinst.task("Convert: Subscription Manager - Disable all repositories")
         subscription.disable_repos()
-
-    # remove non-RHEL packages containing repofiles or affecting variables in the repofiles
-    loggerinst.task("Convert: Remove packages containing .repo files")
-    pkghandler.remove_repofile_pkgs()
 
     # we need to enable repos after removing repofile pkgs, otherwise we don't get backups
     # to restore from on a rollback
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -89,6 +89,9 @@ REGISTRATION_TIMEOUT = 180
 
 _SUBMGR_PKG_REMOVED_IN_CL_85 = "subscription-manager-initial-setup-addon"
 
+# Path to the releasever variable file for dnf.
+DNF_RELEASEVER_FILE = "/etc/yum/vars/releasever"
+
 
 class UnregisterError(Exception):
     """Raised with problems unregistering a system."""
@@ -455,7 +458,8 @@ class RegistrationCommand(object):
         register_server = system_bus.get_object("com.redhat.RHSM1", "/com/redhat/RHSM1/RegisterServer")
         loggerinst.debug("Starting a private DBus to talk to subscription-manager")
         address = register_server.Start(
-            i18n.SUBSCRIPTION_MANAGER_LOCALE, dbus_interface="com.redhat.RHSM1.RegisterServer"
+            i18n.SUBSCRIPTION_MANAGER_LOCALE,
+            dbus_interface="com.redhat.RHSM1.RegisterServer",
         )
 
         try:
@@ -525,7 +529,10 @@ class RegistrationCommand(object):
         finally:
             # Always shut down the private bus
             loggerinst.debug("Shutting down private DBus instance")
-            register_server.Stop(i18n.SUBSCRIPTION_MANAGER_LOCALE, dbus_interface="com.redhat.RHSM1.RegisterServer")
+            register_server.Stop(
+                i18n.SUBSCRIPTION_MANAGER_LOCALE,
+                dbus_interface="com.redhat.RHSM1.RegisterServer",
+            )
 
     def _set_connection_opts_in_config(self):
         """
@@ -776,10 +783,33 @@ def attach_subscription():
 def get_avail_subs():
     """Get list of all the subscriptions available to the user so they are
     accessible by index once the user chooses one.
+
+    .. note::
+        Get multiline string holding all the subscriptions available to the
+        logged-in user
     """
-    # Get multiline string holding all the subscriptions available to the
-    # logged-in user
+    # With the current changes introduced in PR#627, we needed to remove the
+    # non-RHEL packages that contain repofiles before we do any call to
+    # subscription-manager, and since we are doing that, this function starts to
+    # fail because it doesn't have the package that sets the releasever anymore.
+    # For some reason, subscription-manager needs to call libdnf internally to
+    # read the repositories on the system, and since libdnf needs to do
+    # substitutions on those repofiles, mainly the releasever and any other file
+    # placed in /etc/dnf/vars, this command call fails as it cannot replace the
+    # releasever variable anymore.
+    releaver_created = False
+    if system_info.version.major >= 8:
+        if not os.path.exists(DNF_RELEASEVER_FILE):
+            with open(DNF_RELEASEVER_FILE, "w") as handler:
+                handler.write(system_info.original_releasever)
+
+            releaver_created = True
+
     subs_raw, ret_code = utils.run_subprocess(["subscription-manager", "list", "--available"], print_output=False)
+
+    if releaver_created:
+        os.remove(DNF_RELEASEVER_FILE)
+
     if ret_code != 0:
         loggerinst.critical("Unable to get list of available subscriptions:\n%s" % subs_raw)
     return list(get_sub(subs_raw))
@@ -1005,7 +1035,12 @@ def _log_rhsm_download_directory_contents(directory, when_message):
     pkgs = ["<download directory does not exist>"]
     if os.path.isdir(SUBMGR_RPMS_DIR):
         pkgs = os.listdir(SUBMGR_RPMS_DIR)
-    loggerinst.debug("Contents of %s directory %s:\n%s", SUBMGR_RPMS_DIR, when_message, "\n".join(pkgs))
+    loggerinst.debug(
+        "Contents of %s directory %s:\n%s",
+        SUBMGR_RPMS_DIR,
+        when_message,
+        "\n".join(pkgs),
+    )
 
 
 def exit_on_failed_download(paths):
@@ -1036,7 +1071,11 @@ def lock_releasever_in_rhel_repositories():
             "Updating /etc/yum.repos.d/rehat.repo to point to RHEL %s instead of the default latest minor version."
             % system_info.releasever
         )
-        cmd = ["subscription-manager", "release", "--set=%s" % system_info.releasever]
+        cmd = [
+            "subscription-manager",
+            "release",
+            "--set=%s" % system_info.releasever,
+        ]
 
         output, ret_code = utils.run_subprocess(cmd, print_output=False)
         if ret_code != 0:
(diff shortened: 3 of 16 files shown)


convert2rhel/subscription.py
```diff
@@ -179,7 +180,9 @@ def register_system():
         registration_cmd = RegistrationCommand.from_tool_opts(tool_opts)
 
         try:
+            os_release_file.restore(rollback=False)  # RHELC-16
             registration_cmd()
+            os_release_file.remove()  # RHELC-16
```
SpyTec, line 185: Add a comment why we need to do this. If I were to look at this code after a while I wouldn't know why we're doing it without going through an issue

convert2rhel/main.py
```diff
@@ -201,6 +201,12 @@ def pre_ponr_conversion():
         loggerinst.task("Convert: Install RHEL certificates for RHSM")
         system_cert = cert.SystemCert()
         system_cert.install()
+
+    # remove non-RHEL packages containing repofiles or affecting variables in the repofiles
+    loggerinst.task("Convert: Remove packages containing .repo files")
+    pkghandler.remove_repofile_pkgs()
+
+    if not toolopts.tool_opts.no_rhsm:
```
SpyTec, line 209: Add a comment about needing to remove the repofile packages before disabling repos

convert2rhel/backup.py
```diff
@@ -121,10 +121,11 @@ def backup(self):
         else:
             loggerinst.info("Can't find %s.", self.filepath)
 
-    def restore(self):
+    def restore(self, rollback=True):
         """Restore a previously backed up file"""
         backup_filepath = os.path.join(BACKUP_DIR, os.path.basename(self.filepath))
-        loggerinst.task("Rollback: Restoring %s from backup" % self.filepath)
+        if rollback:
+            loggerinst.task("Rollback: Restoring %s from backup" % self.filepath)
```
SpyTec, line 128: Probably an info log would be helpful here
```suggestion
        if rollback:
            loggerinst.task("Rollback: Restoring %s from backup" % self.filepath)
        else:
            loggerinst.info("Restoring %s from backup" % self.filepath)
```

SpyTec (changes requested):

convert2rhel/backup.py
hosekadam: @SpyTec What do you think about changing the other logs [1](https://github.com/oamg/convert2rhel/pull/627/files#diff-8f4622020cde9a253bc82f69bf2b88910bcc5a84943db3c7d3a435becedfd4d5R145), [2](https://github.com/oamg/convert2rhel/pull/627/files#diff-8f4622020cde9a253bc82f69bf2b88910bcc5a84943db3c7d3a435becedfd4d5R151) and [3](https://github.com/oamg/convert2rhel/pull/627/files#diff-8f4622020cde9a253bc82f69bf2b88910bcc5a84943db3c7d3a435becedfd4d5R153) to `info` too? My idea of using the `debug` was to not complicate the output for user, but now I'm thinking if it could be somehow useful if the backup would be there.

SpyTec: @hosekadam The links don't work for me but I assume you mean the other sections with restore complete etc. For those kinds of logs I would say this:
- if it is important to message the user about a completion, it should be `info` for sure. That's why having it when rolling back is good
- if the message is part of any task and isn't critical, the task completion should be a debug as it otherwise is unnecessary clutter

The reason I wanted to change this is that it relates to a task's start. As customers or problem solvers (like us or our support engineers) would like to know what happened to lead to an issue. If we back something up and something fails, without logs we'd make the assumption it was the previous task. Or be let in the dark of what happened without digging into the code itself. As for using `task` vs `info`, when rolling back it is a crucial step to restore backups, so a task make sense there. Though if it's within the intended workflow of converting a system it would be confusing and thus an `info` log is good enough

hosekadam: @SpyTec Ok, so we can leave the rest as `debug`. And thanks for the clarification :+1:

convert2rhel/backup.py
```diff
@@ -314,7 +317,19 @@ def restore(self):
...
+        else:
+            loggerinst.debug("File %s restored." % self.filepath)
+
+    def remove(self):
+        """Remove restored file from original place, backup isn't removed"""
+        try:
+            os.remove(self.filepath)
+            loggerinst.debug("File %s removed." % self.filepath)
+        except (OSError, IOError, TypeError):
+            loggerinst.debug("Could't remove restored file %s" % self.filepath)
```
r0x0d, line 332: Typo 
```suggestion
            loggerinst.debug("Couldn't remove restored file %s" % self.filepath)
```

convert2rhel/backup.py
```diff
@@ -314,7 +317,19 @@ def restore(self):
...
+            loggerinst.info("File %s restored." % self.filepath)
+        else:
+            loggerinst.debug("File %s restored." % self.filepath)
+
+    def remove(self):
+        """Remove restored file from original place, backup isn't removed"""
+        try:
+            os.remove(self.filepath)
+            loggerinst.debug("File %s removed." % self.filepath)
+        except (OSError, IOError, TypeError):
```
r0x0d, line 331: I'm wondering why we need to have the `TypeError` here. Could you elaborate on why it is needed to catch this exception too? 

If it is something way off that you have experienced in your tests, then it might be good to add a comment explaining the reason to it.

hosekadam: @r0x0d The `TypeError` exception would be raised, if the `filepath` would somehow contain `None`. Which I think couldn't happen, but if yes, I've choose that just saying `Couldn't remove restored file` is better than crashing on this.

r0x0d: Right. Makes sense from this point of view, but since we construct the `RestorableFile` object from hard-coded strings in our code, then I don't think we need the `TypeError` catch here, as we will always pass an `filepath` to the constructor of the class.

Catching the `TypeError` in this function, for me, seems unnecessary, as it would be better if we catch that type of mistaken in the class constructor, for example: 

```python
class RestorableFile(object):
    def __init__(self, filepath):
        if not isinstance(filepath, str):
            raise TypeError("filepath must be an instance of str.")
            
        self.filepath = filepath
```

This way we could prevent working with an empty `filepath` in other methods inside the class.

But for now, I would say this is not strictly necessary, as we don't have any case in our code base that could pass an empty `filepath` to this class constructor.

hosekadam: Ok, so I'll remove that.

SpyTec (approved):

hosekadam:
Heh, I used probably a wrong command and I see there is a bit mess...

hosekadam:
There are few things left that needs to be done to get the integration test running properly:
1. Merge this [merge request](https://gitlab.cee.redhat.com/oamg/convert2rhel/convert2rhel-secrets/-/merge_requests/9)
2. In the satellite repo for CentOS7 is the old version of kernel, so we need to wait to sync the repositories

Otherwise the test works, it's just changed the test for offline conversion using satellite to use the new satellite dogfood server

r0x0d:
> There are few things left that needs to be done to get the integration test running properly:
> 
>     1. Merge this [merge request](https://gitlab.cee.redhat.com/oamg/convert2rhel/convert2rhel-secrets/-/merge_requests/9)
> 
>     2. In the satellite repo for CentOS7 is the old version of kernel, so we need to wait to sync the repositories
> 
> 
> Otherwise the test works just for the CentOS 7 so I think I should limit that just for this system and not to run it with others. The second variant is to ask satellite guys to provide servers that we need for another systems. What do you think @SpyTec or @r0x0d?

@hosekadam ping @kokesak to merge your PR. I don't have write access to that repository. 

I don't know how the satellite guys would react if we ask them for more servers for the other distros we use… Furthermore, I don't know if @SpyTec or @bocekm knows how that works, if so, can any of you help us here? 

As far as limiting the test to CentOS 7, I think it's a good idea if your test is only related to C

tests/integration/conftest.py
```diff
@@ -26,7 +26,7 @@
 logging.basicConfig(level="DEBUG" if env.str("DEBUG") else "INFO", stream=sys.stderr)
 logger = logging.getLogger(__name__)
 
-SATELLITE_URL = "satellite.sat.engineering.redhat.com"
```
kokesak, line 29: We are actually trying to use the new satellite URL (https://issues.redhat.com/browse/RHELC-732). The dogfood is going to be removed.

kokesak:
The test was running for all distros so I would keep that.

IIUC, it's the offline-conversion test in which we currently use the env variables to omit bug this PR fixes. However, we need centos and oracle content in the satellite server, currently, we have just centos7 right?. Is this fix going to be in the next release?

I would rather have it tested on all distros before merging.

tests/integration/conftest.py
hosekadam: Yeah, I see. But if you register the system using subscription-manager to get the CentOS 7 satellite server, there are still urls to the dogfood in `/etc/yum.repos.d/redhat.repo`. Maybe adding both of them, so once it would be finally changed, nothing breaks and just be needed to remove the dogfood url from `/etc/dnsmasq.conf`?

hosekadam:
@kokesak yes, I see [there](https://gitlab.cee.redhat.com/satellite/satellite-ansible-setup/-/merge_requests/12/diffs?commit_id=dca405d5de75b63a753a1b871045b20ff012208d) that there is available satellite just for CentOS 7 (and Centos 8 Stream, but it's useless for us I think). 

And this PR should be in the next release (from Jira).

convert2rhel/subscription.py
```diff
@@ -179,7 +180,15 @@ def register_system():
...
         try:
+            # The file /etc/os-release is needed for subscribing the system and is being removed with
+            # <system-name>-release package in one of the steps before
+            # RHELC-16
+            os_release_file.restore(rollback=False)
             registration_cmd()
+            # Need to remove the file, if it would stay there would be leftover /etc/os-release.rpmorig
+            # after conversion
+            # RHELC-16
+            os_release_file.remove()
```
abadger, line 191: This part feels a little unclean but we can fix it in the next release.  I've written up what's unclean about it and how we can fix it later here: https://issues.redhat.com/browse/RHELC-809

convert2rhel/subscription.py
```diff
@@ -179,7 +180,15 @@ def register_system():
         registration_cmd = RegistrationCommand.from_tool_opts(tool_opts)
 
         try:
+            # The file /etc/os-release is needed for subscribing the system and is being removed with
+            # <system-name>-release package in one of the steps before
+            # RHELC-16
```
abadger, line 185:
```suggestion
            # The file /etc/os-release is needed for subscribing the system and is being removed with
            # <system-name>-release package in one of the previous steps:
            # Fixes RHELC-16
            #
            # This code should be cleaned up in the future.  See  https://issues.redhat.com/browse/RHELC-809
            # for what should be done.
```

kokesak:
On centos7 it seems that the satellite repos do not have the kernel we update to during the update system phase:
```
22:17:29         out: CRITICAL - The version of the loaded kernel is different from the latest version in the enabled system repositories.
22:17:29         out:  Latest kernel version available in Satellite_Engineering_CentOS_7_Updates_x86_64: 3.10.0-1160.76.1.el7
22:17:29         out:  Loaded kernel version: 3.10.0-1160.80.1.el7
22:17:29         out: 
22:17:29         out: To proceed with the conversion, update the kernel version by executing the following step:
22:17:29         out: 
22:17:29         out: 1. yum install kernel-3.10.0-1160.76.1.el7 -y
22:17:29         out: 2. reboot
22:17:29         out: [11/18/2022 17:17:29] DEBUG - Traceback (most recent call last):
22:17:29         out:   File "/usr/lib/python2.7/site-packages/convert2rhel/main.py", line 101, in main
22:17:29         out:     checks.perform_system_checks()
22:17:29         out:   File "/usr/lib/python2.7/site-packages/convert2rhel/checks.py", line 127, in perform_system_checks
22:17:29         out:     is_loaded_kernel_latest()
22:17:29         out:   File "/usr/lib/python2.7/site-packages/convert2rhel/checks.py", line 780, in is_loaded_kernel_latest
22:17:29         out:     "2. reboot" % (repos_message, repoid, latest_kernel, loaded_kernel, package_to_check, latest_kernel)
22:17:29         out:   File "/usr/lib/python2.7/site-packages/convert2rhel/logger.py", line 195, in _critical
22:17:29         out:     sys.exit(msg)
22:17:29         out: SystemExit: The version of the loaded kernel is different from the latest version in the enabled system repositories.
22:17:29         out:  Latest kernel version available in Satellite_Engineering_CentOS_7_Updates_x86_64: 3.10.0-1160.76.1.el7
22:17:29         out:  Loaded kernel version: 3.10.0-1160.80.1.el7
22:17:29         out: 
22:17:29         out: To proceed with the conversion, update the kernel version by executing the following step:
22:17:29         out: 
22:17:29         out: 1. yum install kernel-3.10.0-1160.76.1.el7 -y
22:17:29         out: 2. reboot
```

hosekadam:
@kokesak Yes, I've mentioned this problem in the [comment](https://github.com/oamg/convert2rhel/pull/627#issuecomment-1314111669) :D Do you know somebody who we can contact to help us with that? I thought, they (who prepared the satellite server) have some automated way to stay updated with the original repositories. 

When I was testing that, I used the older version of vagrant box and removed the update step from ansible setup.

kokesak:
Hmm, seem that passing the the env var `CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK` did not work. I think I will need to make some other workaround, though it's weird that this envar is ommited

plans/tier1.fmf
```diff
@@ -228,9 +228,20 @@
...
+      distro != centos-7
   discover+:
     test: checks-after-conversion
   prepare+:
+  - name: workaround_kernel
+    how: shell
+    script: "yum install kernel-3.10.0-1160.76.1.el7 -y && yum remove kernel-3.10.0-1160.80.1.el7 -y"
+  - name: reboot machine
+    how: ansible
+    playbook: tests/ansible_collections/roles/reboot/main.yml
```
danmyway, line 244: Do we need to reboot?
Wouldn't `grubby --set-default-kernel` be sufficient?
I would probably add some comment about the kernel workaround being temporary.
Or a `#TODO` to remove it, when the repository gets synced over at Satellite.
But I guess for the sake of saving some time, we can also keep it as is.

tests/integration/tier1/convert-offline-systems/main.fmf
```diff
@@ -1,4 +1,6 @@
-summary: Convert systems that have no access to the Internet
+summary: |
+    Convert systems that have no access to the Internet. This test require having the repositories of original
+    distro available in the satellite server. As of now we only have CentOS7 repositories available.
```
danmyway, line 3: Summary should be shorter, 50 characters if possible.
Description can be longer.
But we can fix this whenever.

danmyway (approved):
Just a couple of minor changes, we can sort this out with another PR.

plans/tier1.fmf
kokesak: The reboot would be needed I believe as the c2r check is done against the running kernel which changes only after reboot.

tests/integration/tier1/convert-offline-systems/main.fmf
kokesak: I created a Jira issue for this so we can tackle all the missing stuff in future fix and not block current PR.
Link to the Jira: https://issues.redhat.com/browse/RHELC-824

convert2rhel/backup.py
```diff
@@ -314,7 +317,19 @@ def restore(self):
             # IOError for py2 and OSError for py3
             loggerinst.warning("Error(%s): %s" % (err.errno, err.strerror))
             return
-        loggerinst.info("File %s restored." % self.filepath)
+
+        if rollback:
+            loggerinst.info("File %s restored." % self.filepath)
+        else:
+            loggerinst.debug("File %s restored." % self.filepath)
```
r0x0d, line 324: This message here is not making too much sense at the moment, I would say... 

If it take as an example the same condition above, we are issuing an `loggerinst.task()` when we have a rollback, and a `loggerinst.info()` when we are not rollbacking. 

I would suggest following the same pattern as above, but this is not a blocker. 

```suggestion
        if rollback:
            loggerinst.task("Rollback: File %s restored." % self.filepath)
        else:
            loggerinst.info("File %s restored." % self.filepath)
```

r0x0d (approved):

convert2rhel/backup.py
SpyTec: I'd disagree. If it's a rollback we'd want to have it as a step, so it would be 
```
[Task] Rollback: Restoring %s from backup
[Info] Rollback succeeded
```

Since at that step we care about the rollback and not about a generic backup we should make a not of it. If it's not a rollback it should be part of the logs as normal, as it's part of the overall task it's part of
```
[Info] Rollback: Restoring %s from backup
[Debug] Rollback succeeded
```

I wouldn't change it to two tasks as that separates it and creates a division that can be understood in the logs as being part of two functions

bocekm: The original non-task way is IMHO correct. We want just the first line to have the TASK prefix, and the "confirmation" message to be just of an info level below it. Example of what we have now:
```
10:34:02                 out: [11/29/2022 05:34:02] TASK - [Rollback: Restoring /etc/system-release from backup] **************
10:34:02                 out: File /etc/system-release restored.
```

With your proposal it would be:
```
10:34:02                 out: [11/29/2022 05:34:02] TASK - [Rollback: Restoring /etc/system-release from backup] **************
10:34:02                 out: [11/29/2022 05:34:02] TASK - [Rollback: File /etc/system-release restored.] **************
```

r0x0d: Makes sense to me. Thanks for the explanation

PR #637, opened 2022-10-19
[RHELC-568] docs: improve failed conversion message

SpyTec:
As brought up in RHELC-568, the message when a conversion was unsuccessful wasn't ideal and needed to be changed. This PR fixes the message to hint at what the system is left in, that convert2rhel can't fix it by itself, and advise the user to restore a backup

Jira Issue: [RHELC-568](https://issues.redhat.com/browse/RHELC-568)

Checklist


convert2rhel/main.py
```diff
@@ -151,7 +151,13 @@ def main():
             # system rollback without user intervention. If a proper rollback
             # solution is necessary it will need to be future implemented here
             # or with the use of other backup tools.
-            loggerinst.warning("Conversion process interrupted and manual user intervention will be necessary.")
+            loggerinst.warning(
+                "The conversion process failed.\n\n"
+                "The system is left in an undetermined state that Convert2RHEL cannot fix. The system might not be"
+                " fully converted, and is reporting itself as a Red Hat Enterprise Linux machine.\n\n"
+                "It is strongly recommended to store the Convert2RHEL logs for the later investigation, and restore"
+                " the system from a backup."
```
bocekm, line 159:
```suggestion
                " fully converted.\n\n"
                "It is strongly recommended to store the Convert2RHEL logs for later investigation, and restore"
                " the system from a backup."
```
The system might not be easy to recognize as RHEL. In the early stage we remove `oraclelinux-release`, `centos-release`, or similar, and we install the `redhat-release` package only later during the package replacement phase. If the conversion fails somewhere after the PONR and before the `redhat-release` installation, there won't be even the `/etc/system-release` file on the system.

bocekm (changes requested):

convert2rhel/main.py
SpyTec: So, the reason I want that sentence in there is due to some people re-running Convert2RHEL and it will say it's an unsupported system. Or those that simply check if it's a false-negative message and that the system was actually converted, so they'll check /etc/system-release. We've seen this before where people assume they're converted and then get strange issues after trying leapp

r0x0d: I would say that Eric suggestion makes sense to me. I prefer giving a hint, even though that hint might not be 100% true, than leaving the user with nothing.

r0x0d (approved):

PR #639, opened 2022-10-25
[RHELC-729] Do not show output from systemctl show during systeminfo gathering.

abadger:
We use systemctl to tell us whether dbus is running on the system. When we do so, we don't want to directly emit the output of the systemctl command to the log as it is cryptic for the end user.

Fixes https://issues.redhat.com/browse/RHELC-729

Jira Issue: [RHELC-729](https://issues.redhat.com/browse/RHELC-729)

Checklist


r0x0d (approved):
Approving, since the change is just adding a new parameter to the `run_subprocess` call and doesn't interfere with anything else.

PR #648, opened 2022-11-02
[RHELC-784] Allow but warn about deprecated env var

abadger:
We've renamed CONVERT2RHEL_UNSUPPORTED_VERSION to CONVERT2RHEL_ALLOW_OLDER_VERSION.  Allow the old env var name to be used but emit a deprecation warning in the log.

Add unittests to show the deprecation message is emitted when we use the old env var rather than the new one to allow the conversion to continue.

Followup to changes introduced for: #566
Jira Issue: [RHELC-784](https://issues.redhat.com/browse/RHELC-784)

Checklist

### DIFF
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -200,7 +200,6 @@ class TestCheckConvert2rhelLatest(object):
         global_system_info.has_internet_access = False
         checks.check_convert2rhel_latest()
 
-        convert2rhel_latest_version_test
         log_msg = "Skipping the check because no internet connection has been detected."
         assert log_msg in caplog.text
 
--- a/convert2rhel/unit_tests/checks_test.py
+++ b/convert2rhel/unit_tests/checks_test.py
@@ -190,179 +190,195 @@ def test_pre_ponr_checks(monkeypatch):
     create_transaction_handler_mock.assert_called_once()
 
 
-@pytest.mark.parametrize(
-    ("convert2rhel_latest_version_test",),
-    ([{"local_version": "0.20", "package_version": "convert2rhel-0:0.22-1.el7.noarch", "pmajor": "7"}],),
-    indirect=True,
-)
-def test_convert2rhel_latest_offline(caplog, convert2rhel_latest_version_test, global_system_info):
-    global_system_info.has_internet_access = False
-    checks.check_convert2rhel_latest()
-
-    convert2rhel_latest_version_test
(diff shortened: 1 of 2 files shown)


r0x0d:
Re-running one integration test as it seems to be failing because of infrastructure problems.

convert2rhel/unit_tests/checks_test.py
```diff
@@ -190,179 +190,195 @@ def test_pre_ponr_checks(monkeypatch):
...
-    ("convert2rhel_latest_version_test",),
-    (
-        [{"local_version": "0.20", "package_version": "convert2rhel-0:0.22-1.el7.noarch", "pmajor": "6"}],
-        [{"local_version": "0.18", "package_version": "convert2rhel-0:1.10-1.el7.noarch", "pmajor": "6"}],
-    ),
-    indirect=True,
-)
-def test_convert2rhel_latest_out_of_date_el6(caplog, convert2rhel_latest_version_test):
-    checks.check_convert2rhel_latest()
+        convert2rhel_latest_version_test
```
r0x0d, line 203: It seems to me that we are not using this line for anything in the test... 

@abadger What would you like to do about this? Maybe create a Jira issue to take care of this point, or adding another assertion to this test to use this data?

abadger: You are right.  The line should be removed.  I assume you are wondering whether to remove it now  or open a ticket  because integration tests take so long to run?

r0x0d: Yeah, well... If we remove the line, then there is no need for the parametrization in that test, but if we use that for asserting, then we can leave if there, just adjust that line to have an `assert ...`.

Either way, I'm fine with any approach.

r0x0d:
@abadger I think you need to link this issue as well: [RHELC-784](https://issues.redhat.com/browse/RHELC-784)

SpyTec:
> @abadger I think you need to link this issue as well: [RHELC-784](https://issues.redhat.com/browse/RHELC-784)

You're right, I changed it to be that one instead

SpyTec (approved):
Seems good to me, we can merge since the integration tests failing seem unrelated

r0x0d:
The test `/tests/integration/tier0/inhibit-if-kmods-is-not-supported` is fixed in upstream, a rebase should fix this, but I would say that's not necessary, since every PR is failing with that. 

All errors seem to not be related to the PR. Merging this right now.

PR #642, opened 2022-10-26
[RHELC-744] Install the redhat-uep.pem certificate if it is needed.

abadger:
We need this certificate for the convert2rhel repo if the user installed using the convert2rhel.repo file:

https://ftp.redhat.com/redhat/convert2rhel/8/convert2rhel.repo

This is a workaround for https://issues.redhat.com/browse/RHELC-744 When we make uninstalling and reinstalling the RHEL subscription-manager packages a single step process, we will no longer need this as there will always be a redhat-uep.pem available from an rpm package.

Jira Issue: [RHELC-744](https://issues.redhat.com/browse/RHELC-744)

TODO:
- [x] unit tests
- [x] integration test

Checklist
- [x] PR meets acceptance criteria specified in the Jira issue
- [x] PR has been tested manually in a VM (either author or reviewer)
- [ ] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [ ] Code and tests are documented properly
- [x] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending`

PR #650, opened 2022-11-03
[RHELC-744, RHELC-762] Backup all CentOS subscription-manager packages

bocekm:
Not removing and backing up the python3-syspurpose and python3-cloud-what packages, which are also built out of the subscription-manager SRPM, was causing a rollback failure on CentOS Linux 8.5 when an older version of these two packages was installed.

The rollback failure caused the following issues during the subsequent second run of convert2rhel:
 1. when the rollback happened before removing the centos-linux-release package, the removed subscription-manager-rhsm-certificates was not re-installed back during the rollback leading to a missing /etc/rhsm/ca/redhat-uep.pem file which is necessary for accessing the official convert2rhel repo on CDN
 2. when the rollback happened before removing the centos-linux-release package, this package was not re-installed back during the rollback and that lead to the $releasever variable being undefined, ultimately causing a traceback when using the DNF python API

Replaces https://github.com/oamg/convert2rhel/pull/642.

Jira Issue: [RHELC-744](https://issues.redhat.com/browse/RHELC-744), [RHELC-762](https://issues.redhat.com/browse/RHELC-762)

Checklist
- [x] `[RHELC-744]` and `[RHELC-762]` are part of the PR title

### DIFF
--- a/tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
+++ b/tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
@@ -55,10 +55,8 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
     """
     assert shell("modprobe -r -v bonding").returncode == 0
 
-    if "oracle-7" in system_version:
+    if "oracle-7" in system_version or "centos-7" in system_version:
         prompt_amount = 3
-    elif "centos-7" in system_version:
-        prompt_amount = 4
     elif "oracle-8" in system_version:
         prompt_amount = 2
     elif "centos-8" in system_version:
--- a/tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
+++ b/tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
@@ -55,12 +55,14 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
     """
     assert shell("modprobe -r -v bonding").returncode == 0
 
-    if "oracle-7" in system_version or "centos-7" in system_version:
+    if "oracle-7" in system_version:
         prompt_amount = 3
+    elif "centos-7" in system_version:
+        prompt_amount = 4
     elif "oracle-8" in system_version:
         prompt_amount = 2
     elif "centos-8" in system_version:
-        prompt_amount = 3
+        prompt_amount = 4
     # If custom module is not loaded the conversion is not inhibited.
     with convert2rhel(
         "--no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug".format(
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -41,7 +41,7 @@ jobs:
     epel-7-x86_64:
       distros: [centos-7, oraclelinux-7]
     epel-8-x86_64:
-      distros: [centos-8.4, centos-8, oraclelinux-8.4, oraclelinux-8.6]
+      distros: [centos-8.4, centos-8-latest, oraclelinux-8.4, oraclelinux-8.6]
   use_internal_tf: True
   trigger: pull_request
 - <<: *tests
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -592,6 +592,10 @@ def remove_original_subscription_manager():
     loggerinst.info("Removing installed subscription-manager/katello-ca-consumer packages.")
     # python3-subscription-manager-rhsm, dnf-plugin-subscription-manager, subscription-manager-rhsm-certificates, etc.
     submgr_pkgs = pkghandler.get_installed_pkg_objects("*subscription-manager*")
+    # The python-syspurpose, python3-syspurpose, and python3-cloud-what packages are also built out of
+    # the subscription-manager SRPM.
+    submgr_pkgs += pkghandler.get_installed_pkg_objects("python*-syspurpose")
+    submgr_pkgs += pkghandler.get_installed_pkg_objects("python3-cloud-what")
     # Satellite-server related package
     submgr_pkgs += pkghandler.get_installed_pkg_objects("katello-ca-consumer*")
     if not submgr_pkgs:
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -919,6 +919,7 @@ def download_rhsm_pkgs():
         loggerinst.info("Skipping due to the use of --keep-rhsm.")
         return
     utils.mkdir_p(_RHSM_TMP_DIR)
+
     pkgs_to_download = [
         "subscription-manager",
         "subscription-manager-rhsm-certificates",
@@ -948,9 +949,20 @@ def download_rhsm_pkgs():
 
 
 def _download_rhsm_pkgs(pkgs_to_download, repo_path, repo_content):
+    _log_rhsm_download_directory_contents(SUBMGR_RPMS_DIR, "before RHEL rhsm packages download")
+
     utils.store_content_to_file(filename=repo_path, content=repo_content)
     paths = utils.download_pkgs(pkgs_to_download, dest=SUBMGR_RPMS_DIR, reposdir=_RHSM_TMP_DIR)
+
+    _log_rhsm_download_directory_contents(SUBMGR_RPMS_DIR, "after RHEL rhsm packages download")
     exit_on_failed_download(paths)
+
+
+def _log_rhsm_download_directory_contents(directory, when_message):
+    pkgs = ["<download directory does not exist>"]
+    if os.path.isdir(SUBMGR_RPMS_DIR):
+        pkgs = os.listdir(SUBMGR_RPMS_DIR)
+    loggerinst.debug("Contents of %s directory %s:\n%s", SUBMGR_RPMS_DIR, when_message, "\n".join(pkgs))
 
 
 def exit_on_failed_download(paths):
--- a/convert2rhel/unit_tests/subscription_test.py
+++ b/convert2rhel/unit_tests/subscription_test.py
@@ -308,54 +308,6 @@ class TestSubscription(unittest.TestCase):
     def test_install_rhel_subscription_manager_unable_to_install(self):
         self.assertRaises(SystemExit, subscription.install_rhel_subscription_manager)
 
-    class StoreContentMocked(unit_tests.MockFunction):
-        def __init__(self):
-            self.called = 0
-            self.filename = None
-            self.content = None
-
-        def __call__(self, filename, content):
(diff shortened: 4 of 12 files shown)


abadger:
Some integration test failures which are suspicious since they don't seem transient.  Re-kicked those tests to see if they are consistent.

The change looks correct.  I think in the future we should look at:

1) The disconnect between downloading sub-man related packages, removing sub-man related packages, and re-installing sub-man related packages during rollback.  The set of packages we deal with for these seems tightly coupled but the code that deals with them doesn't do a good job of relating them.  Fixing these to more tightly couple the package set will help to prevent problems like this.

2) Removing some of the filtering code that was introduced here: https://github.com/oamg/convert2rhel/pull/393 When bocekm and I talked about this,  we thought that python-syspurpose and python*-cloud-what are built as part of subscription-manager so it should be safe to remove them and reinstall from the backup rpms.  json-c probably needs to be filtered as long as we do this in two steps (remove the package, then install the backup rpm) because it might be a dependency of other packages and the removal step will break that.

Neither of these need to be fixed in this PR.  We have two other issues where we can probably address them:

* Refactoring backups of rpm packages to use the BackupController instead.  We can implement an RpmPackageSet RestorableChange which should let us couple the install/reinstall more tightly

* Moving to installing the thel subscription-manager packages using dnf. This might help us do the subscription-manager replacement as a single transaction instead of separate remove and then install (which has the potential to break if the would temporarily removal breaks dependencies)

bocekm:
> 1. The disconnect between downloading sub-man related packages, removing sub-man related packages, and re-installing sub-man related packages during rollback. The set of packages we deal with for these seems tightly coupled but the code that deals with them doesn't do a good job of relating them. Fixing these to more tightly couple the package set will help to prevent problems like this.

Agreed.

> 2. Removing some of the filtering code that was introduced here: https://github.com/oamg/convert2rhel/pull/393 When bocekm and I talked about this, we thought that python-syspurpose and python*-cloud-what are built as part of subscription-manager so it should be safe to remove them and reinstall from the backup rpms. json-c probably needs to be filtered as long as we do this in two steps (remove the package, then install the backup rpm) because it might be a dependency of other packages and the removal step will break that.

Currently we don't remove `json-c` together with the subscription-manager packages and because of that we don't back it up. If the package is installed prior to the conversion, we need to filter it out as we do now (or in other words - not track the `json-c` as a package installed by convert2rhel). Not doing so would lead to a removal of `json-c` upon rollback and thus a missing dependency for many packages (the output of `dnf rq --whatdepends json-c` is huge btw). 

What is related and we could be doing around `json-c` is that:

1. With `json-c` installed prior to the conversion it may happen that when we get to installing subscription-manager from the UBI, the `json-c` package is updated.

  <details>
    <summary>Example from CentOS Linux 8.5</summary>

```
[11/03/2022 19:51:15] DEBUG - Calling command 'yum install -y --setopt=module_platform_id=platform:el8 /usr/share/convert2rhel/subscription-manager/subscription-manager-1.28.29-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/subscription-manager-rhsm-certificates-1.28.29-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/python3-subscription-manager-rhsm-1.28.29-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/dnf-plugin-subscription-manager-1.28.29-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/python3-syspurpose-1.28.29-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/python3-cloud-what-1.28.29-3.el8.x86_64.rpm /usr/share/convert2rhel/subscription-manager/json-c-0.13.1-3.el8.x86_64.rpm'
Last metadata expiration check: 0:00:46 ago on Thu Nov  3 19:50:29 2022.
Dependencies resolved.
================================================================================
 Package                                Arch   Version       Repository    Size
================================================================================
Installing:
 dnf-plugin-subscription-manager        x86_64 1.28.29-3.el8 @commandline 311 k
 python3-cloud-what                     x86_64 1.28.29-3.el8 @commandline 310 k
 python3-subscription-manager-rhsm      x86_64 1.28.29-3.el8 @commandline 388 k
 python3-syspurpose                     x86_64 1.28.29-3.el8 @commandline 327 k
 subscription-manager                   x86_64 1.28.29-3.el8 @commandline 1.2 M
 subscription-manager-rhsm-certificates x86_64 1.28.29-3.el8 @commandline 284 k
Upgrading:
 json-c                                 x86_64 0.13.1-3.el8  @commandline  41 k
```
  </details>

2. We currently don't track the updated package at all, causing the rollback to be incomplete - we leave the system with the package having a newer version.

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
```diff
@@ -55,12 +55,14 @@ def test_do_not_inhibit_if_module_is_not_loaded(shell, convert2rhel):
     """
     assert shell("modprobe -r -v bonding").returncode == 0
 
-    if "oracle-7" in system_version or "centos-7" in system_version:
+    if "oracle-7" in system_version:
         prompt_amount = 3
+    elif "centos-7" in system_version:
+        prompt_amount = 4
```
bocekm, line 61: @danmyway, @kokesak, I needed to bump up the number due to an additional y/n question related to the python-syspurpose package installed under the other tier0 test:
https://github.com/oamg/convert2rhel/blob/v1.0/tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py#L9

It's quite weird. Shouldn't the previous test "clean" the system to not influence this other test?

bocekm:
@danmyway, @kokesak, the CentOS Linux 8.5 system has the `python3-syspurpose` installed: http://artifacts.osci.redhat.com/testing-farm/bab7cec0-40ec-42f1-948a-767436fd9646/work-remove_all_submgr_pkgsVfDPlP/log.txt.

It's weird. How did it get there? Even though it's tier1 test, I have a suspicion that it's a leftover from this tier0 test machine: https://github.com/oamg/convert2rhel/blob/v1.0/tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py#L10.
Is it possible?

tests/integration/tier0/inhibit-if-kmods-is-not-supported/test_kmods_not_supported.py
danmyway: In the last run we are getting past the PONR, so I guess no need to bump up the `prompt_amount` 
```
20:45:34                 out: Successfully validated the yum transaction set.
20:45:34                 out: WARNING - ********************************************************
20:45:34                 out: WARNING - The tool allows rollback of any action until this point.
20:45:34                 out: WARNING - By continuing all further changes on the system will need to be reverted manually by the user, if necessary.
20:45:34                 out: WARNING - ********************************************************
20:45:34                 out: 
20:45:35                 out: Continue with the system conversion? [y/n]: y
```

danmyway:
> @danmyway, @kokesak, the CentOS Linux 8.5 system has the `python3-syspurpose` installed: http://artifacts.osci.redhat.com/testing-farm/bab7cec0-40ec-42f1-948a-767436fd9646/work-remove_all_submgr_pkgsVfDPlP/log.txt.
> 
> It's weird. How did it get there? Even though it's tier1 test, I have a suspicion that it's a leftover from this tier0 test machine: https://github.com/oamg/convert2rhel/blob/v1.0/tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py#L10. Is it possible?

@bocekm I am not sure yet, how it got there.
I have tried to run the test just by itself to ensure a "clean" machine, and it is there as well
http://artifacts.osci.redhat.com/testing-farm/94190db7-71a9-45b0-b160-ad9fb73bed2f

danmyway:
> > @danmyway, @kokesak, the CentOS Linux 8.5 system has the `python3-syspurpose` installed: http://artifacts.osci.redhat.com/testing-farm/bab7cec0-40ec-42f1-948a-767436fd9646/work-remove_all_submgr_pkgsVfDPlP/log.txt.
> > It's weird. How did it get there? Even though it's tier1 test, I have a suspicion that it's a leftover from this tier0 test machine: https://github.com/oamg/convert2rhel/blob/v1.0/tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py#L10. Is it possible?
> 
> @bocekm I am not sure yet, how it got there. I have tried to run the test just by itself to ensure a "clean" machine, and it is there as well http://artifacts.osci.redhat.com/testing-farm/94190db7-71a9-45b0-b160-ad9fb73bed2f

@bocekm I think I have the answer.
It was always there, but what we did so far, was to upgrade said package.
See PR #642 http://artifacts.osci.redhat.com/testing-farm/9a5d1b44-8905-44a4-91d3-b3f4402cacd8/work-inhibit_if_kmods_is_not_supportedMyXHhD/plans/tier0/inhibit_if_kmods_is_not_supported/execute/data/tests/integration/tier0/inhibit-if-kmods-is-not-supported/output.txt
As opposed to this PR, where we remove it and then install the new version.

kokesak:
> > > @danmyway, @kokesak, the CentOS Linux 8.5 system has the `python3-syspurpose` installed: http://artifacts.osci.redhat.com/testing-farm/bab7cec0-40ec-42f1-948a-767436fd9646/work-remove_all_submgr_pkgsVfDPlP/log.txt.
> > > It's weird. How did it get there? Even though it's tier1 test, I have a suspicion that it's a leftover from this tier0 test machine: https://github.com/oamg/convert2rhel/blob/v1.0/tests/integration/tier0/test-check-rollback-handling/test_check_rollback_handling.py#L10. Is it possible?
> > 
> > 
> > @bocekm I am not sure yet, how it got there. I have tried to run the test just by itself to ensure a "clean" machine, and it is there as well http://artifacts.osci.redhat.com/testing-farm/94190db7-71a9-45b0-b160-ad9fb73bed2f
> 
> @bocekm I think I have the answer. It was always there, but what we did so far, was to upgrade said package. See PR #642 http://artifacts.osci.redhat.com/testing-farm/9a5d1b44-8905-44a4-91d3-b3f4402cacd8/work-inhibit_if_kmods_is_not_supportedMyXHhD/plans/tier0/inhibit_if_kmods_is_not_supported/execute/data/tests/integration/tier0/inhibit-if-kmods-is-not-supported/output.txt As opposed to this PR, where we remove it and then install the new version.

It's as a @danmyway says, the package is present on the running VM from the begining:
```
[root@ip-10-29-162-89 ~]# rpm -qa | grep python3-sys
python3-syspurpose-1.27.16-1.el8.x86_64
```
Generally, we should create our own Cent85 image instead of using the "latest" which is just 8.3 and mirror of 1minutetip image.

danmyway (approved):
The failing tests should be disabled for Oracle Linux 8.6 now, merging.

PR #612, opened 2022-09-20
[RHELC-664] Collect breadcrumbs through RHSM custom facts

r0x0d:
This PR introduces the collection of breadcrumbs through subscription-manager, by actually converting the current state of our breadcrumbs into RHSM custom facts and uploading it to the candlepin server tied to the registration. 

Currently, we are introducing this as an MVP to the general idea. In the future, we will stop outputting the file at `/etc/migration-results` and stick only with the one that is located at `/etc/rhsm/facts/*.facts`, but for now, we have both of them placed on the system in case of something going wrong and not having the facts synced with subscription-manager.

### TODO
- [ ] Create tasks on jira to keep improving the existing breadcrumbs collection
- [ ] TODO
 

Jira Issue: [RHELC-664](https://issues.redhat.com/browse/RHELC-664)

Checklist
- [x] PR meets acceptance criteria specified in the Jira issue
- [x] PR has been tested manually in a VM (either author or reviewer)
- [ ] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [x] Code and tests are documented properly
- [x] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending`

PR #660, opened 2022-11-10
Move system info logs to happen after the data collection

r0x0d:
After we introduced the collection of breadcrumbs (#612) as a RHSM facts, we started to use some internal (now public) functions from the systeminfo.py module, which at first, those functions were responsible for logging out some system information, such as: Name, Version, Config file used, and so on... That change caused some duplicated lines in the Convert2RHEL log output.

This commit fixes this by moving the logs to the `resolve_system_info` function instead of placing the logs in each of the facts collection.

Checklist

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -88,6 +88,7 @@ def main():
         # gather system information
         loggerinst.task("Prepare: Gather system information")
         systeminfo.system_info.resolve_system_info()
+        systeminfo.system_info.print_system_information()
         breadcrumbs.breadcrumbs.collect_early_data()
 
         loggerinst.task("Prepare: Clear YUM/DNF version locks")
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -126,6 +126,11 @@ class SystemInfo(object):
         self.has_internet_access = self._check_internet_access()
         self.dbus_running = self._is_dbus_running()
 
+        self.logger.info("%-20s %s" % ("Name:", self.name))
+        self.logger.info("%-20s %d.%d" % ("OS version:", self.version.major, self.version.minor))
+        self.logger.info("%-20s %s" % ("Architecture:", self.arch))
+        self.logger.info("%-20s %s" % ("Config filename:", self.cfg_filename))
+
     @staticmethod
     def get_system_release_file_content():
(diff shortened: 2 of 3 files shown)


SpyTec (dismissed):
LGTM. Are we using the private functions directly outside of systeminfo?

abadger:
It feels like this violates some principle of software design although I can't quite put my finger on it.  Maybe the separation of concerns https://en.wikipedia.org/wiki/Separation_of_concerns ?  In this case, the `SystemInfo` class feels like a large container for separate "plugins" (the individual functions) which each gather a separate piece of information.  It feels a little off for a function that initialized the values in the class to decide what and how to print a subset of the plugin's information.

OTOH, we do want to solve the duplication of data somehow.

Would it be appropriate to print out a summary of all of the data that was gathered in SystemInfo?  Maybe a better design would be to have a separate function that logged a report of all the data in SystemInfo.  And then we call that right after we call `resolve_system_info()`, here: https://github.com/oamg/convert2rhel/blob/1e524492996f61e6674f75aca551edcf35c345ac/convert2rhel/main.py#L90

r0x0d:
> LGTM. Are we using the private functions directly outside of systeminfo?

No, we have a third function that call the internal functions from the SystemInfo class and returns a constructed value for the breadcrumbs.

r0x0d:
> It feels like this violates some principle of software design although I can't quite put my finger on it. Maybe the separation of concerns https://en.wikipedia.org/wiki/Separation_of_concerns ? In this case, the `SystemInfo` class feels like a large container for separate "plugins" (the individual functions) which each gather a separate piece of information. It feels a little off for a function that initialized the values in the class to decide what and how to print a subset of the plugin's information.
> 
> OTOH, we do want to solve the duplication of data somehow.
> 
> Would it be appropriate to print out a summary of all of the data that was gathered in SystemInfo? Maybe a better design would be to have a separate function that logged a report of all the data in SystemInfo. And then we call that right after we call `resolve_system_info()`, here:
> 
> https://github.com/oamg/convert2rhel/blob/1e524492996f61e6674f75aca551edcf35c345ac/convert2rhel/main.py#L90

I like that idea. I will come up with a function for this

SpyTec (approved):
Much better!

abadger (approved):
Looks good!  Approved.

PR #664, opened 2022-11-15
Disable update system role

danmyway:
* it is possible to update to Oracle Linux 8.7 through the system update, disable the ansible role for 8.6 to prevent this


kokesak:
We need to also update the https://github.com/oamg/convert2rhel/blob/main/tests/ansible_collections/roles/oracle-linux-specific/tasks/main.yml.  Add the installation of the non uek kernel for the ol86 from the 86 repository :D

danmyway:
> We need to also update the https://github.com/oamg/convert2rhel/blob/main/tests/ansible_collections/roles/oracle-linux-specific/tasks/main.yml. Add the installation of the non uek kernel for the ol86 from the 86 repository :D

Good catch, done!

PR #667, opened 2022-11-18
Re-enable testing on Oracle Linux 8.6

danmyway:
* OL8.7 is officially out, re-enable 8.6 as a target to run tests on
* the target will get updated using ansible to 8.7 latest
* when 8.7 image is available on testing farm, switch target to 8.7

### DIFF
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -22,7 +22,10 @@ adjust+:
         SYSTEM_RELEASE: oracle-8.4
       when: distro == oraclelinux-8.4
     - environment+:
-        SYSTEM_RELEASE: oracle-8.6
+        #TODO bump distro context to 8.7 when the image is available on the testing farm
+        # Even though the distro context is 8.6 we are setting the SYSTEM_RELEASE to 8.7
+        # to tag the updated system correctly
+        SYSTEM_RELEASE: oracle-8.7
       when: distro == oraclelinux-8.6
 
 prepare:
--- a/tests/integration/tier0/custom-kernel/test_custom_kernel.py
+++ b/tests/integration/tier0/custom-kernel/test_custom_kernel.py
@@ -5,16 +5,17 @@ import pytest
 
 SYSTEM_RELEASE = os.environ.get("SYSTEM_RELEASE")
 
-ORACLE_LATEST_ORIGINAL_KERNEL = os.popen("rpm -q --last kernel | head -1 | cut -d ' ' -f1").read()
+ORIGINAL_KERNEL = os.popen("rpm -q --last kernel | head -1 | cut -d ' ' -f1").read()
 
 DISTRO_KERNEL_MAPPING = {
     "centos-7": {
-        "original_kernel": "kernel-3.10.0-1160.80.1.el7.x86_64",
+        "original_kernel": f"{ORIGINAL_KERNEL}",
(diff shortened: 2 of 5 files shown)


r0x0d (approved):

plans/main.fmf
```diff
@@ -22,7 +22,10 @@ adjust+:
         SYSTEM_RELEASE: oracle-8.4
       when: distro == oraclelinux-8.4
     - environment+:
-        SYSTEM_RELEASE: oracle-8.6
+        #TODO bump distro context to 8.7 when the image is available on the testing farm
+        # Even though the distro context is 8.6 we are setting the SYSTEM_RELEASE to 8.7
+        # to tag the updated system correctly
```
r0x0d, line 27: Maybe something similar to this comment could live in `.packit.yaml` too. I imagine someone (me) looking over the packit.yaml and not getting why we set ol-8.6 and we test for ol-8.7

tests/integration/tier0/custom-kernel/test_custom_kernel.py
```diff
@@ -5,6 +5,8 @@
 
 SYSTEM_RELEASE = os.environ.get("SYSTEM_RELEASE")
 
+ORACLE_LATEST_ORIGINAL_KERNEL = os.popen("rpm -q --last kernel | head -1 | cut -d ' ' -f1").read()
```
r0x0d, line 8: That's clever. Can we update that to the other targets too?

r0x0d (approved):

tests/integration/tier0/custom-kernel/test_custom_kernel.py
danmyway: That would be nicer for sure.
The question is, if we really need to do that. 
I guess for none other than Oracle Linux 8.7 will the kernel be updated very frequently however all except CentOS8.5 could still receive an update IIUIC.
WDYT @r0x0d ?

r0x0d: Well, I guess you're correct. Anything else here will receive an update, so having the static form is good.

kokesak: @danmyway , @r0x0d at firs I thought the same but then I realized that the centos7 is still getting updates (in updates), see: http://mirror.centos.org/centos/7/updates/x86_64/Packages/ , the last kernel `.80` is from 10.11.2022. So we will need to do this dynamically for centos7 and ol7 I presume as well. The only one we can skip is centos8 as it is no longer supported. It can also be handy when alma and rocky comes in place.

danmyway: Yeah, that's what I have meant with receiving the updates, I guess I could have been clearer.
Alright, I will make all original kernel packages load dynamically and keep the CentOS 8.5 static, for one there won't be any update for sure and for second, as you said @kokesak we could use that as a reference for Rocky and Alma.

plans/main.fmf
danmyway: You are right, I wonder why I did not comment in the `.packit.yaml` in the first place :thinking:

kokesak (approved):
Look good, If the test passes feel free to merge it

PR #528, opened 2022-06-22
[RHELC-576, RHELC-596] Merge yum transactions into a single one

r0x0d:
This PR introduces a possibility to merge some yum transactions to execute into a single one during the packages replacement step after the PONR. 

# What are we introducing with this? 

Right now, in our current workflow, we are calling some yum (dnf as well) commands using the CLI. While this work great for some use cases, we are missing two main things that can make our tool more resilient:

## Checking for transaction problems before the PONR

With the current implementation in upstream, we don't have a way implemented to check for transaction problems before the PONR, this implicates that many bugs happens during the package replacement, because we don't have a good way to check if anything will fail prior to this point. Keep in mind that the packages replacement always occur after the PONR.

## Rollback in case of transaction problems

Because the package replacement happens after the PONR, and if anything fails during that step, we don't have any way to rollback the user to the original state, because using the CLI, all the packages are converted up until the point they fail. Thinking in terms of numbers, if the system has 1000 packages, and the replacement fails after the package 900, we left the user with 100~ packages without conversion.

--- 

# What we tested so far?

We have tested the following systems with a bunch of different tweaks and packages: 

- [x] CentOS Linux 7
- [x] CentOS Linux 8
- [x] Oracle Linux 7
- [x] Oracle Linux 8

The tests consisted in a mix of the current resolvers that we parse in the yum output: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/pkghandler.py#L181-L186

And finally, a table showing the tests by system and types 
| Multilib         | Mismatch         | Errors           | Protected        | Required         | Resolve Dependency Integration Tests |
|------------------|------------------|------------------|------------------|------------------|--------------------------------------|
| CentOS Linux 7   | CentOS Linux 7   | CentOS Linux 7   | CentOS Linux 7   | CentOS Linux 7 ✅ | CentOS Linux 7 ✅                     |
| CentOS Linux 8   | CentOS Linux 8   | CentOS Linux 8   | CentOS Linux 8   | CentOS Linux 8   | CentOS Linux 8 ✅                     |
| Oracle Linux 7 ✅ | Oracle Linux 7 ✅ | Oracle Linux 7 ✅ | Oracle Linux 7 ✅ | Oracle Linux 7 ✅ | Oracle Linux 7                       |
| Oracle Linux 8   | Oracle Linux 8 ✅ | Oracle Linux 8 ✅ | Oracle Linux 8 ✅ | Oracle Linux 8 ✅ | Oracle Linux 8                       |
[...]

PR #668, opened 2022-11-23
[RHELC-776] Remove shim-x64 special workaround

r0x0d:
Since the PR #528 landed in upstream, we included the shim-x64 package as part of the excluded packages for Oracle Linux 7, meaning that we remove this package before doing any operation in Convert2RHEL, thus, we don't require the special workaround any more as this package will not be present on the system when it reaches that part of the code.

The special workaround we had was to remove the protected file that the shim-x64 package places on /etc/yum/protected.d when the package is installed, since we remove it, the protection file is not present any more.

Jira Issue: [RHELC-776](https://issues.redhat.com/browse/RHELC-776)

Checklist


abadger (dismissed):
This looks good to me.  Thanks for writing the history of the special case in the description too.  I wouldn't have known what to look for otherwise!

abadger:
Note that the codeql requests look valid but should be easy changes that do not impact the operation of the code.

r0x0d:
> Note that the codeql requests look valid but should be easy changes that do not impact the operation of the code.

I think I can go ahead an fix them right away. They seem easy to fix indeed.

SpyTec (approved):

tests/integration/tier1/handle-special-pkgs-case/test_handle_shim_x64_pkg.py
```diff
@@ -1,34 +0,0 @@
...
-            env.str("RHSM_USERNAME"),
-            env.str("RHSM_PASSWORD"),
-            env.str("RHSM_POOL"),
-        )
-    ) as c2r:
-        pass
-    assert c2r.exitstatus == 0
-
-    # Check that the package is still present on the system
-    assert shell(f"rpm -qi {shim_x64_pkg}").returncode != 0
```
SpyTec, line 34: We should probably keep this test but modify it to work

r0x0d: How exactly we should develop the test for it? Maybe checking if convert2rhel got rid of the package during the conversion? If so, then we already have an test for it. 

We could make sure that the package is not present after the conversion finishes?

SpyTec: Gotcha. Is that an integration test then?

r0x0d: Yeah, there is an integration test that does that. https://github.com/oamg/convert2rhel/pull/668/files#diff-effa0515e248da3fe738adfe9c21f2977701c9e1c8701bcc86c36acb32415a43R97 You just pass the package name that needs to be installed in the tests, and it will check for package is not present after the conversion

SpyTec: Okay. LGTM

SpyTec (approved):

PR #672, opened 2022-11-25
[RHELC-823] chore: add devcontainer

SpyTec:
Since convert2rhel is a volatile tool to run on our own machines without
prior understanding that the tests we are running are mocked out
correctly. We may break our machine. It is better to run within a
container but we then get into issues where debugging, testing, and so
on isn't as integrated with existing tools as we'd like.

This is the first step to adding devcontainer using VSCode Remote
Development which will show VSCode as normal but within a container,
making it much easier to run tests and debug

---

I believe most things work at the moment but there are some tasks that need to be done. This also assumes a Podman environment, I do not think it will work with Docker
- [x] Cannot create files in project root (likely SELinux as project root isn't owned by container user)
- [x] Clean-up development Containerfile

Jira Issue: [RHELC-823](https://issues.redhat.com/browse/RHELC-823) 

Checklist

### DIFF
--- a/.devcontainer/centos8-development.Dockerfile
+++ b/.devcontainer/centos8-development.Dockerfile
@@ -1,0 +1,45 @@
+FROM centos:8 as base
+
+ENV PODMAN_USERNS=keep-id
+ENV PYTHON python3
+ENV PIP pip3
+ENV PYTHONDONTWRITEBYTECODE 1
+
+ENV URL_GET_PIP "https://bootstrap.pypa.io/pip/3.6/get-pip.py"
+ENV APP_DEV_DEPS ".devcontainer/centos8.requirements.txt"
+ENV APP_MAIN_DEPS \
+    python3 \
+    python3-six \
+    python3-dbus \
+    python3-pexpect \
+    git
+
+ENV USERNAME=vscode
+ENV USER_UID=1000
+ENV USER_GID=$USER_UID
+
+WORKDIR /workspaces/convert2rhel
+
+FROM base as install_main_deps
+
+RUN sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Linux-* && \
+    sed -i 's|#baseurl=http://mirror.centos.org|baseurl=https://vault.centos.org|g' /etc/yum.repos.d/CentOS-Linux-*
+
+RUN dnf update -y && dnf install -y $APP_MAIN_DEPS && dnf clean all
+
+FROM install_main_deps as install_dev_deps
+
+RUN curl $URL_GET_PIP | $PYTHON
+COPY $APP_DEV_DEPS $APP_DEV_DEPS
+RUN $PIP install -r $APP_DEV_DEPS
+
+FROM install_dev_deps as install_application
+
+RUN groupadd --gid=$USER_GID -r $USERNAME && \
+    useradd --uid=$USER_UID --home /home/$USERNAME --gid=$USER_GID -m $USERNAME
+
+RUN chown -R $USERNAME:$USERNAME .
+COPY --chown=$USERNAME:$USERNAME . .
+
+COPY .devcontainer/.bashrc /home/$USERNAME
+USER $USERNAME:$USERNAME
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,46 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/docker-existing-dockerfile
+{
+	"name": "CentOS Linux 8",
+	// Sets the run context to one level up instead of the .devcontainer folder.
+	"context": "..",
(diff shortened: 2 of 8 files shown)


SpyTec:
Screenshot from 2022-11-25 17-02-37

r0x0d:
Would be nice to update the README.md or CONTRIBUTING.md with some information on how to use the devcontainer.

SpyTec:
> Would be nice to update the README.md or CONTRIBUTING.md with some information on how to use the devcontainer.

Yep. It required some setting up and there are some assumptions that had to be taken. Like how it most likely doesn't work with Docker. I want to see about fixing the SELinux errors first, optimize it a bit, then see about improving the usability before tackling documentation

SpyTec:
@r0x0d I think everything is addressed now. I had to downgrade a few `pre-commit` things due to it dropping support for 3.6. As well as removing some other things.

Lmk what you think of the documentation - I will go ahead and create a Jira issue for this soon where I can track this. I'll present it tomorrow as well unless anyone wants to show other things

I know there will be places to improve, if you spot anything just shout

SpyTec:
@r0x0d @bocekm This now works as I want it to. With #679 merged it fixed my issue as well. I also found out how to debug code so that is now implemented as well

The `CONTRIBUTING.md` has a section for known issues when it comes to debugging. Problem for us is that the python extension dropped support for py3.6 and versions later than 2022.8.1 now breaks. So we have to make sure to pin the version to that

.devcontainer/centos8-development.Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM centos:8 as base
+
+ENV PODMAN_USERNS=keep-id
+ENV PYTHON python3
+ENV PIP pip3
+ENV PYTHONDONTWRITEBYTECODE 1
+
+ENV URL_GET_PIP "https://bootstrap.pypa.io/pip/3.6/get-pip.py"
```
r0x0d, line 8: Can't we just replace this with `yum install python3-pip`?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+	},
+	// Extensions that are handy when developing
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"ms-python.vscode-pylance",
+				"eamodio.gitlens",
+				"GitHub.vscode-pull-request-github",
+				"Cameron.vscode-pytest",
+				"EditorConfig.EditorConfig",
```
r0x0d, line 41: Do we need the `EditorConfig` extension here? I don't see any `.editorconfig` in the root folder anywhere.

r0x0d (commented):
Pretty nice work, @SpyTec! I tested it locally, and it seems to work like a charm! I just want to mention that (at least, for a fresh vscode install), it seems to detect that this project is supposed to run under dev containers, so, if you don't have the extension installed, it will prompt you to install it, and after that, another button will show to open a new vscode window and run the project inside the container. 

Other than that, I would just like to point out what we previously discussed in google chat: 

- We should make it more explicit that the dev container is supposed to run using podman, not docker. The CONTRIBUTING.MD document has a sentence about that, but I think we could make it more visible, so the reader can spot that easily. 
- It should be good to place in the CONTRIBUTING.MD the piece of configuration that allows vscode to use podman instead of docker
- Not sure what went wrong, but, the first time I tried to debug a unit_test, it didn't work out as it gave me the message about python 3.7 being the minor version it supported, thought, I had the correct version of the python extension listed in the docs. Restarting the vscode fixed that for me.

And again, nice work with all that, @SpyTec! Looks pretty simple and organized.

.devcontainer/centos8-development.Dockerfile
SpyTec: We have it in our other Containerfiles as well so I'm not sure, I'll look into it

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,46 @@
...
+	],
+	// Need to mount manually as we run into SELinux issues otherwise
+	"workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/${localWorkspaceFolderBasename},type=bind,Z",
+	// Need to change home to avoid errors in Podman
+	"containerEnv": {
+		"HOME": "/home/vscode"
+	},
+	// Uncomment to use the Docker CLI from inside the container. See https://aka.ms/vscode-remote/samples/docker-from-docker
+	// "mounts": [ "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind" ],
+	// Uncommented to connect as a non-root user. See https://aka.ms/vscode-remote/containers/non-root
```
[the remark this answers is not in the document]
r0x0d: I think this could apply for the other comments there are here as well... Most of them are not needed to be contained here, as someone who is doing changes can easily find the docs for devcontainers, or, we could place the link for the docs as a top-level comment.

r0x0d (approved):
Approving and merging this as it works locally (tested twice). 

I left a comment about removing the unnecessary comments from the `devcontainer.json` file, but I will not block the PR because of this.

PR #674, opened 2022-11-29
Add a few exceptions class for yum transaction handler

r0x0d:
We discovered that we were missing a few exception classes to catch during the package download. We went ahead and added a couple of exceptions to catch gpg checks and mirror exceptions too.

Checklist

### DIFF
--- a/convert2rhel/pkgmanager/handlers/yum.py
+++ b/convert2rhel/pkgmanager/handlers/yum.py
@@ -216,14 +216,21 @@ class YumTransactionHandler(TransactionHandlerBase):
 
         try:
             self._base.processTransaction()
-        except (
-            pkgmanager.Errors.YumRPMCheckError,
-            pkgmanager.Errors.YumTestTransactionError,
-            pkgmanager.Errors.YumRPMTransError,
-            pkgmanager.Errors.YumDownloadError,
-            pkgmanager.Errors.YumBaseError,
-            pkgmanager.Errors.YumGPGCheckError,
-        ) as e:
+        except pkgmanager.Errors.YumBaseError as e:
+            # We are catching only `pkgmanager.Errors.YumBaseError` as the base
+            # exception here because all of the other exceptions that can be
+            # raised during the transaction process inherit it from
+            # YumBaseError.
+
+            # The following exceptions is the ones we are actually looking for,
+            # but simplified to only catch the YumBaseError:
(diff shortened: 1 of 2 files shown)


convert2rhel/pkgmanager/handlers/yum.py
```diff
@@ -220,6 +220,9 @@ def _process_transaction(self, validate_transaction):
             pkgmanager.Errors.YumRPMCheckError,
             pkgmanager.Errors.YumTestTransactionError,
             pkgmanager.Errors.YumRPMTransError,
+            pkgmanager.Errors.YumDownloadError,
+            pkgmanager.Errors.YumBaseError,
+            pkgmanager.Errors.YumGPGCheckError,
         ) as e:
```
abadger, line 226:
```suggestion
            pkgmanager.Errors.YumBaseError
        ) as e:
```

`YumBaseError` is the base class for all of the yum exceptions so we can just list that to catch all of these exceptions.  Note that this could be made into a single line but github's suggestion feature wouldn't let me select all the way to the `except (` line.  That would look like this:

``` python
        except pkgmanager.Errors.YumBaseError as e:
```

convert2rhel/unit_tests/pkgmanager/handlers/yum_test.py
```diff
@@ -180,6 +180,9 @@ def test_process_transaction_with_exceptions(self, pretend_os, _mock_yum_api_cal
             pkgmanager.Errors.YumRPMCheckError,
             pkgmanager.Errors.YumTestTransactionError,
             pkgmanager.Errors.YumRPMTransError,
+            pkgmanager.Errors.YumDownloadError,
+            pkgmanager.Errors.YumBaseError,
+            pkgmanager.Errors.YumGPGCheckError,
         )
```
abadger, line 186: Do the side_effects need to be given to this unit test via parametrize?  At first glance, it looks like this unit test will only tun once, raising `YumRPMCheckError` and checking that will lead to us calling `loginst.critical()`.

r0x0d: We can simplify that to raising only the YumBaseError

SpyTec (commented):
LGTM. Lets wait for integration tests

SpyTec (approved):

PR #681, opened 2022-12-08
[RHELC-723] Remove RegistrationCommand.args property

pr-watson:
Removed RegistrationCommand.args property and modified the log messages to take the arguments from self rather than the RegistrationCommand.args property https://issues.redhat.com/browse/RHELC-723

Checklist

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -27,8 +27,6 @@ import dbus
 import dbus.connection
 import dbus.exceptions
 
-from six.moves import urllib
-
 from convert2rhel import backup, i18n, pkghandler, utils
 from convert2rhel.redhatrelease import os_release_file
 from convert2rhel.systeminfo import system_info
@@ -368,42 +366,6 @@ class RegistrationCommand(object):
         return cls(**registration_attributes)
 
     @property
-    def args(self):
-        """
-        This property is a list of the command-line arguments that will be passed to
-        subscription-manager to register the system. Set the individual attributes for
-        :attr:`server_url`, :attr:`activation_key`, etc to affect the values here.
-
-        .. note:: :attr:`password` is not passed on the command line. Instead,
-            it is sent to the running subscription-manager process via pexpect.
-        """
-        args = ["register", "--force"]
-
-        if self.connection_opts:
-            if self.rhsm_port:
-                netloc = "%s:%s" % (self.rhsm_hostname, self.rhsm_port)
-            else:
-                netloc = self.rhsm_hostname
-
-            prefix = self.rhsm_prefix if self.rhsm_prefix else ""
-            if prefix.startswith("/"):
-                prefix = prefix[1:]
-
-            server_url = urllib.parse.urlunsplit(("https", netloc, prefix, "", ""))
-            args.append("--serverurl=%s" % server_url)
-
-        if self.activation_key:
-            args.append("--activationkey=%s" % self.activation_key)
-
-        if self.org:
-            args.append("--org=%s" % self.org)
-
-        if self.username:
-            args.append("--username=%s" % self.username)
-
-        return args
-
-    @property
     def connection_opts(self):
         """
         This property is a dbus.Dictionary that contains the connection options for RHSM
@@ -469,7 +431,12 @@ class RegistrationCommand(object):
 
             try:
                 if self.password:
-                    loggerinst.info("Registering via username/password: %s" % " ".join(utils.hide_secrets(self.args)))
+                    if self.org:
+                        loggerinst.info("Organization: %s", self.org)
+                    loggerinst.info("Username: %s", self.username)
+                    loggerinst.info("Password: %s", "*" * 5)
+                    loggerinst.info("Connection Options: %s", self.connection_opts)
+                    loggerinst.info("Locale settings: %s", i18n.SUBSCRIPTION_MANAGER_LOCALE)
                     args = (
                         self.org or "",
                         self.username,
@@ -489,7 +456,10 @@ class RegistrationCommand(object):
                     )
 
                 else:
-                    loggerinst.info("Registering via org/activation_key: %s" % " ".join(utils.hide_secrets(self.args)))
+                    loggerinst.info("Organization: %s", self.org)
+                    loggerinst.info("Activation Key: %s", "*" * 5)
+                    loggerinst.info("Connection Options: %s", self.connection_opts)
+                    loggerinst.info("Locale settings: %s", i18n.SUBSCRIPTION_MANAGER_LOCALE)
                     args = (
                         self.org,
                         [self.activation_key],
--- a/convert2rhel/unit_tests/subscription_test.py
+++ b/convert2rhel/unit_tests/subscription_test.py
@@ -27,8 +27,6 @@ import dbus.exceptions
 import pytest
 import six
 
-from six.moves import urllib
-
 from convert2rhel import backup, pkghandler, subscription, toolopts, unit_tests, utils
 from convert2rhel.systeminfo import system_info
 from convert2rhel.unit_tests import GetLoggerMocked, get_pytest_marker, run_subprocess_side_effect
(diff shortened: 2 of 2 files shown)


pr-watson:
Here is a snippet of the logging messages after the change:

[12/08/2022 21:15:55] DEBUG - Connecting to the private DBus

Registering via username/password:

Organization: None

Username: c2r_main

Password: *****

Register Options: dbus.Dictionary({'force': True}, signature=dbus.Signature('sv'), variant_level=1)

Connection Options: dbus.Dictionary({'host': 'subscription.rhsm.stage.redhat.com'}, signature=dbus.Signature('sv'), variant_level=1)

Locale Setting: C


Setting RHSM connection configuration.

convert2rhel/subscription.py
```diff
@@ -489,7 +465,19 @@ def __call__(self):
...
+                        "Organization: %s\nActivation Key:%s\nRegister Options: %s\n"
+                        "Connection Options: %s\nLocale Setting: %s\n"
+                        % (
+                            self.org,
+                            "*" * 5,
+                            REGISTER_OPTS_DICT,
+                            self.connection_opts,
+                            i18n.SUBSCRIPTION_MANAGER_LOCALE,
+                        )
+                    )
```
r0x0d, line 479: @pr-watson, I would modify this logging a bit and make it a bit simpler to read and to maintain later, maybe something like this: 

 ```suggestion
                    loggerinst.info("Organization: %s", self.org)
                    loggerinst.info("Activation Key: %s", "*" * 5)
                    loggerinst.info("Connection Options: %s", self.connection_opts)
                    loggerinst.info("Locale settings: %s", i18n.SUBSCRIPTION_MANAGER_LOCALE)
```

I would say it's totally fine to have multiple log calls as far as they are easier to ready and maintain... What do you think? 

NOTE: This also applies to the logging text above that you have in the same way.

**Review by r0x0d — COMMENTED** [review]
Thank you for your first PR, @pr-watson! It's looking pretty good, only a couple of small things that could be modified! 

If you can, please, fix the `CodeQL` suggestions in your PR (mainly, removing the unused import of `urllib` as you removed the whole piece of code that made use of them) just so we don't have any more warnings/suggestions to our bucket as the PRs get merged. 

On another note, I read in the description of that issue that this is supposed to use the `utils.hide_secrets()` function to make the output obfuscated for password/activation_key, but unfortunately that function is supposed to work with CLI parameters rather than any other type of text. @abadger do you think it's worth it to enhance that a bit to also accept other type of data, this way we don't need to explicitly call `"*" * 5` here and there whenever we want to work with an obfuscated output. 

In any case, looking good and waiting for your reply, @pr-watson!

**Comment by pr-watson** [comment]
@r0x0d I'll also get rid of the unused import for urllib

File: convert2rhel/subscription.py [hunk 2]
```diff
@@ -469,7 +431,10 @@ def __call__(self):
 
             try:
                 if self.password:
-                    loggerinst.info("Registering via username/password: %s" % " ".join(utils.hide_secrets(self.args)))
+                    loggerinst.info("Organization: %s", self.org)
+                    loggerinst.info("Activation Key: %s", "*" * 5)
```
r0x0d, line 435: I think we want to change it here to use `username` and `password` instead of `activation key`, as it seems that the `activation key` one is only the one that happens in the `else` statement... 

```suggestion
                    loggerinst.info("Username: %s", self.username)
                    loggerinst.info("Password: %s", "*" * 5)
```

convert2rhel/subscription.py
```diff
@@ -469,7 +431,10 @@ def __call__(self):
 
             try:
                 if self.password:
-                    loggerinst.info("Registering via username/password: %s" % " ".join(utils.hide_secrets(self.args)))
+                    loggerinst.info("Organization: %s", self.org)
```
r0x0d, line 434: I'm a bit unsure if we have to print `Organization:` here... Maybe we do because we set that in the line 439 :thinking:  I guess it would be nice to confirm with @abadger since he did the code a while ago.

r0x0d: Also, if we do, I would do conditional to print that only if there is an org present otherwise, I would omit this information as it is not relevant if it's not present, like so: 

```suggestion
                    if self.org:
                        loggerinst.info("Organization: %s", self.org)

```

r0x0d: The indentation may be wrong, though...

convert2rhel/subscription.py
pr-watson: Good catch, i'll change those lines

convert2rhel/subscription.py
pr-watson: I'll wait on @abadger to respond for this change

abadger: It turns out that organization is needed even for username and password if the username is associated with more than one organization.  We have a ticket open to handle that case: https://issues.redhat.com/browse/RHELC-726

So logging the organization is a good idea here.

abadger:
> On another note, I read in the description of that issue that this is supposed to use the `utils.hide_secrets()` function to make the output obfuscated for password/activation_key, but unfortunately that function is supposed to work with CLI parameters rather than any other type of text. @abadger do you think it's worth it to enhance that a bit to also accept other type of data, this way we don't need to explicitly call `"*" * 5` here and there whenever we want to work with an obfuscated output.

I've been contemplating that over the weekend.  I think it doesn't make sense to enhance that particular function.  The reason is that we don't have a good way to tell that the password is used just from a bare variable.  The function is able to do its job on the command line args because they're key value pairs, where the key tells us that we're dealing with a password or other secret.

We could extract the obfuscation string itself into its own function or a global variable... but I lean towards it not being worth it:
``` python
# As a function:
def obfuscate_value(value):
    return "*" * 5

password = "s3cret"
print("Password %s" % obfuscate_value(password))

# As a global:
SECRET_REPLACEMENT_STRING = "*" * 5

print("Password %s" % SECRET_REPLACEMENT_STRING
```

r0x0d:
> > On another note, I read in the description of that issue that this is supposed to use the `utils.hide_secrets()` function to make the output obfuscated for password/activation_key, but unfortunately that function is supposed to work with CLI parameters rather than any other type of text. @abadger do you think it's worth it to enhance that a bit to also accept other type of data, this way we don't need to explicitly call `"*" * 5` here and there whenever we want to work with an obfuscated output.
> 
> I've been contemplating that over the weekend. I think it doesn't make sense to enhance that particular function. The reason is that we don't have a good way to tell that the password is used just from a bare variable. The function is able to do its job on the command line args because they're key value pairs, where the key tells us that we're dealing with a password or other secret.
> 
> We could extract the obfuscation string itself into its own function or a global variable... but I lean towards it not being worth it:
> 
> ```python
> # As a function:
> def obfuscate_value(value):
>     return "*" * 5
> 
> password = "s3cret"
> print("Password %s" % obfuscate_value(password))
> 
> # As a global:
> SECRET_REPLACEMENT_STRING = "*" * 5
> 
> print("Password %s" % SECRET_REPLACEMENT_STRING
> ```

Makes sense to me.... I was trying to figure out another way of doing it, but it turns out that there are not many cases to use such function, and I think most of the time we will use the `hide_secrets()` the way it is.

pr-watson:
This new patch fixes the broken unit tests, combines two unit tests (test_calling_registration_command_activation_key and test_calling_registration_command_password) into one general unit test. This unit test also checks that Organization is properly logged using caplog.
Thanks @abadger for guiding me through the pytest process

SpyTec:
Did a rebase to be up to date with pipeline and also trigger new integration tests

r0x0d:
@danmyway it seems that we still have the problem in our integration tests, judging by the results here :(

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -858,50 +781,112 @@ def test_connection_opts(self, rhsm_hostname, rhsm_port, rhsm_prefix, expected):
...
+    #     )
+    #     mocked_rhsm_call_blocking.assert_called_once_with(
+    #         "com.redhat.RHSM1",
+    #         "/com/redhat/RHSM1/Register",
+    #         "com.redhat.RHSM1.Register",
+    #         "RegisterWithActivationKeys",
+    #         "sasa{sv}a{sv}s",
+    #         args,
+    #         timeout=subscription.REGISTRATION_TIMEOUT,
+    #     )
```
SpyTec, line 803: If we don't need it we should remove it
```suggestion
```

SpyTec (changes requested):
One remark otherwise LGTM. Will move to dev complete

Edit: Ah we decided integration tests are not needed. Just have to make sure the existing ones pass

convert2rhel/unit_tests/subscription_test.py
pr-watson: Just made the change, thanks @SpyTec

SpyTec (approved):

r0x0d (approved):
Looks good to me! Thanks for the work here, @pr-watson

convert2rhel/subscription.py
```diff
@@ -469,7 +431,12 @@ def __call__(self):
 
             try:
                 if self.password:
-                    loggerinst.info("Registering via username/password: %s" % " ".join(utils.hide_secrets(self.args)))
+                    if self.org:
+                        loggerinst.info("Organization: %s", self.org)
+                    loggerinst.info("Username: %s", self.username)
```
bocekm, line 436: Please replace these two with asterisks based on https://github.com/oamg/convert2rhel/pull/725.

You can perhaps define this `obfuscation_string` as a constant on a module level and then use it in this `subscription.py` module as well as in `utils.py`.
https://github.com/oamg/convert2rhel/blob/v1.1/convert2rhel/utils.py#L722

bocekm (changes requested):

convert2rhel/subscription.py
bocekm: Or if this is ready to merge, we can do so and I can update #725.

pr-watson: @bocekm it is ready to merge, it would be great if you could update #725 thanks!

r0x0d:
The failures here looks fine in my opinion. They don't seem to be related to the changes in the PR. I will leave it up to you guys to decide if you want to merge it right away or wait for an investigation regarding the errors.

convert2rhel/subscription.py
bocekm: Done.

PR #682, opened 2022-12-14
[RHELC-839] Port the cert unittests to use pytest instead of unittest.

abadger:
Jira Issue: [RHELC-839](https://issues.redhat.com/browse/RHELC-839)

Checklist

### DIFF
--- a/convert2rhel/unit_tests/cert_test.py
+++ b/convert2rhel/unit_tests/cert_test.py
@@ -16,77 +16,109 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 import os
 import shutil
-import unittest
 
 import pytest
 import six
 
-from convert2rhel import unit_tests
-
-
-six.add_move(six.MovedModule("mock", "mock", "unittest.mock"))
-
-from convert2rhel import cert, utils
+from convert2rhel import cert, unit_tests, utils
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
 
 
-class TestCert(unittest.TestCase):
+# Directory with all the tool data
+BASE_DATA_DIR = os.path.realpath(os.path.join(os.path.dirname(__file__), "../data/"))
 
-    # Certificates for all the supported RHEL variants
-    certs = {
-        "x86_64": {"6": "69.pem", "7": "69.pem", "8": "479.pem"},
-        "ppc64": {
-            "7": "74.pem",
-        },
-    }
 
-    # Directory with all the tool data
-    base_data_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "../data/"))
+@pytest.fixture
+def cert_dir(monkeypatch, request):
+    if request.param["data_dir"] is None:
+        data_dir = unit_tests.NONEXISTING_DIR
+        rhel_cert_dir = os.path.join(data_dir, "rhel-certs")
+    else:
+        data_dir = request.param["data_dir"]
+        rhel_cert_dir = os.path.join(data_dir, "rhel-certs")
 
-    @unit_tests.mock(utils, "DATA_DIR", unit_tests.TMP_DIR)
-    def test_get_cert_path(self):
-        # Check there are certificates for all the supported RHEL variants
-        for arch, rhel_versions in self.certs.items():
-            for rhel_version, pem in rhel_versions.items():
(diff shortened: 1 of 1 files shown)


convert2rhel/unit_tests/cert_test.py
```diff
@@ -16,77 +16,109 @@
...
+            },
+        ),
+        # Missing cert dir
+        (
+            "Error: Could not access %(cert_dir)s.",
+            {
+                "data_dir": None,
+                "arch": "nonexisting_arch",
+            },
+        ),
```
r0x0d, line 76: Have you considered giving an `id` to the test param? This way you don't need to rely on a comment in the code, but actually have it as a named id (I think it shows under the test run too) 

```suggestion
        pytest.param(
            "Error: System certificate (.pem) not found in %(cert_dir)s.",
            {
                "data_dir": unit_tests.TMP_DIR,
                "arch": "arch",
            },
            id="missing-certificate",
        ),
        pytest.param(
            "Error: Could not access %(cert_dir)s.",
            {
                "data_dir": None,
                "arch": "nonexisting_arch",
            },
            id="missing-cert-dir",
        ),
```

abadger: Nice!  I didn't know about that feature.  I like it a lot!

r0x0d: I remembered about it the other day, so I thought it might come in handy for those types of tests that we need to specify what is going on. Probably it would be a good idea to use the same logic in the `test_hide_utils` at some point to…

r0x0d (commented):
It looks good, @abadger! But could you take care of removing the unused import of `six`´? After that, I think it's pretty much it!

abadger:
> It looks good, @abadger! But could you take care of removing the unused import of `six`´? After that, I think it's pretty much it!

Good catch.  Unused import has been removed.

convert2rhel/unit_tests/cert_test.py
```diff
@@ -16,77 +16,108 @@
...
-        # Remove the temporary directory tree
+
+    monkeypatch.setattr(utils, "DATA_DIR", data_dir)
+    monkeypatch.setattr(system_info, "arch", request.param["arch"])
+
+    yield rhel_cert_dir
+
+    # Remove the temporary directory tree if we created it earlier
+    if request.param["data_dir"] is not None:
         shutil.rmtree(os.path.join(utils.DATA_DIR, "rhel-certs"))
```
r0x0d, line 51: I would change this to `data_dir` instead of using the constant in `utils`, because if someone changes this fixture in some way, or we change the constant name, this fixture would start failing.

```suggestion
        shutil.rmtree(data_dir)
```

convert2rhel/unit_tests/cert_test.py
```diff
@@ -16,77 +16,108 @@
...
+        ("x86_64", "6", "69.pem"),
+        ("x86_64", "7", "69.pem"),
+        ("x86_64", "8", "479.pem"),
+        ("ppc64", "7", "74.pem"),
+    ),
+)
+def test_get_cert_path(arch, rhel_version, pem, monkeypatch):
+    monkeypatch.setattr(utils, "DATA_DIR", unit_tests.TMP_DIR)
+    # Check there are certificates for all the supported RHEL variants
+    utils.DATA_DIR = os.path.join(BASE_DATA_DIR, rhel_version, arch)
```
r0x0d, line 100: We could just mock the `DATA_DIR` here with the line `100`, I think... Probably we don't need to mock to a temporary dir and then overwrite that with the value we want, right?

r0x0d (approved):
This is fine to merge from my point of view. The unit tests looks way better and easier to understand now than it was in the past. 

@abadger waiting for you to take a look at my comments and decide if they are worth it fixing right now or later. 

EDIT: The integration tests failure here doesn't mean too much to the PR, as the changes are related only to unit_tests, not code.

convert2rhel/unit_tests/cert_test.py
r0x0d: @abadger, pushed the fix here.

convert2rhel/unit_tests/cert_test.py
r0x0d: @abadger, I have pushed the fix here.

r0x0d (approved):

r0x0d:
Merging this as the results of the integration tests are not important for this PR.

PR #687, opened 2023-01-04
[RHELC-829] Improve repoquery parsing for loaded kernel check

r0x0d:
The repoquery parsing for the loaded kernel check was only filtering out a single line that was present in the output, we were filtering this out and removing it with the intention to clean the output and let only the relevant output of repoquery to be parsed, however, we discovered that different text can appear in the output, thus, leaving us with a broken workflow again.

The solution to fix this problem was simply adding an identifier to the repoquery output, this way we can filter out only the lines we want and ignore the rest (logging them as debug messages in case we miss anything important from that line).

In this commit I'm introducing a small refactoration for that function and it's unit_tests as well.

Jira Issue: [RHELC-829](https://issues.redhat.com/browse/RHELC-829)

Checklist


r0x0d:
@danmyway  I think it would be nice to sync how the integration test for this should be. I have a few ideas, but I would like to make it in a way that is easy to add more preparation steps in case we need to test other type of messages/cases.

convert2rhel/checks.py
```diff
@@ -706,106 +705,122 @@ def is_loaded_kernel_latest():
...
+    # the kernel (or kernel-core) packages, let's just print a warning and
+    # return.
+    if not packages:
+        logger.warning(
+            "Could not find any %s in the available repositories. "
+            "This could mean that the repositories are not enabled, acessible, "
+            "or missing some important information to progress.\n"
+            "Skipping this check as we can't progress with further verification." % package_to_check
+        )
+        return
```
r0x0d, line 798: Removing this to reuse the if condition above to use the length of the packages instead of the length of repoquery output.

SpyTec (approved):

convert2rhel/checks.py
```diff
@@ -706,106 +705,110 @@ def is_loaded_kernel_latest():
...
-                "the %s comparison.\n"
-                "Beware, this could leave your system in a broken state. " % package_to_check
-            )
+    packages = []
+    # We are expecting an repoquery output to be similar to this:
+    #   C2R     1671212820      3.10.0-1160.81.1.el7    updates
+    # We need the `C2R` identifier to be present on the line so we can know for
+    # sure that the line we are working with is the a line that contains
+    # relevant repoquery information to our check, otherwise, we just log the
+    # information as debug and do nothing with it.
```
r0x0d, line 757:
```suggestion
    # We need the `C2R` identifier to be present on the line, so we can know
    # for sure that the line we are working with is the line that contains
    # relevant repoquery information to our check, otherwise, we just log the
    # information as debug and do nothing with it.
```

danmyway:
On a call with @r0x0d and @kokesak, we have decided, that the current state of latest kernel check verification tests is sufficient.

convert2rhel/checks.py
```diff
@@ -706,106 +705,110 @@ def is_loaded_kernel_latest():
         )
         return
 
-    reposdir = get_hardcoded_repofiles_dir()
+    cmd = [
+        "repoquery",
+        "--setopt=exclude=",
+        "--quiet",
+        "--qf",
+        "C2R\\t%{BUILDTIME}\\t%{VERSION}-%{RELEASE}\\t%{REPOID}",
```
r0x0d, line 713: Note for future me: This will conflict with the changes made in https://github.com/oamg/convert2rhel/pull/713

convert2rhel/checks.py
```diff
@@ -715,106 +714,110 @@ def is_loaded_kernel_latest():
...
-            )
+    packages = []
+    # We are expecting an repoquery output to be similar to this:
+    #   C2R     1671212820      3.10.0-1160.81.1.el7    updates
+    # We need the `C2R` identifier to be present on the line so we can know for
+    # sure that the line we are working with is the a line that contains
+    # relevant repoquery information to our check, otherwise, we just log the
+    # information as debug and do nothing with it.
+    for line in repoquery_output.split("\n"):
+        if line.strip() and "C2R" in line:
```
abadger, line 768:
```suggestion
        if line.strip() and line.startswith("C2R"):
```

abadger (dismissed):
One minor change that just tightens up the match for the line (I do not expect that the cornercase would ever be triggered in practice but I just saw that it would be a more exact test).

Given how unlikely that case is, I am approving this whether or not you decide to apply the change.

convert2rhel/checks.py
r0x0d: Nice catch, @abadger, thanks!

convert2rhel/checks.py
```diff
@@ -715,106 +714,110 @@ def is_loaded_kernel_latest():
...
-            )
+    packages = []
+    # We are expecting an repoquery output to be similar to this:
+    #   C2R     1671212820      3.10.0-1160.81.1.el7    updates
+    # We need the `C2R` identifier to be present on the line so we can know for
+    # sure that the line we are working with is the a line that contains
+    # relevant repoquery information to our check, otherwise, we just log the
+    # information as debug and do nothing with it.
+    for line in repoquery_output.split("\n"):
+        if line.strip() and line.startswith("C2R"):
```
abadger, line 768:
```suggestion
        line = line.strip()
        if line.startswith("C2R"):
```

Okay my suggestion broke things because the line has initial whitespace.  This should fix my suggestion and I think it might be better too since:
* It sanitizes the data prior to using it which is the best way to organize things.
* The test for whether  line is not an empty string is a subset of the `line.startswith()` so we don't need to test them both independently.

r0x0d: Tested it locally and it works! Thanks for fixing this for me, @abadger!

abadger (approved):

r0x0d:
I think we have a test stuck... It's been running for 3h now.

r0x0d:
Yay, all green!

abadger (approved):
The current approval isn't affecting the PR for some reason.  Trying to diagnose (and get it approved for real).

abadger (approved):
Approved

PR #695, opened 2023-01-11
Remove black args from pre-commit-config file

r0x0d:
We had pre-commit-config settings the args for black and the
pyproject.toml file as also setting some properties for the tool.

It seems that black override those properties from pyproject.toml since
we used the CLI args for it, but in any case, we should rely only on one
configuration file to tell the tool what property it should override.

Also removed the unnecessary property in pre-commit-config for `isort`.

Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,19 +9,11 @@ repos:
           language: python
           language_version: "python3"
           additional_dependencies: [click==7.1.2, black==21.12b0]
-          args:
-            - "--target-version=py27"
-            - "--target-version=py36"
-            - "--target-version=py37"
-            - "--target-version=py38"
-            - "--target-version=py39"
-            - "--target-version=py310"
 
   - repo: "https://github.com/timothycrosley/isort"
     rev: 5.11.4
     hooks:
       - id: isort
-        language_version: "python3"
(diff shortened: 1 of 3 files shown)


SpyTec:
> Also removed the unnecessary property in pre-commit-config for `isort`.

So it doesn't use it anymore or it will always point to python3 or why is it no longer necessary?

r0x0d:
> > Also removed the unnecessary property in pre-commit-config for `isort`.
> 
> So it doesn't use it anymore or it will always point to python3 or why is it no longer necessary?

Should have clarified this better, but, without the `language_version`, pre-commit will auto-detect the language version that needs to be used for the hook. The `languag_version` only makes sense if we want to override the defaults. Having it set to `python3` would be just the same as not having anything in there, as most of us already have python3 by default on the system. 

Makes sense to keep this property if we want to ensure that it will be python3 all the times.

SpyTec (approved):
That's fine, as long as it's in in this PR I don't want to go through the effort of adding it to the commit as well. Just wanted to know

r0x0d:
> That's fine, as long as it's in in this PR I don't want to go through the effort of adding it to the commit as well. Just wanted to know

Cool! Thanks

PR #702, opened 2023-01-11
[KAIZEN] Make pylint logs usable

abadger:
Make pylint logs usable.
    
The goal is to make pylint output usable for finding regressions in the agreed upon style.
    
* Change the pylint config to disable every check that currently fails.
* Change the Makefile and github workflow to call pylint directly
   instead of the wrapper script which on fails on pylint Error level or
   higher.

Checklist


abadger:
@r0x0d Hey, so I moved pylint into precommit and ran pre-commit run --all-files locally which worked but its failing in pre-commit.ci and pre-commit.ci is truncating the output so I don't know why :-(

If you want to take a look that would be fine.  I'l look again tomorrow.... My guess is that we have to configure pre-commit to install pylint from pypi or something similar.

Makefile
```diff
@@ -72,7 +74,7 @@ tests-locally: install
 	. $(VENV)/bin/activate; pytest $(PYTEST_ARGS)
 
 lint-locally: install
-	. $(VENV)/bin/activate; ./scripts/run_lint.sh
```
SpyTec, line 75: Do we want to remove this file or is it use elsewhere?

r0x0d: I would vote for removing it and rely only on one place

r0x0d:
Alternatively, @abadger, we could try https://pre-commit.ci/lite.html. I think using the pre-commit as an action would output the whole logs for us, although, we lost the speed that pre-commit.ci brings.

SpyTec:
Fixed the error in my commit, wasn't sure about setwinsize so I just told pylint to ignore it

r0x0d:
@abadger now pre-commit.ci is passing again!

Makefile
abadger: The lint-errors Makefile target uses it but we could get rid of that target. (The target does what the old CI test did: fail if pylint flags anything as an error)

tests/integration/tier1/checks-after-conversion/test_enabled_repositories.py
```diff
@@ -1,7 +1,6 @@
+import os
```
abadger, line 1:
```suggestion
import os
import os.path
```

abadger: It's actually an implementation detail that allows `os.path` to work if you only do `import os` so I like to be explicit here.

SpyTec: Problem is that pylint complains about `os.path` not working if we do this

SpyTec: Oh I see you import both `os` and `os.path`. What is the difference here? Just to be verbose and say that we are only using path from os?

abadger: I should remove the `import os`, I thought that something like `os.access()`  must have been used in the file and so pylint would complain about that but I see that only `os.path.exists()` is specified so `import os.path` should be enough.

The pylint error is strange... It only happens for me when run under pre-commit.  If I take what pre-commit should be running for pylint and run that `pylint -sn --rcfile=.pylintrc tests/integration/tier1/checks-after-conversion/test_enabled_repositories.py`  there is no error.

abadger: Okay.  It looks like it's a bug in pylint-2.14.x/astroid-2.11.x.  If I change the pre-commit config to require `pylint==2.15.0, astroid==2.12.13` (the latest major release series for pylint and the earliest astroid version that it requires) then the pylint error (`[E0611 no-name-in-module] No name 'path' in module 'os'` ) goes away.

@r0x0d is there a reason to pin pylint to version 2.14.4?   Should I pin to 2.15.10 instead (the current latest)?

r0x0d: `2.14.4` I think that version broke the CI in the past, IIRC. But since we are moving away from it to use pre-commit, I would say it's fine to pin to that version.

Makefile
abadger: removed in the latest commit

abadger:
Okay, pre-commit ci is now passing.  The changes from last night/this morning are:
* Remove the run-lint.sh script and associated Makefile target.
* Upgrade the required pylint to fix the `import os.path` false positives.
* Set `verbose: false` for the tmt lint pre-commit hook.  This setting should mean that we only get the full output from tmt lint when the test fails.  If it passes, the pre-commit output will not include it.  (That should fix the pre-commit.ci log truncation unless both tmt lint and other hooks fail in the same commit).

requirements/centos7.requirements.txt
```diff
@@ -1,4 +1,3 @@
-pylint==1.9.5
 astroid==1.6.6
```
r0x0d, line 1: I forgot to remove `astroid` from the requirements file. I will open another PR to remove them.

r0x0d (approved):

PR #710, opened 2023-01-20
[RHELC-854] Man/help page to mention config file location

hosekadam:
Small improvement of docustring of `--config-file` option.

Jira Issue: [[RHELC-854](https://issues.redhat.com/browse/RHELC-854)]

Checklist

### DIFF
--- a/convert2rhel/toolopts.py
+++ b/convert2rhel/toolopts.py
@@ -199,10 +199,14 @@ class CLI(object):
         group.add_option(
             "-c",
             "--config-file",
-            help="A configuration file to safely provide either a user password or an activation key for registering"
-            " the system through subscription-manager. Alternatively, passing these values through the"
-            " --activationkey or --password option would leak them through a list of running processes."
-            " Example of this file in /etc/convert2rhel.ini",
+            help="The configuration file is an optional way to safely pass either a user password or an activation key"
+            " to the subscription-manager to register the system. This is more secure than passing these values"
+            " through the --activationkey or --password option, which might leak the values"
+            " through a list of running processes."
+            " You can edit the pre-installed configuration file template at /etc/convert2rhel.ini or create a new"
+            " configuration file at ~/.convert2rhel.ini. The convert2rhel utility loads the configuration from either"
+            " of those locations, the latter having preference over the former. Alternatively, you can specify a path"
+            " to the configuration file using the --config-file option to override other configurations.",
         )
         group.add_option(
             "-a",


convert2rhel/toolopts.py
```diff
@@ -202,7 +202,11 @@ def _register_options(self):
             help="A configuration file to safely provide either a user password or an activation key for registering"
             " the system through subscription-manager. Alternatively, passing these values through the"
             " --activationkey or --password option would leak them through a list of running processes."
-            " Example of this file in /etc/convert2rhel.ini",
+            " Possible locations of the file and it's priority:"
+            " 1) user specified and passed by -c, --config-file option; highest priority"
+            " 2) ~/.convert2rhel.ini; lower priority"
+            " 3) /etc/convert2rhel.ini; the lowest priority"
+            " Example of the file in /etc/convert2rhel.ini",
```
SpyTec, line 209: I think we should do in the opposite order kind of. Mention that `/etc/convert2rhel.ini` file exists and should be used, but that they can overwrite it by using either `-c|--config-file` or `~/.convert2rhel.ini`

hosekadam: @SpyTec Do you mean something like:
```
The example of the file is in /etc/convert2rhel.ini. This file can
be used in this path as the low priority one or in  ~/.convert2rhel.ini
with highest priority. Also is possible to specify different path using 
this option which has the highest priority.
```

abadger: I think Eric basically just wants the list reversed.  Something like:
```
            " Possible locations of the file and it's priority:"
            " 1) /etc/convert2rhel.ini; systemwide, default values"
            " 2) ~/.convert2rhel.ini; overrides any system defaults with user supplied values"
            " 3) user specified and passed by -c, --config-file option; highest priority, options in this file override anything in (1) and (2) "
```

convert2rhel/toolopts.py
```diff
@@ -202,7 +202,11 @@ def _register_options(self):
             help="A configuration file to safely provide either a user password or an activation key for registering"
             " the system through subscription-manager. Alternatively, passing these values through the"
```
mportman12, line 203: The sentence "Alternatively..." is somewhat confusing. Recommend rewriting as:
"This is more secure than passing these values through the --activationkey or --password option, which might leak the values through a list of running processes."

convert2rhel/toolopts.py
hosekadam: Ok, I understand. I've updated the message to your suggestion and changed the sentence alternatively to Miriam's suggestion.

@SpyTec You can re-review that if it's fine

SpyTec: Problem here is though that if I was a new user to convert2rhel reading this I'd assume I have to provide the config file always. So I'd provide `/etc/convert2rhel.ini` even though we already load from there. This should be rewritten a bit to say that the file provided here overwrites the files in etc and user dir and also explain that user dir has higher priority than etc one

hosekadam: OK, I'll try to improve that and than consult it with Miriam.

SpyTec: @hosekadam consult with whomever is easier or faster for you

convert2rhel/toolopts.py
```diff
@@ -199,10 +199,14 @@ def _register_options(self):
...
-            help="A configuration file to safely provide either a user password or an activation key for registering"
-            " the system through subscription-manager. Alternatively, passing these values through the"
-            " --activationkey or --password option would leak them through a list of running processes."
-            " Example of this file in /etc/convert2rhel.ini",
+            help="The configuration file is an optional way to safely pass either a user password or an activation key"
+            " to the subscription-manager to register the system. This is more secure than passing these values"
+            " through the --activationkey or --password option, which might leak the values"
+            " through a list of running processes."
+            " You can edit the pre-installed configuration file template at /etc/convert2rhel.ini or create a new"
+            " configuration file at ~/.convert2rhel.ini. The convert2rhel utility loads the configuration from either"
```
r0x0d, line 207: Not sure if we want to actually write instead:

```suggestion
            " configuration file at ~/.convert2rhel.ini. The Convert2RHEL utility loads the configuration from either"
```

r0x0d (approved):
Just one small comment, but otherwise, I think it's all okay. The wording there sounds clear and very well understanding from a user perspective.

convert2rhel/toolopts.py
SpyTec: That is for the product offering, we use all lowercase for the tool itself

PR #713, opened 2023-01-23
[RHELC-667] Improve package version comparison

pr-watson:
This is a WIP PR for RHELC-667. It aims to improve the current package version comparison with the addition of being able to parse NEVRA, NEVR, NVRA, NVR, ENVRA and ENVR formats. The PR will also separate the current package version comparison and parsing function into two separate functions. 

Jira Issue: [RHELC-667](https://issues.redhat.com/browse/RHELC-667)

Checklist

### DIFF
--- a/convert2rhel/checks.py
+++ b/convert2rhel/checks.py
@@ -26,7 +26,6 @@ import rpm
 from convert2rhel import __version__ as installed_convert2rhel_version
 from convert2rhel import grub, pkgmanager, utils
 from convert2rhel.pkghandler import (
-    _package_version_cmp,
     call_yum_cmd,
     compare_package_versions,
     get_installed_pkg_objects,
@@ -476,7 +475,7 @@ def get_most_recent_unique_kernel_pkgs(pkgs):
             list_of_sorted_pkgs.append(
                 max(
                     distinct_kernel_pkgs[1],
-                    key=cmp_to_key(_package_version_cmp),
+                    key=cmp_to_key(compare_package_versions),
                 )
             )
 
@@ -716,10 +715,10 @@ def is_loaded_kernel_latest():
         return
 
     packages = []
-    # We are expecting an repoquery output to be similar to this:
+    # We are expecting a repoquery output to be similar to this:
     #   C2R     1671212820      3.10.0-1160.81.1.el7    updates
     # We need the `C2R` identifier to be present on the line so we can know for
-    # sure that the line we are working with is the a line that contains
+    # sure that the line we are working with is a line that contains
     # relevant repoquery information to our check, otherwise, we just log the
     # information as debug and do nothing with it.
     for line in repoquery_output.split("\n"):
@@ -757,7 +756,11 @@ def is_loaded_kernel_latest():
 
     uname_output, _ = run_subprocess(["uname", "-r"], print_output=False)
     loaded_kernel = uname_output.rsplit(".", 1)[0]
-    match = compare_package_versions(latest_kernel, str(loaded_kernel))
+    # append the package name to loaded_kernel and latest_kernel so they can be properly processed by
+    # compare_package_versions()
+    latest_kernel_pkg = "%s-%s" % (package_to_check, latest_kernel)
+    loaded_kernel_pkg = "%s-%s" % (package_to_check, loaded_kernel)
+    match = compare_package_versions(latest_kernel_pkg, loaded_kernel_pkg)
 
     if match != 0:
         repos_message = (
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -47,6 +47,27 @@ versionlock_file = RestorableFile(_VERSIONLOCK_FILE_PATH)  # pylint: disable=C01
 # NEVR can contain more than that.  For instance: gcc-c++-4.8.5-44.0.3.el7.x86_64
 PKG_NEVR = r"\b(?:([0-9]+):)?(\S+)-(\S+)-(\S+)\b"
 
+# This regex finds if a package is in ENVR/ENVRA format by searching for the epoch field
+# being the first set of characters in the package string
+ENVRA_ENVR_FORMAT = re.compile(r"^\d:")
+
+# This regex finds if a package is in NEVRA/NEVR format by searching for any digit
+# found between a "-" and a ":"
+NEVRA_NEVR_FORMAT = re.compile(r"-\d:")
+
+# This regex ensures there are no whitespace charcters in the package name
+PKG_NAME = re.compile(r"^[^\s]+$")
+
+# This regex ensures package epoch has only 1 or more digits
+PKG_EPOCH = re.compile(r"^\d+$")
+
+# This regex ensures there are no whitespace charcters or dashes in the package version and release
+PKG_VERSION = re.compile(r"^[^\s-]+$")
+PKG_RELEASE = PKG_VERSION
+
+# Set of valid arches
+PKG_ARCH = ("x86_64", "s390x", "i86", "ppc64le", "aarch64", "noarch")
+
 # It would be better to construct this dynamically but we don't have lru_cache
 # in Python-2.6 and modifying a global after your program initializes isn't a
 # good idea.
@@ -76,13 +97,11 @@ def call_yum_cmd(
     set_releasever=True,
 ):
     """Call yum command and optionally print its output.
-
     The enable_repos and disable_repos function parameters accept lists and they override the default use of repos,
     which is:
     * --disablerepo yum option = "*" by default OR passed through a CLI option by the user
     * --enablerepo yum option = is the repo enabled through subscription-manager based on a convert2rhel configuration
       file for the particular system OR passed through a CLI option by the user
-
     YUM/DNF typically expands the $releasever variable used in repofiles. However it fails to do so after we remove the
     release packages (centos-release, oraclelinux-release, etc.) and before the redhat-release package is installed.
     By default, for the above reason, we provide the --releasever option to each yum call. However before we remove the
@@ -185,7 +204,6 @@ def get_problematic_pkgs(output, excluded_pkgs=frozenset()):
 def find_pkg_names(output, message_key="%s"):
     """
     Find all the package names of a "type" from amongst a string of output from yum.
-
     :arg output: The yum output to parse for package names
     :arg message_key: This function tries to retrieve "types" of packages from the yum output.
         Packages that have multilib problems or dependency errors for instance.
@@ -489,7 +507,6 @@ def print_pkg_info(pkgs):
 
 def get_pkg_nvra(pkg_obj):
     """Get package NVRA as a string: name, version, release, architecture.
-
     Some utilities don't accept the full NEVRA of a package, for example rpm.
     """
     return "%s-%s-%s.%s" % (
@@ -502,7 +519,6 @@ def get_pkg_nvra(pkg_obj):
 
 def get_pkg_nevra(pkg_obj):
     """Get package NEVRA as a string: name, epoch, version, release, architecture.
-
     Epoch is included only when non-zero. However it's on a different position when printed by YUM or DNF:
       YUM - epoch before name: "7:oraclelinux-release-7.9-1.0.9.el7.x86_64"
       DNF - epoch before version: "oraclelinux-release-8:8.2-1.0.8.el8.x86_64"
@@ -527,7 +543,6 @@ def get_pkg_nevra(pkg_obj):
 
 def get_packager(pkg_obj):
     """Get the package packager from the yum/dnf package object.
-
     The packager may not be set for all packages. E.g. Oracle Linux packages have the packager info empty.
     """
     packager = pkg_obj.packager if pkg_obj.packager else "N/A"
@@ -540,7 +555,6 @@ def get_packager(pkg_obj):
 
 def get_vendor(pkg_obj):
     """Get the package vendor from the yum/dnf package object.
-
     The vendor information is provided by the yum/dnf python API on all systems except systems derived from
     RHEL 8.0-8.3 (see bug https://bugzilla.redhat.com/show_bug.cgi?id=1876561).
     """
@@ -574,7 +588,6 @@ def remove_excluded_pkgs():
 def remove_repofile_pkgs():
     """Remove those non-RHEL packages that contain YUM/DNF repofiles (/etc/yum.repos.d/*.repo) or affect variables
     in the repofiles (e.g. $releasever).
-
     Red Hat cannot automatically remove these non-RHEL packages with other excluded packages. While other excluded
     packages must be removed before installing subscription-manager to prevent package conflicts, these non-RHEL
     packages must be present on the system during subscription-manager installation so that the system can access and
@@ -609,7 +622,6 @@ def remove_pkgs_with_confirm(pkgs, backup=True):
 
 def get_system_packages_for_replacement():
     """Get a list of packages in the system to be replaced.
-
     This function will return a list of packages installed on the system by
     using the `system_info.fingerprint_ori_os` signature.
 
@@ -829,12 +841,10 @@ def fix_invalid_grub2_entries():
     """
     On systems derived from RHEL 8 and later, /etc/machine-id is being used to identify grub2 boot loader entries per
     the Boot Loader Specification.
-
     However, at the time of executing convert2rhel, the current machine-id can be different from the machine-id from the
     time when the kernels were installed. If that happens:
     - convert2rhel installs the RHEL kernel, but it's not set as default
     - convert2rhel removes the original OS kernels, but for these the boot entries are not removed
-
     The solution handled by this function is to remove the non-functioning boot entries upon the removal of the original
     OS kernels, and set the RHEL kernel as default.
     """
@@ -926,10 +936,8 @@ def has_duplicate_repos_across_disablerepo_enablerepo_options():
 
 def filter_installed_pkgs(pkg_names):
     """Check if a package is present on the system based on a list of package names.
-
     This function aims to act as a filter for a list of pkg_names to return wether or not a package is present on the
     system.
-
     :param pkg_names: List of package names
     :type pkg_names: list[str]
     :return: A list of packages that are present on the system.
@@ -947,7 +955,6 @@ def filter_installed_pkgs(pkg_names):
 
 def get_pkg_names_from_rpm_paths(rpm_paths):
     """Return names of packages represented by locally stored rpm packages.
-
     :param rpm_paths: List of rpm with filepaths.
     :type rpm_paths: list[str]
     :return: A list of package names extracted from the rpm filepath.
@@ -959,51 +966,15 @@ def get_pkg_names_from_rpm_paths(rpm_paths):
     return pkg_names
 
 
-def compare_package_versions(version1, version2):
-    """Compare two EVR versions against each other.
-
-
-    :param version1: The version to be compared.
-    :type version1: str
-    :param version2: The version to compare against.
-    :type version2: str
-
-    :example:
-
-        >>> match = compare_package_versions("5.14.10-300.fc35", "5.14.15-300.fc35")
-        >>> match # -1
-
-    .. note::
-
-        Since the return type is a int, this could be difficult to understand
-        the meaning of each number, so here is a quick list of the meaning from
-        every possible number:
-            * -1 if the evr1 is less then evr2 version
-            * 0 if the evr1 is equal evr2 version
-            * 1 if the evr1 is greater than evr2 version
-
-    :return: Return a number indicating if the versions match, are less or greater then.
-    :rtype: int
-    """
-
-    evr1 = utils.string_to_version(version1)
-    evr2 = utils.string_to_version(version2)
-
-    return rpm.labelCompare(evr1, evr2)
-
-
 def get_total_packages_to_update(reposdir):
     """Return the total number of packages to update in the system
-
     It uses both yum/dnf depending on whether they are installed on the system,
     In case of RHEL 7 derivative distributions, it uses `yum`, otherwise it uses `dnf`.
 
     To check whether the system is updated or not, we use original vendor repofiles which we ship within the
     convert2rhel RPM. The reason is that we can't rely on the repofiles available on the to-be-converted system.
-
     :param reposdir: The path to the hardcoded repositories for EUS (If any).
     :type reposdir: str | None
-
     :return: The packages that need to be updated.
     :rtype: list[str]
     """
@@ -1032,7 +1003,6 @@ def _get_packages_to_update_yum():
 def _get_packages_to_update_dnf(reposdir):
     """Query all the packages with dnf that has an update pending on the
     system.
-
     :param reposdir: The path to the hardcoded repositories for EUS (If any).
     :type reposdir: str | None
     """
@@ -1070,52 +1040,208 @@ def _get_packages_to_update_dnf(reposdir):
     return packages
 
 
-def _package_version_cmp(pkg_1, pkg_2):
-    """Compare the version key in a given package name.
-
-    Consider the following variables that will be passed to this function::
-
-        pkg_1 = 'kernel-core-0:4.18.0-240.10.1.el8_3.x86_64'
-        pkg_2 = 'kernel-core-0:4.18.0-239.0.0.el8_3.x86_64'
-
-    The output of this will be a tuple containing the package version in a
-    tuple::
-
-        result = _package_version_cmp(pkg_1, pkg_2)
-        print("Result is: %s" % result)
-        # Result is: -1
-
-    The function will ignore the package name as it is not an important
-    information here and will only care about the version that is tied to it's
-    name.
-
-    :param pkg_1: The first package to extract the version
-    :type pkg_1: str
-    :param pkg_2: The second package to extract the version
-    :type pkg_2: str
-    :return: An integer resulting in the package comparision
+def compare_package_versions(version1, version2):
+    """Compare two package versions against each other, including name and arch.
+    This function will receive packages in any of the following formats:
+        * ENVR
+        * ENVRA
+        * NVR
+        * NEVR
+        * NVRA
+        * NEVRA
+    :param version1: The version to be compared.
+    :type version1: str
+    :param version2: The version to compare against.
+    :type version2: str
+    .. example::
+        >>> match = compare_package_versions("kernel-core-5.14.10-300.fc35", "kernel-core-5.14.15-300.fc35")
+        >>> print(match) # -1
+    .. note::
+        Since the return type is a int, this could be difficult to understand
+        the meaning of each number, so here is a list that represents
+        every possible number:
+            * -1 if the version1 is less then version2 version
+            * 0 if the version1 is equal version2 version
+            * 1 if the version1 is greater than version2 version
+    :return: Return a number indicating if the versions match, are less or greater then.
     :rtype: int
     """
+    # call parse_pkg_string to obtain a list containing name, epoch, release, version and arch fields
+    # order of fields returned: name, epoch, version, release, arch
+    version1_components = parse_pkg_string(version1)
+    version2_components = parse_pkg_string(version2)
 
-    # TODO(r0x0d): This function still needs some enhancements code-wise, it
-    # workes perfectly, but the way we are handling the versions is not 100%
-    # complete yet.  will be done in a future work. Right now, all the list of
-    # changes are listed in this comment:
-    # https://github.com/oamg/convert2rhel/pull/469#discussion_r873971400
-    pkg_ver_components = []
-    for pkg in pkg_1, pkg_2:
-        # Remove the package name and split the rest between epoch + version
-        # and release + arch
-        epoch_version, release_arch = pkg.rsplit("-", 2)[-2:]
-        # Separate the (optional) epoch from the version
-        epoch_version = epoch_version.split(":", 1)
-        if len(epoch_version) == 1:
-            epoch = "0"
-            version = epoch_version[0]
+    # ensure package names match, error if not
+    if version1_components[0] != version2_components[0]:
+        raise ValueError(
+            "The package names ('%s' and '%s') do not match. Can only compare versions for the same packages."
+            % (version1_components[0], version2_components[0])
+        )
+
+    # ensure package arches match, error if not
+    if version1_components[4] != version2_components[4] and all(([version1_components[4]], version2_components[4])):
+        raise ValueError(
+            "The arches ('%s' and '%s') do not match. Can only compare versions for the same arches."
+            % (version1_components[4], version2_components[4])
+        )
+
+    # create list containing EVR for comparison
+    evr1 = (version1_components[1], version1_components[2], version1_components[3])
+    evr2 = (version2_components[1], version2_components[2], version2_components[3])
+
+    return rpm.labelCompare(evr1, evr2)
+
+
+def parse_pkg_string(pkg):
+    """
+    This function takes a version string in NEVRA, NEVR, NVRA, NVR, ENVRA, ENVR and decides whether
+    to parse with a yum/dnf module based on the package manager type of the system.
+    :param pkg: The package to be parsed.
+    :type pkg: str
+    :return: Return a Return a list containing name, epoch, version, release, arch
+    :rtype: list[str | None]
+    """
+    if pkgmanager.TYPE == "yum":
+        pkg_ver_components = _parse_pkg_with_yum(pkg)
+    else:
+        pkg_ver_components = _parse_pkg_with_dnf(pkg)
+
+    _validate_parsed_fields(*pkg_ver_components)
+
+    return pkg_ver_components
+
+
+def _validate_parsed_fields(name, epoch, version, release, arch):
+    """
+    Validation for each field contained in pkg_ver_components from the
+    package parsing functions. If one of the fields are invalid then a ValueError is raised
+    :param name: parsed package name
+    :type name: str
+    :param epoch: parsed package epoch
+    :type name: str
+    :param version: parsed package version
+    :type name: str
+    :param release: parsed package release
+    :type name: str
+    :param arch: parsed package arch
+    :type name: str
+    """
+
+    errors = []
+
+    if name is None or not PKG_NAME.match(name):
+        errors.append("name : %s" % name if name else "name : [None]")
+    if epoch is not None and not PKG_EPOCH.match(epoch):
+        errors.append("epoch : %s" % epoch)
+    if version is None or not PKG_VERSION.match(version):
+        errors.append("version : %s" % version if version else "version : [None]")
+    if release is None or not PKG_RELEASE.match(release):
+        errors.append("release : %s" % release if release else "release : [None]")
+    if arch is not None and arch not in PKG_ARCH:
+        errors.append("arch : %s" % arch)
+
+    if errors:
+        raise ValueError("The following field(s) are invalid - %s" % ", ".join(errors))
+
+
+def _parse_pkg_with_yum(pkg):
+    """Parse verison string using yum and rpmUtils splitFilename
+    :param pkg: The package to be parsed.
+    :type pkg: str
+    :return: Return a list containing name, epoch, version, release, arch (may contain null values)
+    :rtype: list[str]
+    """
+
+    # package is in NEVRA/NEVR format
+    if NEVRA_NEVR_FORMAT.findall(pkg):
+        name, epoch_version, release_arch = pkg.rsplit("-", 2)
+        epoch, version = epoch_version.split(":", 1)
+        # package is in NEVRA format
+        if release_arch.endswith(PKG_ARCH):
+            release, arch = release_arch.rsplit(".", 1)
+        # package is in NEVR format
         else:
-            epoch, version = epoch_version
-        # Discard the arch
-        release = release_arch.rsplit(".", 1)[0]
-        pkg_ver_components.append((epoch, version, release))
+            release = release_arch
+            arch = None
 
-    return rpm.labelCompare(pkg_ver_components[0], pkg_ver_components[1])
+    # package is in either ENVR, ENVRA, NVR or NVRA
+    else:
+        # splitFilename doesn't work with packages in NEVRA/NEVR format, that's why we are using it only here.
+        (name, version, release, epoch, arch) = pkgmanager.splitFilename(pkg)
+
+        # splitFilename places part of the release in the arch field when no arch is present
+        # if arch field is invalid, append contents to release field
+        if arch not in PKG_ARCH:
+            temp_release = arch
+            arch = None
+            release = "%s.%s" % (release, temp_release)
+
+    # convert any empty strings to None for consistency
+    pkg_ver_components = tuple((i or None) for i in (name, epoch, version, release, arch))
+
+    return pkg_ver_components
+
+
+def _parse_pkg_with_dnf(pkg):
+    """
+    Parse verison string using hawkey and dnf.
+    :param pkg: The package to be parsed.
+    :type pkg: str
+    :return: Return a list containing name, epoch, version, release, arch (may contain null values)
+    :rtype: list[str]
+    """
+
+    name = epoch = version = release = arch = None
+    no_arch_data = None
+
+    # if format is ENVR/ENVRA, store and remove epoch and evaluate as NVR/NVRA
+    if ENVRA_ENVR_FORMAT.findall(pkg):
+        pkg = pkg.split(":")
+        epoch = pkg[0]
+        pkg = pkg[1]
+
+    # returns generator for every possible nevra
+    subject = pkgmanager.dnf.subject.Subject(pkg)
+    possible_nevra = subject.get_nevra_possibilities(forms=[pkgmanager.hawkey.FORM_NEVRA, pkgmanager.hawkey.FORM_NEVR])
+
+    # loop through each possible set of nevra fields and select the valid one
+    for nevra in possible_nevra:
+
+        # current arch is valid
+        if str(nevra.arch) in PKG_ARCH:
+            name = nevra.name
+            epoch = epoch or nevra.epoch
+            version = nevra.version
+            release = nevra.release
+            arch = nevra.arch
+            break
+
+        # there is no arch present and current arch is None
+        if nevra.arch is None:
+            no_arch_data = nevra
+
+        # arch is not valid, move on to next iteration
+
+    else:  # This else goes with the for loop
+
+        # if no_arch_data is still None by this point, the parser wasn't able to find valid fields
+        # therefore the package entered is invalid and/or in the wrong format
+        if no_arch_data is None:
+            raise ValueError(
+                "Invalid package string - %s, enter a package in one of the formats: NEVRA, NEVR, NVRA, NVR, ENVRA, ENVR."
+                % pkg
+            )
+
+        name = no_arch_data.name
+        epoch = epoch or no_arch_data.epoch
+        version = no_arch_data.version
+        release = no_arch_data.release
+        arch = no_arch_data.arch
+
+    # convert epoch from integer to string for consistency
+    if epoch is not None:
+        epoch = str(epoch)
+
+    pkg_ver_components = (name, epoch, version, release, arch)
+
+    return pkg_ver_components
--- a/convert2rhel/pkgmanager/__init__.py
+++ b/convert2rhel/pkgmanager/__init__.py
@@ -43,7 +43,6 @@ except ImportError as e:
     from dnf import *  # pylint: disable=import-error
     from dnf.callback import Depsolve, DownloadProgress
 
-
     # This is added here to prevent a generic try-except in the
     # `check_package_updates()` function.
     from dnf.exceptions import RepoError
--- a/convert2rhel/pkgmanager/__init__.py
+++ b/convert2rhel/pkgmanager/__init__.py
@@ -23,6 +23,8 @@ from convert2rhel import utils
 loggerinst = logging.getLogger(__name__)
 
 try:
+    # this is used in pkghandler.py to parse version strings in the _parse_pkg_with_yum function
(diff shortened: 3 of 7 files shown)


pr-watson:
This patch separates the previous package comparison and parsing function into two functions - one for parsing and one for package comparison. There are two parsing functions _parse_pkg_with_dnf() and _parse_pkg_with_yum() to handle parsing on both yum and dnf systems. They both accept a package as an argument and return a list with the parsed elements: name, epoch, version, release and arch, with epoch and arch being optional. 

The updated compare_package_versions() function now errors when either the arch or name are not equal. It also doesn't rely on utils.string_to_version() to prepare EVR for comparison using rpm.labelCompare(). Unit tests for checks.py were modified to accommodate the name field being present in the updated compare_package_versions() function.

convert2rhel/pkgmanager/__init__.py
```diff
@@ -16,14 +16,21 @@
...
+    from rpmUtils.miscutils import splitFilename
     from yum import *
 
     # This is added here to prevent a generic try-except in the
     # `check_package_updates()` function.
     from yum.Errors import RepoError  # lgtm[py/unused-import]
 
     TYPE = "yum"
+
+# WARNING: if there is a bug in the yum import section, we might try to import dnf incorrectly
```
r0x0d, line 29: Not a bug, per say, more like "try to import dnf if we fail to import yum on RHEL8+, which is likely to happen"

pr-watson: Wrote this comment because during development I had an incorrect yum import which caused dnf to get imported when we really wanted yum.

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,149 @@ def clean_yum_metadata():
...
+        if nevra.arch is None:
+            no_arch_data = nevra
+
+        # arch is not valid, move on to next iteration
         else:
-            epoch, version = epoch_version
-        # Discard the arch
-        release = release_arch.rsplit(".", 1)[0]
-        pkg_ver_components.append((epoch, version, release))
+            continue
```
abadger, line 1248: If I'm reading the logic correctly, this block (
``` python
        else:
            continue
```

is unnecessary?  (because this is the last statement in the loop so it will continue to the next iteration without this).

pr-watson: Good catch @abadger . I'll remove this block, thanks!

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
     loggerinst.info("Cached yum metadata cleaned successfully.")
 
 
-def _package_version_cmp(pkg_1, pkg_2):
-    """Compare the version key in a given package name.
+def compare_package_versions(version1, version2):
+    """Compare two EVR versions against each other, as well as
+    the name and architectures
```
r0x0d, line 1116: The deal of this is that we are not only comparing `EVR`, right? We are comparing more types as well, even though that seems "invisible" to the caller, correct, @pr-watson?

If so, I would change this description a bit to: 

```suggestion
    """Compare two package versions against each other, including name and arch.

    This function will receive packages in any of the following formats:
        * EVR
        * ENVR
        * ENVRA
        * NVR
        * NEVR
        * NVRA
        * NEVRA
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
-        pkg_1 = 'kernel-core-0:4.18.0-240.10.1.el8_3.x86_64'
-        pkg_2 = 'kernel-core-0:4.18.0-239.0.0.el8_3.x86_64'
+    :param version1: The version to be compared.
+    :type version1: str
+    :param version2: The version to compare against.
+    :type version2: str
 
-    The output of this will be a tuple containing the package version in a
-    tuple::
+    :example:
```
r0x0d, line 1124:
```suggestion
    .. example::
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
 
-        result = _package_version_cmp(pkg_1, pkg_2)
-        print("Result is: %s" % result)
-        # Result is: -1
+        >>> match = compare_package_versions("kernel-core-5.14.10-300.fc35", "kernel-core-5.14.15-300.fc35")
+        >>> match # -1
```
r0x0d, line 1127: This is more like an "accept if you like" type of suggestion, but, I always felt that if you include a `print` statement in the example, it would make much more sense. 

```suggestion
        >>> print(match) # -1
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+    the name and architectures
 
-    Consider the following variables that will be passed to this function::
 
-        pkg_1 = 'kernel-core-0:4.18.0-240.10.1.el8_3.x86_64'
-        pkg_2 = 'kernel-core-0:4.18.0-239.0.0.el8_3.x86_64'
+    :param version1: The version to be compared.
+    :type version1: str
+    :param version2: The version to compare against.
+    :type version2: str
```
r0x0d, line 1122: Should we change those parameters to be `package1` and `package2` instead of `version1` and `version2`? I mean, we are passing full package names here, not only versions to be compared. What do you think, @pr-watson?

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+    version1_components = parse_pkg_string(version1)
+    version2_components = parse_pkg_string(version2)
+
+    # ensure package names match, error if not
+    if version1_components[0] != version2_components[0]:
+        raise ValueError(("%s does not match %s" % (version1_components[0], version2_components[0])))
+
+    # ensure package arches match, error if not
+    if version1_components[4] != version2_components[4] and all(([version1_components[4]], version2_components[4])):
+        raise ValueError(("%s does not match %s" % (version1_components[4], version2_components[4])))
```
r0x0d, line 1152: I think you don't need the extra parenthesis there for both exceptions. 

```suggestion
        raise ValueError("%s does not match %s" % (version1_components[0], version2_components[0]))

    # ensure package arches match, error if not
    if version1_components[4] != version2_components[4] and all(([version1_components[4]], version2_components[4])):
        raise ValueError("%s does not match %s" % (version1_components[4], version2_components[4]))
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+
+    # create list containing EVR for comparison
+    evr1 = (version1_components[1], version1_components[2], version1_components[3])
+    evr2 = (version2_components[1], version2_components[2], version2_components[3])
+
+    return rpm.labelCompare(evr1, evr2)
+
+    # rename prev use of compare_package_versions to compare_package_strings
```
r0x0d, line 1160: is this a TODO comment?

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+
+
+def parse_pkg_string(pkg):
```
r0x0d, line 1163: Just a question to get something out of the way. Do you see this function being used in other places as well, and not only `compare_package_versions`? If so, then we can leave it as a public function, otherwise, I would vote to make it internal to the module. 

I believe that your intention here is to make a generic function that could be used elsewhere, correct, @pr-watson?

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+    """
+    This function takes a version string in NEVRA, NEVR, NVRA, NVR, ENVRA, ENVR and decides whether
+    to parse with a yum/dnf module based on the package manager type of the system
+    """
```
r0x0d, line 1167: Please, document the parameter and the return for this function in this docstring.

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+    if pkgmanager.TYPE == "yum":
+        pkg_ver_components = _parse_pkg_with_yum(pkg)
+    else:
+        pkg_ver_components = _parse_pkg_with_dnf(pkg)
+    return pkg_ver_components
```
r0x0d, line 1172:
```suggestion
        pkg_ver_components = _parse_pkg_with_dnf(pkg)
        
    return pkg_ver_components
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
-        # Separate the (optional) epoch from the version
-        epoch_version = epoch_version.split(":", 1)
-        if len(epoch_version) == 1:
-            epoch = "0"
-            version = epoch_version[0]
+def _parse_pkg_with_yum(pkg):
+    """
+    Parse verison string using yum and rpmUtils splitFilename
+    :return: Return a list containing name, epoch, version, release, arch
+    """
```
r0x0d, line 1179: Can you add the parameter to the docstring as well? 

```suggestion
    """Parse verison string using yum and rpmUtils splitFilename
    
    :return: Return a list containing name, epoch, version, release, arch
    :rtype: list[str | None]
    """
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+
+    # package is in NEVRA/NEVR format
+    if re.findall(r"-\d:", pkg):
+        name, epoch_version, release_arch = pkg.rsplit("-", 2)
+        epoch, version = epoch_version.split(":", 1)
+        # if NEVRA format
```
r0x0d, line 1185:
```suggestion
        # package is in NEVRA format
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+        if re.findall(r".*(x86_64|s390x|i86|ppc64le|aarch64)$", release_arch):
+            release, arch = release_arch.rsplit(".", 1)
+        # if NEVR format
         else:
```
r0x0d, line 1189:
```suggestion
        # package is in NEVR format
        else:
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
-            epoch, version = epoch_version
-        # Discard the arch
-        release = release_arch.rsplit(".", 1)[0]
-        pkg_ver_components.append((epoch, version, release))
+            release = release_arch
+            arch = None
+
+    # package is in ENVR, ENVRA, NVR, NVRA
```
r0x0d, line 1193:
```suggestion
    # package is in either ENVR, ENVRA, NVR or NVRA
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+    else:
+        (name, version, release, epoch, arch) = pkgmanager.splitFilename(pkg)
```
r0x0d, line 1195: This is something that I would do, but that doesn't mean that you need to follow this suggestion, but, I would add a comment above this line mentioning why I'm using `pkgmanager.splitFilename` here and not on the first `if` statement` something like: 

```suggestion
        # splitFilename doesn't work with packages in NEVRA/NEVR format, that's why we are using it only here.
        (name, version, release, epoch, arch) = pkgmanager.splitFilename(pkg)
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+        # splitFilename places part of the release in the arch field when no arch is present
+        # if arch field is invalid, append contents to release field
+        if arch not in ("x86_64", "s390x", "i86", "ppc64le", "aarch64"):
+            temp_release = arch
+            arch = None
+            release = release + "." + temp_release
+
+    pkg_ver_components = [name, epoch, version, release, arch]
+    # convert any empty strings to None for consistency
+    pkg_ver_components = [str(i or None) for i in pkg_ver_components]
```
r0x0d, line 1206: You could do that as well, without the need to shadow and override the variable. 

```suggestion
    # convert any empty strings to None for consistency
    pkg_ver_components = [str(i or None) for i in (name, epoch, version, release, arch)]
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+
+    return pkg_ver_components
+
+
+def _parse_pkg_with_dnf(pkg):
+    """
+    Parse verison string using hawkey and dnf
+    :return: Return a list containing name, epoch, version, release, arch
```
r0x0d, line 1214: Same here as I suggested in the `yum` parse function. 

```suggestion
    Parse verison string using hawkey and dnf
    :return: Return a list containing name, epoch, version, release, arch
    :rtype: list[str | None]
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+    """
+
+    name = epoch = version = release = arch = None
+    no_arch_data = None
```
r0x0d, line 1218: I'm not much a fan myself of this style of declaring variables, but, you could just place the `no_arch_data` on the same line as the others. 

```suggestion
    name = epoch = version = release = arch, no_arch_data = None
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,147 @@ def clean_yum_metadata():
...
+        pkg = pkg.split(":")
+        epoch = pkg[0]
+        pkg = pkg[1]
+
+    # returns generator for every possible nevra
+    subject = pkgmanager.dnf.subject.Subject(pkg)
+    possible_nevra = subject.get_nevra_possibilities(forms=[pkgmanager.hawkey.FORM_NEVRA, pkgmanager.hawkey.FORM_NEVR])
+
+    # loop through each possible set of nevra fields and select the valid one
+    for i, nevra in enumerate(possible_nevra):
```
r0x0d, line 1231: The `i` is not being used 
```suggestion
    for _, nevra in enumerate(possible_nevra):
```

convert2rhel/pkgmanager/__init__.py
r0x0d: Ah, I see. Now it makes sense to me. Will close the conversation as I think this is a valid comment.

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1206,10 +1213,10 @@ def test_is_loaded_kernel_latest_skip_on_not_latest_ol(pretend_os, monkeypatch,
...
         (
-            "1634146676\t3.10.0-1160.45.1.el7\tbaseos",
+            "1634146676\tkernel-3.10.0-1160.45.1.el7\tbaseos",
             "3.10.0-1160.45.1.el7.x86_64",
             0,
-            6,
+            7,
             "kernel",
             False,
             "The currently loaded kernel is at the latest version.",
```
r0x0d, line 1222: We already have a major version `7` on the parameter above, I think it's safe for you to remove this altogether.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1411,21 +1411,90 @@ def test_call_yum_cmd_w_downgrades(monkeypatch, retcode, output):
...
-            "4.6~pre16262021g84ef6bd9-3.fc35",
+            "kernel-core-0:4.6~pre16262021g84ef6bd9-3.fc35",
+            "kernel-core-0:4.6~pre16262021g84ef6bd9-3.fc35",
             0,
         ),
-        ("2:8.2.3568-1.fc35", "2:8.2.3568-1.fc35", 0),
+        ("kernel-core-2:8.2.3568-1.fc35", "kernel-core-2:8.2.3568-1.fc35", 0),
     ),
 )
 def test_compare_package_versions(version1, version2, expected):
```
r0x0d, line 1426: @pr-watson do you think we can include one of each format here? I think we can do some work to make this test more understandable too, like using `pytest.param` for example.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1411,21 +1411,90 @@ def test_call_yum_cmd_w_downgrades(monkeypatch, retcode, output):
...
 
+@pytest.mark.parametrize(
+    ("version1", "version2", "exception_message"),
+    (
+        ("kernel-core-0:390-287.fc36", "kernel-0:390-287.fc36", "kernel-core does not match kernel"),
+        ("kernel-core-0:390-287.fc36.aarch64", "kernel-core-0:391-287.fc36.i86", "aarch64 does not match i86"),
+    ),
+)
+def test_compare_package_versions_warnings(version1, version2, exception_message):
+
```
r0x0d, line 1438:
```suggestion
```

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1411,21 +1411,90 @@ def test_call_yum_cmd_w_downgrades(monkeypatch, retcode, output):
...
+    with pytest.raises(ValueError, match=exception_message):
+        pkghandler.compare_package_versions(version1, version2)
+
+
+def test_parse_pkg_string_dnf(monkeypatch):
+
+    pkg = "kernel-core-0:390-287.fc36.aarch64"
```
r0x0d, line 1445: I feel that this could be converted to a parametrization, and you could include all possible formats too.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1411,21 +1411,90 @@ def test_call_yum_cmd_w_downgrades(monkeypatch, retcode, output):
...
+    parse_pkg_with_dnf_mock = mock.Mock()
+    monkeypatch.setattr(pkghandler, "_parse_pkg_with_dnf", value=parse_pkg_with_dnf_mock)
+    monkeypatch.setattr(pkgmanager, "TYPE", "dnf")
+    pkghandler.parse_pkg_string(pkg)
+    parse_pkg_with_dnf_mock.assert_called_once()
+
+
+def test_parse_pkg_string_yum(monkeypatch):
+
+    pkg = "kernel-core-0:390-287.fc36.aarch64"
```
r0x0d, line 1455: I feel that this could be converted to a parametrization, and you could include all possible formats too.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1411,21 +1411,90 @@ def test_call_yum_cmd_w_downgrades(monkeypatch, retcode, output):
...
+    with pytest.raises(ValueError, match=exception_message):
+        pkghandler.compare_package_versions(version1, version2)
+
+
+def test_parse_pkg_string_dnf(monkeypatch):
+
+    pkg = "kernel-core-0:390-287.fc36.aarch64"
+    parse_pkg_with_dnf_mock = mock.Mock()
+    monkeypatch.setattr(pkghandler, "_parse_pkg_with_dnf", value=parse_pkg_with_dnf_mock)
+    monkeypatch.setattr(pkgmanager, "TYPE", "dnf")
```
r0x0d, line 1448: Have you tried running this without mock and only running those functions under the correct system? I think that would be more valuable here. 

As far as I can tell, we don't reach the database, or internet, or anything that could hang the system. It's just a string parse with the libraries... I would recommend skipping those tests if the package manager is not the one expected here and remove the mocks, letting the function run normally and giving you a real example of what is the output.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1411,21 +1411,90 @@ def test_call_yum_cmd_w_downgrades(monkeypatch, retcode, output):
...
+    pkghandler.parse_pkg_string(pkg)
+    parse_pkg_with_dnf_mock.assert_called_once()
+
+
+def test_parse_pkg_string_yum(monkeypatch):
+
+    pkg = "kernel-core-0:390-287.fc36.aarch64"
+    parse_pkg_with_yum_mock = mock.Mock()
+    monkeypatch.setattr(pkghandler, "_parse_pkg_with_yum", value=parse_pkg_with_yum_mock)
+    monkeypatch.setattr(pkgmanager, "TYPE", "yum")
```
r0x0d, line 1458: The same goes to this one as I commented earlier.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1411,21 +1411,90 @@ def test_call_yum_cmd_w_downgrades(monkeypatch, retcode, output):
...
+    assert pkghandler._parse_pkg_with_yum(package) == expected
+
+
+@pytest.mark.skipif(pkgmanager.TYPE == "yum", reason="cannot test dnf backend if dnf is not present")
+@pytest.mark.parametrize(
+    ("package", "expected"),
+    (test_packages),
+)
+def test_parse_pkg_with_dnf(package, expected):
+    assert pkghandler._parse_pkg_with_dnf(package) == expected
```
r0x0d, line 1495: You can transform that variable to be in all caps. Another thing is, I don't think this name is descriptive enough for what it's holding... But still, I can't figure out what name would suit it better here. 

```suggestion
TEST_PACKAGES = (
    ("kernel-core-0:4.18.0-240.10.1.el8_3.i86", ["kernel-core", "0", "4.18.0", "240.10.1.el8_3", "i86"]),  # NEVRA
    ("kernel-core-0:4.18.0-240.10.1.el8_3", ["kernel-core", "0", "4.18.0", "240.10.1.el8_3", "None"]),  # NEVR
    (
        "1:NetworkManager-1.18.8-2.0.1.el7_9.aarch64",
        ["NetworkManager", "1", "1.18.8", "2.0.1.el7_9", "aarch64"],
    ),  # ENVRA
    ("1:NetworkManager-1.18.8-2.0.1.el7_9", ["NetworkManager", "1", "1.18.8", "2.0.1.el7_9", "None"]),  # ENVR
    (
        "NetworkManager-1.18.8-2.0.1.el7_9.aarch64",
        ["NetworkManager", "None", "1.18.8", "2.0.1.el7_9", "aarch64"],
    ),  # NVRA
    ("NetworkManager-1.18.8-2.0.1.el7_9", ["NetworkManager", "None", "1.18.8", "2.0.1.el7_9", "None"]),  # NVR
)


@pytest.mark.skipif(pkgmanager.TYPE == "dnf", reason="cannot test yum backend if yum is not present")
@pytest.mark.parametrize(
    ("package", "expected"),
    (TEST_PACKAGES),
)
def test_parse_pkg_with_yum(package, expected):
    assert pkghandler._parse_pkg_with_yum(package) == expected


@pytest.mark.skipif(pkgmanager.TYPE == "yum", reason="cannot test dnf backend if dnf is not present")
@pytest.mark.parametrize(
    ("package", "expected"),
    (TEST_PACKAGES),
)
def test_parse_pkg_with_dnf(package, expected):
    assert pkghandler._parse_pkg_with_dnf(package) == expected
```

r0x0d (changes requested):
It looks wonderful, @pr-watson! Thanks for that hard work with this PR! 

I just left a few comments for you (sorry, hehe), but I think they are all fine to fix and go through. Let me know if you need help understanding them or want to go over them together.

convert2rhel/checks.py
```diff
@@ -712,7 +711,13 @@ def is_loaded_kernel_latest():
...
         return
 
-    cmd = ["repoquery", "--setopt=exclude=", "--quiet", "--qf", "%{BUILDTIME}\\t%{VERSION}-%{RELEASE}\\t%{REPOID}"]
+    cmd = [
+        "repoquery",
+        "--setopt=exclude=",
+        "--quiet",
+        "--qf",
+        "%{BUILDTIME}\\t%{NAME}-%{VERSION}-%{RELEASE}\\t%{REPOID}",
+    ]
```
r0x0d, line 720: Note for future @pr-watson: This will (possibly) conflict if https://github.com/oamg/convert2rhel/pull/687 get merged first.

convert2rhel/checks.py
```diff
@@ -769,7 +774,8 @@ def is_loaded_kernel_latest():
         # The loaded kernel version
         uname_output, _ = run_subprocess(["uname", "-r"], print_output=False)
         loaded_kernel = uname_output.rsplit(".", 1)[0]
-        match = compare_package_versions(latest_kernel, str(loaded_kernel))
+        loaded_kernel = package_to_check + "-" + loaded_kernel
```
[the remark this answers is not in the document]
pr-watson: I agree, let me make this change

convert2rhel/pkghandler.py
pr-watson: Looks better, i'll add this

convert2rhel/pkghandler.py
pr-watson: I'll add this thanks

convert2rhel/pkghandler.py
pr-watson: Yes that is more clear to the end users, i'll change this

convert2rhel/unit_tests/pkghandler_test.py
pr-watson: Yes I've done it on a dnf system and the yum tests would fail. I can add the skips and remove the mocks

convert2rhel/pkghandler.py
pr-watson: Noted

convert2rhel/pkghandler.py
pr-watson: yes i'll remove that

convert2rhel/pkghandler.py
pr-watson: Yes, I feel that it will be used elsewhere to parse package strings. One of the motivations for separating the parsing and comparison function into two functions was to be able to parse a package without also comparing them.

convert2rhel/pkghandler.py
pr-watson: noted

convert2rhel/pkghandler.py
pr-watson: Noted, i'll add that

convert2rhel/unit_tests/checks_test.py
pr-watson: Noted

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,165 @@ def clean_yum_metadata():
...
+        name = no_arch_data.name
+        epoch = epoch or no_arch_data.epoch
+        version = no_arch_data.version
+        release = no_arch_data.release
+        arch = no_arch_data.arch
+
+    # convert epoch from integer to string for consistency
+    epoch = str(epoch)
+
+    if epoch == "None":
```
pr-watson, line 1270: This addition is necessary so that PACKAGE_FORMATS in the unit tests all conform to the same types. Hawkey sets the epoch field as an integer while yum splitFilename sets epoch as a string, for consistency, epoch is converted to a string. If the epoch is not present it will be set as None, since we cast it to a string here, we also need to convert the "None" string to None so it will be consistent with how epoch is represented in the PACKAGE_FORMATS.

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,165 @@ def clean_yum_metadata():
...
-    :type pkg_1: str
-    :param pkg_2: The second package to extract the version
-    :type pkg_2: str
-    :return: An integer resulting in the package comparision
+        Since the return type is a int, this could be difficult to understand
+        the meaning of each number, so here is a quick list of the meaning from
+        every possible number:
+            * -1 if the evr1 is less then evr2 version
+            * 0 if the evr1 is equal evr2 version
+            * 1 if the evr1 is greater than evr2 version
```
r0x0d, line 1144:
```suggestion
        Since the return type is a int, this could be difficult to understand
        the meaning of each number, so here is a list that represents
        every possible number:
            * -1 if the version1 is less then version2 version
            * 0 if the version1 is equal version2 version
            * 1 if the version1 is greater than version2 version
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,165 @@ def clean_yum_metadata():
...
+
+    :return: Return a number indicating if the versions match, are less or greater then.
     :rtype: int
     """
+    # call parse_pkg_string to obtain name, epoch, release, version and arch fields
```
r0x0d, line 1149:
```suggestion
    # call parse_pkg_string to obtain a list containing name, epoch, release, version and arch fields
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,165 @@ def clean_yum_metadata():
...
+    # order of fields: name, epoch, version, release, arch
```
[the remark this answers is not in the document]
r0x0d: I see this is already part of the `:return:` in that docstring. Maybe just change a word of two there to say that this is the correct order that will be returned?

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,165 @@ def clean_yum_metadata():
...
+    evr1 = (version1_components[1], version1_components[2], version1_components[3])
+    evr2 = (version2_components[1], version2_components[2], version2_components[3])
+
+    return rpm.labelCompare(evr1, evr2)
+
+
+def parse_pkg_string(pkg):
+    """
+    This function takes a version string in NEVRA, NEVR, NVRA, NVR, ENVRA, ENVR and decides whether
+    to parse with a yum/dnf module based on the package manager type of the system
```
r0x0d, line 1172:
```suggestion
    to parse with a yum/dnf module based on the package manager type of the system.
    
    :param pkg: The package to be parsed. 
    :type pkg: str
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,165 @@ def clean_yum_metadata():
...
+        pkg_ver_components = _parse_pkg_with_yum(pkg)
+    else:
+        pkg_ver_components = _parse_pkg_with_dnf(pkg)
+
+    return pkg_ver_components
+
+
+def _parse_pkg_with_yum(pkg):
+    """Parse verison string using yum and rpmUtils splitFilename
+
```
r0x0d, line 1187:
```suggestion

    :param pkg: The package to be parsed. 
    :type pkg: str
```

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,165 @@ def clean_yum_metadata():
...
+    # convert any empty strings to None for consistency
+    pkg_ver_components = [(i or None) for i in (name, epoch, version, release, arch)]
+
+    return pkg_ver_components
+
+
+def _parse_pkg_with_dnf(pkg):
+    """
+    Parse verison string using hawkey and dnf
+    :return: Return a list containing name, epoch, version, release, arch
```
r0x0d, line 1225:
```suggestion
    Parse verison string using hawkey and dnf.
    
    :param pkg: The package to be parsed. 
    :type pkg: str
    
    :return: Return a list containing name, epoch, version, release, arch
```

convert2rhel/pkghandler.py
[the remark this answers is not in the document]
pr-watson: Noted

convert2rhel/pkghandler.py
```diff
@@ -1144,52 +1111,165 @@ def clean_yum_metadata():
...
+    :rtype: list[str | None]
+    """
+
+    # package is in NEVRA/NEVR format
+    if re.findall(r"-\d:", pkg):
+        name, epoch_version, release_arch = pkg.rsplit("-", 2)
+        epoch, version = epoch_version.split(":", 1)
+        # package is in NEVRA format
+        if re.findall(r".*(x86_64|s390x|i86|ppc64le|aarch64)$", release_arch):
```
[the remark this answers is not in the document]
pr-watson: release_arch contains both release and arch, on the next line the string gets split into the release and arch fields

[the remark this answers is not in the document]
pr-watson: I think using regex is a bit better too since it detects that the arch comes directly after a "."

r0x0d:
@pr-watson, please rebase whenever you have time!

convert2rhel/pkghandler.py
pr-watson: I tried adding this change but since no_arch_data is a different type than the ones above it caused issues when they all were in one line

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1320,21 +1320,115 @@ def test_fix_default_kernel_with_no_incorrect_kernel(self):
...
+@pytest.mark.parametrize(
+    ("package", "expected"),
+    (PACKAGE_FORMATS),
+)
+def test_parse_pkg_string_dnf(package, expected, monkeypatch):
+    assert pkghandler.parse_pkg_string(package) == expected
+    parse_pkg_with_dnf_mock = mock.Mock()
+    monkeypatch.setattr(pkghandler, "_parse_pkg_with_dnf", value=parse_pkg_with_dnf_mock)
+    pkghandler.parse_pkg_string(package)
+    parse_pkg_with_dnf_mock.assert_called_once()
```
r0x0d, line 1397: Not sure if I understood this type of test here... We are calling `parse_pkg_string` and then asserting it with expected, then, we are mocking the `_parse_pkg_with_dnf` and asserting it was called? That doesn't make too much sense to me. 

Probably we could re-write it to something like: 

```suggestion
    assert pkghandler._parse_pkg_with_dnf(package) == expected
```

The function seems simple enough that we could just call it, especially because we have a conditional skip for this test depending on the backend.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1320,21 +1320,115 @@ def test_fix_default_kernel_with_no_incorrect_kernel(self):
...
+    ("package", "expected"),
+    (PACKAGE_FORMATS),
+)
+def test_parse_pkg_string_yum(package, expected, monkeypatch):
+
+    assert pkghandler.parse_pkg_string(package) == expected
+    parse_pkg_with_yum_mock = mock.Mock()
+    monkeypatch.setattr(pkghandler, "_parse_pkg_with_yum", value=parse_pkg_with_yum_mock)
+    pkghandler.parse_pkg_string(package)
+    parse_pkg_with_yum_mock.assert_called_once()
```
r0x0d, line 1411: The same goes for tihs one: 

```suggestion
    assert pkghandler.parse_pkg_string(package) == expected
```

convert2rhel/pkghandler.py
```diff
@@ -1101,52 +1068,173 @@ def clean_yum_metadata():
...
+        epoch = epoch or no_arch_data.epoch
+        version = no_arch_data.version
+        release = no_arch_data.release
+        arch = no_arch_data.arch
+
+    # convert epoch from integer to string for consistency
+    epoch = str(epoch)
+
+    if epoch == "None":
+        epoch = None
```
r0x0d, line 1236: I would rewrite this part here as: 

```suggestion
    if epoch is not None:
        epoch = str(epoch)
```

I know we went over this in our chat, but, I just tested the code, and it seems to me that `epoch` will be either an instance of `int` or `NoneType`. In that case, we could just verify that `epoch` is not None, and if it is not, cast it to string to keep it consistent. Otherwise, it is None already, just leave it as it is.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1320,21 +1320,115 @@ def test_fix_default_kernel_with_no_incorrect_kernel(self):
...
+        ("kernel-core-0:390-287.fc36", "kernel-0:390-287.fc36", "kernel-core does not match kernel"),
+        ("kernel-core-0:390-287.fc36.aarch64", "kernel-core-0:391-287.fc36.i86", "aarch64 does not match i86"),
+    ),
+)
+def test_compare_package_versions_warnings(version1, version2, exception_message):
+    with pytest.raises(ValueError, match=exception_message):
+        pkghandler.compare_package_versions(version1, version2)
+
+
+# format: name, epoch, version, release, arch
```
r0x0d, line 1360: I think this message is not necessary anymore since you're using `pytest.param` 
```suggestion
```

convert2rhel/unit_tests/pkghandler_test.py
r0x0d, line 1411: I think you could unify those methods and have just one that asserts you're going through the correct path. 

They are not adding too much value the way they are right now compared to the `test_parse_pkg_with_yum` and `test_parse_pkg_with_dnf`

Some ideas that could be done: 

1. Remove the `test_parse_pkg_with_[dnf, yum]` and have only those two that will call them (without mocks) and assert the result
2. Have just one test that will assert the path inside the function. This test would be backend independent, you would mock this dynamically based on a parametrization.
3. You can leave the two methods there, but remove the first part where you call `assert pkghandler.parse_pkg_string(package) == expected` as you're calling it twice, but there is no need for this.

If you have more ideas, I would love to hear too.

r0x0d (changes requested):
Thanks for the changes, @pr-watson! I think they are good, just left a few comments for you to take care of. 

I know that sometimes we want to compile the regex expressions earlier, but honestly, I don't know if the cases here are the ones we would care about compiling them first (Probably it is). 

Not only that, but I will let @abadger or @SpyTec give their reviews here to say if that's the case or not.

SpyTec:
> Thanks for the changes, @pr-watson! I think they are good, just left a few comments for you to take care of.
> 
> I know that sometimes we want to compile the regex expressions earlier, but honestly, I don't know if the cases here are the ones we would care about compiling them first (Probably it is).
> 
> Not only that, but I will let @abadger or @SpyTec give their reviews here to say if that's the case or not.

If the function will be run multiple times then the RegEx should be compiled for performance. Looking at the code I think this is one of those cases where it should be compiled beforehand

convert2rhel/pkghandler.py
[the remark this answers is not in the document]
SpyTec: This could be done with groups in RegEx too. Though I'd vote for not having RegEx here and instead go with something like this. Since we're looking at the end of the line
```suggestion
        if release_arch.endswith((x86_64, s390x, i86, ppc64le, aarch64)):
```

convert2rhel/unit_tests/pkghandler_test.py
pr-watson: The third idea is what I had originally so I'll go back to that format, thanks!

convert2rhel/unit_tests/pkghandler_test.py
pr-watson: Noted, I'll remove this comment

convert2rhel/pkghandler.py
pr-watson: This is much better thanks! I'll add this

convert2rhel/pkghandler.py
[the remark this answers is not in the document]
pr-watson: Thanks @SpyTec I'll add this!

convert2rhel/pkghandler.py
```diff
@@ -1101,52 +1068,171 @@ def clean_yum_metadata():
...
+def parse_pkg_string(pkg):
+    """
+    This function takes a version string in NEVRA, NEVR, NVRA, NVR, ENVRA, ENVR and decides whether
+    to parse with a yum/dnf module based on the package manager type of the system.
+
+    :param pkg: The package to be parsed.
+    :type pkg: str
+
+    :return: Return a Return a list containing name, epoch, version, release, arch
+    :rtype: list[str | None]
```
r0x0d, line 1135: I think I did the wrong suggestion here about the return type. The deal with this one that I suggested is that it will be either a `['something']` or `[]`. 

```suggestion
    :rtype: list[str]
```

convert2rhel/pkghandler.py
```diff
@@ -1101,52 +1068,171 @@ def clean_yum_metadata():
...
+
+
+def _parse_pkg_with_dnf(pkg):
+    """
+    Parse verison string using hawkey and dnf.
+
+    :param pkg: The package to be parsed.
+    :type pkg: str
+    :return: Return a list containing name, epoch, version, release, arch
+    :rtype: list[str | None]
```
r0x0d, line 1191:
```suggestion
    :rtype: list[str]
```

r0x0d (approved):
Looks nice! I made two comments there, but honestly, I will not block or require them to be accepted. Just pointing out that I made a mistake in a previous review. 

In any case, looks good for me.

convert2rhel/pkghandler.py
pr-watson: I thought it looked correct, the fields in the list will either be None type or a string

convert2rhel/pkghandler.py
pr-watson: Noted

convert2rhel/pkghandler.py
pr-watson: After some discussion, it turns out lists can only accept one type. Adding this suggestion

r0x0d (approved):

r0x0d (approved):

r0x0d (approved):

convert2rhel/utils.py
```diff
@@ -663,25 +663,14 @@ def find_keyid(keyfile):
...
-    """
-
-    # is there an epoch?
-    components = verstring.split(":")
-    if len(components) > 1:
-        epoch = components[0]
-        components.pop(0)
-    else:
-        epoch = "0"
+def set_locale():
```
abadger, line 666: Looks like this snuck in during a rebase.  This function has recently been removed.

convert2rhel/pkghandler.py
```diff
@@ -1079,52 +1054,170 @@ def _get_packages_to_update_dnf(reposdir):
...
-            version = epoch_version[0]
+    :return: Return a Return a list containing name, epoch, version, release, arch
+    :rtype: list[str | None]
+    """
+    if pkgmanager.TYPE == "yum":
+        pkg_ver_components = _parse_pkg_with_yum(pkg)
+    else:
+        pkg_ver_components = _parse_pkg_with_dnf(pkg)
+
+    return pkg_ver_components
```
abadger, line 1127: Some of this should be moved into `pkgmanager/handlers/{yum.dnf}`.  @r0x0d was going to look into what makes sense to do there going forward, though, so t can be done at a later time.  I just wanted to note it so that it can be thought about when you design functions that have separate backends for yum and dnf in the future.

convert2rhel/utils.py
pr-watson: Noted i'll remove it

convert2rhel/pkghandler.py
```diff
@@ -1079,52 +1054,170 @@ def _get_packages_to_update_dnf(reposdir):
...
     :rtype: int
     """
+    # call parse_pkg_string to obtain a list containing name, epoch, release, version and arch fields
+    # order of fields returned: name, epoch, version, release, arch
+    version1_components = parse_pkg_string(version1)
+    version2_components = parse_pkg_string(version2)
+
+    # ensure package names match, error if not
+    if version1_components[0] != version2_components[0]:
+        raise ValueError("%s does not match %s" % (version1_components[0], version2_components[0]))
```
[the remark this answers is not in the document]
pr-watson: Noted I'll add this

convert2rhel/pkghandler.py
pr-watson: Noted thank you

abadger:
The failing unit test means that one or two things should be changed:
* We need to catch the `ValueError` and use `loginst.critical()` in at least one place where `compare_package_versions()` is called.
* We may need to update the parametrize data to the unittest.  The data that is being used in that unittest may have changed its format in the production code. (The change in what we have repoquery return).

pr-watson:
Yes I noticed them failing after the new repoquery additions. It also looks like I need to change either how data is sent to compare_package_version from is_loaded_kernel_latest() in checks.py, or I need to change the parsing functions (parse_pkg_with_dnf and parse_pkg_with_yum) to be able to accept packages in the VR format

convert2rhel/pkghandler.py
r0x0d: Well noted, @abadger. Maybe I can include this comment in the jira issue where we are tracking this progress. I know that I thought about that the other day, but probably forgot to update comments/jira.

SpyTec:
Unit tests are failing for centos7 here

pr-watson:
> Unit tests are failing for centos7 here

@SpyTec thanks, it was due to the tuple class accepting multiple arguments. Must be due to centos7 using python2. I've put in a fix

SpyTec:
> > Unit tests are failing for centos7 here
> 
> @SpyTec thanks, it was due to the tuple class accepting multiple arguments. Must be due to centos7 using python2. I've put in a fix

Thanks for the fix @pr-watson. I'll take a look shortly, but wanted to check if you have installed pre-commit hooks locally as I noticed that there are commits from pre-commit. Should be able to run `make install` to get it up and running if pre-commit is installed

pr-watson:
> > > Unit tests are failing for centos7 here
> > 
> > 
> > @SpyTec thanks, it was due to the tuple class accepting multiple arguments. Must be due to centos7 using python2. I've put in a fix
> 
> Thanks for the fix @pr-watson. I'll take a look shortly, but wanted to check if you have installed pre-commit hooks locally as I noticed that there are commits from pre-commit. Should be able to run `make install` to get it up and running if pre-commit is installed

I don't believe I did have it installed. I just ran make install and it looks like it downloaded it

SpyTec:
Ahh right. It only installs it but doesn't install the hooks themselves. Run `pre-commit install --install-hooks` and it should be run when you do commits and pushes

pr-watson:
> Ahh right. It only installs it but doesn't install the hooks themselves. Run `pre-commit install --install-hooks` and it should be run when you do commits and pushes

When I ran make install I got these messages:
pre-commit install --install-hooks
pre-commit installed at .git/hooks/pre-commit
touch .pre-commit

So I think it did install the hooks, but I can run that as well

PR #715, opened 2023-01-25
[RHELC-876] Cleanup pylint errors for object inheritance and enable consider-using-with

abadger:
* Evaluate consider-using-f-string: this can only be enabled when we move support up to python-3.6+
* Enable consider-using-with.  Python context managers are used via the with keyword.  these automatcally clean up resources when a context's scope is exited.  Fix the places where we can use existing context managers (file open) to use them and note where we have to wait to drop support for old Python versions to the other places.
* Enable useless-object-inheritance.  Python originally had separate data structures for builtin types and user defined classes.  This changed in Python-2.2 and the new user defined classes were known as new-style classes.  A user defined class had to explicitly inherit from object (`class Foo(object):`) to be a new-style class.  In Python 3, old-style classes were dropped and the simpler syntax which had been reserved for old-style classes now works creates new-style classes (`class Foo:`). This syntax can also be used in Python 2 if the file specifies `__metaclass__ = type` at the top of the file.

  This commit add `__metaclass__ = type` to all of our python files (in case classes are added to any of them in the future) and the class definitions are ported to the simpler syntax.

Jira Issue: [RHELC-876](https://issues.redhat.com/browse/RHELC-876)

Checklist

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,3 +1,6 @@
+[MASTER]
+ignore=
+
 [BASIC]
 
 # Regular expression matching correct function names
@@ -18,11 +21,7 @@ disable=
  arguments-renamed,
  attribute-defined-outside-init,
  broad-except,
- consider-using-dict-items,
- consider-using-f-string,
- consider-using-with,
  cyclic-import,
- deprecated-module,
  duplicate-code,
  empty-docstring,
  expression-not-assigned,
@@ -81,8 +80,6 @@ disable=
  unused-argument,
  unused-import,
  unused-variable,
- useless-else-on-loop,
- useless-object-inheritance,
  useless-option-value,
  useless-return,
  useless-suppression,
--- a/convert2rhel/pkgmanager/handlers/yum/callback.py
+++ b/convert2rhel/pkgmanager/handlers/yum/callback.py
@@ -55,7 +55,9 @@
 #   License above taken from the original code at:
 #
 #       https://github.com/rpm-software-management/yum/blob/master/yum/callbacks.py
-#
+
+__metaclass__ = type
+
 import logging
 
 from convert2rhel import pkgmanager
@@ -66,7 +68,7 @@ loggerinst = logging.getLogger(__name__)
 
 # We need to double inherit here, both from the callback class and the base
 # object class, just to initialize properly with `super`
-class PackageDownloadCallback(pkgmanager.DownloadProgress, object):
+class PackageDownloadCallback(pkgmanager.DownloadProgress, object):  # pylint: disable=useless-object-inheritance
     """Package download callback for YUM transaction."""
 
     def __init__(self):
@@ -111,7 +113,7 @@ class PackageDownloadCallback(pkgmanager.DownloadProgress, object):
         self.last_package_seen = name
 
 
-class TransactionDisplayCallback(pkgmanager.TransactionDisplay, object):
+class TransactionDisplayCallback(pkgmanager.TransactionDisplay, object):  # pylint: disable=useless-object-inheritance
     """Transaction display callback for YUM transaction."""
 
     def __init__(self):
--- a/setup.py
+++ b/setup.py
@@ -65,5 +65,8 @@ setup(
             "convert2rhel = convert2rhel.initialize:run",
         ]
     },
+    install_requires=[
+        "six",
+    ],
     include_package_data=True,
 )
--- a/convert2rhel/__init__.py
+++ b/convert2rhel/__init__.py
@@ -1,1 +1,2 @@
+__metaclass__ = type
 __version__ = "1.4.1"
--- a/convert2rhel/actions/pre_ponr_changes/__init__.py
+++ b/convert2rhel/actions/pre_ponr_changes/__init__.py
@@ -1,0 +1,1 @@
+__metaclass__ = type
--- a/convert2rhel/actions/system_checks/__init__.py
+++ b/convert2rhel/actions/system_checks/__init__.py
@@ -1,0 +1,1 @@
+__metaclass__ = type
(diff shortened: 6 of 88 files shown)


r0x0d:
@abadger can you rebase and fix the conflict? If not, let me know and I can do that for you.

r0x0d (commented):
Changes looks good! A lot of files touches, but the changes were minimal on all of them. Although, we have one CodeQL error in this, could you take a look at it @abadger?

abadger:
> @abadger can you rebase and fix the conflict? If not, let me know and I can do that for you.

Thanks, rebased.  I'll look at the CodeQL error once that's done on the new commit.

r0x0d (dismissed):

.pylintrc
```diff
@@ -19,8 +23,6 @@ disable=
  attribute-defined-outside-init,
  broad-except,
  consider-using-dict-items,
- consider-using-f-string,
```
SpyTec, line 22: But we still require 2.7 due to CentOS 7, will this not cause issues?

.pylintrc
```diff
@@ -82,7 +84,6 @@ disable=
  unused-import,
  unused-variable,
  useless-else-on-loop,
- useless-object-inheritance,
```
SpyTec, line 85: Likewise with this, we need this due to 2.7 as well right?

bocekm: This is about removing the disablement of the check.
@abadger  wrote _"This syntax can also be used in Python 2 if the file specifies __metaclass__ = type at the top of the file."_ which is what he did so enabling this check is fine.

.pylintrc
ferdnyc: @SpyTec It's in the file twice. This is a redundant line, it already appears earlier in the list (line 18):

https://github.com/oamg/convert2rhel/blob/d182c47b5e457c415805568b6b4e881034d6a8d9/.pylintrc#L15-L24

SpyTec: Ohh gotcha. Thanks for letting me know!

.pylintrc
SpyTec: If a class inherits we still use object, can that also be safely removed?

abadger: If I understand what you're asking (what happens when we inherit from an old-style class), then it is not always safe but it is easy to figure out whether we can do it or not.

* In most cases, the class we inherit from probably inherits from object itself.
  * Test this by running `python2.7 -c 'from MODULE import CLASSNAME; print(type(CLASSNAME))'`.  If it prints `<type 'type'>` then it is fine.  If it prints `<type 'classobj'>` then more exploration is needed.
* In other cases, we don't need any features of new-style classes so it does no harm if the class we inherit from is not inheriting from object. (Just two of the features that require a new-style class include using the `@property` decorator and using `super()` inside of a method). If we do need new-style class features then:
* We do need to inherit from both the old-style base class and `object`.  That way our class will be a new-style class and have access to all of the new-style class features.
  * An example is: https://github.com/oamg/convert2rhel/pull/715/files#diff-f82efabf7bb126100afd0472e3d8a5902da6576c6ddf75e3184cce9eb647c52aL218  In Python 2.6, `logging.Formatter` is an old style class and we wantd to use `super()` in our child class.  So we needed to add `object` to the inheritance to make our child class a new-style class.  Note that `logging.Formatter` became a new-style class in Python 2.7 and since we've dropped support for Python-2.6, it is now safe to remove the inheritance from `object`.

SpyTec: Gotcha. Thanks for the info!

bocekm (approved):

abadger:
packit-as-a-service[bot] reacted with eyes emoji

setup.py
```diff
@@ -66,6 +66,9 @@ def get_version():
             "convert2rhel = convert2rhel.initialize:run",
         ]
     },
+    install_requires=[
```
bocekm, line 69: Optional: I think it'd be useful to have a comment in the code explaining why do we have it here. To not require people delve into git history.

SpyTec: Done

SpyTec:
Rebased with upstream and also added commit from commit to `setup.py` about CodeQL

.pylintrc
```diff
@@ -1,3 +1,7 @@
+[MASTER]
+# We did not write this script.  Changes should go upstream.
+ignore=build_manpage.py
```
r0x0d, line 3: From where this script came from? I don't see it in any folder. I see it being imported, but not present on the project 🤔

convert2rhel/pkgmanager/handlers/yum/callback.py
```diff
@@ -66,7 +68,7 @@
 
 # We need to double inherit here, both from the callback class and the base
 # object class, just to initialize properly with `super`
-class PackageDownloadCallback(pkgmanager.DownloadProgress, object):
+class PackageDownloadCallback(pkgmanager.DownloadProgress, object):  # pylint: disable=useless-object-inheritance
```
r0x0d, line 71: We need the object inheritance here because of the first inheritance, that is from yum lib?

r0x0d (approved):

.pylintrc
abadger: Ah yeah, this is from long enough ago that we were using this script but no longer.  So we can remove this portion of the change

convert2rhel/pkgmanager/handlers/yum/callback.py
abadger: Correct.  The yum object is an old-style class on Python2 so we need to also explicitly inherit from object so our class is a new-style class and we can use `super`.

r0x0d (approved):

r0x0d:
@abadger, can you rebase this later? I think everything is good to merge.

r0x0d (approved):

kokesak:
Fails are unrelated to this PR. No need to block this PR because of this.

SpyTec:
@kokesak feel free to merge if so!

r0x0d:
I will merge this in a moment. Just finishing some other reviews.

abadger:
Now merged.  Thank you all!

PR #653, opened 2022-11-07
[RHELC-785] Add system_release and prompt_ammount as envar

danmyway:
* currently we are using python platform, specifically platform.platform() to reach for distribution specific information
* this shows as flaky at times, and pretty much out of our control to be reliable to condition the tests based on this information
* there is an attempt to use pytest fixture which parses `/etc/system-release` or `/etc/os-release`
* these files are however mutable, therefore not reliable as well, we actually remove os-release in one of the tests
* this change utilizes tmt environment adjustment method, therefore provides controllable assignment of the environment variables on which we can rely on

Signed-off-by: Daniel Diblik <ddiblik@redhat.com>

Jira Issue: [RHELC-785](https://issues.redhat.com/browse/RHELC-785)

Checklist
- [x] PR meets acceptance criteria specified in the Jira issue
- [x] PR has been tested manually in a VM (either author or reviewer)
- [x] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [x] Code and tests are documented properly
- [x] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending`

PR #721, opened 2023-02-01
[RHELC-752] Check for kernel boot files after conversion

r0x0d:
Add a later check to warn the user about kernel boot files
    
We added a few later checks to verify and warn the user if there is any
errors with both `initramfs` `vmlinuz` files on their system after the
conversion.
    
This check has the intention to at least inform the user what are the
steps that they need to run in order to fix that problem.
    
Jira Issue: [RHELC-752](https://issues.redhat.com/browse/RHELC-752)

Checklist


r0x0d:
# Steps to reproduce locally (and for testing)
0. Open two terminals, one you will run the conversion inside the vagrant box, the other one you will run a script that will cause the error.
1. Go to the c2r-vagrant project and get a vagrant box, either CentOS 7 or CentOS 8 (this should happen and work on both of them) 
1.1. An easy way of doing it is by pulling the latest changes from that project, and then running `make refresh`
2. On a second terminal, download those scripts to the `convert2rhel` source folder 
2.1.  [Cause scriptlet failure by filling up disk space in /boot](https://gist.github.com/r0x0d/6edda401d120366a4622d8eda7347ec3)
2.2. [Corrupt the initramfs file by echoing a text to it](https://gist.github.com/r0x0d/58e38b5cb88bb618ac87fd174f418683)
3. Now, if you're on the latest version of the `c2r-vagrant` project, you can run `make convert` and a conversion should start
4. When you see the conversion running, go to the second terminal and execute one of those commands
4.1. `vagrant ssh -- -t 'python3 /usr/lib/python3.6/site-packages/convert2rhel/<name-of-script-that-causes-scriptlet-failure>.py'`
4.2. `vagrant ssh -- -t 'python3 /usr/lib/python3.6/site-packages/convert2rhel/<name-of-script-that-causes-corruption>.py'`
5. Wait for the conversion to reach the transaction part, and then, you will that the custom script you ran might stop running, that's intended, and it tells you that it finished and either filled up the disk space or corrupted the file, depending on which script you run.

# Notes

This reproducer probably works with Oracle Linux as well, but you will need to change the kernel version inside of the custom scripts. Keep an eye on that.

convert2rhel/utils.py
```diff
@@ -125,6 +125,123 @@ def restart_system():
...
+                normalized_output = normalized_output.encode()
+
+            normalized_output += line
+        else:
+            normalized_output += line.decode("utf-8")
+
+    return normalized_output
+
+
+def run_piped_subprocess(
```
r0x0d, line 158: @abadger, not sure if that's the correct and best way of coming up with this function, but please, if you review it, let me know what I could improve here.

convert2rhel/utils.py
```diff
@@ -125,6 +125,123 @@ def restart_system():
         loggerinst.warning("In order to boot the RHEL kernel, restart of the system is needed.")
 
 
+def _normalize_subprocess_output(output, skip_stdout_decode):
```
r0x0d, line 128: This function is intended to be used in the `run_subprocess` function too in the future, but that could be a separate PR.

convert2rhel/unit_tests/utils_test.py
```diff
@@ -874,7 +888,298 @@ def communicate(self):
...
 
     def poll(self):
-        return 0
+        return self.return_code
+
+    def close(self):
+        pass
+
+
+@pytest.mark.parametrize(
```
r0x0d, line 897: Also, @abadger, if you could take a good look over this, it would help so much! I feel that those unit tests could be done differently (in a better way), but I'm not sure how to improve it.

convert2rhel/checks.py
```diff
@@ -832,3 +844,88 @@ def check_dbus_is_running():
...
+
+    if not os.path.exists(initramfs_file):
+        return False
+
+    logger.info("Checking if the %s file is not corrupted.", initramfs_file)
+    _, err, return_code = run_piped_subprocess(
+        cmd1=["zcat", initramfs_file],
+        cmd2=["cpio", "-i", "--to-stdout"],
+        skip_stdout_decode=True,
+    )
```
abadger, line 867: Did you want to redirect stdout output to /dev/null?

r0x0d: Yes, I'm fine with the stdout for this to be redirected to `/dev/null` or `pipe`. For that, we only care about stderr and return_code if any, at least for this call.

plans/tier1.fmf
```diff
@@ -370,3 +370,36 @@
...
+/kernel_boot_files_check_boot_files_presence:
+   discover+:
+       test: checks-after-conversion
+   prepare+:
+       - name: main conversion part
+         how: tmt
+         test: tests/integration/tier1/kernel-boot-files/test_verify_file_presence.py
+       - name: reboot after conversion
+         how: ansible
+         playbook: tests/ansible_collections/roles/reboot/main.yml
```
danmyway, line 405:
```suggestion
/kernel_boot_files_missing_boot_files:
   discover+:
       test: checks-after-conversion
   prepare+:
       - name: main conversion part
         how: shell
         script: pytest -svv tests/integration/tier1/kernel-boot-files/test_handle_missing_boot_files.py
       - name: reboot after conversion
         how: ansible
         playbook: tests/ansible_collections/roles/reboot/main.yml

/kernel_boot_files_corrupted_initramfs_file:
   discover+:
       test: checks-after-conversion
   prepare+:
       - name: main conversion part
         how: shell
         script: pytest -svv tests/integration/tier1/kernel-boot-files/test_handle_corrupted_files.py
       - name: reboot after conversion
         how: ansible
         playbook: tests/ansible_collections/roles/reboot/main.yml

/kernel_boot_files_check_boot_files_presence:
   discover+:
       test: checks-after-conversion
   prepare+:
       - name: main conversion part
         how: shell
         script: pytest -svv tests/integration/tier1/kernel-boot-files/test_verify_file_presence.py
       - name: reboot after conversion
         how: ansible
         playbook: tests/ansible_collections/roles/reboot/main.yml
```
For now we stick with the messy test calling.
Should be like this.

abadger (approved):
Main code looks good.  I'm just including a suggestion for how to implement outputing stdout to /dev/null and some spelling and wording changes.

I have looked at but not gone deeply into the unittests yet.  I'm going to approve this because I don't see anything wrong with those but I know you specifically asked if I could see a better way to do them so I will look some more.

(I have not tested in a vm)

convert2rhel/utils.py
r0x0d: Note: I see that the tests/integration/conftest.py could benefit from the `skip_stdout_decode` logic too.

convert2rhel/unit_tests/utils_test.py
```diff
@@ -874,7 +888,312 @@ def communicate(self):
...
+    ("output", "skip_stdout_decode", "expected"),
+    (
+        (DummyPopenOutput([u"test".encode()]), True, b"test"),
+        (DummyPopenOutput([u"testê café".encode("utf-8")]), False, u"testê café"),
+        (DummyPopenOutput([u"testê café".encode("utf-8")]), False, u"testê café"),
+        (DummyPopenOutput([u"test".encode()]), False, "test"),
```
abadger, line 903:
```suggestion
        (DummyPopenOutput([u"test".encode("ascii")]), True, b"test"),
        (DummyPopenOutput([u"test".encode("ascii")]), False, u"test"),
        (DummyPopenOutput([u"testê café".encode("utf-8")]), True, u"testê café".encode("utf-8")),
        (DummyPopenOutput([u"testê café".encode("utf-8")]), False, u"testê café"),
```
* Fix one of the test case's data to test non-ascii data with skip_stdout_decode=True
* Be sure to specify `u` on the ascii-only string when skip_stdout_decode=False.
* Order the output as, contans ascii-only data, non-ascii-only data.
* Always encode() using an explicit encoding. (Note: the encoding calls I added an encoding too could have used utf-8 but it seemed to be clearer what the purpose of the test was to use ascii as the encoding.

convert2rhel/unit_tests/utils_test.py
```diff
@@ -874,7 +888,312 @@ def communicate(self):
...
+    ),
+)
+def test_normalize_subprocess_output(output, skip_stdout_decode, expected):
+    result = utils._normalize_subprocess_output(output.stdout.readline, skip_stdout_decode)
+    assert result == expected
```
abadger, line 908:
```suggestion
    assert result == expected
    assert isinstance(result, type(expected))
```

On python2, ascii-only byte strings and ascii-only unicode strings that contain the same abstract characters are equal.  We actually care about this since we don't want to ensure that this function always returns bytes when skip_stdout_decode is True and return text when skip_stdout_decode is False.  Adding the isinstance assert will catch this cornercase.

convert2rhel/unit_tests/utils_test.py
```diff
@@ -874,7 +888,312 @@ def communicate(self):
...
+        expected,
+        monkeypatch,
+    ):
+        process_mock = DummyPopenOutput(process_stdout, process_stderr)
+        output_mock = DummyPopenOutput(output_stdout, output_stderr, output_return_code)
+
+        monkeypatch.setattr(utils.subprocess, "Popen", mock.Mock(side_effect=[process_mock, output_mock]))
+        out, err, return_code = utils.run_piped_subprocess(cmd1, cmd2)
+
+        assert (out, err, return_code) == expected
```
abadger, line 985: I think I would use less mocking in this test.  Yes, it's not a true unittest if we get rid of the mocking, but I think it's a better test overall.

The heart of what makes `run_piped_subprocess()` different from `run_subprocess()` is that it runs two processes, piping the output from the first to the second.  Injecting mocked output as the output from the second command means that we are not actually testing that the communication between the two processes is working.

```suggestion
    @pytest.mark.parametrize(
        (
            "cmd1",
            "cmd2",
            "expected_return_code",
            "expected",
        ),
        (
            pytest.param(
                ["echo", "first line\ntest line"],
                ["tail", "-1"],
                (u"test line", "", 0),
                id="check-correct-success-output",
            ),
            pytest.param(
                ["echo", "test"],
                ["tail"],
                [u"".encode()],
                [u"test".encode()],
                [u"".encode()],
                [u"test".encode()],
                (u"", u"test", 1),
                id="check-correct-failure-output",
            ),
        ),
    )
    def test_run_piped_subprocess(
        self,
        cmd1,
        cmd2,
        expected,
        monkeypatch,
    ):
        out, err, return_code = utils.run_piped_subprocess(cmd1, cmd2)

        assert (out, err, return_code) == expected
```
Note: I didn't test that my suggestion above works.  I just wanted to write it out since I'm not sure if my explanation was clear.

convert2rhel/unit_tests/utils_test.py
```diff
@@ -874,7 +888,312 @@ def communicate(self):
...
+
+        process_mock = DummyPopenOutput([u"test".encode()], [u"".encode()])
+        output_mock = DummyPopenOutput([u"test".encode()], [u"".encode()], 0)
+
+        monkeypatch.setattr(utils.subprocess, "Popen", mock.Mock(side_effect=[process_mock, output_mock]))
+
+        with open(os.devnull, "w") as devnull:
+            out, err, return_code = utils.run_piped_subprocess(cmd1, cmd2, stdout=devnull)
+
+            assert (out, err, return_code) == ("", "", 0)
```
abadger, line 1196: I think we want to assert that the output did not get printed in this case. You can do that using the `capsysbnary` fixture: https://docs.pytest.org/en/7.1.x/how-to/capture-stdout-stderr.html#accessing-captured-output-from-a-test-function

r0x0d:
@Abadger and I came to a conclusion earlier today that we could simplify the code and fix some of the problems we were having while running the integration tests this week. The problems were related to the piped command we were using `zcat ... | cpio -i --to-stdout`, as initially, we thought that `initramfs-$(uname -r).img` files were always in the `gzip` format, but, it seems that this is not the case as it may contain multiple compression formats in one image. 

Since this kinda breaks our use case, we began to brainstorm and check if we could overcome this situation somehow, and the final idea was also the first one: `Let's use dracut to verify the image`. 

At first, we were a bit reluctant to use dracut (more specifically `/usr/bin/lsinitrd`) as we thought that, since `dracut` can be removed on RHEL8+, this package might not be installed at all, but, further investigation today, we figured out that the `kernel / kernel-core` package does require `dracut`, and if the package is not present on the system, if you install/reinstall the `kernel / kernel-core` package, `dracut` will be installed as well. 

The following is a code that we removed from the PR as it is not being used anymore, since the PR will not use piped commands, nor, Convert2RHEL in any predictable future, but, since things can change, and we may need to run commands like that (for some reason), we will provide a public gist with the old solution, as well, unit tests made for the solution, and also, the comments that @abadger did while reviewing it. 

Link to the gist: https://gist.github.com/r0x0d/5a631e8ab13ebb36078d8d0d0cad1c80

r0x0d:
For folks that will review this PR (I'm guessing @abadger and @hosekadam), keep in mind that this article may be useful for proposing last minute changes: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-verifying_the_initial_ram_disk_image

(now wondering if we should include the link for this article in the warning message :thinking:)

convert2rhel/unit_tests/utils_test.py
r0x0d: Not resolving this comment, but we have this mapped under the gist mentioned in a comment here, so should be fine to resolve it anyway.

convert2rhel/unit_tests/utils_test.py
r0x0d: Not resolving this comment, but we have this mapped under the gist mentioned in a comment here, so should be fine to resolve it anyway.

convert2rhel/unit_tests/utils_test.py
r0x0d: Not resolving this comment, but we have this mapped under the gist mentioned in a comment here, so should be fine to resolve it anyway.

convert2rhel/unit_tests/utils_test.py
r0x0d: Not resolving this comment, but we have this mapped under the gist mentioned in a comment here, so should be fine to resolve it anyway.

tests/integration/tier1/kernel-boot-files/test_verify_file_presence.py
```diff
@@ -0,0 +1,29 @@
...
+        "-y --no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug".format(
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        assert c2r.expect("Checking if vmlinuz file exists on the system.") == 0
+        assert c2r.expect("Checking if initiramfs file exists on the system.") == 0
+        assert c2r.expect("Initramfs and vmlinuz files exists and are valid.") == 0
```
danmyway, line 27: Just a proposal.
Given that this is just a simple conversion with no significant preparation, I would say we can even scrap it as an individual test and run it as an after-conversion-check.
Running this simple conversion could (IMHO) unnecessarily prolong the pipeline execution time by one full conversion runtime.
Maybe a good first use case for the check going over the log file matching required strings we were talking about @r0x0d .
If you would agree with this, I can provide you with a syntax, to skip the check for the test cases with tainted results.
WDYT @r0x0d, @SpyTec, @bocekm ?

SpyTec: Agreed

bocekm: Sounds good.

convert2rhel/checks.py
```diff
@@ -835,3 +841,88 @@ def check_dbus_is_running():
...
+    # detecting the latest kernel by using `rpm` and figuring out which was the
+    # latest kernel installed.
+    latest_installed_kernel = output.split("\n")[0].split(" ")[0]
+    latest_installed_kernel = latest_installed_kernel.split("%s-" % kernel_name)[-1]
+    grub2_config_file = grub.get_grub_config_file()
+
+    logger.info("Checking if vmlinuz file exists on the system.")
+    vmlinuz_exists = _verify_vmlinuz_file(latest_installed_kernel)
+
+    logger.info("Checking if initiramfs file exists on the system.")
```
bocekm, line 912:
```suggestion
    vmlinuz_exists = _verify_vmlinuz_file(latest_installed_kernel)
```
I feel like these two info messages are not necessary. Having the following two messages is totally sufficient:
```
Final: Check kernel boot files
The initramfs and vmlinuz files are valid.
```

bocekm (changes requested):

convert2rhel/checks.py
```diff
@@ -835,3 +841,78 @@ def check_dbus_is_running():
...
+    # point, should be a RHEL kernel. Since we can't get the kernel version
+    # from `uname -r`, as it requires a reboot in order to take place, we are
+    # detecting the latest kernel by using `rpm` and figuring out which was the
+    # latest kernel installed.
+    latest_installed_kernel = output.split("\n")[0].split(" ")[0]
+    latest_installed_kernel = latest_installed_kernel.split("%s-" % kernel_name)[-1]
+    grub2_config_file = grub.get_grub_config_file()
+    initramfs_file = INITRAMFS_FILEPATH % latest_installed_kernel
+
+    vmlinuz_exists = os.path.exists(VMLINUZ_FILEPATH % latest_installed_kernel)
```
SpyTec, line 892: Might be worth outputting the result of vmlinuz in a debug log as we do for initramfs

bocekm: We do so already through `The initramfs and vmlinuz files are valid.` info message. I'd say that's sufficient. What else than the information in this message would you like to print in the debug message?

SpyTec: It is more so if it is not valid, and that message doesn't say specifically which is bad. Only that one of them are. We log for initramfs in its function but nothing for vmlinuz

r0x0d: Something like: 

```suggestion
    latest_installed_kernel = output.split("\n")[0].split(" ")[0]
    latest_installed_kernel = latest_installed_kernel.split("%s-" % kernel_name)[-1]
    grub2_config_file = grub.get_grub_config_file()
    initramfs_file = INITRAMFS_FILEPATH % latest_installed_kernel
    vmlinuz_file = VMLINUZ_FILEPATH % latest_installed_kernel
    
    logger.debug("Checking if the '%s' file exists", vmlinuz_file)
    vmlinuz_exists = os.path.exists(vmlinuz_file)
    logger.debug("Result of vmlinuz presence: %s", vmlinuz_exists)
```

not sure, what you think?

tests/integration/tier1/kernel-boot-files/test_verify_file_presence.py
danmyway: Thanks, implemented

convert2rhel/checks.py
```diff
@@ -835,3 +841,78 @@ def check_dbus_is_running():
...
+def _is_initramfs_file_valid(filepath):
+    """Internal function to verify if a initramfs file is corrupted.
+
+    This method will rely on using lsinitrd to do the validation. If the
+    lsinitrd returns other value that is not 0, then it means that the file is
+    probably corrupted or may cause problems during the next reboot.
+
+    :param filepath: The path to the initramfs file.
+    :type filepath: str
+    :return: A boolean to determinate if the file is corrupted.
```
pr-watson, line 855: do you mean to use determine?

convert2rhel/checks.py
```diff
@@ -835,3 +841,78 @@ def check_dbus_is_running():
...
+    if return_code != 0:
+        logger.info("Couldn't verify initramfs file. It may be corrupted.")
+        logger.debug("Output of lsinitrd: %s", out)
+        return False
+
+    return True
+
+
+def check_kernel_boot_files():
+    """Check if the required kernel files exists and are valid under the boot partition."""
```
pr-watson, line 873:
```suggestion
    """Check if the required kernel files exist and are valid under the boot partition."""
```

convert2rhel/grub.py
```diff
@@ -591,6 +591,23 @@ def post_ponr_set_efi_configuration():
         _log_critical_error(e.message)
 
 
+def get_grub_config_file():
+    """Get the grub config file path.
+
+    This method will return the grub config file, depending if it's is BIOS or
```
pr-watson, line 597:
```suggestion
    This method will return the grub config file, depending if it is BIOS or
```

tests/integration/tier1/kernel-boot-files/main.fmf
```diff
@@ -0,0 +1,15 @@
+summary: Handle kernel boot files
+
+description: |
+    Verify that Convert2RHEL can detect and handle the kernel boot files in
+    case of failure.
+    Verify that, if present, the kernel boot files are not corrupted.
+
+    In any case, the conversion should not be inhibited. An warning will be
```
pr-watson, line 8:
```suggestion
    In any case, the conversion should not be inhibited. A warning will be
```

tests/integration/tier1/kernel-boot-files/test_handle_missing_boot_files.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@pytest.mark.missing_kernel_boot_files
+def test_missing_kernel_boot_files(convert2rhel):
+    """
+    Verify if an output with a warning message is sent to the user in case of
+    the tool can't detect the initramfs and vmlinuz files in /boot.
+
+    This case can happen if the kernel scriptlet fails during the yum/dnf
+    transaction where Convert2RHEL tries to replace the packages. Even thought
```
pr-watson, line 42:
```suggestion
    transaction where Convert2RHEL tries to replace the packages. Even though
```

tests/integration/tier1/kernel-boot-files/test_handle_missing_boot_files.py
```diff
@@ -0,0 +1,65 @@
...
+    the scriptlet can fail, the transaction will still continue and the
+    workflow will continue to be executed. The problem is that, with a
+    scriptlet failure when replacing/installing a kernel, the initramfs and
+    vmlinuz could not be available in the /boot partition, especially if there
+    is no sufficient disk space available in there. This test has the intention
+    to verify that the a warning with the correct steps are provided to the
```
pr-watson, line 48:
```suggestion
    to verify that the warning with the correct steps are provided to the
```

pr-watson: or "a" is also suitable

pr-watson:
I made a few suggestions for proper syntax in the comments. Everything else looks good!

tests/integration/tier1/kernel-boot-files/test_handle_corrupted_files.py
```diff
@@ -0,0 +1,72 @@
...
+    with convert2rhel(
+        "-y --no-rpm-va --serverurl {} --username {} --password {} --pool {} --debug".format(
+            env.str("RHSM_SERVER_URL"),
+            env.str("RHSM_USERNAME"),
+            env.str("RHSM_PASSWORD"),
+            env.str("RHSM_POOL"),
+        )
+    ) as c2r:
+        # Start the watcher as soon as we hit this message.
+        c2r.expect("Convert: List remaining non-Red Hat packages", timeout=600) == 0
```
danmyway:
```suggestion
        c2r.expect("Convert: List remaining non-Red Hat packages")
```
I would scrap the timeout and evaluation.
The run is time outing pretty much right before the expected string.
AD the evaluation, even though I think it should be fine, we mostly using it for asserting, so not needed here.

tests/integration/tier1/kernel-boot-files/test_handle_corrupted_files.py
```diff
@@ -0,0 +1,72 @@
+import os
+import shlex
+import subprocess
+
+from multiprocessing import Pool
+
+import pytest
+
+from envparse import env
+
```
danmyway, line 10:
```suggestion
from conftest import SYSTEM_RELEASE_ENV
```
Let's keep it consistent with #653 please.

tests/integration/tier1/kernel-boot-files/test_handle_corrupted_files.py
```diff
@@ -0,0 +1,72 @@
...
+    Since this could be a real scenario, we prepared this test to assert that,
+    if it happens, Convert2RHEL can detect that partial file there and instruct
+    the user on how to fix the problem.
+
+    .. note::
+        @lnykryn made an reproducer for the `cp` issue that can be seen here: https://gist.github.com/r0x0d/5d6a93c5827bd365e934f3d612fdafae
+    """
+    kernel_name = "kernel"
+    if "centos-8" in system_release or "oracle-8" in system_release:
```
danmyway, line 52:
```suggestion
def test_corrupted_initramfs_file(convert2rhel):
    """
    Verify if an output with a warning message is sent to the user in case of a
    corrupted initramfs file.

    This case can happen when the transaction ran successfully, but, for lack of
    disk space in the /boot partition, the kernel scriptlet will fail to copy
    the uncompressed initramfs file to /boot/initramfs-*.img, thus, leaving the
    file in an partial state and corrupted.

    Since this could be a real scenario, we prepared this test to assert that,
    if it happens, Convert2RHEL can detect that partial file there and instruct
    the user on how to fix the problem.

    .. note::
        @lnykryn made an reproducer for the `cp` issue that can be seen here: https://gist.github.com/r0x0d/5d6a93c5827bd365e934f3d612fdafae
    """
    kernel_name = "kernel"
    if "centos-8" in SYSTEM_RELEASE_ENV or "oracle-8" in SYSTEM_RELEASE_ENV:
```
Addition to the import.

convert2rhel/checks.py
```diff
@@ -835,3 +841,81 @@ def check_dbus_is_running():
...
+    # either case, so we don't care about checking for that here.
+    output, _ = run_subprocess(["rpm", "-q", "--last", kernel_name], print_output=False)
+
+    # We are parsing the latest kernel installed on the system, which at this
+    # point, should be a RHEL kernel. Since we can't get the kernel version
+    # from `uname -r`, as it requires a reboot in order to take place, we are
+    # detecting the latest kernel by using `rpm` and figuring out which was the
+    # latest kernel installed.
+    latest_installed_kernel = output.split("\n")[0].split(" ")[0]
+    latest_installed_kernel = latest_installed_kernel.split("%s-" % kernel_name)[-1]
```
abadger, line 888: I find this easier to understand but I've seen your way of doing this before too.  This is not a required change.

```suggestion
    latest_installed_kernel = latest_installed_kernel[len(kernel_name + "-"):]
```

convert2rhel/checks.py
```diff
@@ -835,3 +841,81 @@ def check_dbus_is_running():
...
+    # We always set this value to True, as a way of having an initial value.
+    # Always assuming the good path.
+    is_initramfs_valid = True
+    if os.path.exists(initramfs_file):
+        is_initramfs_valid = _is_initramfs_file_valid(initramfs_file)
+    else:
+        # Change it to False as if we don't have an initramfs file, then we
+        # shift this value to also reach the below if-statement and output the
+        # warning message.
+        is_initramfs_valid = False
```
abadger, line 906: If you reverse the initial value, you can get rid of the `else:` clause.

```python
    is_initramfs_valid = False
    if os.path.exists(initramfs_file):
        is_initramfs_valid = _is_initramfs_file_valid(initramfs_file)
```

Probably even better, though, would be to move the `os.path.exists()` check into `_is_initramfs_file_valid()` and then the logic here will be:
```suggestion
    is_initramfs_valid = _is_initramfs_file_valid(initramfs_file)

```

convert2rhel/checks.py
```diff
@@ -835,3 +841,81 @@ def check_dbus_is_running():
...
+    This method will rely on using lsinitrd to do the validation. If the
+    lsinitrd returns other value that is not 0, then it means that the file is
+    probably corrupted or may cause problems during the next reboot.
+
+    :param filepath: The path to the initramfs file.
+    :type filepath: str
+    :return: A boolean to determine if the file is corrupted.
+    :rtype: bool
+    """
+    logger.info("Checking if the '%s' file is not corrupted.", filepath)
```
abadger, line 858: See note for: https://github.com/oamg/convert2rhel/pull/721/files#diff-cf19bc90f65d9e26e1b6fa7dcd84a18dfba097f65706e211a4f4ac71b1b467f6R897 to see if you want to apply this suggestion
```suggestion
    logger.info("Checking if the '%s' file is valid.", filepath)
    
    logger.debug("Checking if the '%s' file exists.", filepath)    
    if not os.path.exists(filepath):
        return False
        
    logger.debug("Checking if the '%s' file is not corrupted.", filepath)
```

convert2rhel/checks.py
bocekm: > It is more so if it is not valid, and that message doesn't say specifically which is bad. Only that one of them are. We log for initramfs in its function but nothing for vmlinuz

True, good point.

I'd just add a full stop at the end of the messages. We still have a mix of messages with and without a full stop but from time to time I try to make them consistent and have the full stop (e.g. https://github.com/oamg/convert2rhel/pull/501/files).

Also, to be consistent with checking the initramfs file corruption, we can print only the message saying the vmlinuz file is being present (not that it is present).

```python
    logger.debug("Checking if the '%s' file exists.", vmlinuz_file)
    vmlinuz_exists = os.path.exists(vmlinuz_file)
    if not vmlinuz_exists:
        logger.info("The vmlinuz file is not present.")
```

BTW, again to be consistent, we can switch this initramfs message to a debug level:
```python
logger.info("Checking if the '%s' file is not corrupted.", filepath)
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1612,3 +1612,139 @@ def test_check_dbus_is_running_not_running(caplog, monkeypatch, global_tool_opts
     )
     assert log_msg == caplog.records[-1].message
     assert caplog.records[-1].levelname == "CRITICAL"
+
+
+@pytest.mark.parametrize(
+    ("latest_installed_kernel", "subprocess_output", "expected"),
+    (("6.1.7-200.fc37.x86_64", ("test", 0), True), ("6.1.7-200.fc37.x86_64", ("error", 1), False)),
```
abadger, line 1619:
```suggestion
    (("6.1.7-200.fc37.x86_64", ("test", 0), True),
     ("6.1.7-200.fc37.x86_64", ("error", 1), False),
     ),
```

danmyway (changes requested):
Just a couple of suggestions, otherwise I like it.
Thank you @r0x0d !

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1612,3 +1612,139 @@ def test_check_dbus_is_running_not_running(caplog, monkeypatch, global_tool_opts
...
+)
+def test_is_initramfs_file_valid(latest_installed_kernel, subprocess_output, expected, tmpdir, caplog, monkeypatch):
+    initramfs_file = tmpdir.mkdir("/boot").join("initramfs-%s.img")
+    initramfs_file = str(initramfs_file)
+    with open(initramfs_file % latest_installed_kernel, mode="w") as handler:
+        handler.write(latest_installed_kernel)
```
abadger, line 1625: Because we are monkeypatching `run_subprocess()`, it doesn't matter what is in this file, right?  I would do it like this so that it is apparent that there isn't something special about writing the string in `latest_installed_kernel` to the file.
```suggestion
    open(initramfs_file % latest_installed_kernel, mode="w").close()
```
(or, if one of the linters complains about that):

```suggestion
    with open(initramfs_file % latest_installed_kernel, mode="w"):
        # Just touch the file.  No contents needed.
        pass
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1612,3 +1612,139 @@ def test_check_dbus_is_running_not_running(caplog, monkeypatch, global_tool_opts
...
+    boot_folder = tmpdir.mkdir("/boot")
+    initramfs_file = boot_folder.join("initramfs-%s.img")
+    vmlinuz_file = boot_folder.join("vmlinuz-%s")
+    initramfs_file = str(initramfs_file)
+    vmlinuz_file = str(vmlinuz_file)
+
+    with open(initramfs_file % latest_installed_kernel, mode="w") as handler:
+        handler.write(latest_installed_kernel)
+    with open(vmlinuz_file % latest_installed_kernel, mode="w") as handler:
+        handler.write(latest_installed_kernel)
```
abadger, line 1651:
```suggestion
    open(initramfs_file % latest_installed_kernel, mode="w").close()
    open(vmlinuz_file % latest_installed_kernel, mode="w").close()
```

tests/integration/tier1/kernel-boot-files/test_handle_corrupted_files.py
r0x0d: updated.

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1612,3 +1612,139 @@ def test_check_dbus_is_running_not_running(caplog, monkeypatch, global_tool_opts
...
+            True,
+            ("test", 0),
+            ("kernel-core-6.1.8-200.fc37.x86_64 Wed 01 Feb 2023 14:01:01 -03", 0),
+            "6.1.8-200.fc37.x86_64",
+            id="initramfs-missing",
+        ),
+        pytest.param(
+            True,
+            True,
+            ("test", 1),
```
abadger, line 1691:
```suggestion
            ("error", 1),
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1612,3 +1612,139 @@ def test_check_dbus_is_running_not_running(caplog, monkeypatch, global_tool_opts
...
+                run_piped_subprocess,
+            ]
+        ),
+    )
+    boot_folder = tmpdir.mkdir("/boot")
+    if create_initramfs:
+        initramfs_file = boot_folder.join("initramfs-%s.img")
+        initramfs_file = str(initramfs_file)
+        with open(initramfs_file % latest_installed_kernel, mode="w") as handler:
+            handler.write(latest_installed_kernel)
```
abadger, line 1733:
```suggestion
        open(initramfs_file % latest_installed_kernel, mode="w").close()
```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -1612,3 +1612,139 @@ def test_check_dbus_is_running_not_running(caplog, monkeypatch, global_tool_opts
...
+
+        monkeypatch.setattr(checks, "INITRAMFS_FILEPATH", initramfs_file)
+    else:
+        monkeypatch.setattr(checks, "INITRAMFS_FILEPATH", "/non-existing-%s.img")
+
+    if create_vmlinuz:
+        vmlinuz_file = boot_folder.join("vmlinuz-%s")
+        vmlinuz_file = str(vmlinuz_file)
+        with open(vmlinuz_file % latest_installed_kernel, mode="w") as handler:
+            handler.write(latest_installed_kernel)
```
abadger, line 1743:
```suggestion
        open(vmlinuz_file % latest_installed_kernel, mode="w").close()
```

convert2rhel/unit_tests/grub_test.py
```diff
@@ -602,3 +602,13 @@ def test_update_grub_after_conversion(
     grub.update_grub_after_conversion()
     if expected is not None:
         assert expected in caplog.records[-1].message
+
+
+@pytest.mark.parametrize(
+    ("is_efi", "config_path"), ((False, "/boot/grub2/grub.cfg"), (True, "/boot/efi/EFI/redhat/grub.cfg"))
```
abadger, line 608:
```suggestion
    ("is_efi", "config_path"),
     ((False, "/boot/grub2/grub.cfg"),
     (True, "/boot/efi/EFI/redhat/grub.cfg"),
     )
```

abadger (approved):
A few suggestions that I think will make the code more readable but nothing that changes how the code works.  You can make the changes or not.

Approved.

convert2rhel/checks.py
```diff
@@ -835,3 +841,78 @@ def check_dbus_is_running():
...
+    :param filepath: The path to the initramfs file.
+    :type filepath: str
+    :return: A boolean to determine if the file is corrupted.
+    :rtype: bool
+    """
+    logger.info("Checking if the '%s' file is valid.", filepath)
+
+    logger.debug("Checking if the '%s' file exists.", filepath)
+    if not os.path.exists(filepath):
+        return False
```
bocekm, line 862:
```suggestion
        logger.info("The initramfs file is not present.")
        return False
```

convert2rhel/checks.py
```diff
@@ -835,3 +841,78 @@ def check_dbus_is_running():
...
+    # latest kernel installed.
+    latest_installed_kernel = output.split("\n")[0].split(" ")[0]
+    latest_installed_kernel = latest_installed_kernel[len(kernel_name + "-") :]
+    grub2_config_file = grub.get_grub_config_file()
+    initramfs_file = INITRAMFS_FILEPATH % latest_installed_kernel
+    vmlinuz_file = VMLINUZ_FILEPATH % latest_installed_kernel
+
+    logger.debug("Checking if the '%s' file exists", vmlinuz_file)
+    vmlinuz_exists = os.path.exists(vmlinuz_file)
+    logger.debug("Result of vmlinuz presence: %s", vmlinuz_exists)
```
bocekm, line 901:
```suggestion
    logger.info("Checking if the '%s' file exists.", vmlinuz_file)
    vmlinuz_exists = os.path.exists(vmlinuz_file)
    if not vmlinuz_exists:
        logger.info("The vmlinuz file is not present.")
```
To make it consistent with the initramfs check (info level message, print the result only with failure, full stops).

bocekm (changes requested):

bocekm (approved):

bocekm:
Tested in a VM and it works fine. I've created a followup issue https://issues.redhat.com/browse/RHELC-902.

bocekm:
@danmyway, FYI, I've re-run:

- testing-farm:centos-7-x86_64:tier1
- testing-farm:oraclelinux-7-x86_64:tier1

All issues were infrastructure related.

danmyway:
> @danmyway, FYI, I've re-run:
> 
>     * testing-farm:centos-7-x86_64:tier1
> 
>     * testing-farm:oraclelinux-7-x86_64:tier1
> 
> 
> All issues were infrastructure related.

I'd say waive these and merge if that happens again, if you agree @bocekm . The code is proved working by the tests implemented by @r0x0d.

danmyway (approved):
Looking good.
Failing tests unrelated to the changes.
Great job, thank you @r0x0d!

bocekm:
> Failing tests unrelated to the changes.

We incorrectly identified the test failures as unrelated to this PR. They were related and here's a fix of the test: https://github.com/oamg/convert2rhel/pull/748.

PR #722, opened 2023-02-02
[RHELC-882] Change the label for FILE output to DEBUG.

abadger:
The contents of FILE are the same as DEBUG.  The only difference is that FILE outputs to the log file whereas DEBUG goes to the terminal and the log file.  This means, in the log file, when we run convert2rhel with the --debug flag, the log messages are labelled with DEBUG.  When we run without --debug, the same messages end up in the log file but they have the label FILE.  Changing the label of LogLevelFile to DEBUG will keep the label the same for both of these cases.

Jira Issue: [RHELC-882](https://issues.redhat.com/browse/RHELC-882)

Checklist

### DIFF
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -82,7 +82,9 @@ class LogLevelTask(object):
 
 class LogLevelFile(object):
     level = 5
-    label = "FILE"
+    # Label messages DEBUG as it is contains the same messages as debug, just that they always go
+    # to the log file.
+    label = "DEBUG"
 
 
 def setup_logger_handler(log_name, log_dir):
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -23,7 +23,8 @@ WARNING   (30)    Prints warning message using date/time
 INFO      (20)    Prints info message (no date/time, just plain message)
 TASK      (15)    CUSTOM LABEL - Prints a task header message (using asterisks)
 DEBUG     (10)    Prints debug message (using date/time)
-FILE      (5)     CUSTOM LABEL - Prints only to file handler (using date/time)
+FILE      (5)     CUSTOM LABEL - Outputs with the DEBUG label but only to a file
+                  handle (using date/time)
 """
 import logging
 import os
--- a/convert2rhel/unit_tests/utils_test.py
+++ b/convert2rhel/unit_tests/utils_test.py
@@ -799,7 +799,7 @@ def test_hide_secret_unexpected_input(caplog):
         "--activationkey",
     ]
     assert len(caplog.records) == 1
-    assert caplog.records[-1].levelname == "FILE"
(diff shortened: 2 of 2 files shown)


r0x0d (dismissed):

convert2rhel/logger.py
```diff
@@ -82,7 +82,9 @@ class LogLevelTask(object):
 
 class LogLevelFile(object):
     level = 5
-    label = "FILE"
+    # Label messages DEBUG as it is contains the same messages as debug, just that they always go
+    # to the log file.
+    label = "DEBUG"
```
SpyTec, line 87: Docs at the top of the file needs to change. Seems to be the only thing left

convert2rhel/logger.py
```diff
@@ -82,7 +82,9 @@ class LogLevelTask(object):
 
 class LogLevelFile(object):
```
bocekm, line 83: Can't we remove this class altogether and make DEBUG level the default?

SpyTec: I wouldn't want the customer to default to DEBUG for CLI runs

convert2rhel/logger.py
r0x0d: nice catch, indeed.

convert2rhel/logger.py
abadger: It looks like the original intent was to output debug to the log file always but output debug to the screen only when `--debug` is passed.

convert2rhel/logger.py
abadger: Done.

pr-watson (commented):
The changes look good to me, approved. The failing integration tests appear unrelated to the changes.

convert2rhel/logger.py
bocekm: > I wouldn't want the customer to default to DEBUG for CLI runs. If so we might as well just change all DEBUG to INFO

@SpyTec, the default log level for stdout has been debug all the time:
https://github.com/oamg/convert2rhel/pull/722/files#diff-f82efabf7bb126100afd0472e3d8a5902da6576c6ddf75e3184cce9eb647c52aL116 

However when logging the debug level, there's a condition that diverts the message to a file only if the --debug is not used: https://github.com/oamg/convert2rhel/pull/722/files#diff-f82efabf7bb126100afd0472e3d8a5902da6576c6ddf75e3184cce9eb647c52aL205

I must say that it's unnecessarily convoluted solution.

> It looks like the original intent was to output debug to the log file always but output debug to the screen only when --debug is passed.

That's true. But I don't think we need a special class to make it happen. I'll try drafting a solution without using the LogLevelFile.

bocekm: https://github.com/oamg/convert2rhel/pull/741

PR #724, opened 2023-02-02
[RHELC-134] Refactor yum makecache to be separate from main checks

pr-watson:
This PR will separate the yum makecache call from check_custom_repos_are_valid() in checks.py to a different function, as well as moving clean_yum_metadata() from pkghandler.py. 

Jira Issue: [RHELC-134](https://issues.redhat.com/browse/RHELC-134)

Checklist


pr-watson:
@bocekm moving the new generate_yum_cache function from checks.py to repo.py causes a circular import due to the call_yum_cmd call which lives in pkghandler(). pkghandler() already imports backup.py and backup.py in turn imports repo.py, leading to a circular import. To resolve this we can either move call_yum_cmd out of pkghandler and into __utils__.py because run_subprocess() is already there or we can move get_hardcoded_repofiles_dir into systeminfo.py (or another suitable file). I have left the new generate_yum_cache function in checks.py for the time being. Please let me know how you would like to proceed, thanks!

bocekm:
Thanks for asking. Thinking about it, because it's about yum/dnf "internals", I'd move it to pkgmanager/__init__.py.
WDYT, @r0x0d?

r0x0d:
> `pkgmanager/__init__.py`

I agree with you, since it's related to the package managers itself, `pkgmanager` is the best suitable place for it. I would say that you can move it to where @bocekm point out, although, I don't like the idea to place too many things inside the `__init__.py` file, I will say that it is fine for now. 

We plan to do some type of "refactor" to this module in the future, so this could get moved and handled in another file/module as well. For now, moving it to `pkgmanager/__init__.py` is totally fine.

pr-watson:
Thanks everyone I will move them to __init__.py

convert2rhel/checks.py
```diff
@@ -342,7 +342,7 @@ def check_custom_repos_are_valid():
         return
 
     output, ret_code = call_yum_cmd(
-        command="makecache",
+        command="repolist",
         args=["-v", "--setopt=*.skip_if_unavailable=False"],
         print_output=False,
     )
```
r0x0d, line 348: I always got confused with this function. @bocekm should this check be related to the repos we received through `--enablerepo` in convert2rhel, or, what type of repo? Reading the function docstring it seems to be that we should check the ones that are provided by the user, not the ones that we have in the system.

convert2rhel/checks.py
```diff
@@ -356,6 +356,27 @@ def check_custom_repos_are_valid():
...
+    """
+    Without clearing the metadata cache, the `yum makecache` command may
+    return 0 (everything's ok) even when the baseurl of a repository is not
+    accessible. That would happen when the repository baseurl is changed but
+    yum still uses the previous baseurl stored in its cache.
+    call_yum_cmd(command="clean", args="metadata", print_output=False)
+    """
+
+    output, ret_code = call_yum_cmd(
+        command="makecache", args="-v --setopt=*.skip_if_unavailable=False", print_output=False
```
r0x0d, line 369:
```suggestion
        command="makecache", args=["-v", "--setopt=*.skip_if_unavailable=False"], print_output=False
```

convert2rhel/checks.py
```diff
@@ -356,6 +356,27 @@ def check_custom_repos_are_valid():
...
+    )
+    if ret_code != 0:
+        logger.warning(
+            "Unable to generate the cache for the yum repositories.\n"
+            "For more details, see YUM/DNF output:\n{0}".format(output)
+        )
+    else:
+        logger.info("Yum cache generated succesfully.")
```
r0x0d, line 377:
```suggestion
        logger.info("Repositories metadata cache generated successfully.")
```

convert2rhel/repo.py
```diff
@@ -169,3 +170,25 @@ def _restore_varsdir(path):
...
+    This is to make sure that Convert2RHEL works with up-to-date data from repositories before, for instance, querying
+    whether the system has the latest package versions installed, or before checking whether enabled repositories have
+    accessible URLs.
+    """
+    # We are using run_subprocess here as an alternative to call_yum_cmd
+    # which doesn't apply the correct --enablerepos option because at this point
+    # the tool didn't initialize the necessary functions in SystemInfo.
+    # The viable solution was calling the yum command as a subprocess manually
+    # instead of using that function wrapper.
+    output, ret_code = utils.run_subprocess(("yum", "clean", "metadata", "--quiet"), print_output=False)
```
r0x0d, line 187:
```suggestion
    output, ret_code = utils.run_subprocess(["yum", "clean", "metadata", "--quiet"], print_output=False)
```

convert2rhel/repo.py
```diff
@@ -169,3 +170,25 @@ def _restore_varsdir(path):
...
+    loggerinst.debug("Output of yum clean metadata:\n%s" % output)
+
+    if ret_code != 0:
+        loggerinst.warning("Failed to clean yum metadata:\n%s" % output)
```
r0x0d, line 191: Wondering if we should raise an `critical` here instead of warning. I would imagine that a broken metadata for some reason could be bad for the conversion, as the cache would have mixed states?

convert2rhel/repo.py
```diff
@@ -169,3 +170,25 @@ def _restore_varsdir(path):
...
+        return
+
+    loggerinst.info("Cached yum metadata cleaned successfully.")
```
r0x0d, line 194:
```suggestion
    loggerinst.info("Cached repositories metadata cleaned successfully.")
```

convert2rhel/repo.py
pr-watson: Let me change it to a critical

bocekm: We don't know all the cases where yum returns non-0 code.
For example when you disable all repos (`yum-config-manager --disable "*"`) then `yum clean metadata` returns 1. But is it something we should halt the conversion on? I don't think so. Having no repos enabled is a strange state but it might be valid from the customer's POV and we shouldn't block the conversion on that.

convert2rhel/checks.py
bocekm: This is indeed related to the RHEL repoid(s) provided by the user through the `--enablerepo` option. We check if they are accessible because if the provided repoids point to some custom repo location, let's say to the user's FTP server where they mirror the RHEL repos from Red Hat CDN, we want to make sure they have set it up correctly before proceeding. Or to spot even things like typos in the .repo file they've crafted.

convert2rhel/checks.py
```diff
@@ -342,7 +342,7 @@ def check_custom_repos_are_valid():
         return
 
     output, ret_code = call_yum_cmd(
-        command="makecache",
+        command="repolist",
```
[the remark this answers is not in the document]
bocekm: Also, before calling yum makecache, we need to be sure the cache does not exist otherwise yum would skip making it: https://github.com/oamg/convert2rhel/pull/196#discussion_r609974562.

You could tell me _"but hey, we're cleaning the cache in the `clean_yum_metadata()`"_. That's true but it cleans the cache just for the repos that are *enabled* on the system.

Now consider this scenario:
- I'm on a CentOS Linux 7 system and I'm preparing it for conversion to RHEL 7
- I have a mirror of RHEL 7 repo on an FTP server available through my internal company network
- I create a /etc/yum.repos.d/rhel7-on-our-ftp.repo file where I define a `rhel-7-server-rpms` repoid pointing to the RHEL 7 repo on my on-premise FTP server
- I make the repo disabled by default (`enabled=0`)
- I check whether the repo definition is ok by running `yum info kernel --setopt=reposdir=/home/admin/conversion/ --enablerepo=rhel-7-server-rpms` and it works ok. That command generates a yum cache for this repo
- Couple of hours later I run convert2rhel, not knowing the FTP server is down
- convert2rhel runs `yum clean metadata` but that cleans cache of just the enabled repos on the system and the `rhel-7-server-rpms` is not enabled
- convert2rhel runs this custom repo availability check in which it runs `yum makecache -v --setopt=*.skip_if_unavailable=False --disablerepo="*" --enablerepo=rhel-7-server-rpms` and it returns 0 because it uses the cache generated hours ago

@abadger, do you think is my scenario valid? If it is I'd recommend cleaning the cache either in the `clean_yum_metadata() and using `--enablerepo="*"` or by running another `yum clean metadata` right before this `yum makecache` with using the `--enablerepo` option.

bocekm (dismissed):

convert2rhel/repo.py
pr-watson: Noted

convert2rhel/checks.py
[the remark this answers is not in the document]
pr-watson: I found the suggestion in a gist from the Jira ticket. It would replace the verification of repos that makecache previously did. Is there a better command that can be run it it's place @bocekm ?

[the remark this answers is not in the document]
bocekm: Ah, ok. @r0x0d, why have you decided to replace makecache with repolist in https://gist.github.com/r0x0d/f645a1f7b005ee96e9421954e27528db?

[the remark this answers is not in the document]
r0x0d: Not sure, too, @bocekm. It's been quite a while since I did this gist, but I think that my thinking back then was: 

1. We need to check for repo availability, so, calling repolist could work 
2. If we need to pass the `--enablerepo` and `--disablerepo` to that call, we could extend it to check only that. 
3. Back then, I thought that makecache would not verify the repositories, only generate the caches
4.  I slightly remember we were discussing this over some meeting and realizing that `repolist` might not be the best thing here, probably another command could do the trick for us
5.  I haven't investigated further since then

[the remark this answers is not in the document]
bocekm: Based on https://github.com/oamg/convert2rhel/pull/196#discussion_r588298250 it seems that `repolist` returns 0 even when there's an inaccessible repo. So I'd stick to makecache which we know returns 1 in such a case.
The `--enablerepo` is passed to the `yum makecache` within the `call_yum_cmd()` function.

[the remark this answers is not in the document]
pr-watson: @bocekm so since we are removing the make cache call from check_custom_repos_are_valid() in checks.py, can we just get rid of that function all together?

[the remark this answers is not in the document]
pr-watson: @bocekm and do we no longer want to generate the yum cache after cleaning the yum cache and before the system checks?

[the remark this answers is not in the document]
bocekm: I believe that we don't need to generate the yum cache "manually" because whatever yum/repoquery/yumdownloader command we call later triggers the cache generation automatically.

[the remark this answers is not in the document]
pr-watson: Noted thank you

convert2rhel/checks.py
r0x0d: Ah, I see. 

Yeah, having the enable repos being enabled under the hood is quite confusing sometimes. Maybe in the future we could make it more explicit and bring that up?

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -34,8 +34,8 @@
 )
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
-from convert2rhel.unit_tests import GetLoggerMocked, is_rpm_based_os, run_subprocess_side_effect
-from convert2rhel.unit_tests.conftest import TestPkgObj, all_systems, centos8, create_pkg_obj
+from convert2rhel.unit_tests import GetLoggerMocked, is_rpm_based_os
+from convert2rhel.unit_tests.conftest import TestPkgObj, all_systems, centos7, centos8, create_pkg_obj
```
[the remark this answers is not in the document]
r0x0d: @pr-watson

convert2rhel/pkgmanager/__init__.py
```diff
@@ -58,3 +65,27 @@ def create_transaction_handler():
...
+    whether the system has the latest package versions installed, or before checking whether enabled repositories have
+    accessible URLs.
+    """
+    # We are using run_subprocess here as an alternative to call_yum_cmd
+    # which doesn't apply the correct --enablerepos option because at this point
+    # the tool didn't initialize the necessary functions in SystemInfo.
+    # The viable solution was calling the yum command as a subprocess manually
+    # instead of using that function wrapper.
+    output, ret_code = utils.run_subprocess(
+        ("yum", "clean", "metadata", "--enablerepo=*", "--quiet"), print_output=False
```
r0x0d, line 83: All the other calls we have for run_subprocess are passing a list instead of a tuple. I think it should be good to follow the pattern there and change this one too. 

```suggestion
        ["yum", "clean", "metadata", "--enablerepo=*", "--quiet"], print_output=False
```

r0x0d (approved):
Looks very nice, @pr-watson! Just left one or two small comments.

convert2rhel/pkgmanager/__init__.py
```diff
@@ -63,3 +69,27 @@ def create_transaction_handler():
...
+
+    This is to make sure that Convert2RHEL works with up-to-date data from repositories before, for instance, querying
+    whether the system has the latest package versions installed, or before checking whether enabled repositories have
+    accessible URLs.
+    """
+    # We are using run_subprocess here as an alternative to call_yum_cmd
+    # which doesn't apply the correct --enablerepos option because at this point
+    # the tool didn't initialize the necessary functions in SystemInfo.
+    # The viable solution was calling the yum command as a subprocess manually
+    # instead of using that function wrapper.
```
r0x0d, line 85: The wording in the comment seems odd. I can imagine that the reason we call run_subprocess here is that because at the point in time we are calling this function in `main`, we don't have the `SystemInfo` initialized yet, but, if for some reason I decide to call this function just before validating the main transaction, then `SystemInfo` would be already initialized, and then the constraint here would not be valid anymore. 

```suggestion
    # We are using run_subprocess here as an alternative to call_yum_cmd
    # which doesn't apply the correct --enablerepos option because, if we call this
    # earlier in the conversion the tool didn't initialize the necessary functions in SystemInfo.
    # The viable solution was calling the yum command as a subprocess manually
    # instead of using that function wrapper.
```

What do you think?

r0x0d (commented):
Only a minor comment.

convert2rhel/pkgmanager/__init__.py
pr-watson: That's a good point, I'll add this and change "didn't" to "doesn't"

r0x0d (approved):

PR #723, opened 2023-02-02
[RHELC-849, RHELC-850] Inform+disable telemetry

hosekadam:
The user is informed about the data we upload to Red Hat servers. They are asked to acknowledge the data collection and in case they do not like it we've added the ability to disable it.

Jira Issue: [RHELC-849](https://issues.redhat.com/browse/RHELC-849)
Jira Issue: [RHELC-850](https://issues.redhat.com/browse/RHELC-850)

Checklist
- [x] PR meets acceptance criteria specified in the Jira issue
- [x] PR has been tested manually in a VM (either author or reviewer)
- [x] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [x] Code and tests are documented properly
- [ ] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending`

PR #740, opened 2023-02-17
Fix CodeQL warnings

r0x0d:
This commit introduces a bunch of fixes for the warnings/notices that CodeQL was showing for Convert2RHEL under the the `Security` tab.

Checklist

### DIFF
--- a/convert2rhel/pkgmanager/__init__.py
+++ b/convert2rhel/pkgmanager/__init__.py
@@ -65,10 +65,10 @@ def create_transaction_handler():
         from convert2rhel.pkgmanager.handlers.yum import YumTransactionHandler
 
         return YumTransactionHandler()
-    elif TYPE == "dnf":
-        from convert2rhel.pkgmanager.handlers.dnf import DnfTransactionHandler
 
-        return DnfTransactionHandler()
+    from convert2rhel.pkgmanager.handlers.dnf import DnfTransactionHandler
+
+    return DnfTransactionHandler()
 
 
 def clean_yum_metadata():
--- a/convert2rhel/unit_tests/conftest.py
+++ b/convert2rhel/unit_tests/conftest.py
@@ -16,29 +16,6 @@ if sys.version_info[:2] <= (2, 7):
 else:
     from unittest import mock  # pylint: disable=no-name-in-module
 
-try:
-    from pytest_catchlog import CompatLogCaptureFixture
-
-    class SubCompatLogCaptureFixture(CompatLogCaptureFixture):
-        @property
-        def messages(self):
-            return [r.getMessage() for r in self.records]
-
-    @pytest.fixture
-    def caplog(request):
-        """Access and control log capturing.
-        Captured logs are available through the following properties/methods::
-        * caplog.messages        -> list of format-interpolated log messages
-        * caplog.text            -> string containing formatted log output
-        * caplog.records         -> list of logging.LogRecord instances
-        * caplog.record_tuples   -> list of (logger_name, level, message) tuples
-        * caplog.clear()         -> clear captured records and formatted log output string
-        """
-        return SubCompatLogCaptureFixture(request.node)
-
-except ImportError:
-    pass
-
 # We are injecting a instance of `mock.Mock()` for `Depsolve` class and
 # `callback` module, as when we run the tests under CentOS 7, it fails by saying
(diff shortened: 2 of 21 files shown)


r0x0d:
No rush for reviewing here.

convert2rhel/unit_tests/conftest.py
```diff
@@ -16,29 +16,6 @@
 else:
     from unittest import mock  # pylint: disable=no-name-in-module
 
-try:
-    from pytest_catchlog import CompatLogCaptureFixture
```
bocekm, line 20: Did we put it here for python 2.6 compatibility?

bocekm (approved):

convert2rhel/unit_tests/conftest.py
r0x0d: Yeah, that is correct. Forgot to add a comment here in the PR mentioning it was for 2.6, and now it can be removed. 

I think I added that try-except during the F2F that happened in Michigan. 

But yeah, since we will not run this under 2.6 anymore, this code can be removed.

bocekm (changes requested):
TODO: Remove a failed test identified at https://github.com/oamg/convert2rhel/pull/723#discussion_r1111971167.

convert2rhel/pkgmanager/__init__.py
```diff
@@ -65,10 +65,10 @@ def create_transaction_handler():
         from convert2rhel.pkgmanager.handlers.yum import YumTransactionHandler
 
         return YumTransactionHandler()
-    elif TYPE == "dnf":
-        from convert2rhel.pkgmanager.handlers.dnf import DnfTransactionHandler
 
-        return DnfTransactionHandler()
+    from convert2rhel.pkgmanager.handlers.dnf import DnfTransactionHandler
+
+    return DnfTransactionHandler()
```
abadger, line 71: Just a note, since you were asking about how to rewrite code that requires either dnf or yum in one of our 1::1s.  This can be done at the toplevel  with something like:
``` python
try:
    from yum import *
    # [...]
    from convert2rhel.pkgmanager.handlers.yum import YumTransactionHandler as TransactionHandler
except ImportError:
    from convert2rhel.pkgmanager.handlers.dnf import DnfTransactionHandler as TransactionHandler
```

The naming (`TransactionHandler vs DnTransactionHandler) could be done inside of the specific handlers instead of in this file as well.

(This should not be done in this PR.  Just seeing this reminded me of our discussion.)

man/build_manpage.py
```diff
@@ -142,9 +142,8 @@ def run(self):
         manpage.append(self._write_header())
         manpage.append(self._write_options())
         manpage.append(self._write_footer())
-        stream = open(self.output, "w")
-        stream.write("".join(manpage))
-        stream.close()
+        with open(self.output, mode="w") as stream:
+            stream.write("".join(manpage))
```
abadger, line 146: This build script is just copied into our repository from https://github.com/andialbrecht/build_manpage  In my pylint cleanup patch, I excluded this file instead of fixing it in our copy so that we didn't get out of sync.  OTOH, the upstream looks dead so maybe changing it is the right thing to do...

(No changes needed, just something to think about).

bocekm: I'd not be worried about diverging since as you mentioned it's not maintained anymore.

convert2rhel/pkgmanager/__init__.py
r0x0d: I think that it would cause a recursive import, as the `{Yum,Dnf}TransactionHandler` import from pkgmanager to do a lot of operations. 

But once we have figured up a good way of transitioning to the refactor of yum/dnf, we can take care of that.

r0x0d:
> TODO: Remove a failed check identified at [#723 (comment)](https://github.com/oamg/convert2rhel/pull/723#discussion_r1111971167).

Done.

bocekm (approved):

r0x0d:
Rebased to trigger the tests once more. Probably if they fail for some reason will not be related to the PR itself, but just in case, triggering this again.

r0x0d:
Failures do not seem to be related to the changes made here.

r0x0d:
Merging this as it seems that the failures are not related to the PR, but rather some infrastructure problems.

PR #741, opened 2023-02-17
Remove unnecessary FILE log level

bocekm:
Having a custom log level was unnecessarily complex and the routing of debug messages to the stdout was not intuitive.
Using a Filter to decide whether a debug level message should be printed to stdout based on the use of the --debug option is cleaner.

Checklist


convert2rhel/logger.py
```diff
@@ -81,52 +79,47 @@ class LogLevelTask(object):
...
     # create sys.stdout handler for info/debug
     stdout_handler = logging.StreamHandler(sys.stdout)
     formatter = CustomFormatter("%(message)s")
     formatter.disable_colors(should_disable_color_output())
     stdout_handler.setFormatter(formatter)
+    debug_flag_filter = DebugFlagFilter()
+    stdout_handler.addFilter(debug_flag_filter)
+    # Set the DEBUG level as the lowest allowed level to be printed to stdout.
+    # Whether a debug message is actually printed or not is decided in DebugFlagFilter.
     stdout_handler.setLevel(logging.DEBUG)
```
abadger, line 112: Does it not work to set this to `logging.DEBUG` if `--debug` is given, otherwise set it to `loggin.INFO`?  (If the problem is a circular dependency [Need toolopts to know if `--debug` was given but toolopts requires that logging has been setup in case it needs to log its values, the way that I've setup logging in other projects is to have two phases.  Phase 1 sets it up with default values.  Then parse the command line, read the config file, consult environment variables, etc.  Then Phase 2, reconfigure portions of logging that user settings can affect.

That's a bigger change than this, but I think it's better structure.  And at the risk of premature optimization, it's also less work for the computer.  Adjusting the log level results in less work being done if the log message is at a less important level.  Setting a filter results in more work being done for every log message emitted.

convert2rhel/logger.py
```diff
@@ -143,6 +136,18 @@
     return False
 
 
+class DebugFlagFilter(logging.Filter):
+    """Print debug messages to the stdout only when --debug is used."""
+
+    def filter(self, record):
+        from convert2rhel.toolopts import tool_opts
```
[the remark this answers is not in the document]
abadger: The circular import would also go away with the restructure to have an early log initialization and later log initialization. (If we put the early log initialization into the initialize.py file and leave the second configuration phase with the `import convert2rhel.toolopts` in this file)

convert2rhel/logger.py
```diff
@@ -143,6 +136,18 @@ def should_disable_color_output():
...
+
+        if record.levelno == logging.DEBUG and not tool_opts.debug:
+            # not logging a debug level message if the --debug option hasn't been used
+            return False
+        return True
```
SpyTec, line 148: Ideally I'd want to set this from within toolopts somehow no?

Venefilyn:
@bocekm want to delegate this to someone else?

PR #752, opened 2023-02-27
[RHELC-832] Add ability to do micro releases

pr-watson:

This PR adds the micro release field to the convert2rhel version. All associated unit tests that deal with the convert2rhel version have been modified to accept the new release field. The function in checks.py that checks if convert2rhel is at the latest version has been modified to no longer use a single regex to parse package strings and instead uses the new parsing function made in https://github.com/oamg/convert2rhel/pull/713 . The convert2rhel rpm has been generated with this new micro release field and it did not cause any issues. The spec file and __version__ field in __init__.py have also been updated with the new micro release field

Jira Issue: [RHELC-832](https://issues.redhat.com/browse/RHELC-832)

Checklist


SpyTec:
@pr-watson I did a quick rebase to fix the merge conflict

convert2rhel/unit_tests/checks_test.py
```diff
@@ -334,17 +431,28 @@ def test_c2r_up_to_date_multiple_packages(self, caplog, convert2rhel_latest_vers
         with pytest.raises(SystemExit):
             checks.check_convert2rhel_latest()
 
+        local_version, package_version = convert2rhel_latest_version_test
+        print(len(package_version))
+        if len(package_version) > 110:
+            package_version = package_version[97:103]
+        else:
+            package_version = package_version[93:97]
+        print(package_version)
```
SpyTec, line 440: Debug prints here should be removed
```suggestion

        if len(package_version) > 110:
            package_version = package_version[97:103]
        else:
            package_version = package_version[93:97]

```

convert2rhel/unit_tests/checks_test.py
```diff
@@ -283,13 +311,64 @@ def test_convert2rhel_latest_log_check_env(self, caplog, monkeypatch, convert2rh
...
+        ),
+        indirect=True,
+    )
+    def test_convert2rhel_latest_value_error_yum(self, caplog, convert2rhel_latest_version_test):
+        checks.check_convert2rhel_latest()
+
+        local_version, package_version = convert2rhel_latest_version_test
+        # this is a test for a debug message, there is no other way to tell we had a ValueError in the parsing function
+        log_msg = "The following field(s) are invalid -"
+        assert log_msg in caplog.text
```
SpyTec, line 360: Due to the comment about debug it seems like this test is asserting debug log messages. That is bad practice so we should not do it. Debug messages should not be tested unless they can spew out sensitive stuff we need to check for

pr-watson: @SpyTec I added this test so we have coverage of the try/except blocking catching a ValueError that gets raised in the parsing function around line 166 in checks.py. Codecov will complain if I remove it, is that ok?

SpyTec: We only care about the catch to log to debug anyway so yeah feel free. Code coverage is only a metric to see how we're doing, it is never a goal

convert2rhel/checks.py
```diff
@@ -136,22 +137,58 @@ def check_convert2rhel_latest():
...
+    # relevant repoquery information to our check, otherwise, we just log the
+    # information as debug and do nothing with it.
+    for raw_version in raw_output_convert2rhel_versions:
+        if "C2R" in raw_version:
+            temp_raw_output.append(raw_version.lstrip("C2R "))
+        else:
+            # Mainly for debugging purposes to see what is happening if we got
+            # anything else that does not have the C2R identifier at the start
+            # of the line.
+            logger.debug("Got a line without the C2R identifier: %s" % raw_version)
```
SpyTec, line 158: Same with this, we should not have tests that assert this @pr-watson. If there are any added I didn't see them, but same applies there

convert2rhel/unit_tests/checks_test.py
pr-watson: Understood, thanks! I'll go ahead and remove those tests

convert2rhel/checks.py
pr-watson: I'll remove this as well

SpyTec:
Code seems good to me, does what it is supposed to do. Need int. tests to be updated then we can merge

danmyway:
The tests have been modified @pr-watson

danmyway:
The only test failing is the offline_conversion.
That is due to satellite repos not being synced, having an older version of kernel in the repositories.

SpyTec:
> The only test failing is the offline_conversion.
> That is due to satellite repos not being synced, having an older version of kernel in the repositories.

Ack feel free to merge

danmyway:
> > The only test failing is the offline_conversion.
> > That is due to satellite repos not being synced, having an older version of kernel in the repositories.
> 
> Ack feel free to merge

What about the codecov failing @SpyTec ?

SpyTec:
> > > The only test failing is the offline_conversion.
> > > That is due to satellite repos not being synced, having an older version of kernel in the repositories.
> > 
> > Ack feel free to merge
> 
> What about the codecov failing @SpyTec ?
> 

Shouldn't be seen as a blocker. I explicitly said to not test debug statements, which is why it's failing. It's just to help with development but is not a goal or metric to use

danmyway:
> > > > The only test failing is the offline_conversion.
> > > > That is due to satellite repos not being synced, having an older version of kernel in the repositories.
> > > 
> > > 
> > > Ack feel free to merge
> > 
> > 
> > What about the codecov failing @SpyTec ?
> 
> Shouldn't be seen as a blocker. I explicitly said to not test debug statements, which is why it's failing. It's just to help with development but is not a goal or metric to use

Cool, thanks @SpyTec
@kokesak, can you take a look please, so we can merge?

tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks.py
```diff
@@ -132,10 +152,6 @@ def test_c2r_latest_older_unsupported_version(convert2rhel):
         c2r.sendline("n")
     assert c2r.exitstatus != 0
 
-    # Clean up
-    os.system(f"cp /tmp/__init__.py {PATH_TO_VERSION}")
-    del os.environ["CONVERT2RHEL_ALLOW_OLDER_VERSION"]
```
kokesak, line 137: I would personally keep the `del os.environ["CONVERT2RHEL_ALLOW_OLDER_VERSION"]` here instead of in the `_restore_c2r_version()` as it is not something I would expect the "restore version" to do.

kokesak (commented):
Only one small thing, otherwise it looks nice

tests/integration/tier0/basic-sanity-checks/test_basic_sanity_checks.py
danmyway: That is a valid point, however I would rather keep it in the fixture.
So I will take it out of the `_restore_c2r_version()` function and put it after the function call, WDYT @kokesak ?

danmyway: On a second thought and after a discussion, we will keep the environment variable actions (setting and deleting) inside the test for the sake of readability.
It doesn't really make much sense to have the envar set in the test (or even in the test metadata, which was contemplated for the future) and deleted by the fixture.

danmyway: Final thought.
Create a simple fixture to handle the envar.
In case the test fails, the removal will be handled anyway, not interfering with subsequent tests.

kokesak: Alright, why not @danmyway

PR #754, opened 2023-02-28
[RHELC-869] chore: remove el6 related code

SpyTec:
All current code regarding conversion to RHEL 6 are to be removed in
this commit. Tests have verified to be working, might have missed
something still but at least 90% of everything regarding el6 is gone

Jira Issue: [RHELC-869](https://issues.redhat.com/browse/RHELC-869)

Checklist

### DIFF
--- a/ansible/setup-convert2rhel.yml
+++ b/ansible/setup-convert2rhel.yml
@@ -26,12 +26,6 @@
         msg: "Distribution must be CentOS or Oracle Linux"
       when: ansible_distribution != 'CentOS' and ansible_distribution != 'OracleLinux'
 
-    - name: Validate Release Version (6.10, 7.9, 8.5)
-      fail:
-        msg: "Distribution version must be 6.10, 7.9, or 8.5"
-      when: (ansible_distribution_version != '6.10' and ansible_distribution_version != '7.9'
-             and ansible_distribution_version != '8.5') and skip_os_version_check|bool
-
     - name: Patch repositories if dealing with CentOS8.
       block:
       - name: Comment mirrorlist in all repos.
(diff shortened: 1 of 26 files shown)


bocekm:
@SpyTec, I've added a commit removing a couple more RHEL 6/python 2.6-related bits.

ansible/setup-convert2rhel.yml
```diff
@@ -26,12 +26,6 @@
         msg: "Distribution must be CentOS or Oracle Linux"
       when: ansible_distribution != 'CentOS' and ansible_distribution != 'OracleLinux'
 
-    - name: Validate Release Version (6.10, 7.9, 8.5)
```
bocekm, line 29: I've removed this check as it's difficult to maintain (we forgot to update it many times) and it's not critical.

convert2rhel/utils.py
```diff
@@ -158,10 +158,6 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
         if print_output:
             loggerinst.info(line.rstrip("\n"))
 
-    # Call communicate() to wait for the process to terminate so that we can get the return code by poll().
-    # It's just for py2.6, py2.7+/3 doesn't need this.
-    process.communicate()
-
```
SpyTec, line 164: @bocekm this seems to still be needed for py2.7 and py3. Both unit tests fail due to this removal. Do you know why that is before I restore?

bocekm: True. The comment was wrong. When looking at the documentation of `poll()` it doesn't wait for the process to finish. `wait()` does but there's a potential deadlock when using `stdout=PIPE` which we do, so it's better to wait for the process to finish through calling `communicate()` and then `poll()` to get the return code.

SpyTec: Done

bocekm (approved):

PR #755, opened 2023-03-01
Fix failing system_up_to_date test

danmyway:
* sometimes the test for non-latest kernel inhibition fails, due to `kvm` kernel module getting tainted in the process of downgrading the kernel package on Oracle Linux 8.7
* for the sake of the test, detect if `kvm` module is marked tainted and remove it (and `kvm_intel` as dependent module)

### DIFF
--- a/tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
+++ b/tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
@@ -1,3 +1,5 @@
+import re
+
 from conftest import SYSTEM_RELEASE_ENV
 from envparse import env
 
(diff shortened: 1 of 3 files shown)


tests/integration/tier1/system-up-to-date/test_non_latest_kernel_inhibitor.py
```diff
@@ -29,6 +31,16 @@ def test_non_latest_kernel(shell, convert2rhel):
...
 
+    # Downgrading the kernel on Oracle Linux 8.7 sometimes causes
+    # the kernel module kvm to be marked as tainted.
+    # Check if the kmod is marked tainted and remove it if so.
+    # Remove also kvm_intel as it is dependent on kvm kernel module.
+
+    kmods_output = shell("cat /proc/modules | grep kvm").output
+    tainted_kvm = re.compile(r"(.*\n)*(kvm ).*\(t\)")
+    if "oracle-8" in SYSTEM_RELEASE_ENV and re.match(tainted_kvm, kmods_output):
+        shell("rmmod kvm_intel && rmmod kvm")
```
kokesak, line 42: I would separate those commands into two. If the `kvm_intel` module is not present (probably due to its name, it is only available on x86_64 architectures) then the `kvm` will not get removed.
Another option would be to detect the running arch and if it is `x86_64` then try also to removing the `kvm_intel`

kokesak (changes requested):

danmyway:
The issue is not popping up anymore, closing this Pull Request.

PR #759, opened 2023-03-06
[RHELC-885] Fix Ctrl + C during conversion not starting rollback during certain tasks 

r0x0d:
Fix a bug where we couldn't issue a Ctrl + C (SIGINT) to and start a rollback immediately.
    
This commit introduces the ability to run some functions as a child process to the main process in order to be able to make `rpm` and `yum`

Jira Issue: [RHELC-885](https://issues.redhat.com/browse/RHELC-885)

Checklist

---

# Scripts that can be used for testing purposes: 

- [Script to check if the Ctrl + C handling is working in the main thread](https://gist.github.com/r0x0d/5bbea149e11dc06aefd80282c3dc0d93)
- [Script to send a SIGINT (Ctrl + C) signal to Convert2RHEL](https://gist.github.com/r0x0d/1a96a92f2e5f4b03a02eb77b04cc091b)

The usage of those scripts are not that complicated, but feel free to ping me in order to see how to test it.


convert2rhel/utils.py
```diff
@@ -66,6 +69,141 @@ class Color(object):
...
+        except (Exception, SystemExit) as e:
+            tb = traceback.format_exc()
+            self._cconn.send((e, tb))
+
+    @property
+    def exception(self):
+        """Custom property to access the exception raised by the children."""
+        if self._pconn.poll():
+            self._exception = self._pconn.recv()
+        return self._exception
```
r0x0d, line 100: Not sure if it is necessary to do unit tests for this class. Let me know what you all think about.

convert2rhel/utils.py
```diff
@@ -66,6 +69,141 @@ class Color(object):
...
+        queue = multiprocessing.Queue()
+        process_args = (queue,)
+        if args:
+            args += process_args
+            process_args = args
+
+        # TODO(r0x0d): I don't think that creating the process can cause an
+        # exception that we should handle here, only if we mess something with
+        # the process_args tuple, but that should be know during development
+        # time. In any case, check this later.
```
r0x0d, line 172: @abadger are you aware if this could raise an exception that we should catch? I don't remind of any.

convert2rhel/utils.py
```diff
@@ -66,6 +69,141 @@ class Color(object):
...
+
+        :return: If the Queue is not empty, return anything that on it,
+        otherwise, returns `None`.
+        :rtype: Any
+        """
+        queue = multiprocessing.Queue()
+        process_args = (queue,)
+        if args:
+            args += process_args
+            process_args = args
```
r0x0d, line 167: Checking with @abadger on how to make it better and use the `kwargs`.

convert2rhel/utils.py
```diff
@@ -66,6 +69,141 @@ class Color(object):
...
+        self._pconn, self._cconn = multiprocessing.Pipe()
+        self._exception = None
+
+    def run(self):
+        """Overrides the `run` method to catch exceptions raised in child."""
+        try:
+            multiprocessing.Process.run(self)
+            self._cconn.send(None)
+        # Here, `SystemExit` is inherit from `BaseException`, which is too
+        # broad to catch as it involves for-loop exceptions. The ideal here
```
pr-watson, line 88:
```suggestion
        # broad to catch as it involves for-loop exceptions. The idea here
```
Unsure if you meant ideal or idea

pr-watson:
Looks good to me, after the code covs are handled and the remaining comments are resolved I'd say this PR is ready

convert2rhel/utils.py
r0x0d: It's `idea`. Thanks!

convert2rhel/utils.py
```diff
@@ -158,12 +296,9 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
         if print_output:
             loggerinst.info(line.rstrip("\n"))
 
-    # Call communicate() to wait for the process to terminate so that we can get the return code by poll().
-    # It's just for py2.6, py2.7+/3 doesn't need this.
-    process.communicate()
+    process.wait()
```
bocekm, line 299: Do you think that we're safe from the deadlock described in https://docs.python.org/3.6/library/subprocess.html#subprocess.Popen.wait:

> This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use [Popen.communicate()](https://docs.python.org/3.6/library/subprocess.html#subprocess.Popen.communicate) when using pipes to avoid that.

We use `stdout=PIPE`.

bocekm (changes requested):

convert2rhel/utils.py
r0x0d: Hmm, good catch! I think communicate should still be used. I will revert to that.

convert2rhel/utils.py
```diff
@@ -66,6 +69,141 @@ class Color(object):
...
+        child process and leave the parent one with the original signal
+        handling that is initiated by python itself (or, whatever signal is
+        registered when the conversion starts as well).
+
+    .. warning::
+        This decorator will use a instance of `multiprocessing.Queue` to
+        communicate between the processes, so, if this decorator is being used
+        in a function that should return values, it will need to adapt to the
+        queue workflow, and instead of returning it directly, just put the
+        value inside the queue and the decorator will pop it out once the
```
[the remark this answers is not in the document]
r0x0d: Is the ` a return value` there correct? It should be `return value`, right?

convert2rhel/pkghandler.py
```diff
@@ -1027,15 +1054,34 @@ def get_total_packages_to_update(reposdir):
...
+    """Query all the packages with yum that have an update pending on the system.
+
+    .. warning::
+        This function is being executed in a child process so we will be able
+        to raise SIGINT or any other signal that is sent to the main process.
+
+        The function calls here does not affect the others subprocesses calls
+        that are called after this function during the conversion, but, it does
+        affect the signal handling while the user tries to send that signal
+        while this function is executing.
```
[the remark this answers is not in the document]
r0x0d: Agreed with you, makes more sense your paragraph.

convert2rhel/utils.py
```diff
@@ -66,6 +69,141 @@ class Color(object):
...
+        :type kwargs: dict
+
+        :raises KeyboardInterrupt: Raises a `KeyboardInterrupt` if a SIGINT is
+            catched during the execution of the child process. This exception
+            will be re-raised to the stack once it is catched.
+        :raises Exception: Raise any general exception that can occurs during
+            the execution of the child process.
+
+        :return: If the Queue is not empty, return anything that on it,
+        otherwise, returns `None`.
```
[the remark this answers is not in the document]
r0x0d: For some reason, this comment got out of the UI for me. Applying this manually.

convert2rhel/utils.py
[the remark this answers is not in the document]
bocekm: Ah, sorry, you're right. It should be `return a value`.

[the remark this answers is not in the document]
r0x0d: Just removed that whole paragraph as it was about the old way that should adapt the functions to have a queue parameter, since it's changed, it is not needed anymore.

convert2rhel/pkghandler.py
```diff
@@ -301,8 +301,45 @@ def get_pkg_fingerprint(pkg_obj):
...
+        the closing of the rpmdb instance to prevent further problems when
+        using Yum in a CLI, or, when there is a call to the Yum python API.
+
+    :param pkg_obj: Instace of a package RPM installed on the system.
+    :type pkg_obj: yum.rpmsack.RPMInstalledPackage
+    :return: Return the package signature.
+    :rtype: str
+    """
+    with pkgmanager.rpm_db_lock(pkg_obj):
+        return get_pkg_fingerprint(pkg_obj)
```
SpyTec, line 332: Could we not use `get_pkg_signature()` here given the name of these two methods being nearly the same?

r0x0d: `get_pkg_signature()` is called a lot of times sometimes, especially if we have a lot of packages to verify, and since we are closing the database by dropping all the cache (as well, as a child process), I created an wrapper function just to this specific part. 

Probably the name of the wrapper function could be improved, but not sure of what to call this.

bocekm: We should not change what we output in the breadcrumbs data without a good reason. With using `get_pkg_fingerprint()` instead of `get_pkg_signature()` you'd change the value of the `signature` breadcrumbs field from `kernel RSA/SHA256, Wed 15 Feb 2023 03:47:41 PM CET, Key ID f55ad3fb5323552a` to `f55ad3fb5323552a`.

```suggestion
    with pkgmanager.rpm_db_lock(pkg_obj):
	    return get_pkg_signature(pkg_obj)
```

convert2rhel/pkghandler.py
```diff
@@ -358,7 +395,8 @@ def _get_installed_pkg_objects_yum(name=None, version=None, release=None, arch=N
 
         return yum_base.rpmdb.returnPackages(patterns=[pattern])
 
-    return yum_base.rpmdb.returnPackages()
+    installed_packages = yum_base.rpmdb.returnPackages()
```
bocekm, line 398: Shouldn't we close and delete the yum base here to have the rpm database properly closed as on the other places? The same question applies to the other use of the yum base in the `_get_original_releasever()` function.
```suggestion
    base.close()
    del base

    installed_packages = yum_base.rpmdb.returnPackages()
```

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -274,13 +274,24 @@ def _process_transaction(self, validate_transaction):
...
         Perform the transaction. If the `validate_transaction` parameter set to
         true, it means the transaction will not be executed, but rather verify
         everything and do an early return.
 
+        .. warning::
+            This function is being executed in a child process so we will be
+            able to raise SIGINT or any other signal that is sent to the main
+            process.
+
+            The function calls here does not affect the others subprocesses
```
bocekm, line 290:
```suggestion
            The function calls here do not affect the others subprocess
```

convert2rhel/utils.py
```diff
@@ -66,6 +68,149 @@ class Color(object):
...
+                raise process.exception
+
+            # Only try to terminate a process if it is alive, otherwise, do
+            # nothing.
+            if process.is_alive():
+                process.terminate()
+
+            if not queue.empty():
+                # We don't need to block the I/O as we are mostly done with
+                # the child process and no exceptions was raised, so we can
```
bocekm, line 192:
```suggestion
                # the child process and no exception was raised, so we can
```

convert2rhel/utils.py
```diff
@@ -66,6 +68,149 @@ class Color(object):
 BACKUP_DIR = os.path.join(TMP_DIR, "backup")
 
 
+# Code taken from https://stackoverflow.com/a/63773140
+class Process(multiprocessing.Process):
+    """Overrides the implementation of the multiprocessing.Process class."""
```
bocekm, line 73: Please explain briefly in the docstring why we need to override the standard library class.

bocekm (changes requested):

convert2rhel/pkghandler.py
r0x0d: Good catch, @bocekm! I thought I was using the correct function in there, but it was the wrong one.

convert2rhel/utils.py
abadger: I cannot think of any exception that we should catch either.

convert2rhel/pkghandler.py
```diff
@@ -283,15 +283,41 @@ def get_pkg_fingerprint(pkg_obj):
         return "none"
 
 
-def get_pkg_signature(pkg_obj):
-    """Get information about a package signature from the RPM database."""
+@utils.run_as_child_process
+def get_pkg_signature(pkg_obj, queue):
+    """Get information about a package signature from the RPM database.
+
+    .. warning::
```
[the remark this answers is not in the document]
r0x0d: I will go with `important`... I like it more for this case.

convert2rhel/pkghandler.py
```diff
@@ -1018,15 +1044,32 @@ def get_total_packages_to_update(reposdir):
     return set(packages)
 
 
-def _get_packages_to_update_yum():
-    """Query all the packages with yum that has an update pending on the system."""
+@utils.run_as_child_process
+def _get_packages_to_update_yum(queue):
+    """Use yum to get all the installed packages that have an update available.
+
+    .. warning::
```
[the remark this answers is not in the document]
r0x0d: It makes sense to me to go with `important` rather than `warning`.

convert2rhel/utils.py
r0x0d: Cool, will remove the comment then.

convert2rhel/utils.py
```diff
@@ -158,11 +303,11 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
         if print_output:
             loggerinst.info(line.rstrip("\n"))
 
-    # Call communicate() to wait for the process to terminate so that we can get the return code by poll()
+    # Call coomunicate() to wait for the process to terminate so that we can
+    # get the return code.
     process.communicate()
 
-    return_code = process.poll()
```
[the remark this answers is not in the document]
r0x0d: Thanks!

convert2rhel/pkghandler.py
r0x0d: We are not trying to forcefully close the db on every possible case as it can come down to performance issues, as the closing/deleting the object (which will do the same as close), will clean up the rpmdb, which can slow us down a lot if we start closing it after every call. 

We are trying to only close the rpmdb in cases where we notice that yum/rpm is leaving locks behind and causing weird and strange messages to be seen by the user.

r0x0d: Those functions are fine because they are running within the same process as convert2rhel. We only had to forcefully close the rpmdb in cases where we are running the functions as a child process. 

I can't see a good place to share this information in the code, so I think I will make a note to myself to append that to our wiki, as I really plan to make a very comprehensive document about working with yum/dnf/rpm here.

bocekm (approved):
Code looks good. I haven't tried running the tool manually.

convert2rhel/pkgmanager/__init__.py
```diff
@@ -100,3 +102,36 @@ def clean_yum_metadata():
...
+
+    .. note::
+        This context manager will only work with an instance of
+        `yum.rpmsack.RPMInstalledPackage`, as it will access specific
+        properties inside that class to close the RPM DB.
+
+    .. important::
+        This context manager will be used for both yum and dnf, but only yum
+        needs to do a manual cleanup after the usage. Dnf seems to handle this
+        properly.
```
abadger, line 119:
```suggestion
    .. important::
        This context manager will be used for both yum and dnf, but only yum
        will actually do an explicit cleanup after the usage. Yum seems to need
        to do that explicitly while Dnf handles it properly.
```

convert2rhel/pkgmanager/__init__.py
```diff
@@ -100,3 +102,36 @@ def clean_yum_metadata():
...
+
+
+@contextmanager
+def rpm_db_lock(pkg_obj):
+    """Context manager to handle rpm database termination.
+
+    .. note::
+        This context manager will only work with an instance of
+        `yum.rpmsack.RPMInstalledPackage`, as it will access specific
+        properties inside that class to close the RPM DB.
```
abadger, line 114: Perhaps this `.. note::` is redundant with the `.. important::` admonition that comes after it?  If we want to keep it, maybe a slight rewording:
```suggestion
    .. note::
        This context manager will only do something with an instance of
        `yum.rpmsack.RPMInstalledPackage`, as it will access specific
        properties inside that class to close the RPM DB.  pkg_obj's from
        dnf are fine to pass in but will be a no-op as dnf manages the
        lifetime of the rpm_db lock fine on its own.
```

abadger (approved):
Just some changes requested for the documentation of the context manager.  Otherwise, looks good!

abadger (approved):
Looks great.

tests/integration/tier0/single-yum-transaction-validation/test_single_yum_transaction_validation.py
```diff
@@ -75,7 +78,13 @@ def test_package_download_error(convert2rhel, shell, yum_cache):
     ) as c2r:
         c2r.expect("Validate the {} transaction".format(PKGMANAGER))
         c2r.expect("Adding {} packages to the {} transaction set.".format(SERVER_SUB, PKGMANAGER))
+
+        if "7" in SYSTEM_RELEASE_ENV:
```
danmyway, line 82: Just a heads up, this condition will also apply to `oracle-8.7` (and other clones when available).
I suppose it won't cause much of an issue, just to be careful with the vague conditioning.
Again, I guess it will be fine for the current latest, so no need to implement this, but prepending a dash `-` would be more specific.
```suggestion
        if "-7" in SYSTEM_RELEASE_ENV:
```
or even
```suggestion
        if re.match(r"^(centos|oracle)-7$", SYSTEM_RELEASE_ENV):
```
Other than that it looks good.
Thanks for the resolution!

r0x0d: Thanks for the heads-up! I vaguely remember that we have some sort of function/class that returns a more constructed object with this information too, right?  I think it would be better to use it instead of regex.

danmyway: Indeed, there is the `system_release()` fixture, however I don't like it being dependent on the `/etc/os-release` or `/etc/system-release`
```
    Examples:
        Centos Linux 7.9 => centos-7.9\n
        Oracle Linux 8.6 => oracle-8.6\n
        Alma Linux 8.4 => almalinux-8.4\n
        Rocky Linux 8.5 => rocky-8.5
```

danmyway: Yeah and the `OsRelease` class, but the same story with `/etc/os-release`

convert2rhel/pkghandler.py
```diff
@@ -965,12 +1099,23 @@ def get_pkg_names_from_rpm_paths(rpm_paths):
...
 
+@utils.run_as_child_process
 def get_total_packages_to_update(reposdir):
-    """Return the total number of packages to update in the system
-    It uses both yum/dnf depending on whether they are installed on the system,
-    In case of RHEL 7 derivative distributions, it uses `yum`, otherwise it uses `dnf`.
-    To check whether the system is updated or not, we use original vendor repofiles which we ship within the
-    convert2rhel RPM. The reason is that we can't rely on the repofiles available on the to-be-converted system.
+    """
+    Return the total number of packages to update in the system It uses both
```
pr-watson, line 1105:
```suggestion
    Return the total number of packages to update in the system. It uses both
```

convert2rhel/pkghandler.py
```diff
@@ -1007,19 +1159,17 @@ def _get_packages_to_update_dnf(reposdir):
     packages = []
     base = pkgmanager.Base()
 
-    # If we have an reposdir, that means we are trying to check the packages under
-    # CentOS Linux 8.4 or 8.5 and Oracle Linux 8.4.
-    # That means we need to use our hardcoded repository files instead of the system ones.
+    # If we have an reposdir, that means we are trying to check the packages
```
pr-watson, line 1162:
```suggestion
    # If we have a reposdir, that means we are trying to check the packages
```

pr-watson (approved):
Made some minor edits for code comments. Looks good to me!

convert2rhel/utils.py
```diff
@@ -66,6 +68,196 @@ class Color(object):
...
+
+        By using this decorator, `rpm` will install the signal handler in the
+        child process and leave the parent one with the original signal
+        handling that is initiated by python itself (or, whatever signal is
+        registered when the conversion starts as well).
+
+    .. important::
+        If a function that creates child processes is decorated with this
+        decorator, then it won't be possible to use it right away, since this
+        decorator spawn child processes as a daemon (which are normal
```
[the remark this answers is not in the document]
abadger: I'm a little unclear on the connection between "it won't be possible to use it right away", "because it spawns child processes as a daemon" and "but can't spawn child processes inside that child process".  Are these three things related?  Or two separate limitations and the reason (daemon processes aren't immediately usable and can't spawn children of their own?)?

convert2rhel/utils.py
```diff
@@ -66,6 +68,196 @@ class Color(object):
...
+                # the child process and no exception was raised, so we can
+                # instantly retrieve the item that was in the queue.
+                return queue.get(block=False)
+
+            return None
+        except KeyboardInterrupt:
+            # We have to check if the process if alive, and if it is (most
+            # probably it will be), then we can call for termination. On
+            # Python2 it is mostly likely that some processes (That calls yum
+            # API) will keep executing until they finish its execution and
```
abadger, line 240:
```suggestion
            # Python2 it is most likely that some processes (That calls yum
            # API) will keep executing until they finish their execution and
```

abadger (changes requested):
Found a few things to change.  We can talk about them if I was unclear about what to do.

convert2rhel/pkghandler.py
```diff
@@ -505,6 +592,21 @@ def print_pkg_info(pkgs):
...
+        print_cmd=False,
+        print_output=False,
+    )
+
+    if "C2R" not in output or retcode != 0:
+        return "N/A"
+
+    return output.lstrip("C2R ")
+
+
```
abadger, line 609: `get_pkg_nvra` and `get_pkg_nevra` can be changed to extract the nevra from PackageInfo objects as well as the yum/dnf objects that represent packages.  Maybe with a helper function like this?
``` python
def _get_nevra_from_pkg_obj(pkg_obj):
    if isinstance(pkg_obj, PackageInformation):
        return pkg_obj.nevra
    return PackageNevra(name=pkg_obj.name, 
                        epoch=pkg_obj.epoch,
                        version=pkg_obj.version,
                        release=pkg_obj.release,
                        arch=pkg_obj.arch)

def get_pkg_nvra(pkg_obj):
    nevra = _get_nevra_from_pkg_obj(pkg_obj)
    # [...]
```

convert2rhel/breadcrumbs.py
```diff
@@ -110,7 +110,9 @@ def _set_nevra(self):
 
     def _set_signature(self):
         """Set signature of installed Convert2RHEL"""
-        self.signature = pkghandler.get_pkg_signature(self._pkg_object)
+        # Use full package name with version, release and arch.
```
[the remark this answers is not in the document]
r0x0d: Was a legacy comment that I forgot to remove. 

```suggestion
```

convert2rhel/checks.py
```diff
@@ -621,8 +619,8 @@ def _bad_kernel_package_signature(kernel_release):
     logger.debug("Booted kernel package name: {0}".format(name))
 
     kernel_pkg_obj = get_installed_pkg_objects(name, version, release, arch)[0]
-    kernel_pkg_gpg_fingerprint = get_pkg_fingerprint(kernel_pkg_obj)
-    bad_signature = system_info.cfg_content["gpg_fingerprints"] != kernel_pkg_gpg_fingerprint
+    package = get_package_information(str(kernel_pkg_obj))[0]
+    bad_signature = system_info.cfg_content["gpg_fingerprints"] != package.fingerprint
```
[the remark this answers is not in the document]
r0x0d: Updated

convert2rhel/pkghandler.py
```diff
@@ -409,15 +504,14 @@ def get_installed_pkgs_w_different_fingerprint(fingerprints, name=""):
     list in the fingerprints parameter. The packages can be optionally
     filtered by name.
     """
+    # if fingerprints is None skip this check.
     if not fingerprints:
-        # if fingerprints is None skip this check.
         return []
```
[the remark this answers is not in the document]
r0x0d: leaving it as-is for now

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -330,11 +353,4 @@ def run_transaction(self, validate_transaction=False):
...
         self._process_transaction(validate_transaction)
-
-        # Because we call the same thing multiple times, the rpm database is not
-        # properly closed at the end of it, thus, having the need to call
-        # `self._base.close()` explicitly before we delete the object. If we use
-        # only `del self._base`, it seems that yum is not able to properly clean
-        # everything in the database.
-        self._base.close()
-        del self._base
+        self._close_yum_base()
```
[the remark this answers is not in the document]
r0x0d: That looks better and we can encapsulate everything inside it. updated!

convert2rhel/pkghandler.py
```diff
@@ -505,6 +592,21 @@ def print_pkg_info(pkgs):
     return pkg_table
 
 
+def _get_package_repository(pkg):
+    """Get package repository information."""
+    package = get_pkg_nvra(pkg.nevra) if isinstance(pkg, PackageInformation) else get_pkg_nvra(pkg)
+    output, retcode = utils.run_subprocess(
+        ["repoquery", "--quiet", "-q", package, "--qf", "C2R %{REPOID}"],
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/pkghandler.py
```diff
@@ -476,25 +573,15 @@ def print_pkg_info(pkgs):
...
-            # Bug filed: https://bugzilla.redhat.com/show_bug.cgi?id=1879168
-            from_repo = pkg._from_repo
+        from_repo = _get_package_repository(pkg)
 
         pkg_list += (
             "%-*s  %-*s  %s"
             % (
                 max_nvra_length,
-                get_pkg_nvra(pkg),
+                get_pkg_nvra(pkg.nevra) if isinstance(pkg, PackageInformation) else get_pkg_nvra(pkg),
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/utils.py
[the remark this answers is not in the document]
r0x0d: updated the comment

convert2rhel/pkghandler.py
r0x0d: added the help function

convert2rhel/pkghandler.py
```diff
@@ -272,44 +298,98 @@ def get_installed_pkgs_by_fingerprint(fingerprints, name=""):
...
     for pkg_obj in package_objects:
-        fingerprint = get_pkg_fingerprint(pkg_obj)
-        pkgs_w_fingerprints.append(PkgWFingerprint(pkg_obj, fingerprint))
+        package = get_package_information(str(pkg_obj))
+        if package:
+            package = package[0]
+            if package.fingerprint:
+                pkgs_w_fingerprints.append(package)
 
     return pkgs_w_fingerprints
```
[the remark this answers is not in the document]
r0x0d: removed the function.

convert2rhel/pkghandler.py
```diff
@@ -476,25 +573,15 @@ def print_pkg_info(pkgs):
...
-                from_repo = pkg.yumdb_info.from_repo
-            except AttributeError:
-                # A package may not have the installation repo set in case it was installed through rpm
-                from_repo = "N/A"
-
-        elif pkgmanager.TYPE == "dnf":
-            # There's no public attribute for getting the installation repository.
-            # Bug filed: https://bugzilla.redhat.com/show_bug.cgi?id=1879168
-            from_repo = pkg._from_repo
+        from_repo = _get_package_repository(pkg)
```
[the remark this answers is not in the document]
r0x0d: passing all packages to repoquery and parsing them to print the list.

convert2rhel/unit_tests/pkgmanager/handlers/dnf/dnf_test.py
```diff
@@ -22,7 +22,7 @@
 from convert2rhel.pkgmanager.handlers.dnf import DnfTransactionHandler
 from convert2rhel.pkgmanager.handlers.dnf.callback import DependencySolverProgressIndicatorCallback
 from convert2rhel.systeminfo import system_info
-from convert2rhel.unit_tests.conftest import centos8
+from convert2rhel.unit_tests.conftest import centos8, create_pkg_information
```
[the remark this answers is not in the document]
r0x0d: removed the functions from conftest and moved them to unit_tests/__init__.py

convert2rhel/pkgmanager/handlers/yum/__init__.py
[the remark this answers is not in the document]
abadger: This has been done.

convert2rhel/pkghandler.py
```diff
@@ -273,43 +299,69 @@ def get_installed_pkgs_by_fingerprint(fingerprints, name=""):
...
-        fingerprint = get_pkg_fingerprint(pkg_obj)
-        pkgs_w_fingerprints.append(PkgWFingerprint(pkg_obj, fingerprint))
+def _get_pkg_fingerprint(signature):
+    """Get fingerprint of the key used to sign a package."""
+    fingerprint_match = re.search("Key ID (.*)", signature)
+    return fingerprint_match.group(1) if fingerprint_match else "none"
 
-    return pkgs_w_fingerprints
 
+def get_installed_pkg_information(pkg_name=""):
```
abadger, line 312:
```suggestion
def get_installed_pkg_information(pkg_name="*"):
```

convert2rhel/pkghandler.py
```diff
@@ -273,43 +299,69 @@ def get_installed_pkgs_by_fingerprint(fingerprints, name=""):
...
+        "--qf",
+        "C2R %{PACKAGER}&%{VENDOR}&%{NAME}-%|EPOCH?{%{EPOCH}}:{0}|:%{VERSION}-%{RELEASE}.%{ARCH}&%|DSAHEADER?{%{DSAHEADER:pgpsig}}:{%|RSAHEADER?{%{RSAHEADER:pgpsig}}:{%|SIGGPG?{%{SIGGPG:pgpsig}}:{%|SIGPGP?{%{SIGPGP:pgpsig}}:{(none)}|}|}|}|\n",
+    ]
+
+    if pkg_name:
+        if "*" in pkg_name:
+            cmd.append("-qa")
+        else:
+            cmd.extend(["-q", pkg_name])
     else:
```
abadger, line 333:
```suggestion
    if pkg_name:
        if "*" in pkg_name:
            cmd.append(["-qa", pkg_name])
        else:
            cmd.extend(["-q", pkg_name])
    else:
```

or, if the default of pkg_name is changed to `"*"`, then:
``` diff
-    if pkg_name:
-        if "*" in pkg_name:
-            cmd.append("-qa")
-        else:
-            cmd.extend(["-q", pkg_name])
+    if "*" in pkg_name:
+        cmd.extend(["-qa", pkg_name])
+    else:
+        cmd.extend(["-q", pkg_name])
-    else:
-        cmd.append("-qa")
```

r0x0d: I think that like this would be more correct, no? 

```python
    if pkg_name:
        if "*" in pkg_name:
            cmd.extend(["-qa", pkg_name])
        else:
            cmd.extend(["-q", pkg_name])
    else:
        cmd.append("-qa")
```

r0x0d: updated

convert2rhel/pkghandler.py
```diff
@@ -474,70 +535,135 @@ def print_pkg_info(pkgs):
...
             )
             + "\n"
         )
 
     pkg_table = header + header_underline + pkg_list
-    loggerinst.info(pkg_table)
     return pkg_table
 
 
+def _get_package_repository(pkgs):
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/pkghandler.py
```diff
@@ -474,70 +535,135 @@ def print_pkg_info(pkgs):
...
+    print("package_nevras", package_nevras)
+    print("pkgs", pkgs)
+    packages_with_repos = _get_package_repository(package_nevras)
+    print("packages_with_repos", packages_with_repos)
+    # Update package_info reference with repoid
+    for nevra, repoid in packages_with_repos.items():
+        print("assignment", nevra, repoid)
+        package_info[nevra]["repoid"] = repoid
+
+    print(package_info)
```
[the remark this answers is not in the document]
r0x0d: done

convert2rhel/pkghandler.py
```diff
@@ -620,17 +743,52 @@ def remove_pkgs_with_confirm(pkgs, backup=True):
...
+
+
 def get_system_packages_for_replacement():
-    """Get a list of packages in the system to be replaced.
-    This function will return a list of packages installed on the system by
-    using the `system_info.fingerprint_ori_os` signature.
+    """
+    Get a list of packages in the system to be replaced. This function will
+    return a list of packages installed on the system by using the
+    `system_info.fingerprint_ori_os` signature.
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/pkghandler.py
```diff
@@ -474,70 +535,135 @@ def print_pkg_info(pkgs):
...
+    # In case of repoquery returning an retcode different from 0, let's log the
+    # output as a debug and return N/A for the caller.
+    if retcode != 0:
+        loggerinst.debug("Repoquery exited with return code %s and with output: %s", retcode, " ".join(output))
+        for package in pkgs:
+            repositories_mapping[package] = "N/A"
+    else:
+        for line in output:
+            if "C2R" in line:
+                print("line", line)
```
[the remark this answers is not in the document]
r0x0d: done.

abadger (approved):

danmyway (approved):
Great job @r0x0d, this will help immensely, thank you!

convert2rhel/pkghandler.py
```diff
@@ -942,7 +965,7 @@ def verify_rhel_kernel_installed():
 
 
 def is_rhel_kernel_installed():
-    installed_rhel_kernels = get_installed_pkgs_by_fingerprint(system_info.fingerprints_rhel, name="kernel")
+    installed_rhel_kernels = get_installed_pkgs_by_fingerprint(system_info.fingerprints_rhel, name="kernel*")
```
r0x0d, line 968: Remove the *

convert2rhel/pkghandler.py
```diff
@@ -637,14 +649,25 @@ def get_pkg_nvra(pkg_obj):
     )
 
 
-def get_pkg_nevra(pkg_obj):
-    """Get package NEVRA as a string: name, epoch, version, release, architecture.
-    Epoch is included when it is present. However it's on a different position when printed by YUM or DNF:
-      YUM - epoch before name: "7:oraclelinux-release-7.9-1.0.9.el7.x86_64"
-      DNF - epoch before version: "oraclelinux-release-8:8.2-1.0.8.el8.x86_64"
+def get_pkg_nevra(pkg_obj, include_epoch=False):
```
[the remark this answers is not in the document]
r0x0d: done.

convert2rhel/pkghandler.py
r0x0d: done.

PR #764, opened 2023-03-13
[RHELC-940] Convert Pre Ponr checks to the action framework

r0x0d:
This commit introduces an rewrite of most of the checks in the pre_ponr function as actions for the new action framework.

Jira Issues: [RHELC-940](https://issues.redhat.com/browse/RHELC-940)

Checklist


convert2rhel/actions/handle_packages.py
```diff
@@ -0,0 +1,110 @@
...
+        logger.task("Convert: List third-party packages")
+        pkghandler.list_third_party_pkgs()
+
+
+class _PackageRemoval:
+    def remove_packages_with_confirmation(self, packages):
+        pkghandler.remove_pkgs_with_confirm(packages)
+
+
+class RemoveExcludedPackages(actions.Action, _PackageRemoval):
```
abadger, line 41: I think I misunderstood a question you were asking the other day.  I would probably call `pkghandler.remove_pkgs_with_confirm(packages)` directly from the Action class.  Since it's just a single function call, I don't think we gain much by putting it into a function and a class inside of handle_packages.  Does that make sense?

If you do think it should be grouped with `handle_packages`, I would move it out to module level.  When helper functons don't make use of `self`, we should usually leave them at module scope.  In other languages, classes are the natural way to create namespaces but in Python modules are the natural way to create namespaces so classes should only hold values if they are really needed there.
```suggestion
def _remove_packages_with_confirmation(packages):
    pkghandler.remove_pkgs_with_confirm(packages)


class RemoveExcludedPackages(actions.Action):
```

Note: If you do decide it should remain inside of a class, then you should know that the method resolution order is left to right.  That means if you have one class that's obviously the "base" for that class hierarchy and other classes which are mixins which are there to add specific traits to this particular class, then you would want the mixins to be to the left and the base to be the furthest right.  Mixins should probably also always call `super()` so they don't entirely mask any methods of the same name which could be added to other parents later. So, if you didn't change this to be a toplevel function, it would look like this:
``` python
class _PackageRemoval:
    def remove_packages_with_confirmation(self, packages):
        pkghandler.remove_pkgs_with_confirm(packages)
        super(_PackageRemoval, self).remove_packages_with_confirmaton(packages)


class Remove ExcludedPackages(_PackageRemoval, actions.Action):
```

r0x0d: I revisited that today and I think that the extra class/function will not be necessary at all... Having the `remove_pkgs_with_confirm()` directly in the action class run method is way better than adding this another layer on top of it.

convert2rhel/actions/__init__.py
```diff
@@ -178,6 +182,41 @@ def run(self):
...
+
+        if error_id != _NO_USER_VALUE:
+            self.error_id = error_id
+
+        if message != _NO_USER_VALUE:
+            self.message = message
+
+    def set_status(self, status="SUCCESS", error_id=None, message=None):
+        """
+        Helper method that sets an status for the class.
```
pr-watson, line 207:
```suggestion
        Helper method that sets a status for the class.
```

convert2rhel/actions/handle_packages.py
```diff
@@ -0,0 +1,148 @@
...
+        logger.info("Searching for the following excluded packages:\n")
+        try:
+            pkghandler.remove_pkgs_with_confirm(system_info.excluded_pkgs)
+            # TODO: Handling SystemExit here as way to speedup exception
+            # handling and not refactor contents of the underlying function.
+        except SystemExit as e:
+            # TODO(r0x0d): Places where we raise SystemExit and need to be
+            # changed to something more specific.
+            #   - When we can't remove a package.
+            self.set_result(status="ERROR", error_id="PACKAGE_REMOVAL_FAILED", message=e)
```
pr-watson, line 71:
```suggestion
            self.set_result(status="ERROR", error_id="PACKAGE_REMOVAL_FAILED", message"%s" % e)
```
You may need to change this message to string format since all three params need to be strings

convert2rhel/actions/handle_packages.py
```diff
@@ -0,0 +1,148 @@
...
+        logger.info("Searching for packages containing .repo files or affecting variables in the .repo files:\n")
+        try:
+            pkghandler.remove_pkgs_with_confirm(system_info.repofile_pkgs)
+            # TODO: Handling SystemExit here as way to speedup exception
+            # handling and not refactor contents of the underlying function.
+        except SystemExit as e:
+            # TODO(r0x0d): Places where we raise SystemExit and need to be
+            # changed to something more specific.
+            #   - When we can't remove a package.
+            self.set_result(status="ERROR", error_id="PACKAGE_REMOVAL_FAILED", message=e)
```
pr-watson, line 103:
```suggestion
            self.set_result(status="ERROR", error_id="PACKAGE_REMOVAL_FAILED", message="%s" % e)
```
same here

convert2rhel/actions/subscription.py
```diff
@@ -0,0 +1,151 @@
...
+            # changed to something more specific.
+            #   - If we can't import the gpg key.
+            #   - if directory does not exist or is empty
+            #   - if we can't download a package
+            #   - if we can't install sub-man rpms
+            #   - If sub-man is not installed and --keep-rhsm was used.
+
+            # TODO(r0x0d): This should be refactored to handle each case
+            # individually rather than relying on SystemExit.
+            self.set_result(status="ERROR", error_id="UNKNOWN_ERROR", message=e)
```
pr-watson, line 81:
```suggestion
            self.set_result(status="ERROR", error_id="UNKNOWN_ERROR", message="%s" % e)
```

convert2rhel/actions/subscription.py
```diff
@@ -0,0 +1,151 @@
...
+        except SystemExit as e:
+            # TODO(r0x0d): This should be refactored to handle each case
+            # individually rather than relying on SystemExit.
+
+            # TODO(r0x0d): Places where we raise SystemExit and need to be
+            # changed to something more specific.
+            #   - Maximum sub-man retries reached
+            #   - If the return-code is different from 0 in disabling repos,
+            #     SystemExit is raised.
+            self.set_result(status="ERROR", error_id="UNKNOWN_ERROR", message=e)
```
pr-watson, line 145:
```suggestion
            self.set_result(status="ERROR", error_id="UNKNOWN_ERROR", message="%s" % e)
```

pr-watson:
Looks good to me. You may need to update some of the message params in set_result to be strings

convert2rhel/actions/handle_packages.py
r0x0d: Used `str(e)` instead.

convert2rhel/actions/handle_packages.py
```diff
@@ -0,0 +1,148 @@
...
+            #       - In case we reach max numbers of retry for dependency
+            #         solving.
+            #       - In case of transaction validation failure
+            #
+            #   - Dnf transaction:
+            #       - If we fail to populate repository metadata
+            #       - If we fail to resolve dependencies in the transaction
+            #       - If we fail to download the transaction packages
+            #       - If we fail to validate the transaction
+            self.set_result(status="ERROR", error_id="UNKNOWN_ERROR", message=e)
```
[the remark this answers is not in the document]
r0x0d: Used `str(e)` instead.

convert2rhel/actions/subscription.py
r0x0d: Used `str(e)` instead.

pr-watson:
Changes look good, approved @r0x0d

convert2rhel/unit_tests/actions/handle_packages_test.py
```diff
@@ -0,0 +1,127 @@
...
+    assert remove_repository_files_packages_instance.status == actions.STATUS_CODE["SUCCESS"]
+
+
+def test_remove_repository_files_packages_error(remove_repository_files_packages_instance, monkeypatch):
+    monkeypatch.setattr(system_info, "repofile_pkgs", [])
+    monkeypatch.setattr(pkghandler, "remove_pkgs_with_confirm", mock.Mock(side_effect=SystemExit("Raising SystemExit")))
+
+    remove_repository_files_packages_instance.run()
+
+    unit_tests.assert_actions_result(status="ERROR", error_id="PACKAGE_REMOVAL_FAILED", message="Raising SystemExit")
```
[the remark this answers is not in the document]
r0x0d: Fixed, thanks!

convert2rhel/actions/handle_packages.py
```diff
@@ -0,0 +1,104 @@
...
+        try:
+            pkghandler.remove_pkgs_with_confirm(system_info.repofile_pkgs)
+            # TODO: Handling SystemExit here as way to speedup exception
+            # handling and not refactor contents of the underlying function.
+        except SystemExit as e:
+            # TODO(r0x0d): Places where we raise SystemExit and need to be
+            # changed to something more specific.
+            #   - When we can't remove a package.
+            self.set_result(status="ERROR", error_id="PACKAGE_REMOVAL_FAILED", message=str(e))
+            return
```
pr-watson, line 104: This return can be removed since it's at the end of the file right?

pr-watson:
Made a minor few comments, otherwise looks good!

pr-watson (approved):

convert2rhel/actions/handle_packages.py
r0x0d:
```suggestion
```

convert2rhel/actions/transaction.py
```diff
@@ -0,0 +1,68 @@
...
+            #         solving.
+            #       - In case of transaction validation failure
+            #
+            #   - Dnf transaction:
+            #       - If we fail to populate repository metadata
+            #       - If we fail to resolve dependencies in the transaction
+            #       - If we fail to download the transaction packages
+            #       - If we fail to validate the transaction
+            self.set_result(status="ERROR", error_id="UNKNOWN_ERROR", message=str(e))
+            return
```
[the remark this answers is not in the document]
r0x0d:
```suggestion
```

PR #772, opened 2023-03-17
[RHELC-934] Port check_rhel_compatible_kernel_is_used to Action framework

pr-watson:
This PR ports the check_rhel_compatible_kernel_is_used and other associated functions to an action framework.

Jira Issues: [RHELC-934](https://issues.redhat.com/browse/RHELC-934)

Checklist

### DIFF
--- a/convert2rhel/actions/rhel_compatible_kernel.py
+++ b/convert2rhel/actions/rhel_compatible_kernel.py
@@ -41,6 +41,7 @@ class RhelCompatibleKernel(actions.Action):
         By requesting that, we can be confident that the RHEL kernel will provide the same capabilities as on the
         original system.
         """
+        super(RhelCompatibleKernel, self).run()
         logger.task("Prepare: Check kernel compatibility with RHEL")
         if any(
             (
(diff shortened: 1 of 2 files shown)


convert2rhel/actions/rhel_compatible_kernel.py
```diff
@@ -0,0 +1,141 @@
...
+
+class RhelCompatibleKernel(actions.Action):
+    id = "RHEL_COMPATIBLE_KERNEL"
+
+    def run(self):
+        """Ensure the booted kernel is signed, is standard (not UEK, realtime, ...), and has the same version as in RHEL.
+        By requesting that, we can be confident that the RHEL kernel will provide the same capabilities as on the
+        original system.
+        """
+        logger.task("Prepare: Check kernel compatibility with RHEL")
```
abadger, line 44: We just recently realized that it is a good idea to do this:

```suggestion
        super(RhelCompatibleKernel, self).run()
        logger.task("Prepare: Check kernel compatibility with RHEL")
```

without calling super, the default successful return values are not added to the class attributes.

r0x0d: Placing the decorator on top of `run()` seems elegant, too. 

```
@actions._action_defaults_to_success
def run(self): 
```

But either way, both will work

r0x0d (commented):
Looks good! Just a small comment to be fixed.

PR #779, opened 2023-03-22
[RHELC-937] Port check_dbus_is_running to Action framework

pr-watson:
This PR ports the check_dbus_is_running function from checks.py to an Action framework

Jira Issues: [RHELC-937](https://issues.redhat.com/browse/RHELC-937)

Checklist

### DIFF
--- a/convert2rhel/actions/__init__.py
+++ b/convert2rhel/actions/__init__.py
@@ -315,7 +315,6 @@ def perform_system_checks():
     check_custom_repos_are_valid()
     check_tainted_kmods()
     is_loaded_kernel_latest()
-    check_dbus_is_running()
 
 
 def perform_pre_ponr_checks():
@@ -764,21 +763,3 @@ def is_loaded_kernel_latest():
         )
 
     logger.info("The currently loaded kernel is at the latest version.")
-
-
-def check_dbus_is_running():
-    """Error out if we need to register with rhsm and the dbus daemon is not running."""
-    logger.task("Prepare: Check that DBus Daemon is running")
-
-    if tool_opts.no_rhsm:
-        logger.info("Skipping the check because we have been asked not to subscribe this system to RHSM.")
-        return
-
-    if system_info.dbus_running:
-        logger.info("DBus Daemon is running")
-        return
-
-    logger.critical(
-        "Could not find a running DBus Daemon which is needed to register with subscription manager.\n"
-        "Please start dbus using `systemctl start dbus`"
-    )
--- a/convert2rhel/actions/dbus.py
+++ b/convert2rhel/actions/dbus.py
@@ -1,0 +1,51 @@
+# Copyright(C) 2023 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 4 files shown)


convert2rhel/actions/dbus_is_running.py
```diff
@@ -0,0 +1,52 @@
...
+            logger.info("Skipping the check because we have been asked not to subscribe this system to RHSM.")
+            return
+
+        if system_info.dbus_running:
+            logger.info("DBus Daemon is running")
+            return
+
+        self.set_result(
+            status="ERROR",
+            error_id="DBUS_DAEMON_NOT_FOUND",
```
r0x0d, line 46: I would change this to 
```suggestion
            error_id="DBUS_DAEMON_NOT_RUNNING",
```

I think that `not found` may give the wrong impression to the user, as it could mean that the whole systemd unit for dbus is not present, but I think that's not the case for this check, which has the intention to verify if that is running.

convert2rhel/actions/dbus_is_running.py
```diff
@@ -0,0 +1,52 @@
...
+            message=(
+                "Could not find a running DBus Daemon which is needed to register with subscription manager.\n"
+                "Please start dbus using `systemctl start dbus`"
+            ),
+        )
+        return
```
r0x0d, line 52: I guess this lonely return statement can be taken out of here, there is nothing below this point

convert2rhel/unit_tests/actions/dbus_is_running_test.py
```diff
@@ -0,0 +1,72 @@
...
+):
+    monkeypatch.setattr(actions.dbus_is_running, "tool_opts", global_tool_opts)
+    global_tool_opts.no_rhsm = no_rhsm
+    monkeypatch.setattr(actions.dbus_is_running, "system_info", global_system_info)
+    global_system_info.dbus_running = dbus_running
+
+    assert dbus_is_running_action.run() is None
+    assert caplog.records[-1].message == log_msg
+
+    assert log_msg == caplog.records[-1].message
```
r0x0d, line 50: Ins't that the same check, but in different order?

convert2rhel/unit_tests/actions/dbus_is_running_test.py
```diff
@@ -0,0 +1,72 @@
...
+    global_tool_opts.no_rhsm = False
+    monkeypatch.setattr(actions.dbus_is_running, "system_info", global_system_info)
+    global_system_info.dbus_running = False
+
+    dbus_is_running_action.run()
+
+    unit_tests.assert_actions_result(
+        dbus_is_running_action,
+        status="ERROR",
+        error_id="DBUS_DAEMON_NOT_FOUND",
```
r0x0d, line 66: That would need to be changed if https://github.com/oamg/convert2rhel/pull/779/files#r1146656910 is accepted 

```suggestion
        error_id="DBUS_DAEMON_NOT_RUNNING",
```

r0x0d (changes requested):
Small changes, otherwise, looks pretty nic

r0x0d:
One extra thing, what about changing the name of the module to `dbus.py` or `checks_dbus.py`, having the meaning that all possibly dbus checks in the future would land in this module, rather than having a module of it's own for just this check?

convert2rhel/actions/dbus_is_running.py
pr-watson: Noted i'll update it

convert2rhel/unit_tests/actions/dbus_is_running_test.py
pr-watson: True, this was from the original function, good catch

pr-watson:
I'll rename it to dbus then :)

r0x0d (approved):

PR #781, opened 2023-03-22
[RHELC-724] Add a sanity check for logging duplication

danmyway:
* Verification test
* Verify the lines in log file are unique and not duplicated

Jira Issues: [RHELC-724](https://issues.redhat.com/browse/RHELC-724)

Checklist

### DIFF
--- a/tests/integration/tier1/checks-after-conversion/main.fmf
+++ b/tests/integration/tier1/checks-after-conversion/main.fmf
@@ -46,3 +46,13 @@ link:
         tag+:
             - initramfs-and-vmlinuz-present
         test: pytest -svv -m initramfs_and_vmlinuz_present
+
+/log_lines_not_duplicated:
+    summary+: Logged lines are not duplicated
+    # TODO (danmyway) remove when the duplication of lines with multiple packages with different versions gets resolved
+    result: info
+    description+: |
+        Iterate over the log file and check that the lines are unique and not duplicated.
+    tag+:
+        - log-file
+    test: pytest -svv -m verify_logging_is_not_duplicated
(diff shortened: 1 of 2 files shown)


kokesak (approved):
Looks good :D

danmyway:
> @danmyway, I like the idea of printing NEVRA on one of the places you mentioned - where we skip tracking a package:
> 
> ```
> 21:19:06             out: [03/22/2023 17:19:06] DEBUG - Skipping tracking previously installed package: json-c
> 21:19:06             out: [03/22/2023 17:19:06] DEBUG - Skipping tracking previously installed package: json-c
> ```
> 
> It's not trivial to accomplish because we have just the package name available when printing the message so we'd need to change the `get_installed_pkgs_by_fingerprint()` function to return pkg objects instead of just names, or create a new function that would do that.
> 
> The second occurrence is a bit different:
> 
> ```
> 20:11:17             out: WARNING - Package kernel-uek.x86_64 not available in RHEL repositories.
> 20:11:17             out: WARNING - Package kernel-uek.x86_64 not available in RHEL repositories.
> 20:11:17             out: WARNING - Package kernel-uek.x86_64 not available in RHEL repositories.
> ```
> 
> We are looking for really just the package name (with its arch) in the RHEL repos. We're not looking for a specific version or release, so the way the warning is printed is ok. However it shouldn't be printed three times but just once. For that we can simply return a set instead of a list from the `get_system_packages_for_replacement()` function.
> 
> Would you please file a Jira ticket for that? Till it's implemented, the test can be marked as INFO.

Sure, on it @bocekm

PR #782, opened 2023-03-22
[RHELC-875] Porting out optparse to argparse

r0x0d:
This is the initial work before we deep dive into making changes for the pre conversion framework, as we will introduce subcommands there, we need to at least, have our current CLI module to be using argparse for those features.

The current optparse usage doesn't support subcommands natively.

Jira Issues: [RHELC-875](https://issues.redhat.com/browse/RHELC-875)

Checklist


convert2rhel/toolopts.py
```diff
@@ -79,7 +79,7 @@ def __init__(self):
     def _get_argparser():
         usage = (
             "\n"
-            "  convert2rhel [-h]\n"
+            "  convert2rhel [-h]<n>"
```
abadger, line 82: Is this correct?

convert2rhel/toolopts.py
```diff
@@ -138,28 +139,68 @@ def _register_options(self):
...
+        group.add_argument(
+            "--disable-submgr",
+            action="store_true",
+            help="Replaced by --no-rhsm. Both options have the same effect.",
+        )
+        group.add_argument(
+            "--no-rhsm",
+            action="store_true",
+            help="Do not use the subscription-manager, use custom repositories instead. See --enablerepo/--disablerepo"
+            " options. Without this option, the subscription-manager is used to access RHEL repositories by default."
```
abadger, line 186:
```suggestion
            " options. Without this option, subscription-manager is used to access RHEL repositories by default."
```

abadger (changes requested):
A few minor changes to make to the help strings but otherwise this looks good.

convert2rhel/toolopts.py
r0x0d: Oh, I must have changed it by mistake while tyring out a few things. Thanks for spotting that out! 

```suggestion
            "  convert2rhel [-h]\n"
```

abadger (approved):

PR #784, opened 2023-03-23
[RHELC-926] Stages and dependency order

abadger:
Add Stages to the running of Actions and change dependencies for stages and other needs (string based dependencies).

Jira Issues: [RHELC-926](https://issues.redhat.com/browse/RHELC-926)

Checklist


abadger:
TODO: sort the potential actions before processing dependencies.  That way the order will be stable even though it is not predictable (Where multiple orders will meet the dependency requirements, the results will be the same across runs if the code base hasn't changed.

convert2rhel/actions/__init__.py
```diff
@@ -137,7 +183,7 @@ class Convert2rhelLatest(Action):
         """
 
     #: Override dependencies with a Sequence that has other :class:`Action`s
-    #: that must be run before this one.
+    #: :attr:`Action.id`s that must be run before this one.
```
r0x0d, line 186: Would be nice to specify that the contents of `depdencies` should be in string format, not the class name. 

I think that people can look at it and have the wrong interpretation if they see like that, for instance, if I saw that I would do 

```python
dependencies = (Something.id, )
```

Rather than what we have right now.

convert2rhel/main.py
```diff
@@ -94,14 +88,22 @@ def main():
 
         ### End calls that should be put into pre_ponr_changes(), rollback(), or post_ponr_changes()
 
-        # begin conversion process
-        process_phase = ConversionPhase.PRE_PONR_CHANGES
-        # pre_ponr_conversion()
+        # FIXME: Write a nicely formatted report here.
+        # Write to the screen for this release
+        # Also write to a json formatted file for the next release.
```
r0x0d, line 93: I think it is better to move this to a jira issue rather than a comment in the code if we already know what will be done for now.

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
+            self.next_stage.check_dependencies(actions_so_far)
+
+    def run(self, successes=None, failures=None, skips=None):
+        """
+        Run all the actions in Stage and other linked Stages.
+
+        :keyword successes: Actions which have already run and succeeded.
+        :type: Sequence
+        :keyword failures: Actions which have already run and failed.
+        :type: Sequence
```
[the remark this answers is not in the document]
r0x0d: The `skips` is missing from the docstring

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
+        :return: 2-tuple consisting of two lists.  One with Actions that
+            have succeeded and one of Actions that have failed.  These
+            lists contain the Actions passed in via successes and failures
+            in addition to any that were run in this :class`Stage`.
+
+        :rtype: FinishedActions
+        .. note:: Success is currently defined as an action whose status after
+            running is WARNING or better (WARNING or SUCCESS) and
+            failure as worse than WARNING (OVERRIDABLE, ERROR, FATAL)
```
r0x0d, line 328: What about transforming this into an `.. warning::`? I think this note is a very important thing for people no pay attention to.

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
+
+        if failures is None:
+            failures = []
+        else:
+            failures = list(failures)
+
+        if skips is None:
+            skips = []
+        else:
+            skips = list(skips)
```
r0x0d, line 349: What about making this a bit shorter? I think it looks better and easier to read too

```suggestion
successes = list(successes) if successes else None
failures = list(failures) if failures else None
skips = list(skips) if skips else None
```

convert2rhel/utils.py
```diff
@@ -730,6 +730,28 @@ def hide_secrets(
     return sanitized_list
 
 
+def format_sequence_as_message(sequence_of_items):
+    """
+    Format a sequence of items for display to the user.
+
+    :param sequence_of_items: Sequence of items which should be formatted for
+        a message to be printed to the user.
+    :type Sequence:
```
r0x0d, line 739:
```suggestion
    :type sequence_of_items: Sequence
```

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
+            try:
+                action.run()
+            except (Exception, SystemExit) as e:
+                # Uncaught exceptions are handled by constructing a generic
+                # failure message here that should be reported
+                message = (
+                    "Unhandled exception was caught: %s"
+                    "\nPlease file a bug at https://issues.redhat.com/ to have this"
+                    " fixed or a specific error message added."
+                    "\nTraceback: %s" % (e, traceback.format_exc())
```
r0x0d, line 377: Not a necessary to apply, but I would format the message in this way, as it seems easier to read from the point-of-view that we start reading from left -> right, moving the `\n` and `" "` to the end of the string instead of the beginning. 

But again, not necessary to apply if you don't think it makes sense.

```suggestion
                    "Unhandled exception was caught: %s\n"
                    "Please file a bug at https://issues.redhat.com/ to have this "
                    "fixed or a specific error message added.\n"
                    "Traceback: %s" % (e, traceback.format_exc())
```

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
+                )
+                action.set_result(status=STATUS_CODE["ERROR"], error_id="UNEXPECTED_ERROR", message=message)
+
+            # Categorize the results
+            if action.status <= STATUS_CODE["WARNING"]:
+                successes.append(action)
+
+            if action.status > STATUS_CODE["WARNING"]:
+                failures.append(action)
+
```
r0x0d, line 387: Not sure if that's necessary or not, but should we change this to make sure that anything that is `>= SKIP` is a failure, instead of the currrent behavior of `> WARNING`? 

The categorization is correct here, but, I think it would make more sense from the point of view of the reader if we use `SKIP` rather than `WARNING` here, especially because `WARNING` will be included in the success case. 

Not sure if you agree with that logic, let me know if that doesn't apply here! 

```suggestion
            # Categorize the results
            if action.status <= STATUS_CODE["WARNING"]:
                successes.append(action)

            if action.status > STATUS_CODE["SKIP"]:
                failures.append(action)

```

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
-        depend on their output.
-    :raises Exception: when it is impossible to satisfy a dependency in an Action.
+    :param potential_actions: Sequence of Actions which we need to find the
+        order of.
+    :type: Sequence
+    :param previously_resolved_actions: Sequence of Actions which have already
+        had been resolved into dependency order.
+    :returns: Iterator of Actions sorted so that all dependent Actions are run
+        before actions which depend on them.
+    :raises DependencyError: when it is impossible to satisfy a dependency in
```
r0x0d, line 405:
```suggestion
    :param potential_actions: Sequence of Actions which we need to find the
        order of.
    :type potential_actions: Sequence
    :param previously_resolved_actions: Sequence of Actions which have already
        had been resolved into dependency order.
    :type previously_resolved_actions: Sequence

    :raises DependencyError: when it is impossible to satisfy a dependency in
    
    :returns: Iterator of Actions sorted so that all dependent Actions are run
        before actions which depend on them.
```

convert2rhel/actions/__init__.py
abadger: The actual logic would be:

```python
successes = [] if successes is None else list(successes)
failures = [] if failures is None else list(failures)
skips = [] if skips is None else list(skips)
```

 It is shorter, which I like but is it too much logic for such a short space?  Maybe if I make the comment longer?
```suggestion
# Make a mutable copy of these parameters so we don't overwrite the caller's data.
# If they weren't passed in, default to an empty list.
successes = [] if successes is None else list(successes)
failures = [] if failures is None else list(failures)
skips = [] if skips is None else list(skips)
```
(I would replace the existing comment with this version.)

What do you think?

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
+    def __init__(self, stage_name, task_header=None, next_stage=None):
+        self.stage_name = stage_name
+        self.task_header = task_header if task_header else stage_name
+        self.next_stage = next_stage
+        self._has_run = False
+
+        python_package = importlib.import_module("convert2rhel.actions.%s" % self.stage_name)
+        self.actions = get_actions(python_package.__path__, python_package.__name__ + ".")
+
+    def check_dependencies(self, previous_stage_actions=None):
```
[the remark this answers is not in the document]
abadger: Yeah, this is a bug (it should be used).  I have that fixed locally but haven't pushed it yet.

convert2rhel/actions/__init__.py
abadger: <nod>  `Something.id is technically correct yeah, but it has the problem of listing the same Action class twice that we ran into before.  I could change the code to automatically remove duplicate Action classes which would make both of these work without issues:

``` python
dependencies = ("SOMETHING",)
dependencies = (Something.id,)
```

That kinda seems like it might be the right thing to do?  What do you think?

convert2rhel/actions/__init__.py
r0x0d: Oh, that's looks way better! I like the `important` more than the `warning`! Thanks for finding that out.

convert2rhel/actions/__init__.py
r0x0d: Looks even better to me!

convert2rhel/actions/__init__.py
r0x0d: Your point makes sense to me, with my suggestion we will be losing anything that is between the ranges. 

I would presume that we have a fixed point defined that starts from `WARNING` since it is the first thing we check, for consistency, I would keep the `WARNING` for the second one too.

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
-    resolved_actions = [action for action in potential_actions if not action.dependencies]
-    resolved_action_names = set(action.name for action in resolved_actions)
+    resolved_actions = previously_resolved_actions + [action for action in potential_actions if not action.dependencies]
+
+    resolved_action_ids = set()
+    for action in resolved_actions:
+        # Build up the set of resolved_action_ids and yield all the actions
+        # which did not have dependencies
+        resolved_action_ids.add(action.id)
+        yield action
```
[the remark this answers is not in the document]
abadger: Done.

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
+    # order is stable. (Always yields the same order if the input and
+    # algorithm has not changed)
+    potential_actions = sorted(potential_actions, key=lambda action: action.id)
+
+    # Pre-seed these variables using Actions which have no dependencies
+    previous_number_of_unresolved_actions = len(potential_actions) + len(previously_resolved_actions)
     unresolved_actions = [action for action in potential_actions if action.dependencies]
-    resolved_actions = [action for action in potential_actions if not action.dependencies]
-    resolved_action_names = set(action.name for action in resolved_actions)
+    resolved_actions = previously_resolved_actions + [action for action in potential_actions if not action.dependencies]
```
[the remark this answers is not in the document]
abadger: Done

convert2rhel/actions/__init__.py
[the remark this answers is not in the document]
abadger: Fixed

convert2rhel/actions/__init__.py
abadger: `Something.id` is simply a constant string value which happens to be defined inside of a class.  It feels like using a string literal or a variable that points to a constant string should both be legal here.

I resisted uniquifying the classes at first because I worried that someone might use the same class name twice and the uniquifying code might make that silently pass and be hard to debug.  However,

(1) the Action.id would still have to be unique (we have a unittest which will fail if there are duplicates) and
(2) the Action name would have to be repeated in the same file.  Python will return different Actions if two classes use the same name, just in different files.  So my original worry is probably not valid.

Should there be a best practice of using a string literal or the id inside of the Action's class?  Both have drawbacks... The string literal could be typoed (but we have a unittest which will fail if that happens).  Using Action.id requires importing the Action which is a bit more work for the programmer and for people reading the code and for a bit more work for the computer since it will detect the imported Action and then have to remove it since it is a duplicate.

convert2rhel/actions/__init__.py
```diff
@@ -230,63 +285,199 @@ def get_actions(actions_path, prefix):
...
-def resolve_action_order(potential_actions, failed_actions):
+class Stage:
+    def __init__(self, stage_name, task_header=None, next_stage=None):
+        self.stage_name = stage_name
+        self.task_header = task_header if task_header else stage_name
+        self.next_stage = next_stage
+        self._has_run = False
+
+        python_package = importlib.import_module("convert2rhel.actions.%s" % self.stage_name)
+        self.actions = get_actions(python_package.__path__, python_package.__name__ + ".")
```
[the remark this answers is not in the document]
abadger: Docstring added.

convert2rhel/actions/__init__.py
abadger: I've added code to uniqueify this and changed the docstring to make it clear that a string literal can be used here.  Is that okay?

abadger:
[ ] Rebase now that report has been merged
[ ] Modify to give the data we return to report in main

convert2rhel/unit_tests/actions/actions_test.py
```diff
@@ -84,28 +190,645 @@ def test_found_actions(self, sys_path, test_dir_name, expected_action_names):
...
+                (
+                    ("SUCCESSTEST", "WARNINGTEST"),
+                    ("FATALTEST", "ERRORTEST", "OVERRIDABLETEST"),
+                    ("SKIPSINGLETEST", "SKIPMULTIPLETEST"),
+                ),
+            ),
+            (  # Test that exceptions inside of Actions are detected as errors
+                ("action_exceptions",),
+                (("SUCCESSTEST",), ("DIVIDEBYZEROTEST", "LOGCRITICALTEST"), ()),
+            ),
```
r0x0d, line 284:
```suggestion
            # Test that all action statuses are categorized correctly
            (
                ("all_status_actions",),
                (
                    ("SUCCESSTEST", "WARNINGTEST"),
                    ("FATALTEST", "ERRORTEST", "OVERRIDABLETEST"),
                    ("SKIPSINGLETEST", "SKIPMULTIPLETEST"),
                ),
            ),
            # Test that exceptions inside of Actions are detected as errors
            (
                ("action_exceptions",),
                (("SUCCESSTEST",), ("DIVIDEBYZEROTEST", "LOGCRITICALTEST"), ()),
            ),
```

r0x0d (commented):
Looks good to me, just some small comments made.

convert2rhel/actions/__init__.py
```diff
@@ -225,68 +282,273 @@ def get_actions(actions_path, prefix):
...
+        :type next_stage: str
+
+        Stages are used for ordering only. This is different from
+        Action.dependencies which are used for both ordering and to determine
+        whether an Action can only run if a depended upon Action succeeded.
+
+        A Stage specified in ``next_stage`` will be told about the Actions
+        which have executed successfully and unsuccessfully in ths stage when
+        it is run.
+        """
```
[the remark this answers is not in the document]
abadger: That's a good question.  I think I wonder the same thing any time I want to use more than a summary and parameter list to document a function.  There's not a lot of official documentation that I could find but I did find this section: https://www.sphinx-doc.org/en/master/tutorial/describing-code.html#cross-referencing-python-objects which seems to place the long description after the paraeer list: https://www.sphinx-doc.org/en/master/_images/lumache-py-function-full.png

The stdlib documentation doesn't use the standard sphinx method of documenting parameter lists so it isn't a great deal of help.  FWIW though, they seem to embed descriptions of the parameters in the middle of their long decsriptions.

I think I lean towards having the the parameter description first as it seems like the more quick-reference section and any long description is supplemental to that.  But I don't know that one way or the other is a clear winner.

convert2rhel/main.py
abadger: Done: https://issues.redhat.com/browse/RHELC-975

convert2rhel/actions/__init__.py
```diff
@@ -225,68 +282,273 @@ def get_actions(actions_path, prefix):
...
+                action.set_result(status="ERROR", error_id="UNEXPECTED_ERROR", message=message)
+
+            # Categorize the results
+            if action.status <= STATUS_CODE["WARNING"]:
+                logger.info("%s has succeeded" % action.id)
+                successes.append(action)
+
+            if action.status > STATUS_CODE["WARNING"]:
+                ### FIXME: Probably some code in reports that can do this better
+                logger.error("%s has failed (%s): %s" % (action.id, action.status, action.message))
```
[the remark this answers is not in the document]
abadger: This is actually the part that Michal was wanting to output success and failure as each Action ran.  I am going to see if there's something from report that can do the formatting/some of rhe formatting for this, though.

convert2rhel/actions/__init__.py
r0x0d: That's sounds good to me.

convert2rhel/actions/__init__.py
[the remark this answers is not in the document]
r0x0d: Oh, right! That's fine then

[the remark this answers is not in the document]
abadger: Okay, this piece of code is now upated.  I had to move the `format_report_message()` function from `actions/report.py` to `actions/__init__.py` to avoid a circular import and I moved a little bit of the logic from `sumary()` into `format_report_message()` to make it easier for the two pieces of code to use it together.

r0x0d (approved):

PR #791, opened 2023-03-30
[RHELC-927] Add actions report to be displayed by the end of the analysis

r0x0d:
This commit introduces the report summary that will be displayed to the user by end of the analysis, regardless if they are using doing a normal conversion, or a pre conversion check.

The report is pretty simple as it consist mainly of logging the message to the user in a nice format, ordered by most important first.

The catch here is that if the user is doing a pre conversion analysis, they will get all messages generated by the action classes displayed for them, while, if they are doing a normal conversion, only `WARNINGS` and above will be displayed.

Jira Issues: [RHELC-927](https://issues.redhat.com/browse/RHELC-927)

Checklist

### DIFF
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -23,7 +23,7 @@ from convert2rhel.actions import STATUS_CODE
 logger = logging.getLogger(__name__)
 
 
-_STATUS_NAME_FROM_CODE = dict((value, key) for key, value in actions.STATUS_CODE.items())
+_STATUS_NAME_FROM_CODE = dict((value, key) for key, value in STATUS_CODE.items())
 
 
 def _format_report_message(template, status_name, action_id, error_id, message):
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -23,24 +23,7 @@ from convert2rhel.actions import STATUS_CODE
 logger = logging.getLogger(__name__)
 
 
-def _dictionary_value_lookup(dictionary=None, look_for=None):
-    """Return the key associated with the value.
-
-    .. important:: This function will not do a depth search for the dictionary,
-        it is meant to be used with dictionaries of one level.
-
-    :param dictionary: The dictionary to be used in the revese search.
-    :type dictionary: Mapping
-    :param look_for: The value to be used in the revese search.
-    :type look_for: Any
-
-    :return: The expected key inside the dictionary
-    :rtype: str | None
-    """
-    for key, value in dictionary.items():
-        if value == look_for:
-            return key
-
+_STATUS_NAME_FROM_CODE = dict((value, key) for key, value in actions.STATUS_CODE.items())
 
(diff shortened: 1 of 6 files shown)


convert2rhel/actions/report.py
```diff
@@ -0,0 +1,155 @@
...
+    :param look_for: The value to be used in the revese search.
+    :type look_for: Any
+
+    :return: The expected key inside the dictionary
+    :rtype: str | None
+    """
+    for key, value in dictionary.items():
+        if value == look_for:
+            return key
+
```
abadger, line 43: Since STATUS_CODE is constant and we're going to be looking up the same status multiple times, I think I would implement this by switching keys and values once and then consulting that dictionary for the proper status name.  (Note: In python3, another way to do this would be to use `@functools.lru_cache`)

```suggestion
_STATUS_NAME_FROM_CODE = dict((value, key) for key, value in actions.STATUS_CODE.items())
```

convert2rhel/actions/report.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    # Sort the results in reverse order, this way, the most important messages
+    # will be on top.
+    results = sorted(results.items(), key=lambda item: item[1]["status"], reverse=True)
+
+    has_report_message = False
+    template = "({STATUS}) {ACTION_ID}"
+
+    for action_id, result in results:
+        status_name = _dictionary_value_lookup(dictionary=STATUS_CODE, look_for=result["status"])
```
abadger, line 142:
```suggestion
        status_name = _STATUS_NAME_FROM_CODE[result["status"]]
```

convert2rhel/actions/report.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if include_all_reports:
+            has_report_message = True
+            logger.info(message)
+        elif result["status"] >= STATUS_CODE["WARNING"]:
+            has_report_message = True
+            logger.info(message)
+
+    # If there is no warning or higher sent to the user, then we just give a
+    # happy message to them.
```
abadger, line 153:
```suggestion
    # If there is no other message sent to the user, then we just give a
    # happy message to them.
```

abadger (commented):
Just one comment about how to retrieve the status name from the status code.  Other than that, this looks ready to merge.

convert2rhel/unit_tests/actions/report_test.py
```diff
@@ -0,0 +1,165 @@
...
+import pytest
+
+from convert2rhel.actions import report
+
+
+@pytest.mark.parametrize(
+    ("dictionary", "look_for", "expected"), (({"zero": 0}, 0, "zero"), ({"zero": 0, "one": 1}, 0, "zero"))
+)
+def test_dictionary_value_lookup(dictionary, look_for, expected):
+    assert report._dictionary_value_lookup(dictionary, look_for) == expected
```
r0x0d, line 27:
```suggestion
```

abadger (approved):

PR #803, opened 2023-04-06
Refactor the catchall exception in main() to two catchall exceptions.

abadger:
There are really two separate catchall proceedures that differ depending on how far we've gotten into the conversion:

* Anytime prior to the PONR, we want the system to be returned to the state it was in before we started running convert2rhel.
* Anytime after the PONR, we want to log a helpful an error message.

Splitting the exception blocks to handle those seperately makes that distinction clear.  It also creates the logical place to exit from the analyze subcommand once that change is merged.

Checklist

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -77,6 +77,7 @@ def main():
 
     # handle command line arguments
     toolopts.CLI()
+
     try:
         process_phase = ConversionPhase.POST_CLI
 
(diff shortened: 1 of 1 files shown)


abadger:
I am going to close this PR.  There have been some large updates to main.py and this is a very low priority item.  This might make the code more readable but the current code works and we (those who have written code for main.main) understand it.  So there isn't a desperate need for this at this time.

Probably something to consider as a starting point when someone finds that it isn't easy for them to read main.main anymore.

PR #808, opened 2023-04-12
[RHELC-878] Remove 8.4 from supported EUS versions

pr-watson:
This PR removes the CentOS 8.4 EUS repos and their links to integration tests

Jira Issues: [RHELC-878](https://issues.redhat.com/browse/RHELC-878)

Checklist


pr-watson:
@danmyway since this PR removes RHEL 8.4 repos I believe the integration test for centos-8.4 need to be updated to use centos-8.6.

r0x0d:
@pr-watson, I think you can drop the 8.4 versions from packit too. @danmyway, do you agree?

danmyway:
> @pr-watson, I think you can drop the 8.4 versions from packit too. @danmyway, do you agree?

I agree with dropping it from the config.
However, I have a question about a timeline for this PR - will this get merged ASAP? 
Are we dropping the upstream verifications for 8.4 right now?
I suppose we won't do regression testing for the `1.3` given the RHEL 8.4 goes of out of Extended Update Support (EUS) by the time of the release.
But IMO keeping the 8.4 verifications live until May, 31st seems reasonable to me.
Again, remove it from the config in this PR at your will, I am not opposed to that at all.
Thoughts @SpyTec @bocekm ?

pr-watson:
> > @pr-watson, I think you can drop the 8.4 versions from packit too. @danmyway, do you agree?
> 
> I agree with dropping it from the config. However, I have a question about a timeline for this PR - will this get merged ASAP? Are we dropping the upstream verifications for 8.4 right now? I suppose we won't do regression testing for the `1.3` given the RHEL 8.4 goes of out of Extended Update Support (EUS) by the time of the release. But IMO keeping the 8.4 verifications live until May, 31st seems reasonable to me. Again, remove it from the config in this PR at your will, I am not opposed to that at all. Thoughts @SpyTec @bocekm ?

My PR for removing the 8.4 repos from Dogfood servers has been postponed for merging till May 31, 2023, so I assume the same would apply for this PR.

SpyTec:
For dogfood servers it makes sense as we want to be able to test against it until last second. However, I don't think we need to block this pr. We have the release branch if we need to release something urgently 

Packit is also related to testing so I'd want to keep that for now, at least until we drop support for 8.4. But I assume that would cause issues for all PRs right?

danmyway:
> For dogfood servers it makes sense as we want to be able to test against it until last second. However, I don't think we need to block this pr. We have the release branch if we need to release 
something urgently 
> 
> Packit is also related to testing so I'd want to keep that for now, at least until we drop support for 8.4. But I assume that would cause issues for all PRs right? 

If we would merge this PR, we need to remove the 8.4 from Packit config, otherwise most of the CentOS 8.4 pipeline will fail as you say @SpyTec.
So in case we want to drop the upstream verifications now, let me know and I will also remove all the 8.4 related code from tests.

SpyTec:
> If we would merge this PR, we need to remove the 8.4 from Packit config, otherwise most of the CentOS 8.4 pipeline will fail as you say 

That's what I feared. Then we should wait with this PR too and do Packit then

danmyway:
> That's what I feared. Then we should wait with this PR too and do Packit together with this PR

Cool, thank you for the clarification.
I can take care of it @pr-watson, and I will also remove the 8.4 related test code.

r0x0d:
There is a PR that removes the 8.4 support from our internal vagrant workflow. The PR will be merged once this one is merged, so, this is a future warning for myself.

tests/integration/tier1/removed-pkgs-centos-85/ansible/install_removed_pkgs_from_centos_85_repos.yml
danmyway, line 1: Hey, @pr-watson, I will revert changes in this file.
We need to add and enable the 8.4 repos, because we are looking for a package, which is not present in 8.5 repos anymore.

plans/tier1.fmf
```diff
@@ -28,7 +28,7 @@
         - enabled: false
           # TODO EUS Bump disabled version
           when: >
-            distro != centos-8.4
+            distro != centos-8.6
```
danmyway, line 31: There's no CentOS 8.6

danmyway: I will disable the plan until we have 8.6 in pipeline

danmyway:
I have rebased and modified the tests @pr-watson

tests/integration/tier0/latest-kernel-check/test_latest_kernel_check.py
```diff
@@ -14,7 +14,7 @@ def tainted_repository(shell):
     repofile = "broken_repo"
     centos_custom_reposdir = "/usr/share/convert2rhel/repos/"
 
-    # For CentOS we are working with hardcoded repos in /usr/share/convert2rhel/repos/centos-8.{4,5}
+    # For CentOS, we are working with hardcoded repos in /usr/share/convert2rhel/repos/centos-8.{4,5}
```
kokesak, line 17: the 8.4 are removed no?
```suggestion
    # For CentOS, we are working with hardcoded repos in /usr/share/convert2rhel/repos/centos-8.5
```

kokesak (commented):
Just one small thing in the comment. Otherwise, the QE side looks good!

tests/integration/tier0/latest-kernel-check/test_latest_kernel_check.py
danmyway: Thanks for spotting this

convert2rhel/unit_tests/pkgmanager/handlers/dnf/dnf_test.py
```diff
@@ -70,7 +70,7 @@ def test_set_up_base(self, pretend_os):
         instance._set_up_base()
 
         assert isinstance(instance._base, pkgmanager.Base)
-        assert instance._base.conf.substitutions["releasever"] == "8.4"
+        assert instance._base.conf.substitutions["releasever"] == "8.6"
```
r0x0d, line 73:
```suggestion
        assert instance._base.conf.substitutions["releasever"] == "8.5"
```

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -443,6 +442,6 @@ def test_print_system_information(pretend_os, caplog):
     system_info.print_system_information()
 
     assert "CentOS Linux" in caplog.records[-4].message
-    assert "8.4" in caplog.records[-3].message
+    assert "8.6" in caplog.records[-3].message
```
r0x0d, line 445:
```suggestion
    assert "8.5" in caplog.records[-3].message
```

tests/ansible_collections/roles/add-custom-repos/main.yml
```diff
@@ -8,5 +8,3 @@
   # "8.6" is to be moved to the condition below related to enabling RHEL 8 EUS repos
 - import_playbook: rhel8-repos.yml
   when: ansible_facts['distribution_version'] == "8.5" or ansible_facts['distribution_version'] == "8.7"
-- import_playbook: rhel8-eus-repos.yml
-  when: ansible_facts['distribution_version'] == "8.4"
```
r0x0d, line 12: I think this part should include the 8.6 instead of just removing the rhel8-eus-repos, right? 

```yaml
- import_playbook: rhel8-eus-repos.yml
  when: ansible_facts['distribution_version'] == "8.6"
```

Since 8.6 will be our lowest minor version supported for EUS, we should enable it for that version, look at this comment: https://github.com/oamg/convert2rhel/pull/808/files#diff-f3b2429fb7abe33e7f90d47b80809bf722ae4459862b257f57138761a4d03140R8

@danmyway my reasoning here is correct?

pr-watson:
@r0x0d thanks for the suggestions

tests/ansible_collections/roles/add-custom-repos/main.yml
danmyway: You're right @r0x0d, I will put it back

danmyway: Thanks for noticing

danmyway:
Rebased, conflicts resolved, ready to merge

r0x0d (dismissed):

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -1407,6 +1406,10 @@ def test_lock_releasever_in_rhel_repositories(pretend_os, subprocess, expected,
         "run_subprocess",
         value=run_subprocess_mock,
     )
+
+    major_version = EUS_MINOR_VERSIONS[0][0]
+    minor_version = EUS_MINOR_VERSIONS[0][2]
+    monkeypatch.setattr(system_info, "version", namedtuple("Version", ["major", "minor"])(major_version, minor_version))
```
r0x0d, line 1412: I think you could also do something like: 

```suggestion
    eus_version = EUS_MINOR_VERSION[0].split(".")
    major_version = eus_version[0]
    minor_version = eus_version[1]
    monkeypatch.setattr(system_info, "version", namedtuple("Version", ["major", "minor"])(major_version, minor_version))
```

Would look a bit cleaner and easier to understand. No need to change right now, just writing down the suggestion.

r0x0d (approved):
Looks good!

convert2rhel/unit_tests/subscription_test.py
pr-watson: It does look better! I can add that in, wondering if we should wait till the first version is merged or not.

r0x0d: Since everything passed, I would say to merge this first and then later do this change. No worries.

PR #812, opened 2023-04-15
RHSM: do not use the 'force' D-Bus registration option

ptoscano:
The 'force' D-Bus registration option used to simply be not implemented, effectively being a no-op; because of that, the system is manually unregistered before the actual registration.

Newer versions of subscription-manager (in 8.8 and 9.2) actually do implement the 'force' option; OTOH, since the system was unregistered before the registration attempt, having the option there does not make any difference. Hence, simply drop the 'force' option altogether.

Improve the comment regarding the unregistration to mention the status of the 'force' option, and the general reason for unconditionally unregistering at that point.

Checklist

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -145,10 +145,13 @@ def register_system():
         )
 
         # Force the system to be unregistered
-        # The subscription-manager DBus API has a force parameter but there's
-        # a bug in susbcription-manager where that doesn't take effect.
-        # Explicitly unregister here to workaround that.
-        # Sub-man bug: https://bugzilla.redhat.com/show_bug.cgi?id=2118486
+        # The subscription-manager D-Bus API has a 'force' parameter, however
+        # it is not implemented (and thus it does not work)
+        # - in RHEL 7 and earlier
+        # - in RHEL 8 before 8.8: https://bugzilla.redhat.com/show_bug.cgi?id=2118486
+        # - in RHEL 9 before 9.2: https://bugzilla.redhat.com/show_bug.cgi?id=2121350
+        # Explicitly unregister here to workaround that in any version,
+        # to not have to do version checks, keeping things simpler.
         loggerinst.info("Unregistering the system to clear the server's state for our registration.")
         try:
             unregister_system()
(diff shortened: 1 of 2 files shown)


pr-watson:
Change looks good to me, just waiting on integration tests to pass

pr-watson (approved):
Change looks good to me!

abadger:
>
> It's been a while since i looked at this (and I'm at pycon until the 28th
so i can't look closely at it) but i think the unregister was just a
workaround for force not working.  So we'd rather keep force and remove the
unregistration.

Of course, that might not be sufficient since we never tested that with
force working via dbus.  and I don't know if that will work on all the
versions of rhel clones (latest 7, latest centos 7, latest 8, and the eus 8
releases, latest 9 and eus 9 releases) that we care about.  So relying
solely on force might not be enough.

@mbocek might also remember this discussion and be able to decide if we
want to do it this way while i am at pycon.

bocekm:
This change is the outcome of these two BZs about the force option not working through the dbus API:
RHEL 8: https://bugzilla.redhat.com/show_bug.cgi?id=2118486
RHEL 9: https://bugzilla.redhat.com/show_bug.cgi?id=2121350

There's no BZ for RHEL 7 and I believe that even if there was one it would be rejected. RHEL 7 is not being worked on anymore except critical bugs.

As the sub-man is not going to be fixed on RHEL 7, we need to keep un-registering the system ourselves until we stop allowing conversions of RHEL 7 in the tool altogether.

To improve the user experience at least on RHEL 8+, we can use the force option on RHEL 8+ to save one dbus API call, leading to a (perhaps) faster conversion and less info log messages. @abadger, WDYT about that?

ptoscano:
> To improve the user experience at least on RHEL 8+, we can use the force option on RHEL 8+ to save one dbus API call,

Considering that the current `unregister_system()` executes `subscription-manager unregister`, there is no wasted _D-Bus_ call ;-) 

> leading to a (perhaps) faster conversion and less info log messages.

Nah, it does not really buy you anything. The slowest parts in the runtime of subscription-manager are the network interactions with the registration server (whose time you can count in seconds), you can try yourself how much time a `subscription-manager register` (to RHSM) takes to complete on an unregistered system (maybe it'll be slightly faster in a local Katello/Satellite). Because of this, an extra command invocation or D-Bus call on the system is almost "noise" in the time spent.

If you really wanted to use the `force` D-Bus option, you would need to do a "multiple version check" dance, which is not pretty and easy to maintain. I did it in the `redhat_subscription` Ansible role, and it's hard to get right. Not to mention it doubles your testing, as you have two code paths ("old sub-man" and "new sub-man") that you need to test.

LT;DR: keep the current way, it's easier for convert2rhel.

bocekm:
> Considering that the current unregister_system() executes subscription-manager unregister, there is no wasted D-Bus call ;-)

Ah, you got me :)

> LT;DR: keep the current way, it's easier for convert2rhel.

I'm ok with that.

ptoscano:
friendly ping ;)

bocekm (approved):

PR #818, opened 2023-05-05
[RHELC-895] Fix error message when repoquery fails to retrieve information on kmod packages

### LINKED ISSUE #610: CentOS7 >> RHEL 7 Converting with RH Satellite 6.11.2
Hi,

I'm getting the following errors migrating CentOS7 to RHEL 7

_""CRITICAL - No packages containing kernel modules available in the enabled repositories (rhel-7-server-rpms).""_

And the Host Output says:
output_type: stdout
    output: !ruby/string:Sequel::SQL::Blob 'fatal: [ec2-3-67-227-205.eu-central-1.compute.amazonaws.com]:
      FAILED! => {"changed": true, "cmd": ["convert2rhel", "-y", "--activationkey",
      "convert2rhel_rhel7", "--org", "kimmlingen", "--keep-rhsm"], "delta": "0:03:25.087522",
      "end": "2022-09-14 06:51:54.367869", "msg": "non-zero return code", "rc": 1,
      "start": "2022-09-14 06:48:29.280347", "stderr": "", "stderr_lines": [], "stdout":

Does anyone have a clue where I'm going to fail here?
Thanks!

DYNFLOW Output Satellite Server
https://pastebin.com/eTCNBLBU

abadger:
When repoquery fails to run (for instance, when repoquery decides that there isn't enough disk space to download metadata from the repository), we were printing an error message that there were no kmod containing packages.  With this change, we will print an error message that says repoquery failed.

Fixes https://issues.redhat.com/browse/RHELC-895
Fixes #610

Jira Issues: [RHELC-895](https://issues.redhat.com/browse/RHELC-895)

Checklist

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
+++ b/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
@@ -36,6 +36,10 @@ class RHELKernelModuleNotFound(Exception):
     pass
 
 
+class PackageRepositoryError(Exception):
+    pass
+
+
 class EnsureKernelModulesCompatibility(actions.Action):
     id = "ENSURE_KERNEL_MODULES_COMPATIBILITY"
     dependencies = ("SUBSCRIBE_SYSTEM",)
@@ -59,6 +63,12 @@ class EnsureKernelModulesCompatibility(actions.Action):
 
     def _get_rhel_supported_kmods(self):
         """Return set of target RHEL supported kernel modules."""
+        precache = [
+            "yum",
+            "makecache",
+            "--releasever=%s" % system_info.releasever,
+            "--setopt=*.skip_if_unavailable=False",
+        ]
         basecmd = [
             "repoquery",
             "--releasever=%s" % system_info.releasever,
@@ -66,9 +76,28 @@ class EnsureKernelModulesCompatibility(actions.Action):
 
         if system_info.version.major >= 8:
             basecmd.append("--setopt=module_platform_id=platform:el" + str(system_info.version.major))
+            precache.append("--setopt=module_platform_id=platform:el" + str(system_info.version.major))
 
+        active_repos = ["--disablerepo=*"]
         for repoid in system_info.get_enabled_rhel_repos():
-            basecmd.extend(("--repoid", repoid))
+            active_repos.extend(("--enablerepo", repoid))
+        precache.extend(active_repos)
+        basecmd.extend(active_repos)
+
+        # Retrieve the yum metadata for the repos we use.  yum makecache will
+        # error out if there is a problem downloading the metadata (for instance,
+        # no disk space left to save it) whereas repoquery will return no results
+        # if there is already some metadata available.
+
+        precache_out, precache_exit_code = run_subprocess(precache, print_output=False)
+        if precache_exit_code != 0:
+            raise PackageRepositoryError(
+                "We were unable to download the repository metadata for (%s) to"
+                " determine packages containing kernel modules.  Can be caused by"
+                " not enough disk space in /var/cache or too little memory.  The"
+                " yum output below may have a clue for what went wrong in this"
+                " case:\n\n%s" % (", ".join(system_info.get_enabled_rhel_repos()), precache_out)
+            )
 
         cmd = basecmd[:]
         cmd.append("-f")
@@ -77,8 +106,15 @@ class EnsureKernelModulesCompatibility(actions.Action):
         # Without the release package installed, dnf can't determine the
         # modularity platform ID. get output of a command to get all
         # packages which are the source of kmods
-        kmod_pkgs_str, _ = run_subprocess(cmd, print_output=False)
+        kmod_pkgs_str, repoquery_exit_code = run_subprocess(cmd, print_output=False)
 
+        if repoquery_exit_code != 0:
+            raise PackageRepositoryError(
+                "We were unable to query the repositories (%s) to"
+                " determine packages containing kernel modules."
+                " Output from the failed repoquery command:\n\n%s"
+                % (", ".join(system_info.get_enabled_rhel_repos()), kmod_pkgs_str)
+            )
         # from these packages we select only the latest one
         kmod_pkgs = self._get_most_recent_unique_kernel_pkgs(kmod_pkgs_str.rstrip("\n").split())
         if not kmod_pkgs:
@@ -245,6 +281,14 @@ class EnsureKernelModulesCompatibility(actions.Action):
                 return
 
             logger.debug("All loaded kernel modules are available in RHEL.")
+        except PackageRepositoryError as e:
+            self.set_result(
+                level="ERROR",
+                id="PROBLEM_WITH_PACKAGE_REPO",
+                title="Could not download package metadata",
+                description="There was an error retrieving the package metadata to verify kernel modules are available in RHEL.",
+                diagnosis=str(e),
+            )
         except RHELKernelModuleNotFound as e:
             self.set_result(
                 level="ERROR",
@@ -260,5 +304,5 @@ class EnsureKernelModulesCompatibility(actions.Action):
                 id="CANNOT_COMPARE_PACKAGE_VERSIONS",
                 title="Error while comparing packages",
                 description="There was an error while detecting the kernel package which corresponds to the kernel modules present on the system.",
-                diagnosis="Package comparison failed: %s" % e,
+                diagnosis="Package comparison failed: %s" % str(e),
             )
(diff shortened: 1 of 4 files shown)


convert2rhel/checks.py
```diff
@@ -443,8 +443,15 @@ def get_rhel_supported_kmods():
...
-    kmod_pkgs_str, _ = run_subprocess(cmd, print_output=False)
+    kmod_pkgs_str, repoquery_exit_code = run_subprocess(cmd, print_output=False)
 
+    if repoquery_exit_code != 0:
+        logger.critical(
+            "We were unable to query the repositories (%s) to"
+            " determine packages containing kernel modules."
+            " Output from the failed repoquery command:\n\n%s"
+            % (", ".join(system_info.get_enabled_rhel_repos()), kmod_pkgs_str)
+        )
```
SpyTec, line 454: We should unit test this due to it being a critical exit

SpyTec (dismissed):

convert2rhel/checks.py
abadger: <nod>.  Done.

SpyTec (approved):

convert2rhel/checks.py
```diff
@@ -443,8 +443,15 @@ def get_rhel_supported_kmods():
     #   platform ID.
     # get output of a command to get all packages which are the source
     # of kmods
-    kmod_pkgs_str, _ = run_subprocess(cmd, print_output=False)
+    kmod_pkgs_str, repoquery_exit_code = run_subprocess(cmd, print_output=False)
 
+    if repoquery_exit_code != 0:
+        logger.critical(
+            "We were unable to query the repositories (%s) to"
+            " determine packages containing kernel modules."
```
bocekm, line 451: What about adding a note that the reason might be having not enough free disk space? Without any hint like the users would could get easily frustrated, not knowing how to proceed further.

abadger: The repoquery output which is included on line 452 should include that:
``` 
# repoquery --setopt=module_platform_id=platform:el8 -f '/lib/modules/*.ko*'
Main config did not have a module_platform_id attr. before setopt
Insufficient space in download directory /var/cache/yum/x86_64/7/updates
    * free  4.0 M
    * needed 21 M
```

I could add a little bit that says disk space is one possible reason in case the repoquery output doesn't include that summary.  However, if he repoquery output doesn't include it, my thought is it would be something different than the disk space issue..

What do you think?

abadger: Hmmm... I haven't yet.  I did see that the error went to stderr which I thought might have caused problems but `run_subprocess()` directs stderr to stdout.  That should have output the repoquery message before though, so I'll double check whether it's at least working with this PR now.

abadger: Actually... not just not-enabled... not configured at all. there were no repo files referencing those cache directories.

abadger: And now I'm getting the opposite result (no repodata gives the out of space error) ... However, this makes more sense.... If I have repodata downloaded that says the file belongs to package foo in repo bar but repo bar is disabled, that could just be filtered out, then I could see repoquery saying there are no packages... it's sort of saying "there are no packages with these files [in any of the repositories I have information on]".  OTOH, if there is no repodata downloaded, then repoquery is saying "I need to download data from at least one repository to do anything at all.  You need to free up some space for me to work".

Inside of convert2rhel's run, we will download the repodata for the source os as part of the early checks.  Then, after we enable the RHEL repos, we will do the kmod packages check.  That means that we will always have some repodata when we perform the kmod package check but it won't be for the repository we need.  Thus w will always fall into the first case ("there are no packages with these files [in the repositories I have downloaded metadata for]").

abadger: Okay, this has now been reworked to run yum makecache before running hte repoquery.  yum makecache will fail if there is not enough disk space with a helpful error message which we can then show to the user.  So hopefully this will work better than repoquery alone.

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
```diff
@@ -67,8 +76,26 @@ def _get_rhel_supported_kmods(self):
         if system_info.version.major == 8:
             basecmd.append("--setopt=module_platform_id=platform:el8")
 
+        active_repos = ["--disablerepo=*"]
```
bocekm, line 79: This is tricky. What if the user uses the `--disablerepo` c2r option?

We handle it in the `call_yum_cmd` (https://github.com/oamg/convert2rhel/blob/v1.3.2/convert2rhel/pkghandler.py#L123).

bocekm (changes requested):

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
abadger: I think `--disablerepo` from the c2r options is safe to ignore as `--disablerepo=*` is a superset of anything that the user might have specified.

OTOH, prior to this change, we didn't disable any repos when running repoqery in this function.  I'm not sure how that worked... It seems like we would be running repoquery using both the vendor's repos and the RHEL repos in that case, but we only want to find kernel modules that are in packages in the RHEL repos here?

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
```diff
@@ -67,8 +76,26 @@ def _get_rhel_supported_kmods(self):
         if system_info.version.major == 8:
             basecmd.append("--setopt=module_platform_id=platform:el8")
```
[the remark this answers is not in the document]
abadger: <nod>  Do you think we could simply add the setopt to the yum makecache and it would be enough to mitigate any potential problems or do we need to figure out what problems could exist and whether doing that is safe and effective? 
```suggestion
            basecmd.append("--setopt=module_platform_id=platform:el8")
            precache.append("--setopt=module_platform_id=platform:el8")            
```

abadger:
@bocekm Does the latest set of changes address your concerns?

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
```diff
@@ -245,6 +281,14 @@ def run(self):
                 return
 
             logger.debug("All loaded kernel modules are available in RHEL.")
+        except PackageRepositoryError as e:
+            self.set_result(
+                level="ERROR",
+                id="PROBLEM_WITH_PACKAGE_REPO",
+                title="Could not download package metadata",
+                description="There was an error retrieving the package metadata to verify kernel modules are available in RHEL.",
+                diagnosis="%s" % str(e),
```
r0x0d, line 290: Out of curiosity, why not just str(e)?

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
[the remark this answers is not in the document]
bocekm: I think simply setting it is fine. Thanks.

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
bocekm: > I think --disablerepo from the c2r options is safe to ignore as --disablerepo=* is a superset of anything that the user might have specified.

Ah, right, that makes sense.

> It seems like we would be running repoquery using both the vendor's repos and the RHEL repos in that case, but we only want to find kernel modules that are in packages in the RHEL repos here?

Yes, we want to query just the RHEL repos. Previously we've been using the `--repoid` option.

```
--repoid=<repo>
     Specify  which  repository  to query. Using this option disables all repositories not explicitly enabled with --repoid option (can be used multiple times). By default repoquery uses whatever repositories are enabled in YUM configuration.
```
So we were really querying just the RHEL repos. Now given that yum doesn't have this option and you've switched to using `--enablerepo` then passing `--disablerepo=*` makes sense.

bocekm (approved):

abadger:
This is Dev complete.  Just needs integration tests.

r0x0d:
Probably its best to rebase this and get the newest changes? Making sure the other targets will run the tests properly as well

abadger:
Rebased and also squashed into one commit.

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
abadger: Now fixed to use str(e).

SpyTec (approved):

SpyTec:
@abadger Needs a rebase then we can merge

PR #828, opened 2023-05-11
Handle empty arch in package information

r0x0d:
Handle the cases where the `arch` will be empty when getting the installed packages information.

This rare case happens most of the times when we are querying the `gpg-pubkeys` package, which does not contain architecture, thus, the function returns an `.(none)` instead.

Checklist


r0x0d:
> Seems good to me but we should attach this to a Jira since it handles an edge-case

Added link to Jira ticket.

PR #832, opened 2023-05-18
Do not honor CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK when in analyze mode.

abadger:
`analyze` depends on being able to rollback all changes to the system to get back to the initial state.  If INCOMPLETE_ROLLBACK is specified, we wouldn't be able to do that, which would leave the user with a system which can't be converted without manual intervention after running `convert2rhel analyze.`  So we need to prevent that from happening.

Note that having the check for the environment variable and analyze activity in utils.download_pkg() isn't great.  It means that we will report a FAILURE when we get there whereas it would be more accurate to report a SKIP.  In the future we should redesign this so download_pkgs() raises an exception in this case and a higher level catches the exception and decides what to do based on the environment variable and the activity.

Checklist


convert2rhel/utils.py
```diff
@@ -717,14 +717,45 @@ def download_pkg(
...
             loggerinst.critical(
                 "Couldn't download the %s package. This means we will not be able to do a"
                 " complete rollback and may put the system in a broken state.\n"
                 "Check to make sure that the %s repositories are enabled"
                 " and the package is updated to its latest version.\n"
                 "If you would rather ignore this check set the environment variable"
-                " 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK'." % (pkg, system_info.name)
+                " 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK'. Note that the"
+                " environment variable is only consulted when actually converting a"
+                " system. It is disabled when performing a pre-conversion analysis." % (pkg, system_info.name)
```
SpyTec, line 758: This is redundant I feel. I do not think we need to specify this as the analyze command is separate workflow in a way

r0x0d: Agreed with @SpyTec, it seems a bit redundant in this case.

abadger: Currently this error message is output in both the analyze and convert workflow.  So without this portion of the message, the user would run `CONVERT2RHEL_EXPERIMENTAL_ANALYSIS=1 convert2rhel` , get the error message, rerun it as `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK=1 CONVERT2RHEL_EXPERIMENTAL_ANALYSIS=1 convert2rhel` and then be left wondering why they got the same error message telling them to use `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK=1`.

We could probably give a different error message when running under analyze than from convert... We probably do need to mention that `INCOMPLETE_ROLLBACK` isn't supported in the analyze case but we could forgo adding this in the convert case.

abadger: This has been updated so that there is a separate error message for analysis and conversions

convert2rhel/utils.py
```diff
@@ -717,23 +717,59 @@
...
+        #     have to download a package in order for insights or
+        #     subscription-manager to run. In those cases, we either cannot use
+        #     this function or we might show the user a misleading message).
+        # (3) Functions in utils should be free of other dependencies within
+        #     convert2rhel.  That allows us to use utils with no fear of
+        #     circular dependency issues.
+        # (4) Making the choices here mean that when used inside of the Action
+        #     framework, we are limited to returning a FAILURE for the Action
+        #     plugin whereas returning SKIP would be more accurate.
+        from convert2rhel import toolopts
```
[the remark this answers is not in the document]
abadger: AFAICS, this is not an error because the import is happening inside of this function and the function is not being used inside of `toolopts`.

abadger:
As this is a fix for a new feature in 1.3, this should not need a changelog entry if it is merged for 1.3.

convert2rhel/utils.py
SpyTec: LGTM

convert2rhel/utils.py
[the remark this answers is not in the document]
SpyTec: Likely just checking if import exists in both files, which is valid as we are playing with fire a bit. But I concur, not an issue here

SpyTec (approved):

convert2rhel/utils.py
```diff
@@ -717,23 +717,59 @@ def download_pkg(
...
-                " complete rollback and may put the system in a broken state.\n"
-                "'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK' environment variable detected, continuing conversion."
-                % (pkg)
+            loggerinst.critical(
+                "Couldn't download the %s package which is needed to do a rollback of this action."
+                " Check to make sure that the %s repositories are enabled and the package is"
+                " updated to its latest version.\n"
+                "Note that you can choose to ignore this check when actually running a conversion by"
+                " setting the environment variable 'CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK'"
+                " but not during pre-conversion analysis."
```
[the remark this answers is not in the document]
abadger: Děkuji, Dan!

tests/integration/conftest.py
```diff
@@ -327,3 +327,48 @@ def required_packages(shell):
 
     except KeyError:
         raise
+
+
+@pytest.fixture(scope="function")
+def repositories(shell):
```
kokesak, line 333: I would name it something like `backup_system_repositories` or something like that which is more self-explaining.

tests/integration/conftest.py
```diff
@@ -327,3 +327,48 @@ def required_packages(shell):
...
+    assert shell(f"rm -rf {backup_dir}")
+    # Return EUS repositories to their original location
+    if "centos-8" in SYSTEM_RELEASE_ENV:
+        assert shell(f"mv {backup_dir_eus}/* /usr/share/convert2rhel/repos/").returncode == 0
+        assert shell(f"rm -rf {backup_dir_eus}")
+
+
+@pytest.fixture(autouse=True)
+def missing_centos_release_workaround(system_release, shell):
+    # TODO(danmyway) remove when/if the issue gets fixed
```
kokesak, line 363: Do we have a link to jira to something that would point to the issue? If so please add it here.

kokesak (changes requested):

kokesak (approved):
The requested changed will be added in a separate fix

tests/integration/conftest.py
danmyway: Will update in separate PR to unblock merge

tests/integration/conftest.py
danmyway: Will update in separate PR to unblock merge

abadger:
Thanks @danmyway and @kokesak !

PR #835, opened 2023-05-19
chore: update vscode directory settings

SpyTec:
This removes the settings file and also adds a suggested extensions file

Checklist

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,15 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"ms-python.python",
+		"ms-vscode-remote.remote-containers",
+		"github.vscode-pull-request-github",
+		"eamodio.gitlens",
+		"ms-python.isort",
+		"github.vscode-github-actions",
+	],
+	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
+	"unwantedRecommendations": []
+}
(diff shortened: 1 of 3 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,15 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"ms-python.vscode-pylance",
```
r0x0d, line 6: Maybe instead of just pylance we could use the `ms-python.python` instead? Which comes with everything related to python, including pylance.

.vscode/extensions.json
```diff
@@ -0,0 +1,15 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
```
r0x0d, line 5: What about including `aaron-bond.better-comments`? We have a lot of comments in the code, and I think this extension helps a lot to make them more visible...

r0x0d (approved):
Looks lovely to me! Just two comments that I will decide if you want to add them or not.

.vscode/extensions.json
SpyTec: Updated!

.vscode/extensions.json
SpyTec: I want to keep all nice to have things that are purely aesthetically to each user, as some may not want colors on certain things that causes them to stand out etc.

But if you want it in we can add it as well

PR #841, opened 2023-05-24
[RHELC-1042] Updates to the action framework

pr-watson:
This PR contains updates to the action framework so it can accomdate returning multiple errors as well as reporting info and warning messages in the analyze report.

Jira Issues: [RHELC-1042](https://issues.redhat.com/browse/RHELC-1042)

Checklist

### DIFF
--- a/convert2rhel/actions/__init__.py
+++ b/convert2rhel/actions/__init__.py
@@ -22,6 +22,7 @@ import importlib
 import itertools
 import logging
 import pkgutil
+import re
 import traceback
 
 from functools import wraps
@@ -37,13 +38,13 @@ logger = logging.getLogger(__name__)
 #: Status code of an Action.
 #:
 #: When an action completes, it may have succeeded or failed.  We set the
-#: `Action.status` attribute to one of the following values so that we know
-#: what happened.  This mapping lets us use a symbolic name for the status
+#: `Action.level` attribute to one of the following values so that we know
+#: what happened.  This mapping lets us use a symbolic name for the level
 #: for readability but that is mapped to a specific integer for consumption
 #: by other tools.
 #:
 #: .. note:: At the moment, we only make use of SUCCESS and ERROR.  Other
-#:      statuses may be used in future releases as we refine this system
+#:      levels may be used in future releases as we refine this system
 #:      and start to use it with console.redhat.com
 #:
 #: :SUCCESS: no problem.
@@ -59,10 +60,10 @@ logger = logging.getLogger(__name__)
 #:      tests can be run.
 #:
 #: .. warning::
-#:      * Do not change the numeric value of these statuses once they
+#:      * Do not change the numeric value of these levels once they
 #:        have been in a public release as external tools may be depending on
 #:        the value.
-#:      * Actions should not set a status to ``SKIP``.  The code which
+#:      * Actions should not set a level to ``SKIP``.  The code which
 #:        runs the Actions will set this.
 STATUS_CODE = {
     "SUCCESS": 0,
@@ -102,8 +103,8 @@ def _action_defaults_to_success(func):
     def wrapper(self, *args, **kwargs):
         return_value = func(self, *args, **kwargs)
 
-        if self.status is None:
-            self.status = STATUS_CODE["SUCCESS"]
+        if self.result is None:
+            self.result = ActionResult("SUCCESS")
 
         return return_value
 
@@ -116,6 +117,10 @@ _NO_USER_VALUE = object()
 
 class ActionError(Exception):
     """Raised for errors related to the Action framework."""
+
+
+class InvalidMessageError(ActionError):
+    """Raised when invalid parameters are passed to set_result() or add_message()."""
 
 
 class DependencyError(ActionError):
@@ -137,7 +142,7 @@ class DependencyError(ActionError):
         self.resolved_actions = kwargs.pop("resolved_actions", [])
 
 
-#: Contains Actions which have run, separated into categories by status.
+#: Contains Actions which have run, separated into categories by level.
 #:
 #: :param successes: Actions which have run successfully
 #: :type: Sequence
@@ -183,10 +188,9 @@ class Action:
         They represent whether the Change succeeded or failed and if it failed,
         gives useful information to the user.
         """
-        self.status = None
-        self.message = None
-        self.error_id = None
         self._has_run = False
+        self._result = None
+        self.messages = []
 
     @_action_defaults_to_success
     @abc.abstractmethod
@@ -194,9 +198,8 @@ class Action:
         """
         The method that performs the action.
 
-        .. note:: This method should set :attr:`status`, :attr:`message`, and
-            attr:`error_id` before returning.  The @action_defaults_to_success
-            decorator takes care of setting a default success status but you
+        .. note:: This method should set self.result before returning.  The @action_defaults_to_success
+            decorator takes care of setting a default success level but you
             can either add more information (for instance, a message to
             display to the user) or make additional changes to return an error
             instead.
@@ -206,25 +209,128 @@ class Action:
 
         self._has_run = True
 
-    def set_result(self, status=_NO_USER_VALUE, error_id=_NO_USER_VALUE, message=_NO_USER_VALUE):
+    @property
+    def result(self):
         """
-        Helper method that sets the resulting values for status, error_id and message.
+        Once the :class:`Action` is run, this holds the `result`.  The result is an :class:`ActionResult`
+        which holds information of whether the Action succeeded or failed and why.
+        """
+        return self._result
 
-        :param status: The status to be set.
-        :type: status: str
-        :param error_id: The error_id to identify the error.
-        :type error_id: str
+    @result.setter
+    def result(self, action_message):
+        """
+        Make sure the correct policy for result messages is used.
+        """
+        if not isinstance(action_message, ActionResult):
+            raise TypeError()
+
+        self._result = action_message
+
+    def set_result(self, level, id=None, message=None):
+        """
+        Helper method that sets the resulting values for level, id and message.
+
+        :param id: The id to identify the result.
+        :type id: str
+        :param level: The status_code of the result .
+        :type: level: str
         :param message: The message to be set.
         :type message: str | None
         """
-        if status != _NO_USER_VALUE:
-            self.status = STATUS_CODE[status]
+        if level not in ("ERROR", "OVERRIDABLE", "SKIP", "SUCCESS"):
+            raise KeyError("The level of result must be FAILURE, OVERRIDABLE, SKIP, or SUCCESS.")
 
-        if error_id != _NO_USER_VALUE:
-            self.error_id = error_id
+        self.result = ActionResult(level, id, message)
 
-        if message != _NO_USER_VALUE:
-            self.message = message
+    def add_message(self, level, id=None, message=None):
+        """
+        Helper method that adds a new informational message to display to the user.
+        The passed in values for level, id and message of a warning or info log message are
+        saved in an :class:`ActionMessage` which is appended to ``self.messages``
+
+        :param id: The id to identify the message.
+        :type id: str
+        :param level: The level to be set.
+        :type: level: str
+        :param message: The message to be set.
+        :type message: str | None
+        """
+        msg = ActionMessage(level, id, message)
+        self.messages.append(msg)
+
+
+class ActionMessageBase:
+    """
+    Parent class for setting message and result variables and returning
+    those values in a dictionary format
+
+    :keyword id: The id to identify the message.
+    :type id: str
+    :keyword level: The level to be set, defaulting to SUCCESS.
+    :type: level: str
+    :keyword message: The message to be set.
+    :type message: str | None
+    """
+
+    def __init__(self, level="SUCCESS", id=None, message=""):
+        self.id = id
+        self.level = STATUS_CODE[level]
+        self.message = message
+
+    def to_dict(self):
+        """
+        Returns a dictionary representation of the :class:`ActionMessageBase`.
+        :returns: The attributes of :class:`ActionBase` expressed as a dictionary
+        :rtype: dict
+        """
+        return {
+            "id": self.id,
+            "level": self.level,
+            "message": self.message,
+        }
+
+
+class ActionMessage(ActionMessageBase):
+    """
+    A class that defines the contents and rules for messages set through :meth:`Action.add_message`.
+    """
+
+    def __init__(self, level=None, id=None, message=None):
+        if not (id and level and message):
+            raise InvalidMessageError("Messages require id, level and message fields")
+
+        # None of the result status codes are legal as a message.  So we error if any
+        # of them were given here.
+        if not (STATUS_CODE["SUCCESS"] < STATUS_CODE[level] < STATUS_CODE["SKIP"]):
+            raise InvalidMessageError("Invalid level '%s', set for a non-result message" % level)
+
+        super(ActionMessage, self).__init__(level, id, message)
+
+
+class ActionResult(ActionMessageBase):
+    """
+    A class that defines content and rules for messages set through :meth:`Action.set_result`.
+    """
+
+    def __init__(self, level="SUCCESS", id=None, message=""):
+
+        if STATUS_CODE[level] >= STATUS_CODE["SKIP"]:
+            if not id and not message:
+                raise InvalidMessageError("Non-success results require an id and a message")
+
+            if not id:
+                raise InvalidMessageError("Non-success results require an id")
+
+            if not message:
+                raise InvalidMessageError("Non-success results require a message")
+
+        elif STATUS_CODE["SUCCESS"] < STATUS_CODE[level] < STATUS_CODE["SKIP"]:
+            raise InvalidMessageError(
+                "Invalid level '%s', the level for result must be SKIP or more fatal or SUCCESS." % level
+            )
+
+        super(ActionResult, self).__init__(level, id, message)
 
 
 def get_actions(actions_path, prefix):
@@ -250,7 +356,7 @@ def get_actions(actions_path, prefix):
                                      system_checks.__name__ + ".")
         for action in resolve_action_order(action_classes):
             action.run()
-            if action.status = STATUS_CODE["SUCCESS"]:
+            if action.level == STATUS_CODE["SUCCESS"]:
                 successful_actions.append(action)
             else:
                 failed_actions.append(action)
@@ -345,7 +451,7 @@ class Stage:
             in addition to any that were run in this :class`Stage`.
 
         :rtype: FinishedActions
-        .. important:: Success is currently defined as an action whose status after
+        .. important:: Success is currently defined as an action whose level after
             running is WARNING or better (WARNING or SUCCESS) and
             failure as worse than WARNING (OVERRIDABLE, ERROR)
         """
@@ -382,7 +488,7 @@ class Stage:
                     to_be,
                 )
 
-                action.set_result(status="SKIP", error_id="SKIP", message=message)
+                action.set_result(level="SKIP", id="SKIP", message=message)
                 skips.append(action)
                 failed_action_ids.add(action.id)
                 logger.error("Skipped %s. %s" % (action.id, message))
@@ -400,15 +506,17 @@ class Stage:
                     " fixed or a specific error message added.\n"
                     "Traceback: %s" % (e, traceback.format_exc())
                 )
-                action.set_result(status="ERROR", error_id="UNEXPECTED_ERROR", message=message)
+                action.set_result(level="ERROR", id="UNEXPECTED_ERROR", message=message)
 
             # Categorize the results
-            if action.status <= STATUS_CODE["WARNING"]:
+            if action.result.level <= STATUS_CODE["WARNING"]:
                 logger.info("%s has succeeded" % action.id)
                 successes.append(action)
 
-            if action.status > STATUS_CODE["WARNING"]:
-                message = format_action_status_message(action.status, action.id, action.error_id, action.message)
+            if action.result.level > STATUS_CODE["WARNING"]:
+                message = format_action_status_message(
+                    action.result.level, action.id, action.result.id, action.result.message
+                )
                 logger.error(message)
                 failures.append(action)
                 failed_action_ids.add(action.id)
@@ -433,13 +541,10 @@ def resolve_action_order(potential_actions, previously_resolved_actions=None):
         an Action.
     :returns: Iterator of Actions sorted so that all dependent Actions are run
         before actions which depend on them.
-
     .. note::
-
         * The sort is stable but not predictable. The order will be the same
           as long as the Actions given has not changed but adding or
           subtracting Actions or dependencies can alter the order a lot.
-
         * The returned actions do not include ``previously_resolved_actions``.
           It is up to the caller to processs those before the actions returned
           by this function.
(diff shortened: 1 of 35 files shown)


convert2rhel/unit_tests/actions/actions_test.py
```diff
@@ -40,6 +40,9 @@ class _ActionForTesting(actions.Action):
 
     def __init__(self, **kwargs):
         super(_ActionForTesting, self).__init__()
+        action_id = kwargs.pop("action_id")
+        self.id = action_id
```
abadger, line 44:
```suggestion
```

convert2rhel/unit_tests/actions/actions_test.py
```diff
@@ -317,15 +320,21 @@ class TestResolveActionOrder:
...
-                    _ActionForTesting(id="One"),
-                    _ActionForTesting(id="Two", dependencies=("One",)),
-                    _ActionForTesting(id="Three", dependencies=("Two",)),
-                    _ActionForTesting(id="Four", dependencies=("Three",)),
+                    _ActionForTesting(action_id="One"),
+                    _ActionForTesting(action_id="Two", dependencies=("One",)),
+                    _ActionForTesting(action_id="Three", dependencies=("Two",)),
+                    _ActionForTesting(action_id="Four", dependencies=("Three",)),
                 ],
                 ["One", "Two", "Three", "Four"],
```
abadger, line 339: You can leave these as `id`

convert2rhel/unit_tests/actions/actions_test.py
```diff
@@ -641,37 +650,37 @@ class TestRunActions:
...
-                    "One": {"error_id": None, "message": None, "status": STATUS_CODE["SUCCESS"]},
+                    "One": {"id": None, "message": None, "level": STATUS_CODE["SUCCESS"]},
                 },
             ),
             (
                 actions.FinishedActions(
                     [
                         _ActionForTesting(
-                            id="One", status=STATUS_CODE["WARNING"], error_id="DANGER", message="Warned about danger"
+                            action_id="One", level=STATUS_CODE["WARNING"], id="DANGER", message="Warned about danger"
```
abadger, line 666:
```suggestion
                            id="One", result=ActionResult(level=STATUS_CODE["WARNING"], id="DANGER", message="Warned about danger")
```

abadger:
I've changed the jira link for this PR.  This PR will mostly cover the change to the framework to accomodate multiple messages.  Examining messages and outputting messages for all of the warning and some of the info messages can be done in this or another PR (I would suggest another PR as multiple people can work on that after this one is merged).

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,94 @@ def run(self):
 
         self._has_run = True
 
-    def set_result(self, status=_NO_USER_VALUE, error_id=_NO_USER_VALUE, message=_NO_USER_VALUE):
+    @property
+    def result(self):
```
r0x0d, line 215: No basic docstrings here

convert2rhel/actions/__init__.py
```diff
@@ -59,11 +60,12 @@
...
 #:
 #: .. warning::
-#:      * Do not change the numeric value of these statuses once they
+#:      * Do not change the numeric value of these levels once they
 #:        have been in a public release as external tools may be depending on
 #:        the value.
-#:      * Actions should not set a status to ``SKIP``.  The code which
+#:      * Actions should not set a level to ``SKIP``.  The code which
 #:        runs the Actions will set this.
+# add info status code as 25
```
r0x0d, line 68: This will be done as part of this PR?

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,94 @@ def run(self):
...
+        :param level: The status_code of the result .
+        :type: level: str
+        :param message: The message to be set.
+        :type message: str | None
+        """
+        if level not in ("ERROR", "OVERRIDABLE", "SKIP", "SUCCESS"):
+            raise KeyError("The level of result must be FAILURE, OVERRIDABLE, SKIP, or SUCCESS.")
+
+        result = ActionResult(level, id, message)
+        self.result = result
```
r0x0d, line 243: Not a necessary change, but I would lean towards doing something like: 


```suggestion
        self.result = ActionResult(level, id, message)
```

Personally, I don't like setting the result (or instance, in that case) of something to a variable and then doing an assigment adds too much value to the code, instead, on cases like this, I would assign it directly to the thing we want to assign.

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,94 @@ def run(self):
...
+
+    def add_message(self, level, id=None, message=None):
```
r0x0d, line 245: Should we call this `set_message` instead? Just to apply the same naming as `set_result`

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,94 @@ def run(self):
...
+            "id": self.id,
+            "level": self.level,
+            "message": self.message,
+        }
+
+
+class ActionMessage(ActionMessageBase):
+    def __init__(self, level, id=None, message=""):
+        if not (id and level and message):
+            raise InvalidMessage("Messages require id, level and message fields")
```
r0x0d, line 277: Since all three are required at this point, I believe you could remove the assigment for the param. 

```suggestion
    def __init__(self, level, id, message):
        if not (id and level and message):
            raise InvalidMessage("Messages require id, level and message fields")
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,94 @@ def run(self):
...
+
+        if not (STATUS_CODE["SUCCESS"] < STATUS_CODE[level] < STATUS_CODE["SKIP"]):
```
r0x0d, line 279: Wouldn't be better to do something like: 


```suggestion
        if level not in STATUS_CODES.keys():
```

Assuming that we are checking if a level set is not present in the possibilities of `STATUS_CODES`.

r0x0d: With that suggestion, if we add a new status_code in the future, we won't need to take care of changing this statement to reflect it.

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,94 @@ def run(self):
...
+                raise InvalidMessage("Non-success results require an id and a message")
+
+            if not id:
+                raise InvalidMessage("Non-success results require an id")
+
+            if not message:
+                raise InvalidMessage("Non-success results require a message")
+
+        elif STATUS_CODE["SUCCESS"] < STATUS_CODE[level] < STATUS_CODE["SKIP"]:
+            raise InvalidMessage("The level for result must be SKIP or more fatal or SUCCESS.")
```
r0x0d, line 299: The message here is kind of confusing... It is not clear if the level should be between the levels `SUCCESS` and `SKIP`, or , if it is only allowed `SUCCESS` or anything higher than `SKIP`. 

I would suggest doing the same as the previous class, both in the if statement as the message, to make it more explicit of the intentions here. 

If this will be changed, I would suggest creating a method in `ActionMessageBase` that does this validation, as it is something common for both classes that inherit it.

convert2rhel/actions/__init__.py
```diff
@@ -250,7 +324,7 @@ def get_actions(actions_path, prefix):
                                      system_checks.__name__ + ".")
         for action in resolve_action_order(action_classes):
             action.run()
-            if action.status = STATUS_CODE["SUCCESS"]:
+            if action.level = STATUS_CODE["SUCCESS"]:
```
r0x0d, line 327:
```suggestion
            if action.level == STATUS_CODE["SUCCESS"]:
```

convert2rhel/actions/__init__.py
```diff
@@ -423,26 +499,13 @@ def resolve_action_order(potential_actions, previously_resolved_actions=None):
...
-        * The returned actions do not include ``previously_resolved_actions``.
-          It is up to the caller to processs those before the actions returned
-          by this function.
+    :arg potential_actions: Sequence of Actions which should be run.
+    :arg failed_actions: A set of failed Actions.  resolve_action_order() does not modify this but
+        does expect that its contents will be updated between iterations if there are more
+        failures.
+    :returns: A list of Actions sorted so that all dependent Actions are run before actions which
+        depend on their output.
+    :raises Exception: when it is impossible to satisfy a dependency in an Action.
```
r0x0d, line 508: Why the docstring here was entirely changed?

convert2rhel/actions/__init__.py
```diff
@@ -540,17 +603,23 @@ def run_actions():
...
-    #                 "error_id": "$error_id",
-    #                 "message": "" or "$message"},
+    # {"$Action_id": {
+    #  "messages" : ["level": int,
+    #                "id": "$id",
+    #                "message": "" or "$message"}],
+    #  "result" : {
+    #      "level": int,
+    #      "id": "$id",
+    #      "message": "" or "$message"},
```
r0x0d, line 613:
```suggestion
    # Format results as a dictionary:
    # {
    #     "$Action_id": {
    #         "messages": [
    #             "level": int,
    #             "id": "$id",
    #             "message": "" or "$message"
    #         ]
    #     },
    #     "result": {
    #         "level": int,
    #         "id": "$id",
    #         "message": "" or "$message"
    #     },
    # }
```

r0x0d: There was an error in the JSON were the order of closing the brackets was inverse. 

I updated that and formatted the JSON to look better in the comment.

convert2rhel/actions/__init__.py
```diff
@@ -567,34 +636,35 @@ def find_actions_of_severity(results, severity):
         # matched_actions will contain all actions which were skipped
         # or failed while running.
     """
-    matched_actions = [action for action in results.items() if action[1]["status"] >= STATUS_CODE[severity]]
-
+    print("find_actions_of_severity")
+    matched_actions = [message for message in results.items() if key(message) >= STATUS_CODE[severity]]
+    print(matched_actions)
```
r0x0d, line 641: No prints statement needed


```suggestion
    matched_actions = [message for message in results.items() if key(message) >= STATUS_CODE[severity]]
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,94 @@ def run(self):
...
-            self.status = STATUS_CODE[status]
+        msg = ActionMessage(level, id, message)
+        self.messages.append(msg)
 
-        if error_id != _NO_USER_VALUE:
-            self.error_id = error_id
 
-        if message != _NO_USER_VALUE:
-            self.message = message
+class ActionMessageBase:
```
r0x0d, line 260: Missing docstring for `ActionMessageBase` and the classes that inherit it.

convert2rhel/actions/__init__.py
pr-watson: Noted

convert2rhel/actions/__init__.py
pr-watson: Not as part of this PR, once we finish the PR to evaluate if we have any info messages to add or not we will decide whether we need to add it or not

convert2rhel/actions/__init__.py
pr-watson: I'll add this

convert2rhel/actions/__init__.py
pr-watson: It would be better to have the same naming scheme, I'll change this

convert2rhel/actions/__init__.py
pr-watson: This is checking that the level is WARNING for a message, in the future if we add INFO, it will also fall between SUCCESS and SKIP and remain valid. We basically only want to have WARNING as a valid level for messages, and INFO in the future

convert2rhel/actions/__init__.py
pr-watson: This message is saying that the level cannot be anything between success and skip, in other words no WARNING or INFO messages (if we add that)

pr-watson: So for example if we add a result with a warning level then this error will be raised to say the level needs to be either SKIP, more fatal than SKIP or SUCCESS

pr-watson: I can add what level caused the error

convert2rhel/actions/__init__.py
pr-watson: Not sure, maybe I had an older version of the docstring, I will revert it

convert2rhel/actions/__init__.py
pr-watson: Will remove these thanks

convert2rhel/actions/__init__.py
pr-watson: I will add these

convert2rhel/actions/__init__.py
```diff
@@ -118,6 +120,10 @@ class ActionError(Exception):
     """Raised for errors related to the Action framework."""
 
 
+class InvalidMessage(ActionError):
```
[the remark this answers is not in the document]
pr-watson: Hmm this is a tough one, I see your point about having Error in the name. I can change it for now and make edits later on

convert2rhel/actions/__init__.py
```diff
@@ -540,17 +641,28 @@ def run_actions():
...
+    #             "id": "$id",
+    #             "message": "" or "$message"
+    #         ]
+    #     },
+    #     "result": {
+    #         "level": int,
+    #         "id": "$id",
+    #         "message": "" or "$message"
+    #     },
     # }
```
abadger, line 657:
```suggestion
    #     "$Action_id": {
    #         "messages": [
    #             "level": int,
    #             "id": "$id",
    #             "message": "" or "$message"
    #         ]
    #         "result": {
    #             "level": int,
    #             "id": "$id",
    #             "message": "" or "$message"
    #         },
    #     },
    # }
```

The structure is a tiny bit different than this.  `result` and `messages` are both cildren of `$Action_id`.

convert2rhel/actions/__init__.py
abadger, line 68:
```suggestion
```
It's probably better to track this in [RHELC-976](https://issues.redhat.com/browse/RHELC-976?focusedId=22507957&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22507957)

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
 
         self._has_run = True
 
-    def set_result(self, status=_NO_USER_VALUE, error_id=_NO_USER_VALUE, message=_NO_USER_VALUE):
+    @property
+    def result(self):
+        """
+        Helper method to save result variable in a different variable, so it does
+        not get overwritten by the __init__ method of the Action class
```
abadger, line 218:
```suggestion
        Once the :class:`Action` is run, this holds the `result`.  The result is an :class:`ActionResult`
        which holds information of whether the Action succeeded or failed and why.
```

abadger: The docstring on the `@property` decorated method ends up being used as the docstring for the property itself so it should be written as if it is describing the attribute rather than the method.

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
+        return self._result
+
+    @result.setter
+    def result(self, action_message):
+        """
+        Make sure the correct policy for result messages is used.
         """
-        Helper method that sets the resulting values for status, error_id and message.
+        if not isinstance(action_message, ActionResult):
+            raise Exception()
```
abadger, line 228: This should raise a specific exception instead of a generic `Exception`.  `TypeError` is probably appropriate since we're expecting a specific type of parameter, an `ActionResult`.

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
         :type message: str | None
         """
-        if status != _NO_USER_VALUE:
-            self.status = STATUS_CODE[status]
+        if level not in ("ERROR", "OVERRIDABLE", "SKIP", "SUCCESS"):
+            raise KeyError("The level of result must be FAILURE, OVERRIDABLE, SKIP, or SUCCESS.")
+
+        self.result = ActionResult(level, id, message)
+
+    def set_message(self, level, id=None, message=None):
```
abadger, line 248: This shouldn't be named `set_` because it doesn't set the message attribute (`message = FOO`).

`add_` was used before which is similar to `set` but there could be other prefixes that work as well.

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
+        :param message: The message to be set.
+        :type message: str | None
+        """
+        msg = ActionMessage(level, id, message)
+        self.messages.append(msg)
+
+
+class ActionMessageBase:
+    """
+    Parent class for setting message and result variables and returning the
```
abadger, line 265:
```suggestion
    Parent class for setting message and result variables and returning
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
+    those values in a dictionary format
+    :param id: The id to identify the message.
```
abadger, line 267:
```suggestion
    those values in a dictionary format
    
    :param id: The id to identify the message.
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
+    :type id: str
+    :param level: The level to be set, defaulting to SUCCESS.
```
abadger, line 269:
```suggestion
    :keyword level: The level to be set, defaulting to SUCCESS.
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
+    :type: level: str
+    :param message: The message to be set.
```
abadger, line 271:
```suggestion
    :keyword message: The message to be set.
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
+    """
+
+    def __init__(self, level="SUCCESS", id=None, message=""):
+        self.id = id
+        self.level = STATUS_CODE[level]
+        self.message = message
 
-        if error_id != _NO_USER_VALUE:
-            self.error_id = error_id
+    def to_dict(self):
```
abadger, line 282:
```suggestion
    def to_dict(self):
        """
        Returns a dictionary representation of the :class:`ActionMessageBase`.
        """
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
+            "message": self.message,
+        }
 
-        if message != _NO_USER_VALUE:
-            self.message = message
+
+class ActionMessage(ActionMessageBase):
+    """
+    Class inherited from ActionMessageBase that defines the rules
+    for messages set through set_message
```
abadger, line 293:
```suggestion
    A class that defines the contents and rules for messages set through Action.set_message().
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +211,122 @@ def run(self):
...
+        if not (STATUS_CODE["SUCCESS"] < STATUS_CODE[level] < STATUS_CODE["SKIP"]):
+            raise InvalidMessageError("Invalid level '%s', set for a non-result message" % level)
+
+        super(ActionMessage, self).__init__(level, id, message)
+
+
+class ActionResult(ActionMessageBase):
+    """
+    Class inherited from ActionMessageBase that defines the rules
+    for messages set through set_result
```
abadger, line 309:
```suggestion
    A class that defines content and rules for messages set through :meth:`Action.set_result`.
```

convert2rhel/actions/__init__.py
```diff
@@ -540,17 +641,28 @@ def run_actions():
...
     formatted_results = {}
     for action in itertools.chain(*results):
-        formatted_results[action.id] = {"status": action.status, "error_id": action.error_id, "message": action.message}
+        msgs = [msg.to_dict() for msg in action.messages]
+        formatted_results[action.id] = {"messages": msgs, "result": action.result.to_dict()}
     return formatted_results
 
 
-def find_actions_of_severity(results, severity):
+def find_actions_of_severity(results, severity, key):
```
abadger, line 665: Add documentation for `key` to the docstring as well.

convert2rhel/actions/__init__.py
pr-watson: Noted

convert2rhel/actions/__init__.py
pr-watson: I will change this

convert2rhel/actions/__init__.py
pr-watson: add probably works best for now

convert2rhel/actions/__init__.py
```diff
@@ -118,6 +119,10 @@ class ActionError(Exception):
     """Raised for errors related to the Action framework."""
 
 
+class InvalidMessageError(ActionError):
+    """Raised for errors related to invalid passes to set_result or add_message"""
```
abadger, line 123:
```suggestion
    """Raised when invalid parameters are passed to set_result() or add_message()."""
```

convert2rhel/actions/__init__.py
```diff
@@ -206,25 +210,124 @@ def run(self):
...
+        """
+        if level not in ("ERROR", "OVERRIDABLE", "SKIP", "SUCCESS"):
+            raise KeyError("The level of result must be FAILURE, OVERRIDABLE, SKIP, or SUCCESS.")
+
+        self.result = ActionResult(level, id, message)
+
+    def add_message(self, level, id=None, message=None):
         """
-        Helper method that sets the resulting values for status, error_id and message.
+        Helper method that adds the resulting values for level, id and message of a warning or info log message.
```
abadger, line 249:
```suggestion
        Helper method that adds a new informational message to display to the user.
        The passed in values for level, id and message of a warning or info log message are
        saved in an :class:`ActionMessage` which is appended to ``self.messages``
```

convert2rhel/actions/__init__.py
abadger: Note for people looking at this in the future, this function appends a message to the list of messages to be returned to the user while set_result() sets the one and only result from running the Action.  Therefore we kept this as add_message() to differentiate it for now.

abadger (changes requested):
All the things i saw have been noted.  Nothing major but i do think they should be made before the PR is merged.  Note that some unit_tests will need to be updated with these changes and I haven't modified those.

convert2rhel/actions/report.py
```diff
@@ -43,6 +43,10 @@
 }
 
 
+def level_for_combined_action_data(message):
+    return message[1]["level"]
+
+
```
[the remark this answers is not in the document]
pr-watson: I'll make this change

abadger (approved):
All requested changes have been made.  From a code review standpoint, this looks good.  I noticed that all the integration tests had failed but I'm not sure yet if that's a problem with the PR or with the test infrastructure.

abadger:
At least some of the integration test failures look like they're unrelated problems.  https://github.com/oamg/convert2rhel/pull/841/checks?check_run_id=14845110458  is failing due to convert2rhel not honoring Ctrl-C:
> ^CException KeyboardInterrupt in <bound method Popen.__del__ of <subprocess.Popen object at 0x7fe9fd98edd0>> ignored
> 
> Skipping the execution of 'rpm -Va'.

Subsequent tests in that test plan are failing because the first test did not rollback so the environment is polluted.

PR #843, opened 2023-05-25
tests/hotfix/Download certificate for c2r repo

danmyway:
* curl failing, add the certs


kokesak (approved):
LGTM, just please re-run the failed tests. Seems like only some internal server error when subscribing the machine, but to make sure it's ok

danmyway:
This seems to be unnecessary at this point, closing.

PR #846, opened 2023-05-30
hotfix/tests/Remove --pool option from tests

danmyway:
* we are hitting issues with attaching the pool with the one specified with --pool option
* remove the option from tests to mitigate the issues


danmyway:
I don't think this is needed at this point.
Might be revisited in the future.

PR #850, opened 2023-06-06
[RHELC-975] Output the pre-assessment as a json file.

abadger:
For integration with insights and other tools, we need to output the assessment report as a json file.

Jira Issues: [RHELC-975](https://issues.redhat.com/browse/RHELC-975)

### Open questions:
- [x] What file should it write?
  - Rodolfo thinks it makes more sense alongside the breadcrumbs output.  I've moved it to `/etc/convert2rhel-assessment.json` for now.
- [x] Is action.report the right place for the function to live? 
  - Rodolfo agrees that this location makes sense. 
- [x] Is writing the file after the results are returned the right place or should it be done nearer to where the breadcrumbs are written?  (maybe in a finally: in main.py))
  - Rodolfo thinks that `finally:` in main.py makes more sense so I will make this change. 

Checklist

### DIFF
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -15,6 +15,7 @@
 
 __metaclass__ = type
 
+import copy
 import json
 import logging
 import textwrap
@@ -22,6 +23,7 @@ import textwrap
 from convert2rhel import utils
 from convert2rhel.actions import (
     _STATUS_HEADER,
+    _STATUS_NAME_FROM_CODE,
     find_actions_of_severity,
     format_action_status_message,
     level_for_combined_action_data,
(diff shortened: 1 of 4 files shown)


abadger:
\cc @r0x0d

r0x0d (approved):
Looks good to me!

convert2rhel/actions/report.py
```diff
@@ -26,6 +26,9 @@
 
 logger = logging.getLogger(__name__)
 
+#: The filename to store the results of running preassessment
+CONVERT2RHEL_JSON_RESULTS = "/etc/convert2rhel-assessment.json"
```
bocekm, line 30: I can see that for Leapp the txt report is being created at this path `/var/log/leapp/leapp-report.txt`. I suspect they generate the json to the same folder. They also generate the migration-results file to /etc/ but that serves quite a different purpose - it's in /etc/ to make it more likely to survive an attempt to purge the converted system from all-things-conversion. So I'd put our report under `/var/log/`.

Also, we use the word report instead of assessment everywhere, so I'd name it convert2rhel-report.json.

bocekm (changes requested):

convert2rhel/actions/report.py
bocekm: Ok, let's wait for the answer from the Leapp team what are their plans around a post-upgrade report.

Because we want to add the post-conversion report (https://issues.redhat.com/browse/RHELC-902), it might be beneficial to differentiate the pre- and post- ones. Unless we use just one file for both reports.

abadger:
For the file name, what about  `convert2rhel-pre-conversion-report.json` and `convert2rhel-post-conversion-report.json`?

abadger:
Rebased

PR #857, opened 2023-06-28
[RHELC-976] Add info and warning level messages to report

pr-watson:
This PR adds warning and info messages that need to be included in the pre-conversion analysis report. 

Jira Issues: [RHELC-976](https://issues.redhat.com/browse/RHELC-976)

Checklist

### DIFF
--- a/convert2rhel/actions/__init__.py
+++ b/convert2rhel/actions/__init__.py
@@ -83,6 +83,7 @@ _STATUS_NAME_FROM_CODE = dict((value, key) for key, value in STATUS_CODE.items()
 #: what the results mean
 _STATUS_HEADER = {
     0: "Success (No changes needed)",
+    25: "Info (No changes needed)",
     51: "Warning (Review and fix if needed)",
     101: "Skip (Could not be checked due to other failures)",
     152: "Overridable (Review and either fix or ignore the failure)",
--- a/convert2rhel/actions/__init__.py
+++ b/convert2rhel/actions/__init__.py
@@ -48,6 +48,7 @@ logger = logging.getLogger(__name__)
 #:      and start to use it with console.redhat.com
 #:
 #: :SUCCESS: no problem.
+#: :INFO: informational message to the user, no problem.
 #: :WARNING: the problem is just a warning displayed to the user. (unused,
 #:      warnings are currently emitted directly from the Action).
 #: :SKIP: the action could not be run because a dependent Action failed.
@@ -67,6 +68,7 @@ logger = logging.getLogger(__name__)
 #:        runs the Actions will set this.
 STATUS_CODE = {
     "SUCCESS": 0,
+    "INFO": 25,
     "WARNING": 51,
     "SKIP": 101,
     "OVERRIDABLE": 152,
@@ -109,10 +111,6 @@ def _action_defaults_to_success(func):
         return return_value
 
     return wrapper
-
-
-#: Used as a sentinel value for Action.set_result() method.
-_NO_USER_VALUE = object()
 
 
 class ActionError(Exception):
@@ -277,6 +275,22 @@ class ActionMessageBase:
         self.id = id
         self.level = STATUS_CODE[level]
         self.message = message
+
+    def __eq__(self, other):
+        if self.level == other.level and self.id == other.id and self.message == other.message:
+            return True
+        return False
+
+    def __hash__(self):
+        return hash((self.level, self.id, self.message))
+
+    def __repr__(self):
+        return "%s(level=%s, id=%s, message=%s)" % (
+            self.__class__.__name__,
+            _STATUS_NAME_FROM_CODE[self.level],
+            self.id,
+            self.message,
+        )
 
     def to_dict(self):
         """
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -41,7 +41,7 @@ class ListThirdPartyPackages(actions.Action):
             message = (
                 "Only packages signed by %s are to be"
                 " replaced. Red Hat support won't be provided"
-                " for the following third party packages:\n%s" % (system_info.name, ", ".join(pkg_list))
+                " for the following third party packages:\n%s" % (system_info.name, pkg_list)
             )
             logger.warning(message)
             self.add_message(
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -54,7 +54,7 @@ class ListThirdPartyPackages(actions.Action):
 
     def extract_packages(self, pkg):
         """Key function to extract the package name from third_party_pkgs"""
-        return pkg.pkg_obj.name
+        return pkg.nevra.name
 
 
 class RemoveExcludedPackages(actions.Action):
@@ -84,6 +84,9 @@ class RemoveExcludedPackages(actions.Action):
 
         # shows which packages were not removed, if false, all packages were removed
         pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+        pkgs_not_removed = [str(pkg) for pkg in pkgs_not_removed]
+        pkgs_removed = [str(pkg) for pkg in pkgs_removed]
+
         if pkgs_not_removed:
             message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
             logger.warning(message)
@@ -140,6 +143,10 @@ class RemoveRepositoryFilesPackages(actions.Action):
 
         # shows which packages were not removed, if false, all packages were removed
         pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+
+        pkgs_not_removed = [str(pkg) for pkg in pkgs_not_removed]
+        pkgs_removed = [str(pkg) for pkg in pkgs_removed]
+
         if pkgs_not_removed:
             message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
             logger.warning(message)
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -53,6 +53,7 @@ class ListThirdPartyPackages(actions.Action):
             logger.info("No third party packages installed.")
 
     def extract_packages(self, pkg):
+        """Key function to extract the package name from third_party_pkgs"""
         return pkg.pkg_obj.name
 
 
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -84,18 +84,20 @@ class RemoveExcludedPackages(actions.Action):
         # shows which packages were not removed, if false, all packages were removed
         pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
         if pkgs_not_removed:
-            logger.warning("The following packages were not removed: %s" % ", ".join(pkgs_not_removed))
+            message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
+            logger.warning(message)
             self.add_message(
                 level="WARNING",
                 id="EXCLUDED_PACKAGES_NOT_REMOVED",
-                message="The following packages were not removed: %s" % ", ".join(pkgs_not_removed),
+                message=message,
             )
         else:
-            logger.info("The following packages were removed: %s" % ", ".join(pkgs_removed))
+            message = "The following packages were removed: %s" % ", ".join(pkgs_removed)
+            logger.info(message)
             self.add_message(
                 level="INFO",
                 id="EXCLUDED_PACKAGES_REMOVED",
-                message="The following packages were removed: %s" % ", ".join(pkgs_removed),
+                message=message,
             )
 
 
@@ -138,16 +140,18 @@ class RemoveRepositoryFilesPackages(actions.Action):
         # shows which packages were not removed, if false, all packages were removed
         pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
         if pkgs_not_removed:
-            logger.warning("The following packages were not removed: %s" % ", ".join(pkgs_not_removed))
+            message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
+            logger.warning(message)
             self.add_message(
                 level="WARNING",
                 id="REPOSITORY_FILE_PACKAGES_NOT_REMOVED",
-                message="The following packages were not removed: %s" % ", ".join(pkgs_not_removed),
+                message=message,
             )
         else:
-            logger.info("The following packages were removed: %s" % ", ".join(pkgs_removed))
+            message = "The following packages were removed: %s" % ", ".join(pkgs_removed)
+            logger.info(message)
             self.add_message(
                 level="INFO",
                 id="REPOSITORY_FILE_PACKAGES_REMOVED",
-                message="The following packages were removed: %s" % ", ".join(pkgs_removed),
+                message=message,
             )
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -95,7 +95,8 @@ class RemoveExcludedPackages(actions.Action):
                 id="EXCLUDED_PACKAGES_NOT_REMOVED",
                 message=message,
             )
-        else:
+            return
+        if pkgs_removed:
             message = "The following packages were removed: %s" % ", ".join(pkgs_removed)
             logger.info(message)
             self.add_message(
@@ -103,6 +104,7 @@ class RemoveExcludedPackages(actions.Action):
                 id="EXCLUDED_PACKAGES_REMOVED",
                 message=message,
             )
+            return
 
 
 class RemoveRepositoryFilesPackages(actions.Action):
@@ -155,7 +157,8 @@ class RemoveRepositoryFilesPackages(actions.Action):
                 id="REPOSITORY_FILE_PACKAGES_NOT_REMOVED",
                 message=message,
             )
-        else:
+            return
+        if pkgs_removed:
             message = "The following packages were removed: %s" % ", ".join(pkgs_removed)
             logger.info(message)
             self.add_message(
@@ -163,3 +166,4 @@ class RemoveRepositoryFilesPackages(actions.Action):
                 id="REPOSITORY_FILE_PACKAGES_REMOVED",
                 message=message,
             )
+            return
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -37,14 +37,23 @@ class ListThirdPartyPackages(actions.Action):
         logger.task("Convert: List third-party packages")
         third_party_pkgs = pkghandler.get_third_party_pkgs()
         if third_party_pkgs:
-            logger.warning(
+            pkg_list = pkghandler.format_pkg_info(sorted(third_party_pkgs, key=self.extract_packages))
+            message = (
                 "Only packages signed by %s are to be"
                 " replaced. Red Hat support won't be provided"
-                " for the following third party packages:\n" % system_info.name
+                " for the following third party packages:\n%s" % (system_info.name, ", ".join(pkg_list))
             )
-            pkghandler.print_pkg_info(third_party_pkgs)
+            logger.warning(message)
+            self.add_message(
+                level="WARNING",
+                id="THIRD_PARTY_PACKAGE_DETECTED",
+                message=message,
+            )
         else:
             logger.info("No third party packages installed.")
+
+    def extract_packages(self, pkg):
+        return pkg.pkg_obj.name
 
 
 class RemoveExcludedPackages(actions.Action):
@@ -60,8 +69,8 @@ class RemoveExcludedPackages(actions.Action):
         logger.task("Convert: Remove excluded packages")
         logger.info("Searching for the following excluded packages:\n")
         try:
-            pkgs_to_remove = pkghandler._get_packages_to_remove(system_info.excluded_pkgs)
-            pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove)
+            pkgs_to_remove = sorted(pkghandler._get_packages_to_remove(system_info.excluded_pkgs))
+            pkgs_removed = sorted(pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove))
 
             # TODO: Handling SystemExit here as way to speedup exception
             # handling and not refactor contents of the underlying function.
@@ -69,7 +78,25 @@ class RemoveExcludedPackages(actions.Action):
             # TODO(r0x0d): Places where we raise SystemExit and need to be
             # changed to something more specific.
             #   - When we can't remove a package.
-            self.set_result(level="ERROR", id="PACKAGE_REMOVAL_FAILED", message=str(e))
+            self.set_result(level="ERROR", id="EXCLUDED_PACKAGE_REMOVAL_FAILED", message=str(e))
+            return
+
+        # shows which packages were not removed, if false, all packages were removed
+        pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+        if pkgs_not_removed:
+            logger.warning("The following packages were not removed: %s" % ", ".join(pkgs_not_removed))
+            self.add_message(
+                level="WARNING",
+                id="EXCLUDED_PACKAGES_NOT_REMOVED",
+                message="The following packages were not removed: %s" % ", ".join(pkgs_not_removed),
+            )
+        else:
+            logger.info("The following packages were removed: %s" % ", ".join(pkgs_removed))
+            self.add_message(
+                level="INFO",
+                id="EXCLUDED_PACKAGES_REMOVED",
+                message="The following packages were removed: %s" % ", ".join(pkgs_removed),
+            )
 
 
 class RemoveRepositoryFilesPackages(actions.Action):
@@ -96,8 +123,8 @@ class RemoveRepositoryFilesPackages(actions.Action):
         logger.task("Convert: Remove packages containing .repo files")
         logger.info("Searching for packages containing .repo files or affecting variables in the .repo files:\n")
         try:
-            pkgs_to_remove = pkghandler._get_packages_to_remove(system_info.repofile_pkgs)
-            pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove)
+            pkgs_to_remove = sorted(pkghandler._get_packages_to_remove(system_info.repofile_pkgs))
+            pkgs_removed = sorted(pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove))
 
             # TODO: Handling SystemExit here as way to speedup exception
             # handling and not refactor contents of the underlying function.
@@ -105,4 +132,22 @@ class RemoveRepositoryFilesPackages(actions.Action):
             # TODO(r0x0d): Places where we raise SystemExit and need to be
             # changed to something more specific.
             #   - When we can't remove a package.
-            self.set_result(level="ERROR", id="PACKAGE_REMOVAL_FAILED", message=str(e))
+            self.set_result(level="ERROR", id="REPOSITORY_FILE_PACKAGE_REMOVAL_FAILED", message=str(e))
+            return
+
+        # shows which packages were not removed, if false, all packages were removed
+        pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+        if pkgs_not_removed:
+            logger.warning("The following packages were not removed: %s" % ", ".join(pkgs_not_removed))
+            self.add_message(
+                level="WARNING",
+                id="REPOSITORY_FILE_PACKAGES_NOT_REMOVED",
+                message="The following packages were not removed: %s" % ", ".join(pkgs_not_removed),
+            )
+        else:
+            logger.info("The following packages were removed: %s" % ", ".join(pkgs_removed))
+            self.add_message(
+                level="INFO",
+                id="REPOSITORY_FILE_PACKAGES_REMOVED",
+                message="The following packages were removed: %s" % ", ".join(pkgs_removed),
+            )
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -24,6 +24,10 @@ from convert2rhel.systeminfo import system_info
 logger = logging.getLogger(__name__)
 
 
+def _get_pkg_names(packages):
+    return [pkghandler.get_pkg_nevra(pkg_obj, include_zero_epoch=True) for pkg_obj in packages]
+
+
 class ListThirdPartyPackages(actions.Action):
     id = "LIST_THIRD_PARTY_PACKAGES"
 
@@ -38,17 +42,23 @@ class ListThirdPartyPackages(actions.Action):
         third_party_pkgs = pkghandler.get_third_party_pkgs()
         if third_party_pkgs:
             pkg_list = pkghandler.format_pkg_info(sorted(third_party_pkgs, key=self.extract_packages))
-            message = (
+            warning_message = (
                 "Only packages signed by %s are to be"
                 " replaced. Red Hat support won't be provided"
-                " for the following third party packages:\n%s" % (system_info.name, pkg_list)
+                " for the following third party packages:\n" % system_info.name
             )
-            logger.warning(message)
+            logger.warning(warning_message)
+            logger.info(pkg_list)
+            self.set_result(
+                level="SUCCESS",
+                id="THIRD_PARTY_PACKAGE_DETECTED",
+            )
             self.add_message(
                 level="WARNING",
-                id="THIRD_PARTY_PACKAGE_DETECTED",
-                message=message,
+                id="THIRD_PARTY_PACKAGE_DETECTED_MESSAGE",
+                message=warning_message + ", ".join(_get_pkg_names(third_party_pkgs)),
             )
+            return
         else:
             logger.info("No third party packages installed.")
 
@@ -84,11 +94,10 @@ class RemoveExcludedPackages(actions.Action):
 
         # shows which packages were not removed, if false, all packages were removed
         pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
-        pkgs_not_removed = [str(pkg) for pkg in pkgs_not_removed]
-        pkgs_removed = [str(pkg) for pkg in pkgs_removed]
 
         if pkgs_not_removed:
-            message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
+            pkg_names = _get_pkg_names(pkgs_not_removed)
+            message = "The following packages were not removed: %s" % ", ".join(pkg_names)
             logger.warning(message)
             self.add_message(
                 level="WARNING",
@@ -146,11 +155,9 @@ class RemoveRepositoryFilesPackages(actions.Action):
         # shows which packages were not removed, if false, all packages were removed
         pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
 
-        pkgs_not_removed = [str(pkg) for pkg in pkgs_not_removed]
-        pkgs_removed = [str(pkg) for pkg in pkgs_removed]
-
         if pkgs_not_removed:
-            message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
+            pkg_names = _get_pkg_names(pkgs_not_removed)
+            message = "The following packages were not removed: %s" % ", ".join(pkg_names)
             logger.warning(message)
             self.add_message(
                 level="WARNING",
--- a/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
+++ b/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
@@ -215,6 +215,15 @@ class EnsureKernelModulesCompatibility(actions.Action):
                     " We will continue the conversion with the following kernel modules unavailable in RHEL:\n"
                     "{kmods}\n".format(kmods="\n".join(unsupported_kmods))
                 )
+                self.add_message(
+                    level="WARNING",
+                    id="ALLOW_UNAVAILABLE_KERNEL_MODULES",
+                    message=(
+                        "Detected 'CONVERT2RHEL_ALLOW_UNAVAILABLE_KMODS' environment variable."
+                        " We will continue the conversion with the following kernel modules unavailable in RHEL:\n"
+                        "{kmods}\n".format(kmods="\n".join(unsupported_kmods))
+                    ),
+                )
                 return
 
             # If there is any unsupported kmods found, set the result to error
--- a/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
+++ b/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
@@ -215,14 +215,16 @@ class EnsureKernelModulesCompatibility(actions.Action):
                     " We will continue the conversion with the following kernel modules unavailable in RHEL:\n"
                     "{kmods}\n".format(kmods="\n".join(unsupported_kmods))
                 )
-                self.add_message(
-                    level="WARNING",
-                    id="ALLOW_UNAVAILABLE_KERNEL_MODULES",
-                    message=(
+                message = (
                         "Detected 'CONVERT2RHEL_ALLOW_UNAVAILABLE_KMODS' environment variable."
                         " We will continue the conversion with the following kernel modules unavailable in RHEL:\n"
                         "{kmods}\n".format(kmods="\n".join(unsupported_kmods))
-                    ),
+                    )
+                logger.warning(message)
+                self.add_message(
+                    level="WARNING",
+                    id="ALLOW_UNAVAILABLE_KERNEL_MODULES",
+                    message=message,
                 )
                 return
 
--- a/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
+++ b/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
@@ -210,16 +210,11 @@ class EnsureKernelModulesCompatibility(actions.Action):
             # Check if we have the environment variable set, if we do, send a
             # warning and return.
             if "CONVERT2RHEL_ALLOW_UNAVAILABLE_KMODS" in os.environ:
-                logger.warning(
+                message = (
                     "Detected 'CONVERT2RHEL_ALLOW_UNAVAILABLE_KMODS' environment variable."
                     " We will continue the conversion with the following kernel modules unavailable in RHEL:\n"
                     "{kmods}\n".format(kmods="\n".join(unsupported_kmods))
                 )
-                message = (
-                        "Detected 'CONVERT2RHEL_ALLOW_UNAVAILABLE_KMODS' environment variable."
-                        " We will continue the conversion with the following kernel modules unavailable in RHEL:\n"
-                        "{kmods}\n".format(kmods="\n".join(unsupported_kmods))
-                    )
                 logger.warning(message)
                 self.add_message(
                     level="WARNING",
--- a/convert2rhel/actions/pre_ponr_changes/subscription.py
+++ b/convert2rhel/actions/pre_ponr_changes/subscription.py
@@ -31,11 +31,12 @@ class PreSubscription(actions.Action):
         super(PreSubscription, self).run()
 
         if toolopts.tool_opts.no_rhsm:
-            logger.warning("Detected --no-rhsm option. Skipping.")
+            message = "Detected --no-rhsm option. Skipping."
+            logger.warning(message)
             self.add_message(
                 level="WARNING",
                 id="PRE_SUBSCRIPTION_CHECK_SKIP",
-                message="Detected --no-rhsm option. Skipping.",
+                message=message,
             )
             return
 
@@ -93,11 +94,12 @@ class SubscribeSystem(actions.Action):
         super(SubscribeSystem, self).run()
 
         if toolopts.tool_opts.no_rhsm:
-            logger.warning("Detected --no-rhsm option. Skipping.")
+            message = "Detected --no-rhsm option. Skipping."
+            logger.warning(message)
             self.add_message(
                 level="WARNING",
                 id="PRE_SUBSCRIPTION_CHECK_SKIP",
-                message="Detected --no-rhsm option. Skipping.",
+                message=message,
             )
             return
 
--- a/convert2rhel/actions/pre_ponr_changes/subscription.py
+++ b/convert2rhel/actions/pre_ponr_changes/subscription.py
@@ -32,6 +32,11 @@ class PreSubscription(actions.Action):
 
         if toolopts.tool_opts.no_rhsm:
             logger.warning("Detected --no-rhsm option. Skipping.")
+            self.add_message(
+                level="WARNING",
+                id="PRE_SUBSCRIPTION_CHECK_SKIP",
+                message="Detected --no-rhsm option. Skipping.",
+            )
             return
 
         try:
@@ -89,6 +94,11 @@ class SubscribeSystem(actions.Action):
 
         if toolopts.tool_opts.no_rhsm:
             logger.warning("Detected --no-rhsm option. Skipping.")
+            self.add_message(
+                level="WARNING",
+                id="PRE_SUBSCRIPTION_CHECK_SKIP",
+                message="Detected --no-rhsm option. Skipping.",
+            )
             return
 
         try:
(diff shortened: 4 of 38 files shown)


pr-watson:
Unit tests for the messages added still need to be put in place, waiting on https://github.com/oamg/convert2rhel/pull/841 to be merged to properly test the messages

convert2rhel/actions/__init__.py
```diff
@@ -226,6 +226,27 @@ def set_result(self, status=_NO_USER_VALUE, error_id=_NO_USER_VALUE, message=_NO
...
+        """
+        # if level != _NO_USER_VALUE:
+        #     self.level = level
+        #
+        # if id != _NO_USER_VALUE:
+        #     self.id = id
+        #
+        # if message != _NO_USER_VALUE:
+        #     self.message = message
+        pass
```
r0x0d, line 248: Why is this commented out? It will not be used?

convert2rhel/actions/pre_ponr_changes/subscription.py
```diff
@@ -135,3 +143,11 @@ def run(self):
...
                 message="One or more combinations were missing for subscription-manager parameters: %s" % e,
             )
+
+        if non_available_repos:
+            self.add_message(
+                level="WARNING",
+                id="NON_AVAILABLE_REPOSITORIES_DETECTED",
+                message="%s repositories are not available - some packages"
+                " may not be replaced and thus not supported." % ", ".join(non_available_repos),
+            )
```
r0x0d, line 153: I think that to be on the safe side, we should move this condition to be inside the try-except, below line 110.

Since the other except statements don't have any return statement, my fear is that after setting the result in one of the exceptions, we will fall to this if statement and the variable might not exist at that point.

convert2rhel/actions/__init__.py
pr-watson: Commented out because that method is just a placeholder until the actual add_message code is introduced from https://github.com/oamg/convert2rhel/pull/841

convert2rhel/actions/pre_ponr_changes/subscription.py
pr-watson: Ok I will move it

convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
```diff
@@ -129,6 +137,16 @@ def run(self):  # pylint: disable= too-many-return-statements
...
 
+        self.add_message(
+            level="WARNING",
+            id="UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK_DETECTED",
+            message=(
+                "Detected 'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' environment variable, we will skip "
+                "the %s comparison.\n"
+                "Beware, this could leave your system in a broken state. " % package_to_check
+            ),
+        )
```
r0x0d, line 148: I think this is in the wrong part of the code, it should be on line 82

convert2rhel/actions/system_checks/package_updates.py
```diff
@@ -54,12 +62,15 @@ def run(self):
...
-                "There was an error while checking whether the installed packages are up-to-date. Having an updated system is"
-                " an important prerequisite for a successful conversion. Consider verifyng the system is up to date manually"
-                " before proceeding with the conversion."
+            self.add_message(
+                level="WARNING",
+                id="PACKAGE_UP_TO_DATE_CHECK_FAIL",
+                message=(
+                    "There was an error while checking whether the installed packages are up-to-date. Having an updated system is"
+                    " an important prerequisite for a successful conversion. Consider verifyng the system is up to date manually"
+                    " before proceeding with the conversion. %s" % e
```
r0x0d, line 71: Probably we are missing the `str(e)` here?

convert2rhel/subscription.py
```diff
@@ -878,9 +879,15 @@ def check_needed_repos_availability(repo_ids_needed):
...
             )
+            non_avail_repos.append(repo_id)
             all_repos_avail = False
+
     if all_repos_avail:
         loggerinst.info("Needed RHEL repositories are available.")
+    if non_avail_repos:
+        return non_avail_repos
+    else:
+        return None
```
r0x0d, line 890: Maybe we could just simplify this and return `non_avail_repos`

```suggestion
    return non_avail_repos
```

In whatever function that calls this, the checking condition would still be the same, like `if not ....`

convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
pr-watson: I will move this

convert2rhel/actions/system_checks/package_updates.py
pr-watson: I'll add that

convert2rhel/subscription.py
pr-watson: Noted i'll add this

abadger (changes requested):
Some requested changes but let's ask the rest of the team whether they think we should keep the logger.warnings too.  From michal's remarks on past PRs, I think so but it is good to confirm.

.github/workflows/coverage.yml
```diff
@@ -2,9 +2,9 @@ name: test_coverage
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, pre-check ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, pre-check ]
```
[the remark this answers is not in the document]
pr-watson: Noted

abadger (approved):
Some things that could be changed for slightly better output and better readabillity down the road.  But it's also okay as it is.  You have my approval when you are happy with this.

convert2rhel/actions/pre_ponr_changes/subscription.py
```diff
@@ -99,7 +109,15 @@ def run(self):
             rhel_repoids = repo.get_rhel_repoids()
 
             logger.task("Convert: Subscription Manager - Check required repositories")
-            subscription.check_needed_repos_availability(rhel_repoids)
+            non_available_repos = subscription.check_needed_repos_availability(rhel_repoids)
+            if non_available_repos:
+
+                self.add_message(
```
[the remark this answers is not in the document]
pr-watson: This check will be removed in https://github.com/oamg/convert2rhel/pull/878 so we can ignore this log message

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -96,13 +118,29 @@ def run(self):
...
+        # shows which packages were not removed, if false, all packages were removed
+        pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+        if pkgs_not_removed:
+            self.add_message(
+                level="WARNING",
+                id="REPOSITORY_FILE_PACKAGES_NOT_REMOVED",
+                message="The following packages were not removed: %s" % ", ".join(pkgs_not_removed),
+            )
+        else:
+            self.add_message(
```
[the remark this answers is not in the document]
pr-watson: Yes I will add this

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -60,16 +66,32 @@ def run(self):
...
+        # shows which packages were not removed, if false, all packages were removed
+        pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+        if pkgs_not_removed:
+            self.add_message(
+                level="WARNING",
+                id="EXCLUDED_PACKAGES_NOT_REMOVED",
+                message="The following packages were not removed: %s" % ", ".join(pkgs_not_removed),
+            )
+        else:
+            self.add_message(
```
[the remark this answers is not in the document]
pr-watson: Yes I will add this

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -37,15 +37,26 @@ def run(self):
         logger.task("Convert: List third-party packages")
         third_party_pkgs = pkghandler.get_third_party_pkgs()
         if third_party_pkgs:
-            logger.warning(
+            pkg_list = pkghandler.format_pkg_info(sorted(third_party_pkgs, key=self.extract_packages(third_party_pkgs)))
```
SpyTec, line 40: Unit tests are failing here
```
     def extract_packages(self, pkg):
    
>       return pkg.pkg_obj.name
E       AttributeError: 'list' object has no attribute 'pkg_obj'
```

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -37,15 +37,24 @@ def run(self):
...
+            self.add_message(
+                level="WARNING",
+                id="THIRD_PARTY_PACKAGE_DETECTED",
+                message=message,
             )
-            pkghandler.print_pkg_info(third_party_pkgs)
         else:
             logger.info("No third party packages installed.")
 
+    def extract_packages(self, pkg):
```
r0x0d, line 55: docstring here

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -60,16 +69,34 @@ def run(self):
...
+
+        # shows which packages were not removed, if false, all packages were removed
+        pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+        if pkgs_not_removed:
+            logger.warning("The following packages were not removed: %s" % ", ".join(pkgs_not_removed))
+            self.add_message(
+                level="WARNING",
+                id="EXCLUDED_PACKAGES_NOT_REMOVED",
+                message="The following packages were not removed: %s" % ", ".join(pkgs_not_removed),
+            )
```
r0x0d, line 92:
```suggestion
            message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="EXCLUDED_PACKAGES_NOT_REMOVED",
                message=message,
            )
```

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -60,16 +69,34 @@ def run(self):
...
+        else:
+            logger.info("The following packages were removed: %s" % ", ".join(pkgs_removed))
+            self.add_message(
+                level="INFO",
+                id="EXCLUDED_PACKAGES_REMOVED",
+                message="The following packages were removed: %s" % ", ".join(pkgs_removed),
+            )
```
r0x0d, line 99:
```suggestion
            message = "The following packages were removed: %s" % ", ".join(pkgs_removed)
            logger.info(message)
            self.add_message(
                level="INFO",
                id="EXCLUDED_PACKAGES_REMOVED",
                message=message,
            )
```

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -96,13 +123,31 @@ def run(self):
...
+                id="REPOSITORY_FILE_PACKAGES_NOT_REMOVED",
+                message="The following packages were not removed: %s" % ", ".join(pkgs_not_removed),
+            )
+        else:
+            logger.info("The following packages were removed: %s" % ", ".join(pkgs_removed))
+            self.add_message(
+                level="INFO",
+                id="REPOSITORY_FILE_PACKAGES_REMOVED",
+                message="The following packages were removed: %s" % ", ".join(pkgs_removed),
+            )
```
r0x0d, line 153:
```suggestion
            message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="REPOSITORY_FILE_PACKAGES_NOT_REMOVED",
                message=message,
            )
        else:
            message = "The following packages were removed: %s" % ", ".join(pkgs_removed)
            logger.info(message)
            self.add_message(
                level="INFO",
                id="REPOSITORY_FILE_PACKAGES_REMOVED",
                message=message,
            )
```

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
```diff
@@ -215,6 +215,15 @@ def run(self):
...
                 )
+                self.add_message(
+                    level="WARNING",
+                    id="ALLOW_UNAVAILABLE_KERNEL_MODULES",
+                    message=(
+                        "Detected 'CONVERT2RHEL_ALLOW_UNAVAILABLE_KMODS' environment variable."
+                        " We will continue the conversion with the following kernel modules unavailable in RHEL:\n"
+                        "{kmods}\n".format(kmods="\n".join(unsupported_kmods))
+                    ),
+                )
```
r0x0d, line 226: Can't get the lines above this (logger.warning), but will leave the suggestion anyway:

```suggestion
                message = (
                        "Detected 'CONVERT2RHEL_ALLOW_UNAVAILABLE_KMODS' environment variable."
                        " We will continue the conversion with the following kernel modules unavailable in RHEL:\n"
                        "{kmods}\n".format(kmods="\n".join(unsupported_kmods))
                    )
                logger.warning(message)
                self.add_message(
                    level="WARNING",
                    id="ALLOW_UNAVAILABLE_KERNEL_MODULES",
                    message=message,
                )
```

convert2rhel/actions/pre_ponr_changes/subscription.py
```diff
@@ -32,6 +32,11 @@ def run(self):
 
         if toolopts.tool_opts.no_rhsm:
             logger.warning("Detected --no-rhsm option. Skipping.")
+            self.add_message(
+                level="WARNING",
+                id="PRE_SUBSCRIPTION_CHECK_SKIP",
+                message="Detected --no-rhsm option. Skipping.",
+            )
```
r0x0d, line 39:
```suggestion
            message = "Detected --no-rhsm option. Skipping."
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="PRE_SUBSCRIPTION_CHECK_SKIP",
                message=message,
            )
```

convert2rhel/actions/pre_ponr_changes/subscription.py
```diff
@@ -89,6 +94,11 @@ def run(self):
...
```
r0x0d, line 101:
```suggestion
            message = "Detected --no-rhsm option. Skipping."
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="PRE_SUBSCRIPTION_CHECK_SKIP",
                message=message,
            )
```

convert2rhel/actions/system_checks/convert2rhel_latest.py
```diff
@@ -60,6 +60,11 @@ def run(self):
 
         if not system_info.has_internet_access:
             logger.warning("Skipping the check because no internet connection has been detected.")
+            self.add_message(
+                level="WARNING",
+                id="CONVERT2RHEL_LATEST_CHECK_SKIP_NO_INTERNET",
+                message="Skipping the check because no internet connection has been detected.",
+            )
```
r0x0d, line 67:
```suggestion
            message = "Skipping the check because no internet connection has been detected."
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="CONVERT2RHEL_LATEST_CHECK_SKIP_NO_INTERNET",
                message=message,
            )
```

convert2rhel/actions/system_checks/convert2rhel_latest.py
```diff
@@ -91,6 +96,14 @@ def run(self):
...
                 "repoquery failed with the following output:\n%s" % (raw_output_convert2rhel_versions)
             )
+            self.add_message(
+                level="WARNING",
+                id="CONVERT2RHEL_LATEST_CHECK_SKIP",
+                message=(
+                    "Couldn't check if the current installed Convert2RHEL is the latest version.\n"
+                    "repoquery failed with the following output:\n%s" % (raw_output_convert2rhel_versions)
+                ),
+            )
```
r0x0d, line 106:
```suggestion
            message = (
                "Couldn't check if the current installed Convert2RHEL is the latest version.\n"
                "repoquery failed with the following output:\n%s" % (raw_output_convert2rhel_versions)
            )
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="CONVERT2RHEL_LATEST_CHECK_SKIP",
                message=message,
            )
```

r0x0d: Remember to remove the logger.warning above

convert2rhel/actions/system_checks/convert2rhel_latest.py
```diff
@@ -164,12 +177,29 @@ def run(self):
...
-
+                    self.add_message(
+                        level="WARNING",
+                        id="DEPRECATED_ENVIRONMENT_VARIABLE",
+                        message=(
+                            "You are using the deprecated 'CONVERT2RHEL_UNSUPPORTED_VERSION'"
+                            " environment variable.  Please switch to 'CONVERT2RHEL_ALLOW_OLDER_VERSION'"
+                            " instead."
+                        ),
+                    )
```
r0x0d, line 188:
```suggestion
                    message = (
                            "You are using the deprecated 'CONVERT2RHEL_UNSUPPORTED_VERSION'"
                            " environment variable.  Please switch to 'CONVERT2RHEL_ALLOW_OLDER_VERSION'"
                            " instead."
                        )
                    logger.warning(message)
                    self.add_message(
                        level="WARNING",
                        id="DEPRECATED_ENVIRONMENT_VARIABLE",
                        message=message,
                    )
```

Remove the logger.warning above

convert2rhel/actions/system_checks/convert2rhel_latest.py
```diff
@@ -178,6 +208,15 @@ def run(self):
...
                     )
+                    self.add_message(
+                        level="WARNING",
+                        id="OUTDATED_CONVERT2RHEL_VERSION",
+                        message=(
+                            "You are currently running %s and the latest version of Convert2RHEL is %s.\n"
+                            "We encourage you to update to the latest version."
+                            % (installed_convert2rhel_version, latest_available_version[1])
+                        ),
+                    )
```
r0x0d, line 219:
```suggestion
                    message = (
                            "You are currently running %s and the latest version of Convert2RHEL is %s.\n"
                            "We encourage you to update to the latest version."
                            % (installed_convert2rhel_version, latest_available_version[1])
                        )
                    logger.warning(message)
                    self.add_message(
                        level="WARNING",
                        id="OUTDATED_CONVERT2RHEL_VERSION",
                        message=message,
                    )
```

convert2rhel/actions/system_checks/custom_repos_are_valid.py
```diff
@@ -39,6 +39,12 @@ def run(self):
 
         if not tool_opts.no_rhsm:
             logger.info("Skipping the check of repositories due to the use of RHSM for the conversion.")
+
+            self.add_message(
+                level="INFO",
+                id="CUSTOM_REPOSITORIES_ARE_VALID_CHECK_SKIP",
+                message="Skipping the check of repositories due to the use of RHSM for the conversion.",
+            )
```
r0x0d, line 47:
```suggestion
            message = "Skipping the check of repositories due to the use of RHSM for the conversion."
            logger.info(message)
            self.add_message(
                level="INFO",
                id="CUSTOM_REPOSITORIES_ARE_VALID_CHECK_SKIP",
                message=message,
            )
```

convert2rhel/actions/system_checks/dbus.py
```diff
@@ -35,6 +35,11 @@ def run(self):
 
         if tool_opts.no_rhsm:
             logger.info("Skipping the check because we have been asked not to subscribe this system to RHSM.")
+            self.add_message(
+                level="WARNING",
+                id="DBUS_IS_RUNNING_CHECK_SKIP",
+                message="Skipping the check because we have been asked not to subscribe this system to RHSM.",
+            )
```
r0x0d, line 42:
```suggestion
            message = "Skipping the check because we have been asked not to subscribe this system to RHSM."
            logger.info()
            self.add_message(
                level="WARNING",
                id="DBUS_IS_RUNNING_CHECK_SKIP",
                message=message,
            )
```

convert2rhel/actions/system_checks/efi.py
```diff
@@ -78,6 +78,14 @@ def run(self):
...
                 " file located on local EFI System Partition (ESP)." % efiboot_info.current_bootnum
             )
+            self.add_message(
+                level="WARNING",
+                id="UEFI_BOOTLOADER_MISMATCH",
+                message=(
+                    "The current UEFI bootloader '%s' is not referring to any binary UEFI"
+                    " file located on local EFI System Partition (ESP)." % efiboot_info.current_bootnum
+                ),
+            )
```
r0x0d, line 88: Remember to remove the logger.warning above
```suggestion
            message = (
                    "The current UEFI bootloader '%s' is not referring to any binary UEFI"
                    " file located on local EFI System Partition (ESP)." % efiboot_info.current_bootnum
                )
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="UEFI_BOOTLOADER_MISMATCH",
                message=message,
            )
```

convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
```diff
@@ -55,6 +55,11 @@ def run(self):  # pylint: disable= too-many-return-statements
         reposdir = get_hardcoded_repofiles_dir()
         if reposdir and not system_info.has_internet_access:
             logger.warning("Skipping the check as no internet connection has been detected.")
+            self.add_message(
+                level="WARNING",
+                id="IS_LOADED_KERNEL_LATEST_CHECK_SKIP",
+                message=("Skipping the check as no internet connection has been detected."),
+            )
```
r0x0d, line 62:
```suggestion
            message = "Skipping the check as no internet connection has been detected."
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="IS_LOADED_KERNEL_LATEST_CHECK_SKIP",
                message=message,
            )
```

convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
```diff
@@ -80,6 +85,15 @@ def run(self):  # pylint: disable= too-many-return-statements
...
             )
+            self.add_message(
+                level="WARNING",
+                id="UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK_DETECTED",
+                message=(
+                    "Detected 'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' environment variable, we will skip "
+                    "the %s comparison.\n"
+                    "Beware, this could leave your system in a broken state." % package_to_check
+                ),
+            )
```
r0x0d, line 96: Remember to remove the logger.warning above
```suggestion
            message = (
                    "Detected 'CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK' environment variable, we will skip "
                    "the %s comparison.\n"
                    "Beware, this could leave your system in a broken state." % package_to_check
                )
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK_DETECTED",
                message=message,
            )
```

convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
```diff
@@ -96,6 +110,14 @@ def run(self):  # pylint: disable= too-many-return-statements
...
                 "the repositories. Skipping the loaded kernel check."
             )
+            self.add_message(
+                level="WARNING",
+                id="UNABLE_TO_FETCH_RECENT_KERNELS",
+                message=(
+                    "Couldn't fetch the list of the most recent kernels available in "
+                    "the repositories. Skipping the loaded kernel check."
+                ),
+            )
```
r0x0d, line 120: Remember to remove the logger.warning above

```suggestion
            message = (
                    "Couldn't fetch the list of the most recent kernels available in "
                    "the repositories. Skipping the loaded kernel check."
                )
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="UNABLE_TO_FETCH_RECENT_KERNELS",
                message=message,
            )
```

convert2rhel/actions/system_checks/package_updates.py
```diff
@@ -39,12 +39,25 @@ def run(self):
...
 
         reposdir = get_hardcoded_repofiles_dir()
 
         if reposdir and not system_info.has_internet_access:
             logger.warning("Skipping the check as no internet connection has been detected.")
+            self.add_message(
+                level="WARNING",
+                id="PACKAGE_UPDATES_CHECK_SKIP_NO_INTERNET",
+                message="Skipping the check as no internet connection has been detected.",
+            )
```
r0x0d, line 60:
```suggestion
            message = "Skipping the check as no internet connection has been detected."
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="PACKAGE_UPDATES_CHECK_SKIP_NO_INTERNET",
                message=message,
            )
```

convert2rhel/actions/system_checks/package_updates.py
```diff
@@ -39,12 +39,25 @@ def run(self):
...
                 % (system_info.name, system_info.version.major, system_info.version.minor)
             )
+            self.add_message(
+                level="INFO",
+                id="PACKAGE_UPDATES_CHECK_SKIP_NO_PUBLIC_REPOSITORIES",
+                message=(
+                    "Skipping the check because there are no publicly available %s %d.%d repositories available."
+                    % (system_info.name, system_info.version.major, system_info.version.minor)
+                ),
+            )
```
r0x0d, line 49: Remember to remove the logger.info above

```suggestion
            message = (
                    "Skipping the check because there are no publicly available %s %d.%d repositories available."
                    % (system_info.name, system_info.version.major, system_info.version.minor)
                )
            logger.info(message)
            self.add_message(
                level="INFO",
                id="PACKAGE_UPDATES_CHECK_SKIP_NO_PUBLIC_REPOSITORIES",
                message=message,
            )
```

convert2rhel/actions/system_checks/package_updates.py
```diff
@@ -60,6 +73,15 @@ def run(self):
...
             )
             logger.warning(str(e))
+            self.add_message(
+                level="WARNING",
+                id="PACKAGE_UP_TO_DATE_CHECK_FAIL",
+                message=(
+                    "There was an error while checking whether the installed packages are up-to-date. Having an updated system is"
+                    " an important prerequisite for a successful conversion. Consider verifyng the system is up to date manually"
+                    " before proceeding with the conversion. %s" % str(e)
+                ),
```
r0x0d, line 83:
```suggestion
                message=message + " %s" % str(e),
```

r0x0d: To avoid duplicating the message

convert2rhel/actions/system_checks/package_updates.py
```diff
@@ -75,5 +97,16 @@ def run(self):
...
+                level="WARNING",
+                id="OUT_OF_DATE_PACKAGES",
+                message=(
+                    "The system has %s package(s) not updated based %s.\n"
+                    "List of packages to update: %s.\n\n"
+                    "Not updating the packages may cause the conversion to fail.\n"
+                    "Consider updating the packages before proceeding with the conversion."
+                    % (len(packages_to_update), repos_message, " ".join(packages_to_update))
+                ),
+            )
```
r0x0d, line 110: Remember to remove the logger.warning above

```suggestion
            message = (
                    "The system has %s package(s) not updated based %s.\n"
                    "List of packages to update: %s.\n\n"
                    "Not updating the packages may cause the conversion to fail.\n"
                    "Consider updating the packages before proceeding with the conversion."
                    % (len(packages_to_update), repos_message, " ".join(packages_to_update))
                )
            logger.warning(message)
            self.add_message(
                level="WARNING",
                id="OUT_OF_DATE_PACKAGES",
                message=message,
            )
```

convert2rhel/unit_tests/actions/pre_ponr_changes/handle_packages_test.py
```diff
@@ -53,48 +53,101 @@ def test_list_third_party_packages_no_packages(list_third_party_packages_instanc
 
 def test_list_third_party_packages(list_third_party_packages_instance, monkeypatch, caplog):
     monkeypatch.setattr(pkghandler, "get_third_party_pkgs", unit_tests.GetInstalledPkgsWFingerprintsMocked())
-    monkeypatch.setattr(pkghandler, "print_pkg_info", PrintPkgInfoMocked())
-
+    monkeypatch.setattr(pkghandler, "format_pkg_info", PrintPkgInfoMocked(["pytest", "ruby", "shim"]))
+    monkeypatch.setattr(system_info, "name", "Centos7")
```
r0x0d, line 57: I think you can use @centos7 instead

convert2rhel/actions/system_checks/dbus.py
```diff
@@ -34,7 +34,13 @@ def run(self):
         logger.task("Prepare: Check that DBus Daemon is running")
 
         if tool_opts.no_rhsm:
-            logger.info("Skipping the check because we have been asked not to subscribe this system to RHSM.")
+            message = "Skipping the check because we have been asked not to subscribe this system to RHSM."
+            logger.info(message)
+            self.add_message(
+                level="WARNING",
+                id="DBUS_IS_RUNNING_CHECK_SKIP",
```
r0x0d, line 41: I'm not sure if that `id` makes sense in this context... Sure, we are skipping the check, but it is not because we _know it is running_, it is because the user specified the `no_rhsm` option in the CLI.

I would rename this to 
```suggestion
                id="DBUS_NO_RHSM_CHECK_SKIP",
```

or even 

```suggestion
                id="DBUS_NO_RHSM_SKIP",
```

does that makes sense to you?

convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
```diff
@@ -54,7 +54,13 @@ def run(self):  # pylint: disable= too-many-return-statements
 
         reposdir = get_hardcoded_repofiles_dir()
         if reposdir and not system_info.has_internet_access:
-            logger.warning("Skipping the check as no internet connection has been detected.")
+            message = "Skipping the check as no internet connection has been detected."
+            logger.warning(message)
+            self.add_message(
+                level="WARNING",
+                id="IS_LOADED_KERNEL_LATEST_CHECK_SKIP",
```
r0x0d, line 61: Here I would be more specific on why this is being skipped as well 

```suggestion
                id="IS_LOADED_KERNEL_LATEST_NO_INTERNET_ACCESS",
```

r0x0d (commented):
A few comments during the review, but nothing too hard to change. 

Also, noticed that a few new introduced IDs have the name of the Action in their name, is that intentional? I don't see too much value in prefixing the ID with the name of the action, since a report will be generated and will contain both information.

convert2rhel/actions/system_checks/dbus.py
pr-watson: Yes we can go with this name instead

pr-watson:
> A few comments during the review, but nothing too hard to change.
> 
> Also, noticed that a few new introduced IDs have the name of the Action in their name, is that intentional? I don't see too much value in prefixing the ID with the name of the action, since a report will be generated and will contain both information.

Yes that is a valid point, we can rename them

convert2rhel/actions/pre_ponr_changes/subscription.py
```diff
@@ -88,7 +94,13 @@ def run(self):
         super(SubscribeSystem, self).run()
 
         if toolopts.tool_opts.no_rhsm:
-            logger.warning("Detected --no-rhsm option. Skipping.")
+            message = "Detected --no-rhsm option. Skipping."
+            logger.warning(message)
+            self.add_message(
+                level="WARNING",
+                id="NO_RHSM_SKIP",
```
bocekm, line 101: @abadger, what's the main purpose of the `id`? Shouldn't it be unique to make it possible to aggregate results from multiple machines? To make it possible to say: _"This one message was hit on machines x,y,x."_

convert2rhel/actions/system_checks/dbus.py
```diff
@@ -34,7 +34,13 @@ def run(self):
         logger.task("Prepare: Check that DBus Daemon is running")
 
         if tool_opts.no_rhsm:
-            logger.info("Skipping the check because we have been asked not to subscribe this system to RHSM.")
+            message = "Skipping the check because we have been asked not to subscribe this system to RHSM."
+            logger.info(message)
+            self.add_message(
+                level="WARNING",
```
bocekm, line 40: It's kind of an internal thing - convert2rhel needs dbus to work with RHSM. So this should be just info level because the customer doesn't care whether we're checking dbus when they are not using RHSM for the conversion.
```suggestion
                level="INFO",
```

bocekm (changes requested):

convert2rhel/actions/pre_ponr_changes/subscription.py
r0x0d: I think it would be a combination of action_id + id, in which the action_id is unique across every action and the id itself could be repeated, as for example; 

```
PreSubscription.NO_RHSM_SKIP
DbusCheck.NO_RHSM_KIP
SomeOtherCheck.UNKNOWN_ERROR
AVeryImportantCheck.UNKNOWN_ERROR
```

If I remember correctly, we are using UNKNOWN_ERROR now to say that something happened during the execution of the action, and we didn't expect that to happen (mainly logger.critical)

convert2rhel/actions/system_checks/dbus.py
pr-watson: I will change this thanks

convert2rhel/unit_tests/__init__.py
```diff
@@ -448,12 +448,17 @@ def assert_actions_result(instance, level=_NO_USER_VALUE, id=_NO_USER_VALUE, mes
     """Helper function to assert result set by Actions Framework."""
 
     if level != _NO_USER_VALUE:
+        print(instance.result.level)
+        print(STATUS_CODE[level])
```
SpyTec, line 452:
```suggestion
```

convert2rhel/unit_tests/__init__.py
```diff
@@ -448,12 +448,17 @@ def assert_actions_result(instance, level=_NO_USER_VALUE, id=_NO_USER_VALUE, mes
...
         assert instance.result.level == STATUS_CODE[level]
 
     if id != _NO_USER_VALUE:
         assert instance.result.id == id
 
     if message != _NO_USER_VALUE:
+        print(instance.result.message)
+        print("\n")
+        print(message)
```
SpyTec, line 461:
```suggestion
```

SpyTec (changes requested):

convert2rhel/actions/pre_ponr_changes/subscription.py
```diff
@@ -99,7 +111,15 @@ def run(self):
...
             logger.task("Convert: Subscription Manager - Check required repositories")
-            subscription.check_needed_repos_availability(rhel_repoids)
+            non_available_repos = subscription.check_needed_repos_availability(rhel_repoids)
+            if non_available_repos:
+
+                self.add_message(
+                    level="WARNING",
+                    id="NON_AVAILABLE_REPOSITORIES_DETECTED",
+                    message="%s repositories are not available - some packages"
+                    " may not be replaced and thus not supported." % ", ".join(non_available_repos),
```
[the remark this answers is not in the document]
pr-watson: We are removing this message since it comes from the check_needed_repos_availibility function which is removed in later PRs

convert2rhel/actions/system_checks/package_updates.py
```diff
@@ -54,12 +67,17 @@ def run(self):
...
             # Beware that the `RepoError` exception is based on the `pkgmanager` module and the message sent to the output
             # can differ depending if the code is running in RHEL7 (yum) or RHEL8 (dnf).
-            logger.warning(
+            warning_message = (
                 "There was an error while checking whether the installed packages are up-to-date. Having an updated system is"
                 " an important prerequisite for a successful conversion. Consider verifyng the system is up to date manually"
                 " before proceeding with the conversion."
             )
-            logger.warning(str(e))
+            logger.warning(warning_message + " %s" % str(e))
```
abadger, line 75: Note: Better to way to do this is to assemble this in the warning_message rather than both here and in the add_message() call.  (No need to do this now, though.)

abadger (approved):
Code looks good.  If this is passing integration tests (or approval from QE) then this is ready to merge.

abadger:
I examined the integration test and there is a fix needed:
In report.py you need to add an entry for INFO (25):
```
   42 _STATUS_TO_COLOR = {
   43     # SUCCESS
   44     0: "OKGREEN",
   45     # WARNING
   46     51: "WARNING",
   47     # SKIP
   48     101: "FAIL",
   49     # OVERRIDABLE
   50     152: "FAIL",
   51     # ERROR
   52     202: "FAIL",
   53 }
```

convert2rhel/actions/system_checks/package_updates.py
pr-watson: Adding this

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -37,15 +37,25 @@ def run(self):
         logger.task("Convert: List third-party packages")
         third_party_pkgs = pkghandler.get_third_party_pkgs()
         if third_party_pkgs:
-            logger.warning(
+            pkg_list = pkghandler.format_pkg_info(sorted(third_party_pkgs, key=self.extract_packages))
+            message = (
                 "Only packages signed by %s are to be"
                 " replaced. Red Hat support won't be provided"
-                " for the following third party packages:\n" % system_info.name
+                " for the following third party packages:\n%s" % (system_info.name, pkg_list)
```
r0x0d, line 44: I have separated the `pkg_list` in this message, since it will be used to log as a warning, and the formatted table of pkgs would appear in the same line and with the color yellow for the user. 

https://github.com/oamg/convert2rhel/pull/888/files#diff-7bab0f1619d739dd7e18e3643a6814d444189bb25fc420206ae8e3a0b3ddde85R51-R52

We can leave this as-is and follow-up the change in the split four fields PR.

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -60,16 +70,41 @@ def run(self):
         logger.task("Convert: Remove excluded packages")
         logger.info("Searching for the following excluded packages:\n")
         try:
-            pkgs_to_remove = pkghandler._get_packages_to_remove(system_info.excluded_pkgs)
-            pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove)
+            pkgs_to_remove = sorted(pkghandler._get_packages_to_remove(system_info.excluded_pkgs))
+            pkgs_removed = sorted(pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove))
```
r0x0d, line 74: This will fail if the `remove_pkgs_unless_from_redhat` returns None (which could happen). 

A fix for that is implemented in https://github.com/oamg/convert2rhel/pull/888/files#diff-7bab0f1619d739dd7e18e3643a6814d444189bb25fc420206ae8e3a0b3ddde85R85, maybe we can get away in this PR by having the fix in the next one?

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -60,16 +70,41 @@ def run(self):
...
             # changed to something more specific.
             #   - When we can't remove a package.
-            self.set_result(level="ERROR", id="PACKAGE_REMOVAL_FAILED", message=str(e))
+            self.set_result(level="ERROR", id="EXCLUDED_PACKAGE_REMOVAL_FAILED", message=str(e))
+            return
+
+        # shows which packages were not removed, if false, all packages were removed
+        pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+        pkgs_not_removed = [str(pkg) for pkg in pkgs_not_removed]
+        pkgs_removed = [str(pkg) for pkg in pkgs_removed]
```
r0x0d, line 88: This greatly diverge from https://github.com/oamg/convert2rhel/pull/888/files#diff-7bab0f1619d739dd7e18e3643a6814d444189bb25fc420206ae8e3a0b3ddde85R105, which one should we stick to?

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -96,13 +131,39 @@ def run(self):
         logger.task("Convert: Remove packages containing .repo files")
         logger.info("Searching for packages containing .repo files or affecting variables in the .repo files:\n")
         try:
-            pkgs_to_remove = pkghandler._get_packages_to_remove(system_info.repofile_pkgs)
-            pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove)
+            pkgs_to_remove = sorted(pkghandler._get_packages_to_remove(system_info.repofile_pkgs))
+            pkgs_removed = sorted(pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove))
```
r0x0d, line 135: Same as the comment above

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -96,13 +131,39 @@ def run(self):
...
             #   - When we can't remove a package.
-            self.set_result(level="ERROR", id="PACKAGE_REMOVAL_FAILED", message=str(e))
+            self.set_result(level="ERROR", id="REPOSITORY_FILE_PACKAGE_REMOVAL_FAILED", message=str(e))
+            return
+
+        # shows which packages were not removed, if false, all packages were removed
+        pkgs_not_removed = sorted(frozenset(pkgs_to_remove).difference(pkgs_removed))
+
+        pkgs_not_removed = [str(pkg) for pkg in pkgs_not_removed]
+        pkgs_removed = [str(pkg) for pkg in pkgs_removed]
```
r0x0d, line 150: Same as comment above

convert2rhel/pkghandler.py
```diff
@@ -594,13 +605,13 @@ def remove_pkgs_unless_from_redhat(pkgs_to_remove, backup=True):
     """
     if not pkgs_to_remove:
         loggerinst.info("\nNothing to do.")
-        return
+        return pkgs_to_remove
```
r0x0d, line 608: I don't think I understood the reasoning behind this change. Why should we return the `pkgs_to_remove`? 

Instead, can't we return an empty dict, meaning that there was no package removed? The idea of this function is to return the packages that were removed, not the ones we asked for removal.

convert2rhel/unit_tests/__init__.py
```diff
@@ -420,19 +420,30 @@ def wrapped(*args, **kwargs):
...
+            ),
+            PackageInformation(
+                nevra=PackageNevra(name="gpg-pubkey", epoch="0", version="d38b4796", release="570c8cd3", arch=None),
+                fingerprint="199e2f91fd431d51",
+                signature="",
+                packager="",
+                vendor="",
+            ),
+        ]
         return pkgs
```
r0x0d, line 447: Instead of doing this, we could get the `GetInstalledPkgsWFingerprintsMocked` that lives in the pkghandler_test.py and move it here. 

I have done that in the split messages to four fields PR, so, we probably can get away with it in this one.

r0x0d (commented):
A few comments left

convert2rhel/actions/pre_ponr_changes/handle_packages.py
pr-watson: Sounds good

convert2rhel/actions/pre_ponr_changes/handle_packages.py
pr-watson: I put a fix for this in the latest commit, remove_pkgs_unless_from_redhat will return an empty list if the pkgs_to_remove is empty

convert2rhel/pkghandler.py
pr-watson: This will return an empty list, so later a sort isn't done on a NoneType object

convert2rhel/actions/pre_ponr_changes/handle_packages.py
pr-watson: I think we should stick to the one in this PR

convert2rhel/unit_tests/__init__.py
pr-watson: In this test we wanted to use the actual data instead of a mock because it was messing with the integration tests

convert2rhel/actions/pre_ponr_changes/handle_packages.py
abadger: What does `str(pkg)` look like?  The output of `get_pkg_nevra(include_zero_epoch=True)` is `0:pkg-2-3.x86_64`.

convert2rhel/actions/pre_ponr_changes/handle_packages.py
abadger: Note, in addition to the color, the formatting for the message in `add_message()` will differ.  `pkg_list` in this PR will be in tabular form.  In #888, it's a comma separated list based on `third_party_pkgs`.  We probably want to change that so it returns the formatted `pkg_list` as well.

convert2rhel/pkghandler.py
abadger: Even though they're equivalent, I think @r0x0d 's question about why we're returning `pkgs_to_remove` here points out a readability problem.  Returning a new empty list won't be ay different in practice but it will be easier to understand that the empty list is different than the input.

Additionally, since pkgs_to_remove is a list, it is a mutable value.  Returning the input list as the output runs the risk that the caller will mutate one or the other after this function runs, not knowing that they're really the same list.  That could have unexpected consequences.  So it is better to return a new empty list for robustness as well.

abadger:
```suggestion
        return []
```

convert2rhel/unit_tests/__init__.py
abadger: I think in a followup PR we probably want to get rid of the  `GetInstalledPkgsWFingerprintsMocked` in `pkghandler_test.py` and have the code there use this one?  If the code that uses it there is bug free, then it should "just work" since this one now uses the actual data structures that the real code does.

convert2rhel/actions/pre_ponr_changes/handle_packages.py
r0x0d: Here, in this case, we are not using get_pkg_nevra, I believe... We are just calling str to the value we are iterating through...

convert2rhel/pkghandler.py
pr-watson: Ok I will make this change

convert2rhel/logger.py
```diff
@@ -171,6 +171,7 @@ def _debug(self, msg, *args, **kwargs):
 
 class bcolors:
     OKGREEN = "\033[92m"
+    INFO = "\033[0m"
```
r0x0d, line 174:
```suggestion
    INFO = "\033[94m"
```

convert2rhel/actions/pre_ponr_changes/handle_packages.py
r0x0d: For example, this is the output: 

```
[93m(WARNING) REMOVE_EXCLUDED_PACKAGES::EXCLUDED_PACKAGES_NOT_REMOVED - The

    following packages were not removed: PackageInformation(packager='CentOS

    Buildsys <bugs@centos.org>', vendor='CentOS',

    nevra=PackageNevra(name='centos-gpg-keys', epoch='1', version='8',

    release='3.el8', arch='noarch'), fingerprint='05b555b38483c65d',

    signature='RSA/SHA256, Tue Sep 14 21:14:43 2021, Key ID 05b555b38483c65d')[0m
```

r0x0d: > What does str(pkg) look like?

This is not being formatted correctly, we have to use something like: 

```python
[pkghandler.get_pkg_nevra(pkg_obj, include_zero_epoch=True) for pkg_obj in packages]
```

which is defined here: https://github.com/oamg/convert2rhel/pull/888/files#diff-7bab0f1619d739dd7e18e3643a6814d444189bb25fc420206ae8e3a0b3ddde85R27

pr-watson: Ok I will add this change

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -60,16 +80,40 @@ def run(self):
...
+            return
+        if pkgs_removed:
+            message = "The following packages were removed: %s" % ", ".join(pkgs_removed)
+            logger.info(message)
+            self.add_message(
+                level="INFO",
+                id="EXCLUDED_PACKAGES_REMOVED",
+                message=message,
+            )
+            return
```
SpyTec, line 116: I notice we have return in both the if statements. Is it possible that we remove some packages but not all, so that both should add a message?

convert2rhel/pkghandler.py
```diff
@@ -594,13 +605,13 @@ def remove_pkgs_unless_from_redhat(pkgs_to_remove, backup=True):
     """
     if not pkgs_to_remove:
         loggerinst.info("\nNothing to do.")
-        return
+        return []
```
SpyTec, line 608: Codecov claims not covered by tests. It is a branch we should make sure we enter

convert2rhel/actions/pre_ponr_changes/handle_packages.py
pr-watson: Before we had an if/else instead of two if statements so only one of the messages would be displayed to the user. Later found that the else statement would be entered if pkgs_removed is empty which would be a message like this  "The following packages were removed: " with no packages listed. Toshio said it is fine to either display only one of the messages since that was the previous way of displaying. We can change it so both are displayed.

bocekm:
Moved to https://github.com/oamg/convert2rhel/pull/888.

convert2rhel/actions/system_checks/dbus.py
bocekm: Ouch, this PR was closed in favor of introducing the changes under #888 but this requested change hasn't been moved over there. Now when the #888 is merged I'm seeing this warning when running the conversion with the system pre-registered (which will be the Insights Experience use case):
```
(WARNING) DBUS_IS_RUNNING::DBUS_IS_RUNNING_CHECK_SKIP - Skipping the dbus is running check
     Description: Skipping the check because we have been asked not to subscribe this system to RHSM.
     Diagnosis: [No further information given]
     Remediation: [No further information given]
```

PR #868, opened 2023-07-18
[RHELC-1040] Fix package back up during an offline conversion

bocekm:
When the main transaction validation finds a package dependency problem, we try to remove the package that causes the problem. Before removing it we are backing it up so that we can install it back during a rollback.
This backup was failing on systems not connected to the internet.

Jira Issues: [RHELC-1040](https://issues.redhat.com/browse/RHELC-1040)

Checklist

### DIFF
--- a/convert2rhel/backup.py
+++ b/convert2rhel/backup.py
@@ -349,46 +349,37 @@ class RestorablePackage(object):
         :type reposdir: str
         """
         loggerinst.info("Backing up %s." % self.name)
-        if os.path.isdir(BACKUP_DIR):
-            # If we detect that the current system is an EUS release, then we
-            # proceed to use the hardcoded_repofiles, otherwise, we use the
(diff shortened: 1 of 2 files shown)


bocekm:
This PR does not help with the issue where:

1. There are two activation keys defined in Satellite - one for the original vendor repos and the second for RHEL repos.
2. pPior to executing convert2rhel, the system is registered to Satellite with the first activation key.
3. The second activation key is passed to convert2rhel.

The repos available through the Satellite's activation keys are auto-generated in the /etc/yum.repos.d/redhat.repo. We don't back up the redhat repo (`if repo.endswith(".repo") and repo != "redhat.repo":`). It'd be futile anyway because once we unregister the system to register with the second activation key during the conversion, whatever repo was in this redhat.repo would not be reachable (likely).

We'll try to find an alternative solution.

PR #683, opened 2022-12-15
[RHELC-750, RHELC-1005] Rewrite the log messages in download_pkg() to not use "unsupported"

Andrew-ang9:
Convert2RHEL is moving away from using the word unsupported, and we will gradually go away from using this terminology. This PR also goes through the log message to better explain how the conversion can fail.
* Add sanity test to verify the deprecated `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK` envar is still allowed

Jira Issues: 
- [RHELC-750](https://issues.redhat.com/browse/RHELC-750)
- [RHELC-1005](https://issues.redhat.com/browse/RHELC-1005)

Checklist
- [ ] PR meets acceptance criteria specified in the Jira issue
- [ ] PR has been tested manually in a VM (either author or reviewer)
- [x] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [ ] Code and tests are documented properly
- [ ] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending`

PR #875, opened 2023-08-04
[RHELC-855] Improve the rollback

hosekadam:
This PR solves problem with not backing up changed package files. When the MD5 checksum of file related to package (got by `rpm -Va`) differs then is backed up and restored during rollback. If the file is missing, during rollback is deleted.

Jira Issues: [RHELC-855](https://issues.redhat.com/browse/RHELC-855)

Checklist


convert2rhel/backup.py
```diff
@@ -459,3 +459,4 @@ def remove_epoch_from_yum_nevra_notation(package_nevra):
 
 changed_pkgs_control = ChangedRPMPackagesController()  # pylint: disable=C0103
 backup_control = BackupController()
+package_files_changes = []  # control of changed package files
```
hosekadam, line 462: I had problems with storing this list in the action framework. So since it's related to backup I put it there.

hosekadam:
Just some high-level review is enough. I'll improve the WIP unit test later. Just if you could check if the structure of the functionality makes sense.

convert2rhel/backup.py
abadger: Once we're storing state, it feels like we should be using a `RestorableChange` (Writing a new subclass of that) for this instead of adding a new global.  It might be easiest to port `backup.RestorableFile`'s functionality into a `RestorableChange` and then for each file that we determine we need to backup and restore, create a RestorableFile and push that onto the backup_control instance.

hosekadam: @abadger I totally agree we should use the `RestorableChange`. But I didn't want to port the `RestorableFile` in this PR, since it's in a different task. My idea was to write it using the old way and then port it in the task for porting (which sounds terrible to me when I'm writing it :D). 

What just came to my mind is to port the `RestorableFile` there and leave the old one too. Later we can port the things that are using the old way to the new `RestorableFile` which would be already prepared. So from my point of view the "new" `RestorableFile` can be under this PR since it would be used just for this feature.

Another idea is to have PR just for porting the `RestorableFile` class, but without porting the usage of it - this could be done in some other PR. Maybe this would be the best to have the PRs simple - we can just rebase the dependent ones. This is mostly the same as the previous, but maybe more systematic?

My motivation for this is not to do many changes and have the PR as simple as can be. And also not falling into some cherry-picking etc. between PRs (I remember it from the PRs before last merge freeze which I reviewed) and have to worry about some sync between them.

What do you think? What would be the best way to do this?

abadger: Your second idea is the cleanest: 

> Another idea is to have PR just for porting the RestorableFile class, but without porting the usage of it - this could be done in some other PR. Maybe this would be the best to have the PRs simple - we can just rebase the dependent ones. This is mostly the same as the previous, but maybe more systematic?

But your first idea is okay as well:

> What just came to my mind is to port the RestorableFile there and leave the old one too. Later we can port the things that are using the old way to the new RestorableFile which would be already prepared. So from my point of view the "new" RestorableFile can be under this PR since it would be used just for this feature.

The problems in the last release stemmed from starting to do the second one but not keeping track of where the changes needed to go, which PR needed to be reviewed first, and not concentrating on getting the base PR merged ASAP.  We had a secondary problem of the dependent PRs all bein fairly large changes in and of themselves.

I think either of your plans for this one will avoid those pitfalls.

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -61,3 +63,93 @@ def run(self):
...
+                # If the MD5 checksum differs, the content of the file differs
+                file["backup"] = backup.RestorableFile(file["path"])
+                file["backup"].backup()
+
+    def _get_rpm_va(self):
+        """Run the rpm -Va command to get changes made to package files.
+        Return them as a list of dict, for example:
+        [{"status":"S5T", "file_type":"c", "path":"/etc/yum.repos.d/CentOS-Linux-AppStream.repo"}]
+        """
+        cmd = ["rpm", "-Va"]
```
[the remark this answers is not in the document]
r0x0d: My opinion about deprecating `--no-rpm-va`:

I think this will greatly impact in the wait time for the conversion in some use cases where the user has lots and lots of files, modifications to the system etc... Maybe the last suggestion would be more ideal, i.e: "We could make this Action depend on --no-rpm-va not being set.". 

+1 to parse the initial output of `rpm -Va` instead of running it a second time. I don't think the parsing needs to be there, it could be in this action as long as it checks if there is anything to parse first. Seems to me that the only case where we care about parsing the contents of that output is here (correct me if I'm wrong).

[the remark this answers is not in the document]
bocekm: @abadger, great points. Indeed we should not run the `rpm -Va` twice as it is super slow. We shall reuse the output of the first invocation.

I'd vote for keeping the --no-rpm-va switch functionality, i.e. it would prevent running rpm -Va, but because that can lead to an incomplete rollback I'd newly require the user to set the `CONVERT2RHEL_INCOMPLETE_ROLLBACK=1` env var (after #683 is merged). What do you think, @r0x0d, @abadger?

The second best option for me would be to _"change the help documentation to note that --no-rpm-va only disables the rpm -Va which is run after the conversion"_.

convert2rhel/backup.py
hosekadam: The second idea is done there: https://github.com/oamg/convert2rhel/pull/935

convert2rhel/main.py
```diff
@@ -23,6 +23,7 @@
 from convert2rhel import logger as logger_module
 from convert2rhel import pkghandler, pkgmanager, redhatrelease, repo, subscription, systeminfo, toolopts, utils
 from convert2rhel.actions import level_for_raw_action_data, report
+from convert2rhel.actions.pre_ponr_changes.backup_system import BackupPackageFiles
```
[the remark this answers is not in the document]
r0x0d: That would mean that both the action and the other module have shared data... This must be controlled very well on the other module side to prevent errors if the BackupPackageFiles does not run. 

I think that the BackupPackageFiles can be a action that only writes to that shared data structure and the other module (probably pkghandler (or backup?)) does the rollback if necessary... 

Does that introduce more risks than it is intended? I don't think see another solution other than that to use with the action framework... Well, maybe another idea would be a way to tell the action framework like: "Run this action in case of failure", kinda of a new stage that is only triggered in case of failure, with that in mind, we would have two actions: 
1. BackupPackageFiles - to backup the packages and have it in a shared data structure
2. RollbackPackageFiles - to restore what the BackupPackageFiles saved

But that would be too much work for now, right? Probably the first idea sounds more straightforward to do.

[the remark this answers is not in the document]
r0x0d: I also see that this PR has a dependency on https://github.com/oamg/convert2rhel/pull/935, so my comment may not be appropriate at all.

convert2rhel/actions/pre_ponr_changes/backup_system.py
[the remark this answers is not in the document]
r0x0d: +1 to that, @bocekm

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -63,3 +67,96 @@ def run(self):
...
+
+        # Regex explanation:
+        # Match missing or SM5DLUGTP (letters can be replaced by dots or ?):
+        #   (missing|([S\.\?][M\.\?][5\.\?][D\.\?][L\.\?][U\.\?][G\.\?][T\.\?][P\.\?]))
+        # Match whitespace: \s+
+        # Match type of file, can be replaced by any whitecharacter:
+        #   [cdlr\s+]
+        # Match unix path:
+        #   [\/\\](?:(?!\.\s+)\S)+(\.)?
+        regex = r"^(missing|([S\.\?][M\.\?][5\.\?][D\.\?][L\.\?][U\.\?][G\.\?][T\.\?][P\.\?]))\s+[cdlr\s+]\s+[\/\\](?:(?!\.\s+)\S)+(\.)?$"
```
r0x0d, line 127: Since we are calling this in a for, maybe we should compile it outside the module first...

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -63,3 +67,96 @@ def run(self):
...
+        input = input.strip()
+        lines = input.split("\n")
+        data = []
+
+        for line in lines:
+            parsed_line = self._parse_line(line.strip())
+            if parsed_line:
+                data.append(parsed_line)
+
+        return data
```
r0x0d, line 114: What if we merge those two functions and have just one that do the parsing? 

```suggestion
    def _parse_rpm_va(self):
        """Get the output from rpm -Va command from generated during resolving system info
        to get changes made to package files.
        Return them as a list of dict, for example:
        [{"status":"S5T", "file_type":"c", "path":"/etc/yum.repos.d/CentOS-Linux-AppStream.repo"}]
        """
        path = os.path.join(LOG_DIR, PRE_RPM_VA_LOG_FILENAME)
        with open(path, "r") as f:
            output = f.read().strip()

        lines = output.split("\n")
        data = []

        for line in lines:
            parsed_line = self._parse_line(line.strip())
            if parsed_line:
                data.append(parsed_line)

        return data
```

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -63,3 +67,96 @@ def run(self):
...
+        #   [\/\\](?:(?!\.\s+)\S)+(\.)?
+        regex = r"^(missing|([S\.\?][M\.\?][5\.\?][D\.\?][L\.\?][U\.\?][G\.\?][T\.\?][P\.\?]))\s+[cdlr\s+]\s+[\/\\](?:(?!\.\s+)\S)+(\.)?$"
+
+        ret = re.match(regex, line)
+
+        if not ret:  # line not matching the regex
+            if line.strip() != "":
+                # Line is not empty string
+                loggerinst.debug("Skipping invalid output %s" % line)
+            return
```
r0x0d, line 135: I'm wondering if we should return a dictionary here with some default values to avoid the ` Explicit returns mixed with implicit (fall through) returns` from codeql. Seems to me that it would be a benefit to have the same return type in this case as well, but with default values, like, `None`.

r0x0d: We can filter it out in the function above that depends on this result to only append data that has something other than the default value we don't care.

hosekadam:
What was changed:
- the backup now uses the restorable file from https://github.com/oamg/convert2rhel/pull/935
- added the backup as dependency to some actions. I tried to add it also to `package_updates` but it caused an error, that some dependencies weren't satisfied. It's caused by having this in different folder (checks vs pre_ponr_changes) - so it's not necessary to have it there as dependency since it's just a check.
- if user uses the `--no-rpm-va` the `CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK` is required. I used this one with "unsupported" to have it working with the code we have now. It should accordingly to https://github.com/oamg/convert2rhel/pull/683 when it will be merged
- changed to use the output of `rpm -Va` from system info (load the content of generated file in systeminfo)
- updated the unit tests

I have question about printing the messages there. Now it works like this:
- during standard run is printed `TASK - [Prepare: Backup package files]` and then messages which files were backed up.
- during rollback is printed e.g. this:
```
[2023-10-20T10:31:34+0000] TASK - [Rollback: Restore /usr/lib/python2.7/site-packages/convert2rhel/utils.pyc from backup] 
File /usr/lib/python2.7/site-packages/convert2rhel/utils.pyc restored.
```
It corresponds from my point of view with the other rollback tasks and restoring, where restoring every file is a task. E.g.
```
[2023-10-20T10:31:34+0000] TASK - [Rollback: Restore /etc/system-release from backup] ****************
File /etc/system-release restored.
```
But there are many lines about that. Is it okay, or should it be changed to something else? I'm not sure about printing just initial TASK and then just list of the files since we are using stack in backupcontroller, and I'm not sure if solution of this problem wouldn't be too complicated.

convert2rhel/actions/pre_ponr_changes/backup_system.py
hosekadam: Yes, good idea

convert2rhel/actions/pre_ponr_changes/backup_system.py
hosekadam: Yeah, it may be more efficient, we won't need to call the function and save a little of time

convert2rhel/actions/pre_ponr_changes/backup_system.py
hosekadam: Makes sense, it will be more systematic

hosekadam:
Applied the suggestions @r0x0d

SpyTec:
> But there are many lines about that. Is it okay, or should it be changed to something else? I'm not sure about printing just initial TASK and then just list of the files since we are using stack in backupcontroller, and I'm not sure if solution of this problem wouldn't be too complicated.

@hosekadam I assume you mean one task with many lines for what is rolled back? I t? Ihink it might be a bit out of scope for the time being if we already have the one solution done. But the UX would be an improvement, so we can create a separate task for that IMO. Perfecting this isn't necessary as it conveys the same message to the user just in a condensed way

convert2rhel/toolopts.py
```diff
@@ -187,6 +187,7 @@ def _register_options(self):
             " 'rpm -Va'. By default it's performed before and after the conversion with the output"
             " stored in log files %s and %s. At the end of the conversion, these logs are compared"
             " to show you what rpm files have been affected by the conversion."
+            " For enabling this option is required CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK se to 1."
```
SpyTec, line 190:
```suggestion
            " For enabling this option is required CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK set to 1."
```

SpyTec: WDYT of this suggestion?

convert2rhel/backup.py
```diff
@@ -348,6 +348,79 @@ def restore(self):
...
+        # Prevent multiple backup
+        if self.enabled:
+            return
+
+        loggerinst.info("Backing up %s." % self.filepath)
+        if os.path.isfile(self.filepath):
+            try:
+                loggerinst.debug("Copying %s to %s." % (self.filepath, BACKUP_DIR))
+                shutil.copy2(self.filepath, BACKUP_DIR)
+
```
SpyTec, line 367: I'd probably move this around so that we have a debug message for successfully backing up the file after copying it, unless we do this somewhere else then consistency is better

convert2rhel/backup.py
```diff
@@ -348,6 +348,79 @@ def restore(self):
...
+            loggerinst.info("%s hasn't been backed up." % self.filepath)
+            return
+
+        try:
+            shutil.copy2(backup_filepath, self.filepath)
+        except (OSError, IOError) as err:
+            # Do not call 'critical' which would halt the program. We are in
+            # a rollback phase now and we want to rollback as much as possible.
+            # IOError for py2 and OSError for py3
+            loggerinst.warning("Error(%s): %s" % (err.errno, err.strerror))
```
SpyTec, line 400: Not a blocker. I'm curious if we can use critical_no_exit here, but we shouldn't experiment atm

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -63,3 +79,81 @@ def run(self):
...
+        loggerinst.task("Prepare: Backup package files")
+
+        super(BackupPackageFiles, self).run()
+
+        package_files_changes = self._parse_rpm_va()
+
+        for file in package_files_changes:
+            if file["status"] == "missing":
+                pass  # no action needed now
+            elif re.search(r"[5]", file["status"]):
```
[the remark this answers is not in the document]
hosekadam: My motivation there to use the regex was the ability to easily add other symbols. But who knows if it will be needed and for now, this is fine. I'll change it

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -63,3 +79,81 @@ def run(self):
...
+                # If the MD5 checksum differs, the content of the file differs
+                restorable_file = backup.NewRestorableFile(file["path"])
+                backup.backup_control.push(restorable_file)
+
+    def _parse_rpm_va(self):
```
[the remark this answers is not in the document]
hosekadam: Agree, more readable without need to know what rpm -va does :)

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -63,3 +79,81 @@ def run(self):
...
+
+    def _parse_line(self, line):
+        """Return {"status":"S5T", "file_type":"c", "path":"/etc/yum.repos.d/CentOS-Linux-AppStream.repo"}"""
+        ret = re.match(RPM_VA_REGEX, line)
+
+        if not ret:  # line not matching the regex
+            if line.strip() != "":
+                # Line is not empty string
+                loggerinst.debug("Skipping invalid output %s" % line)
+            return {"status": None, "file_type": None, "path": None}
```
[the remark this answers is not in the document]
hosekadam: @SpyTec What about adding to the condition which checks if the path is not empty another one to check the path too. So it would like:

`if parsed_line["path"] and parsed_line["status"] :`
on this line: https://github.com/oamg/convert2rhel/pull/875/files#diff-5d146d6a82d3b246af03b014ec5d8aa5686de4e19801aaddf68d18d33fe5549eR119 (the one before appending the parsed_line)

convert2rhel/toolopts.py
hosekadam: I like your suggestion, seems better

convert2rhel/backup.py
hosekadam: I'll change this in this PR which is related to that: https://github.com/oamg/convert2rhel/pull/935

convert2rhel/actions/pre_ponr_changes/backup_system.py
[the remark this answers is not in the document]
SpyTec: Sounds good to me!

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -63,3 +79,81 @@ def run(self):
...
+            path = line[1]
+
+        else:
+            file_type = line[1]
+            path = line[2]
+
+        return {"status": status, "file_type": file_type, "path": path}
+
+    @staticmethod
+    def rollback_files():
```
SpyTec, line 151: Is this tested anywhere?

convert2rhel/toolopts.py
```diff
@@ -384,7 +385,11 @@ def _process_cli_options(self):
             tool_opts.activation_key = None
 
         if parsed_opts.no_rpm_va:
-            tool_opts.no_rpm_va = True
+            if os.getenv("CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK", None):
+                tool_opts.no_rpm_va = True
+            else:
+                message = "For running without rpm -Va the CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK=1 is required."
```
bocekm, line 391:
```suggestion
                message = ("We need to run the rpm -Va command for a reliable and complete rollback of changes done to"
                           " the system during the pre-conversion analysis. If you accept the risk of an incomplete"
                           " rollback, set CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK=1. Otherwise, remove the
                           " --no-rpm-va option.")
```

bocekm, line 391:
```suggestion
                message = ("We need to run the 'rpm -Va' command to be able to perform a complete rollback of changes"
                           " done to the system during the pre-conversion analysis. If you accept the risk of an"
                           " incomplete rollback, set the CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK=1 environment"
                           " variable. Otherwise, remove the --no-rpm-va option.")
```

bocekm (changes requested):

convert2rhel/actions/pre_ponr_changes/backup_system.py
hosekadam: @SpyTec It's not used anywhere, probably some leftover from previous implementation without the `NewRestorableFile`. But then it seems to me that I don't handle the missing files - not removing them. I'll take a look if I'm right

danmyway (changes requested):

SpyTec (approved):
Overall LGTM!

danmyway (approved):
Tests run from #1007 
Looks good from my POV

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -63,3 +89,93 @@
...
+            if os.environ.get("CONVERT2RHEL_UNSUPPORTED_INCOMPLETE_ROLLBACK", None):
+                loggerinst.debug("Skipping backup of the package files." " CONVERT2RHEL_INCOMPLETE_ROLLBACK detected.")
+                # Return empty list results in no backup of the files
+                return data
+            else:
+                # The file should be there
+                # If missing conversion is in unknown state
+                loggerinst.critical("Missing file {rpm_va_output} in it's location".format(rpm_va_output=path))
+
+        output = output.strip()
```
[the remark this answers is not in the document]
danmyway: Filed an [issue](https://issues.redhat.com/browse/RHELC-1284) for that, dismissing for now.

PR #879, opened 2023-08-08
[RHELC-1107] Schema to validate the pre-conversion assessment json format 

abadger:
This change adds a schema for the pre-conversion assessment's json format.  This branch has a few extra commits that update the json output to be compliant with the schema, however, those are somewhat hacky and could be implemented in a separate PR if needed.

Jira Issues: [RHELC-1107](https://issues.redhat.com/browse/RHELC-1107)

To be done: 
- [ ] Handle the four fields for messages in `actions/__init__.py` and `actions/report.py`

Checklist:

### DIFF
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -131,10 +131,10 @@ def summary(results, include_all_reports=False, with_colors=True):
                     {
                         "level": int,
                         "id": "$id",
-                        "title": None or "$title",
-                        "description": None or "$description",
-                        "diagnosis": None or "$diagnosis",
-                        "remediation": None or "$remediation"
+                        "title": "" or "$title",
+                        "description": "" or "$description",
+                        "diagnosis": "" or "$diagnosis",
+                        "remediation": "" or "$remediation"
                     }
                 ],
                 "result" : {
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -208,9 +208,9 @@ def summary(results, include_all_reports=False, with_colors=True):
         combined_results_and_message = find_actions_of_severity(
             combined_results_and_message, "WARNING", level_for_combined_action_data
         )
+
     terminal_size = utils.get_terminal_size()
-    # word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=terminal_size[0], replace_whitespace=False)
-    word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=240, replace_whitespace=False)
+    word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=terminal_size[0], replace_whitespace=False)
 
     # Sort the results in reverse order, this way, the most important messages
     # will be on top.
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -127,11 +127,23 @@ def summary(results, include_all_reports=False, with_colors=True):
     .. note:: Expected results format is as following
         {
             "$Action_id": {
-                "messages" : [{"level": int, "id": "$id", "message": None or "$message"}],
+                "messages" : [
+                    {
+                        "level": int,
+                        "id": "$id",
+                        "title": None or "$title",
+                        "description": None or "$description",
+                        "diagnosis": None or "$diagnosis",
+                        "remediation": None or "$remediation"
+                    }
+                ],
                 "result" : {
                     "level": int,
                     "id": "$id",
-                    "message": "" or "$message"
+                    "title": "" or "$title",
+                    "description": "" or "$description",
+                    "diagnosis": "" or "$diagnosis",
+                    "remediation": "" or "$remediation"
                 }
             },
         }
@@ -197,7 +209,9 @@ def summary(results, include_all_reports=False, with_colors=True):
             combined_results_and_message, "WARNING", level_for_combined_action_data
         )
     terminal_size = utils.get_terminal_size()
-    word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=terminal_size[0], replace_whitespace=False)
+    # word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=terminal_size[0], replace_whitespace=False)
+    word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=240, replace_whitespace=False)
+
     # Sort the results in reverse order, this way, the most important messages
     # will be on top.
     combined_results_and_message = sorted(combined_results_and_message, key=lambda item: item[1]["level"], reverse=True)
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -134,7 +134,8 @@ def summary(results, include_all_reports=False, with_colors=True):
                         "title": "" or "$title",
                         "description": "" or "$description",
                         "diagnosis": "" or "$diagnosis",
-                        "remediation": "" or "$remediation"
+                        "remediation": "" or "$remediation",
+                        "variables": None or "$variables",
                     }
                 ],
                 "result" : {
@@ -143,7 +144,8 @@ def summary(results, include_all_reports=False, with_colors=True):
                     "title": "" or "$title",
                     "description": "" or "$description",
                     "diagnosis": "" or "$diagnosis",
-                    "remediation": "" or "$remediation"
+                    "remediation": "" or "$remediation",
+                    "variables": None or "$variables",
                 }
             },
         }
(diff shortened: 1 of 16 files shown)


danmyway:
@abadger should I also move the [c2r_facts_schema.json](https://github.com/oamg/convert2rhel/blob/main/tests/integration/common/checks-after-conversion/artifacts/c2r_facts_schema.json) and [c2r_migration_results_schema.json](https://github.com/oamg/convert2rhel/blob/main/tests/integration/common/checks-after-conversion/artifacts/c2r_migration_results_schema.json) from the [current directory](https://github.com/oamg/convert2rhel/tree/main/tests/integration/common/checks-after-conversion/artifacts) to `schemas`?

convert2rhel/actions/report.py
```diff
@@ -162,7 +180,9 @@ def summary(results, include_all_reports=False, with_colors=True):
             combined_results_and_message, "WARNING", level_for_combined_action_data
         )
     terminal_size = utils.get_terminal_size()
-    word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=terminal_size[0], replace_whitespace=False)
+    # word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=terminal_size[0], replace_whitespace=False)
+    word_wrapper = textwrap.TextWrapper(subsequent_indent="    ", width=240, replace_whitespace=False)
```
r0x0d, line 184: @abadger, I increased the value here just to be able to let the tests pass for now, and will try to figure it out later. 

TODO (@r0x0d): Get this working again later.

r0x0d:
> @abadger should I also move the [c2r_facts_schema.json](https://github.com/oamg/convert2rhel/blob/main/tests/integration/common/checks-after-conversion/artifacts/c2r_facts_schema.json) and [c2r_migration_results_schema.json](https://github.com/oamg/convert2rhel/blob/main/tests/integration/common/checks-after-conversion/artifacts/c2r_migration_results_schema.json) from the [current directory](https://github.com/oamg/convert2rhel/tree/main/tests/integration/common/checks-after-conversion/artifacts) to `schemas`?

@danmyway, I would say that those two should be separated from the one introduced here, since the one in our root repo is for everyone else to consult and make sure they can parse the correct format of our JSON based on that schema. The ones you mentioned are more for the internal tests, while it is true someone could use that for parsing those files, I doubt that would be the case for now, as those files are pretty much generated for specific services that I think there won't be more dependents.

@abadger, correct me here if I'm wrong.

danmyway:
> > @abadger should I also move the [c2r_facts_schema.json](https://github.com/oamg/convert2rhel/blob/main/tests/integration/common/checks-after-conversion/artifacts/c2r_facts_schema.json) and [c2r_migration_results_schema.json](https://github.com/oamg/convert2rhel/blob/main/tests/integration/common/checks-after-conversion/artifacts/c2r_migration_results_schema.json) from the [current directory](https://github.com/oamg/convert2rhel/tree/main/tests/integration/common/checks-after-conversion/artifacts) to `schemas`?
> 
> @danmyway, I would say that those two should be separated from the one introduced here, since the one in our root repo is for everyone else to consult and make sure they can parse the correct format of our JSON based on that schema. The ones you mentioned are more for the internal tests, while it is true someone could use that for parsing those files, I doubt that would be the case for now, as those files are pretty much generated for specific services that I think there won't be more dependents.
> 
> @abadger, correct me here if I'm wrong.

Understood, thank you, @r0x0d !

convert2rhel/actions/report.py
```diff
@@ -174,13 +211,14 @@ def summary(results, include_all_reports=False, with_colors=True):
...
-            entry = colorize(entry, _STATUS_TO_COLOR[combined_result["level"]])
-        report.append(entry)
+        entry = format_action_status_message(combined_result["level"], message_id[0], message_id[1], combined_result)
+        paragraphs = wrap_paragraphs(entry, width=terminal_size[0])
+
+        for paragraph in paragraphs:
+            entry = word_wrapper.fill(paragraph)
+            if with_colors:
+                entry = colorize(entry, _STATUS_TO_COLOR[combined_result["level"]])
+            report.append(entry)
```
abadger, line 221: I haven't looked at the output, but I think this needs to be changed a little:

- Each report entry should correspond to one log message rather than one paragraph.
- We can color for one entry rather than for each paragraph.
- I thnk it then makes sense to have wrap_paragraphs do the `word_wrapper.fill()` and return the wrapped string (then its API becomes single string in, single string out).  So this part would look like:

```suggestion
        entry = wrap_paragraphs(entry, width=terminal_size[0])

        if with_colors:
            entry = colorize(entry, _STATUS_TO_COLOR[combined_result["level"]])
        report.append(entry)
```

convert2rhel/actions/report.py
```diff
@@ -89,6 +89,23 @@ def summary_as_json(results, json_file=CONVERT2RHEL_JSON_RESULTS):
...
+    This solution was taken from
+    https://github.com/python/cpython/issues/46167#issuecomment-1093406764,
+    which is a solution to textwrap not properly respecting line breaks inside
+    strings.
+    """
+    return [
+        line
+        for para in text.splitlines()
+        for line in textwrap.wrap(para, width, subsequent_indent="    ", **kwargs) or [""]
+    ]
```
abadger, line 106: (Note: This is coupled with a suggestion/justification in the summary() function where this is called from.)
``` python
def format(text, width, **kwargs):
    output = []
    first = True
    indent = ""
    subsequent_indent = "    "

    for paragraph in text.splitlines():
        for line in textwrap.wrap(paragraph, width, initial_indent=indent,
subsequent_indent=subsequent_indent, **kwargs) or [""]:
            output.append(line)

        if first:
            indent = subsequent_indent
            subsequent_indent = ""
            first = False

    return "\n".join(output)
```

r0x0d: copied this over to #888

convert2rhel/actions/report.py
r0x0d: copied this over to #888

pr-watson (approved):
Schema looks good to me, approved. Just need integration tests passing

hosekadam:
I'm testing this PR and trying to validate the output data in json. And I'm getting `Required properties are missing from object: action_results.` after validation.

How to reproduce:

- Run the conversion: `convert2rhel analyze --no-rpm-va --serverurl <REPLACE_WITH_URL> --username test --password test --pool a_pool --debug` It is the command from `tests/integration/tier0/non-destructive/assessment-report/test_assessment_report.py`
- Take the output of pre-assesment from `/var/log/convert2rhel/convert2rhel-pre-conversion.json`
- Use the schema from `schemas/assessment-schema-1.0.json` and get the error

r0x0d:
> I'm testing this PR and trying to validate the output data in json. And I'm getting `Required properties are missing from object: action_results.` after validation.
> 
> How to reproduce:
> 
>     * Run the conversion: `convert2rhel analyze --no-rpm-va --serverurl <REPLACE_WITH_URL> --username test --password test --pool a_pool --debug` It is the command from `tests/integration/tier0/non-destructive/assessment-report/test_assessment_report.py`
> 
>     * Take the output of pre-assesment from `/var/log/convert2rhel/convert2rhel-pre-conversion.json`
> 
>     * Use the schema from `schemas/assessment-schema-1.0.json` and get the error
> 
> 
> EDIT: I see the same error in the integration test http://artifacts.osci.redhat.com/testing-farm/6e0aac16-6bcc-4ca8-8452-d03b2519c6e9/ `jsonschema.exceptions.ValidationError: 'action_results' is a required property`
> 
> Is the schema valid for the json we are outputting?

It should be valid! I will try to provide a fix asap.

abadger:
Okay, the json report from convert2rhel and the schema should now be in agreement.

hosekadam (approved):
I think the schema is fine. Fingers crossed to pass the integration test

tests/integration/tier0/non-destructive/assessment-report/test_assessment_report.py
```diff
@@ -1,15 +1,44 @@
...
+def _validate_report():
+    """
+    Helper function.
+    Verify the report is created in /var/log/convert2rhel/convert2rhel-pre-conversion.json,
+    and it corresponds to its respective schema.
+    """
+    assert os.path.exists(PRE_CONVERSION_REPORT)
+
+    with open(PRE_CONVERSION_REPORT, "r") as data:
+        data_json = json.load(data)
```
kokesak, line 25: Use the `_load_json_schema()` function. No need to wait for the new test results as it is already passing

kokesak (changes requested):

bocekm (approved):

kokesak (approved):

PR #890, opened 2023-08-22
Application lock; uses standard Linux lockfile method [RHELC-1099]

jochapma:
…nd unit test.

Jira Issues: [RHELC-](https://issues.redhat.com/browse/RHELC-1099)

Checklist

### DIFF
--- a/convert2rhel/applock.py
+++ b/convert2rhel/applock.py
@@ -1,0 +1,170 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright(C) 2023 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import atexit
+import errno
+import io
+import logging
+import os
+import time
+
+loggerinst = logging.getLogger(__name__)
+
+class ApplicationLockedError(Exception):
+    """Raised when this application is locked."""
+
+    def __init__(self, message):
+        super(ApplicationLockedError, self).__init__(message)
+        self.message = message
+
+
+class ApplicationLock(object):
+    """Holds a lock for a particular application.
+
+    The implementation uses a standard Linux PID file, though that
+    fact (and all implementation details) are hidden from the caller.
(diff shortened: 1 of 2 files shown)


convert2rhel/applock.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def set_lock_dir(self, new_tmp_dir):
+        """Set the lock directory. Used only for testing.
+
+        :param new_tmp_dir: the directory to be used for the lock file
+        """
+        self._lock_dir = new_tmp_dir
+        self._pidfile = os.path.join(self._lock_dir, self._name + ".pid")
+
+    def _try_creat(self):
```
r0x0d, line 74: typo here 

```suggestion
    def _try_create(self):
```

r0x0d:
@jochapma, I think we forgot to call this class in main, right? Not exactly sure where in main it would be called... 

Actually, should this be called in the `initialize.py` module instead? This way, that would be the first thing we could create/verify before actually getting to anything else for the conversion. What do you think?

convert2rhel/applock.py
```diff
@@ -0,0 +1,171 @@
...
+        encoded = (str(self._pid) + "\n").encode("ascii")
+        os.write(file_desc, encoded)
+        os.close(file_desc)
+        loggerinst.debug("%s." % self)
+        return True
+
+    def is_locked(self):
+        """Test whether this object is locked."""
+        return self._locked
+
```
r0x0d, line 95: What about converting this as a property? 


```suggestion
    @property
    def is_locked(self):
        """Test whether this object is locked."""
        return self._locked

```

convert2rhel/applock.py
```diff
@@ -0,0 +1,171 @@
...
+    @staticmethod
+    def _pid_exists(pid):
+        """Test whether a particular process exists."""
+        try:
+            os.kill(pid, 0)
+        except ProcessLookupError:
+            return False
+        except PermissionError:
+            pass
```
r0x0d, line 104: Maybe a little comment here explaining with there is a permission error. Just to make code scanning happy, as this is easily doable. 

Also, might not be intuitive to everyone why PermissionError is raised here when checking for PIDs, so that might be useful to other people reading the code at some point.

convert2rhel/applock.py
```diff
@@ -0,0 +1,171 @@
...
+            file_desc = os.open(self._pidfile, os.O_CREAT | os.O_WRONLY | os.O_EXCL, 0o755)
+        except OSError as exc:
+            if exc.errno == errno.EEXIST:
+                return False
+            raise exc
+        encoded = (str(self._pid) + "\n").encode("ascii")
+        os.write(file_desc, encoded)
+        os.close(file_desc)
+        loggerinst.debug("%s." % self)
+        return True
```
r0x0d, line 90: I'm curious on why not using a contextmanager here with `with open(self._pidfile...)`, is it because `os.write` or other calls could raise an exception and mess with that check? 

At least, looking at the code for the first time, that's the impression I have.

jochapma:
Sorry, I clicked the wrong button and sent this out too early. Will generate a new and improved PR today.

PR #896, opened 2023-08-31
Add expected_returncode param to the fixture

danmyway:
* expected_returncode defaults to 0, if any other returncode is expected specify the parameter

Checklist

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -127,17 +127,25 @@ def convert2rhel(shell):
         c2r_runtime.logfile_read = sys.stdout
         try:
             yield c2r_runtime
-        except Exception:
-            c2r_runtime.close()
-            raise
-        else:
             c2r_runtime.expect(pexpect.EOF)
             c2r_runtime.close()
+        finally:
+            # Check if child is still alive, if so, send SIGINT
+            # this handles the TIMEOUT exception - if the process is still alive,
+            # the EOF is not raised, the process gets terminated.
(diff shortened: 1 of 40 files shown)


Venefilyn:
@danmyway What do you wanna do with this PR?

danmyway:
@Venefilyn keeping this in a backlog for now.
I might revisit until the end of this Q if everything else goes well.

tests/integration/conftest.py
```diff
@@ -113,7 +116,9 @@ def factory(
         options: str,
         timeout: int = 60 * 60,
         unregister: bool = False,
+        expected_exitcode: Union[List[int], int] = 0,
```
r0x0d, line 119: The idea of this accepting a list of expected return codes is that we can say for example: 

```python
expected_exitcode = [0, 1]
```

And the underlaying framework (pexpect `exitstatus`) will check if that return code is inside our list of expected, right?

If my understanding is right, I kinda have a few questions around this: 

1. Could we mess up tests by saying that `expected_exitcode = [0, 1, 2]` and not verify correctly that the tool is giving the real expected exit code? Of course, we have reviews and such, by I can also see this being used to avoid checking exit codes at all.
2. I haven't seen any case in the current commit that uses a list of exit codes to check the return (might be because I overlooked), but couldn't we start simpler and use just only one `expected_exitcode`?

danmyway:
Closing this for now, might revisit later in the future.

PR #897, opened 2023-08-31
[RHELC-676] Unify parsing of the system release string

bookwar:
Jira Issues: [RHELC-676](https://issues.redhat.com/browse/RHELC-676)

There are several places where the system release string is parsed via regular expressions. These regexps are different and adjusting them requires refactoring in multiple places.

This change introduces one parsing function which does a full match of the system-release string and reads all possible data points from it, and provides the result as a dictionary for later use by other tools.

The function replaces three other helper methods: _get_system_name, _get_system_version and _get_system_distribution_id, which are removed.

NOTE1: The parsing function could be a static method or even implemented as a utility function outside the class. The only reason why it is in the class is the logger object.

NOTE2: the function overlaps with the get_system_release_info function. We should check if this needs to be further unified.

Checklist

### DIFF
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -117,10 +117,13 @@ class SystemInfo:
     def resolve_system_info(self):
         self.logger = logging.getLogger(__name__)
         self.system_release_file_content = self.get_system_release_file_content()
-        self.name = self._get_system_name()
-        self.id = self.name.split()[0].lower()
-        self.distribution_id = self._get_system_distribution_id()
-        self.version = self._get_system_version()
+
+        system_release_data = self.parse_system_release_content()
+        self.name = system_release_data["name"]
+        self.id = system_release_data["id"]
+        self.distribution_id = system_release_data["distribution_id"]
+        self.version = system_release_data["version"]
+
         self.arch = self._get_architecture()
 
         self.cfg_filename = self._get_cfg_filename()
@@ -152,49 +155,89 @@ class SystemInfo:
 
         return redhatrelease.get_system_release_content()
 
-    def _get_system_name(self, system_release_content=None):
-        content = self.system_release_file_content if not system_release_content else system_release_content
-        name = re.search(r"(.+?)\s?(?:release\s?)?\d", content).group(1)
-        return name
+    def parse_system_release_content(self, system_release_content=None):
+        """Parse the content of the system release string
 
-    def _get_system_version(self, system_release_content=None):
-        """Return a namedtuple with major and minor elements, both of an int type.
+        If system_release_content is not provided we use the content from the self.system_release_file_content.
 
-        Examples:
-        Oracle Linux Server release 7.8
-        CentOS Linux release 7.6.1810 (Core)
-        CentOS Linux release 8.1.1911 (Core)
+        :param system_release_content: The contents of the system_release file if needed.
+        :type system_release_content: str
+        :returns: A dictionary containing the system release information
+        :rtype: dict[str, str]
         """
+
         content = self.system_release_file_content if not system_release_content else system_release_content
-        match = re.search(r".+?(\d+)\.(\d+)\D?", content)
+
+        match = re.match(
+            # We assume that the /etc/system-release content follows the pattern:
+            # "<name> release <full_version> <Beta> (<dist_id>)"
+            # Here
+            # - <name>, parsed as a named group '(?P<name>.+?)',
+            #   is a non-empty string of arbitrary symbols.
+            # - "release ", parsed as (?:release\s)?,
+            #   is a literal string. It is optional, and it is dropped when parsed, so this group doesn't have a name.
+            # - <full_version>, parsed as a named group '(?P<full_version>[.\d]+)',
+            #   is a string of arbitrary length containing numbers and dots, for example 8.1.1911 or 7.9.
+            #   (For now we do not have examples with version strings containing letters or other symbols)
+            # - <Beta>, just the word Beta which may appear in some cases, parsed as '(?:\sBeta)?'.
+            # - <dist_id>, parsed as a named group (?P<dist_id>.+), is optional and when it is present, it must appear
+            #   in brackets. Thus, the named group is nested under an unnamed group which starts from a space and then
+            #   has brackets '(\s\( ...here goes the nested group... \))?'
+            #
+            # Example:
+            #
+            #   CentOS Stream release 8
+            #   <    name   > <      ><full_version>
+            #
+            #   CentOS Linux release 8.1.1911        (Core     )
+            #   <    name  > <      ><full_version><  <dist_id>>
+            #   CentOS Linux release 8.1.1911      Beta       (Core     )
+            #   <    name  > <      ><full_version><    ><  <dist_id>>
+            r"^(?P<name>.+?)\s(?:release\s)?(?P<full_version>[.\d]+)(?:\sBeta)?(\s\((?P<dist_id>.+)\))?$",
+            content,
+        )
+
         if not match:
-            self.logger.critical("Couldn't get system version from the content string: %s" % content)
-        version = Version(int(match.group(1)), int(match.group(2)))
+            self.logger.critical_no_exit("Couldn't parse the system release content string: %s" % content)
+            return {}
 
-        return version
+        name = match.group("name")
+        system_id = name.split()[0].lower()
 
-    def _get_system_distribution_id(self, system_release_content=None):
-        """Return the distribution id from the system release file.
+        distribution_id = match.group("dist_id")
 
-            .. note::
-                This distribution id differs from the property `id` we have in the SystemInfo class
-                as this id is the last thing that appears on the system-release file as noted by
-                the example below.
+        full_version = match.group("full_version")
+        version_numbers = full_version.split(".")
+        major = int(version_numbers[0])
 
-        Examples:
-        Oracle Linux Server release 7.8      <- None
-        CentOS Linux release 7.6.1810 (Core) <- Core
-        CentOS Linux release 8.1.1911 (Core) <- Core
+        # We assume that distributions separate major and minor versions with a dot. If the full_version split by a dot
+        # has at least two items - we have a major and minor versions.
 
-        :returns: The distribution id from the system release file if any.
-        :rtype: str | None
-        """
-        content = self.system_release_file_content if not system_release_content else system_release_content
-        match = re.search(r"(?<=\()[^)]*(?=\))", content)
-        if not match:
-            return None
+        if len(version_numbers) > 1:
+            minor = int(version_numbers[1])
+        else:
+            # In case there are no minor versions specified in the release string, we assume that we are using CentOS
+            # Stream or a similar distribution, which is continuosly going slightly ahead of the latest released minor
+            # version of RHEL up until the end of its lifecycle. Thus its "ephemeral" minor version is always higher
+            # than RHEL X.0-X.9.
 
-        return match.group()
+            # The Stream lifecycle stops with the RHEL X.10 minor release, when the Stream goes EOL and RHEL catches
+            # up with it. After that the Stream-like system can be converted to RHEL X.10 without a downgrade.
+
+            # Therefore to enable the simplest conversion from CentOS Stream at its EOL date, we hardcode the
+            # CentOS Stream minor version to 10.
+
+            minor = 10
+
+        version = Version(major, minor)
+
+        return {
+            "name": name,
+            "id": system_id,
+            "version": version,
+            "distribution_id": distribution_id,
+            "full_version": full_version,
+        }
 
     def _get_architecture(self):
         arch, _ = utils.run_subprocess(["uname", "-i"], print_output=False)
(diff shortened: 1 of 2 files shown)


convert2rhel/systeminfo.py
```diff
@@ -132,51 +141,51 @@ def get_system_release_file_content():
...
+        """Parse the content of the system release string
 
-    def _get_system_version(self, system_release_content=None):
-        """Return a namedtuple with major and minor elements, both of an int type.
+        If system_release_content is not provided we use the content from the self.system_release_file_content.
 
-        Examples:
-        Oracle Linux Server release 7.8
-        CentOS Linux release 7.6.1810 (Core)
-        CentOS Linux release 8.1.1911 (Core)
```
abadger, line 146: I think we should keep the examples near the regex (perhaps as a comment above it instead of in the docstring).  regexes are hard to maintain and being able to look at the various cases that the regex has been designed to solve while reading it and attempting to fix or modify it is invaluable.

convert2rhel/systeminfo.py
```diff
@@ -132,51 +141,51 @@ def get_system_release_file_content():
...
+        :param system_release_content: The contents of the system_release file if needed.
```
abadger, line 149:
```suggestion
        :keyword system_release_content: The contents of the system_release file if needed.
```

abadger (commented):
Nice restructuring.  Much appreciated.  Just had a few comments on the Draft.

convert2rhel/systeminfo.py
bookwar: Good point. I added a more detailed explanation of the regexp. By pure luck I still remembered some of the details :)

I hope it is more helpful now

convert2rhel/systeminfo.py
SpyTec: To me it seems like `param` is the correct here

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -346,19 +259,26 @@ def test_corresponds_to_rhel_eus_release_eus_override(major, minor, expected, mo
 
 
 @pytest.mark.parametrize(
-    ("system_release_content", "expected"),
+    ("system_release_content,key,value"),
```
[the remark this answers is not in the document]
r0x0d: +1 to that. I think all tests have this syntax

convert2rhel/systeminfo.py
```diff
@@ -152,49 +155,74 @@ def get_system_release_file_content():
...
-        match = re.search(r"(?<=\()[^)]*(?=\))", content)
-        if not match:
-            return None
+        full_version = match.group("full_version")
+        version_numbers = full_version.split(".")
+        major = int(version_numbers[0])
+        if len(version_numbers) > 1:
+            minor = int(version_numbers[1])
+        else:
+            minor = 10  # Stream 8 EOL version maps to 8.10.
```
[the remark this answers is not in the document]
bookwar: In case of CentOS Stream we do not have version_numbers[1] at all. We could put "None", but that would lead to cascading changes further in the code and I wanted to keep the change compatible with the current workflow.

10 is the "largest RHEL minor version ever released". By nature of the CentOS Stream, it runs ahead of the current RHEL minor release for its entire lifecycle, so CentOS Stream is always newer than 8.4 or 8.5 or 8.8. Fully updated CentOS Stream 8 must have higher version number than any released RHEL 8 version, all the way until RHEL 8.10. At RHEL 8.10 release RHEL catches up while CentOS Stream freezes, and that freezed state is .10.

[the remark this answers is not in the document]
SpyTec: What about all RHEL-likes though? Alma, Rocky etc. I'm just worried that this is cut short by only thinking about this specific scenario of CentOS Stream conversions

Is it guaranteed to only be set to 10 when we cannot find the minor version?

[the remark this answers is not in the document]
hosekadam: Or I would add better comment describing what's going on there. And in case it would cause issues in the feature, it will be at least easy to find where is the problem

convert2rhel/systeminfo.py
```diff
@@ -152,49 +155,74 @@ def get_system_release_file_content():
...
-        Oracle Linux Server release 7.8
-        CentOS Linux release 7.6.1810 (Core)
-        CentOS Linux release 8.1.1911 (Core)
+        :param system_release_content: The contents of the system_release file if needed.
+        :type system_release_content: str
+        :returns: A dictionary containing the system release information
+        :rtype: dict[str, str]
         """
+
         content = self.system_release_file_content if not system_release_content else system_release_content
```
hosekadam, line 169: This is mainly used for easier unit testing, right? I was wondering if it's needed, but I see this usecase

convert2rhel/systeminfo.py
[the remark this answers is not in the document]
bookwar: I have updated the PR with a more detailed comment.

convert2rhel/systeminfo.py
bookwar: Yes, I think it is not actively used, but I wanted to keep it, as it simplifies testing and generally allows to tweak the perception of a system-release without editing the file on the filesystem.

Could be useful in scenarios when you want to keep the system read-only.

convert2rhel/systeminfo.py
```diff
@@ -152,49 +155,89 @@ def get_system_release_file_content():
...
+            #
+            #   CentOS Stream release 8
+            #   <    name   > <      ><full_version>
+            #
+            #   CentOS Linux release 8.1.1911        (Core     )
+            #   <    name  > <      ><full_version><  <dist_id>>
+            #   CentOS Linux release 8.1.1911      Beta       (Core     )
+            #   <    name  > <      ><full_version><    ><  <dist_id>>
+            r"^(?P<name>.+?)\s(?:release\s)?(?P<full_version>[.\d]+)(?:\sBeta)?(\s\((?P<dist_id>.+)\))?$",
+            content,
```
SpyTec, line 197: I get this warning locally

image

convert2rhel/systeminfo.py
```diff
@@ -152,49 +155,89 @@ def get_system_release_file_content():
...
+        )
+
         if not match:
-            self.logger.critical("Couldn't get system version from the content string: %s" % content)
-        version = Version(int(match.group(1)), int(match.group(2)))
+            self.logger.critical_no_exit("Couldn't parse the system release content string: %s" % content)
```
r0x0d, line 201: Why do we changed this to no_exit? It seems to me very critical to inform that the system release file is not correct. 

Also, I think we will raise an exception while resolving system info, right? Is it fine if we just bubble up the exception?

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -346,19 +259,26 @@ def test_corresponds_to_rhel_eus_release_eus_override(major, minor, expected, mo
...
+        ("Oracle Linux Server release 7.8", "distribution_id", None),
+        ("CentOS Stream release 8", "id", "centos"),
+        ("CentOS Linux release 8.1.1911 (Core)", "version", Version(8, 1)),
+        ("CentOS Stream release 8", "version", Version(8, 10)),
+        ("Red Hat Enterprise Linux release 8.10 Beta (Ootpa)", "version", Version(8, 10)),
     ),
 )
-def test_get_system_distribution_id(system_release_content, expected):
-    assert system_info._get_system_distribution_id(system_release_content) == expected
+def test_parse_system_release_content_from_string(system_release_content, key, value):
```
r0x0d, line 273: I think we should, at least, have more parameters here to validate the bad scenario for the other parts, for example, what happens if there is no `version` and `name` in the content string?

r0x0d: It is little likely that this scenario would happen on a real system, and if that happens, it is probably better to just tell the user that their system is beyond broken, but still, I think that it would be good to catch that scenario, at least so if we identify we can give the user a good response like "hey, your system is beyond broken because it doesn't have X,Y or Z. Please fix it"

convert2rhel/systeminfo.py
r0x0d:
```
[2024-04-05T13:49:33+0000] TASK - [Prepare: Gather system information] *******************************
CRITICAL - Couldn't parse the system release content string: CentOS Linux release 

Writing breadcrumbs to '/etc/migration-results'.
[2024-04-05T13:49:33+0000] DEBUG - No RHSM facts folder found at '/etc/rhsm/facts'. Creating a new one...
Writing RHSM custom facts to '/etc/rhsm/facts/convert2rhel.facts'.
[2024-04-05T13:49:33+0000] DEBUG - Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 108, in main_locked
    gather_system_info()
  File "/usr/lib/python3.6/site-packages/convert2rhel/main.py", line 263, in gather_system_info
    systeminfo.system_info.resolve_system_info()
  File "/usr/lib/python3.6/site-packages/convert2rhel/systeminfo.py", line 122, in resolve_system_info
    self.name = system_release_data["name"]
KeyError: 'name'

No changes were made to the system.
[2024-04-05T13:49:33+0000] DEBUG - /var/run/lock/convert2rhel.pid PID 34285 unlocked.
make: *** [Makefile:25: analyze] Error 1
```

bookwar: My intent is that parse_.. function is a simple utility function. It should not make decisions on should we stop the execution or not. If parses what it can find, and  then the higher-level structure may decide what to do with this information. In this case it still fails, and that's expected.

It maybe eventually we would be overriding these parameters from a config file or whatever.

I could have raise a custom exception here instead of passing None as a reply. But it seems like an overkill.

convert2rhel/unit_tests/systeminfo_test.py
bookwar: I have added a separate `test_fail_to_parse_system_release_content_from_string()` test below for a failure scenario. Is it what you are referring to? Or you expect something else?

We are using a single regular expression to match the entire string, so I can not easily tell the user which part of the string is missing. I can say no matches found. And that's what I do in that test.

convert2rhel/systeminfo.py
bookwar: Can you check now? I think it was caused by a naming collision so I simply changed `match` to `matched` for a variable name to not overlap with `re.match()`

SpyTec: Seems like it's just the type hints from pylance. Since I don't know how to annotate in comments lets just ignore 

```suggestion
            # We assume that the /etc/system-release content follows the pattern:
            # "<name> release <full_version> <Beta> (<dist_id>)"
            # Here
            # - <name>, parsed as a named group '(?P<name>.+?)',
            #   is a non-empty string of arbitrary symbols.
            # - "release ", parsed as (?:release\s)?,
            #   is a literal string. It is optional, and it is dropped when parsed, so this group doesn't have a name.
            # - <full_version>, parsed as a named group '(?P<full_version>[.\d]+)',
            #   is a string of arbitrary length containing numbers and dots, for example 8.1.1911 or 7.9.
            #   (For now we do not have examples with version strings containing letters or other symbols)
            # - <Beta>, just the word Beta which may appear in some cases, parsed as '(?:\sBeta)?'.
            # - <dist_id>, parsed as a named group (?P<dist_id>.+), is optional and when it is present, it must appear
            #   in brackets. Thus, the named group is nested under an unnamed group which starts from a space and then
            #   has brackets '(\s\( ...here goes the nested group... \))?'
            #
            # Example:
            #
            #   CentOS Stream release 8
            #   <    name   > <      ><full_version>
            #
            #   CentOS Linux release 8.1.1911        (Core     )
            #   <    name  > <      ><full_version><  <dist_id>>
            #   CentOS Linux release 8.1.1911      Beta       (Core     )
            #   <    name  > <      ><full_version><    ><  <dist_id>>
            r"^(?P<name>.+?)\s(?:release\s)?(?P<full_version>[.\d]+)(?:\sBeta)?(\s\((?P<dist_id>.+)\))?$",
            content,  # type: ignore
```

convert2rhel/systeminfo.py
r0x0d: Alright, got the idea behind... And I agree that `parse_*` functions should all have that same intention, but at the same time, I feel that we are making some type of "regression". Right now, when this fails, it will show to the user a red message with the exact error, making it less scary if something goes wrong. 

With this change, we are going away from a message that seems informative about the error, and just raising an base exception that, sometimes, may not be useful to the user. 

What if we handle this type of exception in a higher level function, just so we can maintain the same UX for the user, without them noticing that we changed how we parse the content string?

convert2rhel/unit_tests/systeminfo_test.py
r0x0d: Oh, now I see. I think I have missed that test (or you introduced it recently), but that sounds good to me. I will resolve this comment.

r0x0d (commented):
Looks good to me. 

Only one comment about critical vs critical_no_exit, but other than that, couldn't find anything that would block the PR for me.

SpyTec (approved):
Integration test failures seem unrelated, merging

PR #883, opened 2023-08-12
[RHELC-1102] Enable el9 conversions

SpyTec:
This is the PR for enabling el9 conversions. This is done through the el9-conversions branch, so if you have anything you want to fix, open a PR to that branch.

Until we get confirmation that everything works fine we will keep this in draft.

Fixes #867, #498.
Follows up on #860.

Jira Issues: [RHELC-1102](https://issues.redhat.com/browse/RHELC-1102)

Checklist

- [ ] PR has been tested manually in a VM (either author or reviewer)
- [x] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [ ] GitHub label has been added to help with Release notes 
- [ ] PR title explains the change from the user's point of view
- [ ] Code and tests are documented properly
- [ ] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending` if relevant

PR #898, opened 2023-09-01
[RHELC-1102] Add CentOS 9 unit tests

bookwar:
- Adds CentOS 9 container for unit tests
  The centos9.requirements.txt currently only has pytest. It seems it is enough for existing unit tests.
- Adds removal of .build-image* files to the 'make clean' target.

This pull-request overlaps with https://github.com/oamg/convert2rhel/pull/883
But I think it s better to split small chinks out of that big change. And it seems that unit tests for centos 9 can be enabled independently from the rest of the change.

Jira Issues: [RHELC-1102](https://issues.redhat.com/browse/RHELC-1102)

Checklist


SpyTec (approved):

Containerfiles/centos9.Containerfile
```diff
@@ -0,0 +1,32 @@
+FROM centos:stream9 as base
+
```
SpyTec, line 2: Actually, I don't think we should use CentOS Stream here. It might be better to use Oracle, Alma, or Rocky Linux 9

SpyTec (changes requested):

SpyTec:
Closing as we have it duplicated in the el9-conversions PR. Looks good though 🎉

SpyTec:
After we talked about it we are going to try splitting into small PRs to make things easier; Reopening the PR @bookwar and will leave some comments in general (and we can move from things from the PR here)

Containerfiles/centos9.Containerfile
SpyTec: My suggestion would be (now that Alma and Rocky conversions are enabled for the product offering) that we use Alma Linux 9 as base

Here is the one I setup for Oracle before
https://github.com/oamg/convert2rhel/blob/el9-conversions/Containerfiles/rpmbuild.oracle9.Containerfile

bookwar: Not sure about the choice here.

I can see how for integration or e2e testing we should better support the systems which are the targets for the conversion. But in case of unit testing - the important part here is not the OS but rather the version of the Python library and the base dependencies.

As these unit tests run at the development phase, and do not block the delivery of the release, here we would benefit from looking slightly ahead and more continuously integrating with CentOS Stream than to pin to certain snapshot of the already released system.

SpyTec: True. Sounds like a good idea

We treated it as an early indication of the OS itself doing something unusual and just easier align with the systems we're converting. But this was established before integration tests were developed

SpyTec (approved):

PR #905, opened 2023-09-07
Fix sub-man installation on Alma and Rocky 8

bocekm:
If the python3-syspurpose package is installed on the system prior to the conversion, it has a different NEVRA than subscription-manager requires. The subscription-manager is strict about it - the syspurpose VRA (version, release, arch) needs to match exactly the sub-man VRA. And because we're downloading the sub-man packages from UBI, the VRA doesn't match.

Previously we thought this issue affects just CentOS Linux 8 but it does affect Alma and Rocky 8 as well.

Follows-up on #869.

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -769,9 +769,8 @@ def _dependencies_to_update(pkg_list):
     if not pkg_list:
         return []
 
-    # Only apply this kludge on centos-8.  We assume that all other vendors
-    # will have dependencies of the needed version in their repositories.
-    if not (system_info.id == "centos" and system_info.version.major == 8):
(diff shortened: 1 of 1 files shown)


abadger (approved):
Works for me.

The only drawback if this captures too many distros is that on rollback, the system could have the python-syspurpose and json-c packages from UBI instead of (an older version) from the original vendor.

bocekm:
I've manually verified it works on Alma 8.8 and OL 8.latest (8.8).
Alma 8.8 had python3-syspurpose and json-c pre-installed and both got downloaded and updated together with installation of sub-man.
OL 8.latest had json-c pre-installed and it got downloaded and updated together with installation of sub-man.

PR #918, opened 2023-09-15
Unittest to pytest

abadger:
This PR ports most of the remaining unit tests which use the unittest framework to use pytest instead.

The only ones remaining are in pkghandler.py.  They used a subclass of unittest.TestCase which I didn't see when I first evaluated this. I'll look into updating those in a separate PR as there is a lot of them.

Checklist


convert2rhel/unit_tests/actions/system_checks/efi_test.py
```diff
@@ -161,21 +174,31 @@ def test_check_efi_efi_detected_nofile_entry(self):
                         "The current UEFI bootloader '0002' is not referring to any binary UEFI"
                         " file located on local EFI System Partition (ESP)."
                     ),
-                    remediation=None,
+                    remediation="",
```
SpyTec, line 177: Did this pass before? 🤔

convert2rhel/unit_tests/actions/system_checks/readonly_mounts_test.py
```diff
@@ -17,109 +17,123 @@
...
-
-    @unit_tests.mock(readonly_mounts, "logger", GetLoggerMocked())
-    @unit_tests.mock(
+    readonly_mounts_mnt.run()
+
+    unit_tests.assert_actions_result(
+        readonly_mounts_mnt,
+        level="SUCCESS",
+    )
+    assert "Read-only /mnt mount point not detected." in caplog.text
```
SpyTec, line 60: I wonder if we can assert something else than debug

SpyTec (approved):
All tests methods changed seems to be in pytest. Assertions do the same thing too just in different format. LGTM

convert2rhel/unit_tests/actions/system_checks/efi_test.py
abadger: Yeah, it was.  I looked into it and the reason is that we're using a real `actions.ActionMessage` here.  That class makes sure to turn `None` values for parameters into empty string since it is later used in the json data which we will pass on to the Tasks App.

convert2rhel/unit_tests/actions/system_checks/readonly_mounts_test.py
abadger: This is actually an info message :-).

We probably can rely on the `unit_tests.assert_actions_result()` rather than checking for the logging output at all, though.  The result is equivalent to the return value of a function and return values are normally what you would assert in unittests.  Depending on how much information we pass to `assert_action_result` (Just passing level nad id, for instance) it would also fix the annoyance of having to fix these unittests anytime the log message strings are reworded.

Since we're using logging to display information to the end user, though, I am not sure if we want to go that route or not (so it's probably better to talk about it in cabal, make a decision, and then implement it in one or more other PRs).

abadger:
@danmyway @kokesak I'll wait for you to confirm that since this only contains changes to unittests, that it's okay to merge now and then I'll merge it.

kokesak:
> @danmyway @kokesak I'll wait for you to confirm that since this only contains changes to unittests, that it's okay to merge now and then I'll merge it.

Since this is only unit test changes I'm OK with merging this one

abadger:
Thanks @kokesak and @SpyTec !

PR #919, opened 2023-09-15
[RHELC-179] Port pkghandler from unittest to pytest.

abadger:
* Port pkghandler_test from the unittest library to pytest.
* Remove ExtendedTestCase from unit_tests.__init__ now that it is no longer used.
* Write several Mocked functions that make use of mock.Mock.  (Using mock.Mock everywhere makes for a nice standard for the parameters on the mock functions.  For instance, our home-grown MockFunction uses `called` to record how many times it has been called which Mock uses `call_count`.  Some of our own mock functions do not record the call count at all.
* Change some old test functions that ran the tested function multiple times to use pytest.mark.parametrize()

It's intended that there be no changes to convert2rhel runtime code, only to unittests so that it doesn't cause any more work for QE or introduce potential bugs into the 1.5 and 1.6 releases.

This is a partial fix for several issues:

* Document use of pytest will be made easier once everything is ported to pytest as people won't cargo cult and unittest code: https://issues.redhat.com/browse/RHELC-410
* Together with https://github.com/oamg/convert2rhel/pull/918 This will complete the migrate to pytest Epic: https://issues.redhat.com/browse/RHELC-179

Jira Issues: [RHELC-179](https://issues.redhat.com/browse/RHELC-179)

Checklist

### DIFF
--- a/convert2rhel/unit_tests/pkghandler_test.py
+++ b/convert2rhel/unit_tests/pkghandler_test.py
@@ -18,10 +18,8 @@ __metaclass__ = type
 
 
 import glob
-import logging
 import os
 import re
-import sys
 
 from collections import namedtuple
 
@@ -40,7 +38,11 @@ from convert2rhel.pkghandler import (
 from convert2rhel.systeminfo import Version, system_info
 from convert2rhel.toolopts import tool_opts
 from convert2rhel.unit_tests import (
-    GetLoggerMocked,
+    CallYumCmdMocked,
+    DownloadPkgMocked,
+    RemovePkgsMocked,
+    RunSubprocessMocked,
+    SysExitCallableObject,
     TestPkgObj,
     create_pkg_information,
     create_pkg_obj,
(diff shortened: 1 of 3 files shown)


convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -421,487 +358,343 @@ def test_call_yum_cmd_with_repo_overrides(self):
...
-        self.assertTrue(len(pkghandler.logging.getLogger.info_msgs), 1)
-        self.assertTrue(len(pkghandler.logging.getLogger.warning_msgs), 1)
-        self.assertIn(
-            "Detected leftover boot kernel, changing to RHEL kernel", pkghandler.logging.getLogger.warning_msgs[0]
+
+        warning_msgs = [r for r in caplog.records if r.levelname == "WARNING"]
+        assert warning_msgs
+        assert "Detected leftover boot kernel, changing to RHEL kernel" in warning_msgs[-1].message
+
+        (filename, content), dummy_ = utils.store_content_to_file.call_args
```
abadger, line 667:
```suggestion
        (filename, content), _dummy = utils.store_content_to_file.call_args
```

abadger:
I have a better idea about how to implement the Mocked functions.  Turning into a draft while I implement that.

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -421,487 +358,343 @@ def test_call_yum_cmd_with_repo_overrides(self):
...
-    @unit_tests.mock(pkghandler.logging, "getLogger", GetLoggerMocked())
-    def test_fix_invalid_grub2_entries_not_applicable(self):
         pkghandler.fix_invalid_grub2_entries()
-        self.assertFalse(len(pkghandler.logging.getLogger.info_msgs), 1)
 
-        system_info.version = Version(8, 0)
-        system_info.arch = "s390x"
-        pkghandler.fix_invalid_grub2_entries()
-        self.assertFalse(len(pkghandler.logging.getLogger.info_msgs), 1)
+        assert not [r for r in caplog.records if r.levelname != "DEBUG"]
```
SpyTec, line 591: Not sure I follow this fully but it is also 3 am so

SpyTec (approved):
LGTM, didn't spot any unit tests from deleted lines that didn't port over to pytest and didn't see any new unit tests. Some minor cleanup here and there

convert2rhel/unit_tests/pkghandler_test.py
abadger: The new version is slightly stricter.  Although, neither one is very good.  The test is checking that if the platform never needs to try to fix up the grub records, then the `fix_invalid_grub2_entries()` function will exit before it has a chance to do anything.  In both cases, it does that by looking to see if log messages that the function would normally output were emitted.  The old test only looked for an info message (the first statement after the early exit is: `loggerinst.info("Fixing GRUB boot loader entries.")`.)  The new test looks for anything above debug.

If I was changing convert2rhel runtime code in this PR, I would probably have changed `fix_invalid_grub2_entries()` to emit a message when it exits early. Then look for the presence of that rather than looking for the absence of any message.

PR #920, opened 2023-09-18
[RHELC-474] Deduplicate mocking classes

abadger:
Deduplicates many mock classes and ports away from unit_tests.MockFunction to unit_tests.MockFunctionObject.

Note that there are still functions which are using Mocked Functions inside of the specific unit_test files instead of the shared unit_tests.  It doesn't look like many of these are shared, though.

There are also a great many tests using one-off mocks like this:
```
dummy_mock = mock.Mock()
PkgName = namedtuple("PkgNames", ["name"])
transaction_pkgs = [PkgName(package) for package in packages]

monkeypatch.setattr(pkgmanager.Base, "read_all_repos", value=dummy_mock)
monkeypatch.setattr(pkgmanager.Base, "fill_sack", value=dummy_mock)
monkeypatch.setattr(pkgmanager.Base, "upgrade_all", value=dummy_mock)
monkeypatch.setattr(pkgmanager.Base, "resolve", value=dummy_mock)
```
Most of those are low complexity (just a few lines, perhaps just `mock.Mock()` plus a `return_value` so the main benefit to consolidating them would be to use autospec with them (So that the parameters will be checks for too many positional args or kwargs of unknown names).

We can decide whether to leave RHELC-474 open to take care of those or not.

This PR only touches unittests so it should be mergable without adding work to QE.

Jira Issues: [RHELC-474](https://issues.redhat.com/browse/RHELC-474)

Checklist

### DIFF
--- a/convert2rhel/unit_tests/actions/system_checks/convert2rhel_latest_test.py
+++ b/convert2rhel/unit_tests/actions/system_checks/convert2rhel_latest_test.py
@@ -20,14 +20,9 @@ __metaclass__ = type
 import os
 
 import pytest
-import six
 
 from convert2rhel import actions, systeminfo, unit_tests, utils
 from convert2rhel.actions.system_checks import convert2rhel_latest
-
-
-six.add_move(six.MovedModule("mock", "mock", "unittest.mock"))
-from six.moves import mock
 
 
 @pytest.fixture
--- a/convert2rhel/unit_tests/actions/system_checks/convert2rhel_latest_test.py
+++ b/convert2rhel/unit_tests/actions/system_checks/convert2rhel_latest_test.py
@@ -43,7 +43,7 @@ def convert2rhel_latest_version_test(monkeypatch, tmpdir, request, global_system
     marker = request.param
     monkeypatch.setattr(convert2rhel_latest, "installed_convert2rhel_version", marker["local_version"])
 
-    run_subprocess_mocked = mock.Mock(spec=utils.run_subprocess, return_value=(marker["package_version"], 0))
+    run_subprocess_mocked = unit_tests.RunSubprocessMocked(return_string=marker["package_version"])
 
     monkeypatch.setattr(utils, "run_subprocess", run_subprocess_mocked)
     monkeypatch.setattr(global_system_info, "version", systeminfo.Version(marker["pmajor"], 0))
@@ -307,7 +307,9 @@ class TestCheckConvert2rhelLatest:
     def test_c2r_up_to_date_repoquery_error(
         self, caplog, monkeypatch, convert2rhel_latest_action, convert2rhel_latest_version_test
     ):
-        monkeypatch.setattr(utils, "run_subprocess", mock.Mock(return_value=("Repoquery did not run", 1)))
+        monkeypatch.setattr(
+            utils, "run_subprocess", unit_tests.RunSubprocessMocked(return_value=("Repoquery did not run", 1))
+        )
         expected = set(
             (
                 actions.ActionMessage(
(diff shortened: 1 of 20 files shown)


abadger:
Draft right now because unit_tests have to be run on RHEL7 (yum-based, python-2.7 based) and there's likely to be some bugs to be fixed there before it is  ready.

abadger:
Rebased.

convert2rhel/unit_tests/actions/system_checks/convert2rhel_latest_test.py
```diff
@@ -43,7 +43,7 @@ def convert2rhel_latest_version_test(monkeypatch, tmpdir, request, global_system
     marker = request.param
     monkeypatch.setattr(convert2rhel_latest, "installed_convert2rhel_version", marker["local_version"])
 
-    run_subprocess_mocked = mock.Mock(spec=utils.run_subprocess, return_value=(marker["package_version"], 0))
+    run_subprocess_mocked = unit_tests.RunSubprocessMocked(return_string=marker["package_version"])
```
r0x0d, line 46: Commenting here since it's the only line changed. 

`from six.moves import mock` in line 30 is not being used anymore.

convert2rhel/unit_tests/actions/system_checks/rhel_compatible_kernel_test.py
```diff
@@ -254,7 +254,7 @@ def test_bad_kernel_package_signature_success(
     monkeypatch,
     pretend_os,
 ):
-    run_subprocess_mocked = mock.Mock(spec=run_subprocess, return_value=(kernel_pkg, 0))
+    run_subprocess_mocked = RunSubprocessMocked(return_string=kernel_pkg)
```
r0x0d, line 257: Seems that `get_installed_pkg_objects` is not being used

convert2rhel/unit_tests/actions/system_checks/convert2rhel_latest_test.py
abadger: Fixed, thanks

convert2rhel/unit_tests/actions/system_checks/rhel_compatible_kernel_test.py
abadger: Yeah... not sure why it was added in the first place (it was used like this originally: `mock.Mock(spec=get_installed_pkg_objects, [...])`) but it is definitely not needed now.

Removed.

convert2rhel/unit_tests/actions/system_checks/rhel_compatible_kernel_test.py
```diff
@@ -314,7 +314,7 @@ def test_bad_kernel_package_signature_invalid_signature(
...
```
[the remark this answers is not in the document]
abadger: Removed from here as well.

convert2rhel/unit_tests/backup_test.py
```diff
@@ -579,28 +563,28 @@ def test_restore_uninstall(self, run_subprocess_with_empty_rpmdb, rpm_key):
 
         # Check that the beginning of the run_subprocess call starts with the command to remove
         # the key (The arguments our fixture has added to use the empty rpmdb come after that)
-        assert run_subprocess_with_empty_rpmdb.called_with[-1][0][0][0:3] == ["rpm", "-e", "gpg-pubkey-fd431d51"]
+        assert run_subprocess_with_empty_rpmdb.call_args_list[-1][0][0][0:3] == ["rpm", "-e", "gpg-pubkey-fd431d51"]
```
SpyTec, line 566: This before and after is quite confusing with the array fetching, the `[-1][0][0][0:3]` is difficult to parse all at once

But nothing to do as part of this PR

SpyTec (approved):
LGTM

abadger:
Thanks Eric!  And yeah, I agree about call_args_list.  That could be
improved in the future with some intermediate variables.

PR #922, opened 2023-09-18
Fix for race condition when deleting temporary directory.

abadger:
The directory that we store gpg public keys in when checking package signature is subject to a race condition.  Work harder to workaround the race and do not fail if we cannot.

The error (which shows up from time to time in CI) looks like this:
```
_______________ TestFindKeys.test_find_keyid_race_in_gpg_cleanup _______________
[..]
        finally:
            try:
                # Remove the temporary keyring.  We can't use the context manager
                # for this because it isn't available on Python-2.7 (RHEL7)
>               shutil.rmtree(temporary_dir)

convert2rhel/utils.py:872: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <convert2rhel.unit_tests.utils_test.TestFindKeys.MockedRmtree object at 0x7f068529ee10>
args = ('/tmp/tmpu_e6f2nb',), kwargs = {}

    def __call__(self, *args, **kwargs):
        # Fail on the first call
        self.called += 1
        if self.called == 1:
>           raise self.exception
E           FileNotFoundError: [Errno 2] File not found

convert2rhel/unit_tests/utils_test.py:278: FileNotFoundError

During handling of the above exception, another exception occurred:

self = <convert2rhel.unit_tests.utils_test.TestFindKeys object at 0x7f0685f9b828>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f068529ee48>

    def test_find_keyid_race_in_gpg_cleanup(self, monkeypatch):
        real_rmtree = shutil.rmtree
        monkeypatch.setattr(shutil, "rmtree", self.MockedRmtree(OSError(2, "File not found"), real_rmtree))
    
>       assert utils.find_keyid(self.gpg_key) == "fd431d51"

convert2rhel/unit_tests/utils_test.py:293: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
convert2rhel/utils.py:881: in find_keyid
    shutil.rmtree(temporary_dir)
convert2rhel/unit_tests/utils_test.py:280: in __call__
    return self.real_rmtree(*args, **kwargs)
/usr/lib64/python3.6/shutil.py:486: in rmtree
    _rmtree_safe_fd(fd, path, onerror)
/usr/lib64/python3.6/shutil.py:444: in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[..]
            else:
                try:
>                   os.unlink(name, dir_fd=topfd)
E                   FileNotFoundError: [Errno 2] No such file or directory: 'S.gpg-agent.browser'

/usr/lib64/python3.6/shutil.py:442: FileNotFoundError
```

Checklist


convert2rhel/utils.py
```diff
@@ -866,21 +866,33 @@ def find_keyid(keyfile):
...
             else:
-                raise
+                break
+        else:
+            # If we get here, we tried and failed to rmtree five times
+            # Don't make this fatal but do let the user know so they can clean
+            # it up themselves.
+            loggerinst.info(
+                "Failed to remove temporary directory %s that held Red Hat gpg public keys." % temporary_dir
+            )
```
r0x0d, line 895: Few questions that I will wrap all of them in this block to make it easier. 

1. Wouldn't be cleaner to use `tempfile.TemporaryDirectory` instead of trying 5 times? The context manager would take care of cleaning out the directory at the end of the execution too. (Maybe I'm missing something in this implementation, but, just thought to ask this anyway)
2. Do we need to try 5 times? Seen like a huge amount of tries for that... Wondering from where this number came from
3. Wondering about the log info at the `else` part, does leaving behind the temporary directory with the public keys imposes any risk? I guess not, since its `info`, but if it would impose a risk, then I think we should at least change it to warning.. What do you think?

abadger:
\packit test --labels tier0

SpyTec (approved):

PR #903, opened 2023-09-06
Fix colors in the text-based assessment log being reversed from what it should be.

abadger:
convert2rhel.logger uses a boolean to tell whether to turn colors *off* while we were using a boolean to turn colors *on* in the assessment log.  Made it so both turn colors off now.

Jira Issues: [RHELC-](https://issues.redhat.com/browse/RHELC-)

Checklist

- [ ] PR has been tested manually in a VM (either author or reviewer)
- [ ] Jira issue has been made public if possible
- [ ] `[RHELC-]` is part of the PR title 
- [ ] GitHub label has been added to help with Release notes 
- [ ] PR title explains the change from the user's point of view
- [ ] Code and tests are documented properly
- [ ] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending` if relevant

PR #924, opened 2023-09-19
[RHELC-1055] Print pre-conversion report to txt file

hosekadam:
How it works:
- it generates the /var/log/convert2rhel/convert2rhel-pre-conversion.txt every time when the summary is printed
- contains everything except SUCCESS (INFO, WARNING, SKIP, OVERRIDABLE and ERROR are included)
- content is ordered from ERROR to INFO. So the ERROR will be on the top of the file
- in the file is only info from the last run

Example of the file under Jira issue

Jira Issues: [RHELC-1055](https://issues.redhat.com/browse/RHELC-1055)

Depends on https://github.com/oamg/convert2rhel/pull/903

Checklist

### DIFF
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -289,6 +289,6 @@ def summary_as_txt(results):
 
     txt_result = txt_result.strip()
 
-    # We need info from the last run, any old results needed
+    # We need info from the last run, any old results are discarded
     with open(CONVERT2RHEL_TXT_RESULTS, "w") as file:
         file.write(txt_result)
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -35,6 +35,7 @@ logger = logging.getLogger(__name__)
 
 #: The filename to store the results of running preassessment
 CONVERT2RHEL_JSON_RESULTS = "/var/log/convert2rhel/convert2rhel-pre-conversion.json"
+CONVERT2RHEL_TXT_RESULTS = "/var/log/convert2rhel/convert2rhel-pre-conversion.txt"
 
 #: Map Status codes (from convert2rhel.actions.STATUS_CODE) to color name (from
 #: convert2rhel.logger.bcolor)
@@ -118,6 +119,31 @@ def wrap_paragraphs(text, width=70, **kwargs):
     return "\n".join(output)
 
 
+def get_combined_results_and_message(results):
+    combined_results_and_message = {}
+
+    for action_id, action_value in results.items():
+        combined_results_and_message[(action_id, action_value["result"]["id"])] = {
+            "level": action_value["result"]["level"],
+            "title": action_value["result"]["title"],
+            "description": action_value["result"]["description"],
+            "remediation": action_value["result"]["remediation"],
+            "diagnosis": action_value["result"]["diagnosis"],
+            "variables": action_value["result"]["variables"],
+        }
+        for message in action_value["messages"]:
+            combined_results_and_message[(action_id, message["id"])] = {
+                "level": message["level"],
+                "title": message["title"],
+                "description": message["description"],
+                "remediation": message["remediation"],
+                "diagnosis": message["diagnosis"],
+                "variables": message["variables"],
+            }
+
+    return combined_results_and_message
+
+
 def summary(results, include_all_reports=False, disable_colors=False):
     """Output a summary regarding the actions execution.
 
@@ -183,27 +209,9 @@ def summary(results, include_all_reports=False, disable_colors=False):
     :type include_all_reports: bool
     """
     logger.task("Pre-conversion analysis report")
-    combined_results_and_message = {}
     report = []
 
-    for action_id, action_value in results.items():
-        combined_results_and_message[(action_id, action_value["result"]["id"])] = {
-            "level": action_value["result"]["level"],
-            "title": action_value["result"]["title"],
-            "description": action_value["result"]["description"],
-            "remediation": action_value["result"]["remediation"],
-            "diagnosis": action_value["result"]["diagnosis"],
-            "variables": action_value["result"]["variables"],
-        }
-        for message in action_value["messages"]:
-            combined_results_and_message[(action_id, message["id"])] = {
-                "level": message["level"],
-                "title": message["title"],
-                "description": message["description"],
-                "remediation": message["remediation"],
-                "diagnosis": message["diagnosis"],
-                "variables": message["variables"],
-            }
+    combined_results_and_message = get_combined_results_and_message(results)
 
     if include_all_reports:
         combined_results_and_message = combined_results_and_message.items()
@@ -254,3 +262,33 @@ def format_report_section_heading(status_code):
 
     heading = "{highlight} {status_header} {highlight}".format(highlight=highlight, status_header=status_header)
     return heading
+
+
+def summary_as_txt(results):
+    """
+    Print the report to txt file. Used mainly by Satellite.
+    Accepts the data preformatted by summary function.
+
+    There is no special formatting needed, just the output of the checks.
+    The data are sorted from ERROR to INFO, SUCCESS aren't included.
+    """
+    txt_result = ""
+
+    combined_results_and_message = get_combined_results_and_message(results)
+
+    combined_results_and_message = find_actions_of_severity(
+        combined_results_and_message, "INFO", level_for_combined_action_data
+    )
+    combined_results_and_message = sorted(combined_results_and_message, key=lambda item: item[1]["level"], reverse=True)
+
+    for message_id, combined_result in combined_results_and_message:
+        entry = format_action_status_message(combined_result["level"], message_id[0], message_id[1], combined_result)
+        entry = colorize(entry, _STATUS_TO_COLOR[combined_result["level"]])
+        entry += "\n"
+        txt_result += entry
+
+    txt_result = txt_result.strip()
+
+    # We need info from the last run, any old results needed
+    with open(CONVERT2RHEL_TXT_RESULTS, "w") as file:
+        file.write(txt_result)
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -205,6 +205,7 @@ def main_locked(process_phase):
         # parse and act upon it.
         if pre_conversion_results:
             actions.report.summary_as_json(pre_conversion_results)
+            actions.report.summary_as_txt(pre_conversion_results)
 
     return 0
 
--- a/convert2rhel/unit_tests/actions/report_test.py
+++ b/convert2rhel/unit_tests/actions/report_test.py
@@ -716,7 +716,6 @@ def test_messages_summary_with_long_message(long_message, caplog):
     ),
 )
 def test_results_summary_ordering(results, include_all_reports, expected_results, caplog):
-
     report.summary(results, include_all_reports, disable_colors=True)
 
     # Prove that all the messages occurred and in the right order.
@@ -953,7 +952,6 @@ def test_results_summary_ordering(results, include_all_reports, expected_results
     ),
 )
 def test_messages_summary_ordering(results, include_all_reports, expected_results, caplog):
-
     report.summary(results, include_all_reports, disable_colors=True)
 
     # Filter informational messages and empty strings out of message.splitlines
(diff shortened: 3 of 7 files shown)


convert2rhel/actions/report.py
```diff
@@ -205,6 +206,8 @@ def summary(results, include_all_reports=False, with_colors=True):
                 "variables": message["variables"],
             }
 
+    report_to_txt(combined_results_and_message)
+
```
r0x0d, line 210: I think we should call this function in the `finally` block inside main (https://github.com/hosekadam/convert2rhel/blob/c6715e0d267e4ccb4f6156d0df206c65e8c1ddb5/convert2rhel/main.py#L204-L206) instead of here. 

I think that the idea of having this type of fucntion is to serve as a "plugin" (additional formats output to the summary). 

To my knowledge, this function is no different from the `summary_as_json` in report.py module.

hosekadam:
Finally working unit test for this, had problems with order of ansi color codes and new line characters.
```
'\x1b[91m(ERROR) ErrorAction::ERROR - Error\n'\n ' Description: Action error\n'\n ' Diagnosis: User error\n'\n ' Remediation: move on\n'\n '\x1b[0m\n'\n '\x1b[91m(ERROR)

vs

'\x1b[91m(ERROR) ErrorAction::ERROR - Error\n'\n ' Description: Action error\n'\n ' Diagnosis: User error\n'\n ' Remediation: move on\x1b[0m\n'\n '\n'\n '\x1b[91m(ERROR)
```

convert2rhel/actions/report.py
hosekadam: The reason why I'm calling it there is to reuse a maximum of already generated data. But it makes sense to have it in the same place as the `summary_as_json`. I'll prepare some solution for this.

r0x0d: It makes total sense! But the data that `summary` receives is the raw data after the action framework execution is done... So, in theory, is correct that you have the `report_to_txt` here, but I think that the idea behind this function when we first did it was to output the general summary to the user, and not use it as a "entrypoint" for other format "plugins" (As far as I remember, of course). 

Moving it to be next to `summary_as_json` will give the sense that this is something external, something that we are plugging to the output and will generate whatever format we tell the function to output. I would envision that the `summary` function would keep clean and have its own way to output the summary to the user without interference from external sources. 
-----

@abadger, I think we have spoken about this before, but do you think that there is room for the action framework to have a module output system? Something that we can do once and append/add new formats when we need/want. 

Currently, we only have: 
1. the summary that goes to the user
2. summary as json
3. summary as text (for satellite)
4. ???

I don't know that if we will ever have a new format that we need to output to someone else, but if that would be the case, maybe we should think about a strategy to plug those different formats into some sort of system. What do you think?

hosekadam:
@r0x0d I moved the function call, so you can re-review that.

I have question about mocking the `summary_as_json`. Why it is under `# Mock the rollback calls` and not in the first group? Example: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/unit_tests/main_test.py#L281

convert2rhel/unit_tests/actions/report_test.py
```diff
@@ -1100,6 +1098,358 @@ def test_messages_summary_ordering(results, include_all_reports, expected_result
...
+def test_report_to_txt(results, text_lines, file_exists, tmpdir, monkeypatch):
+    convert2rhel_txt_results = tmpdir.join("convert2rhel-pre-conversion.txt")
+    monkeypatch.setattr(report, "CONVERT2RHEL_TXT_RESULTS", str(convert2rhel_txt_results))
+
+    convert2rhel_txt_results = tmpdir.join("convert2rhel-pre-conversion.txt")
+
+    if file_exists:
+        convert2rhel_txt_results.write("test")
+
+    monkeypatch.setattr(report, "CONVERT2RHEL_TXT_RESULTS", str(convert2rhel_txt_results))
```
r0x0d, line 1326: I may be wrong here, but I think that you don't need to redo that part. Since your function will create the txt file in the end, it might as well be okay when you executed you don't have the file present. The idea, I think, is to test that your function will create the file and will put some contents in it. 

If I may suggest, I would remove this part here, as it is mostly the same thing that is being done above: 

```suggestion
```

And for your assertions, I would make that the file exists and has your content in it. Something like: 

```python
with open(convert2rhel_txt_results, "r") as file: 
    data = file.read()

# assert data == ...
```

r0x0d: Or as I just saw below, you could just reuse `convert2rhel_txt_results.read()` instead of opening a new file handler for that.

convert2rhel/unit_tests/main_test.py
```diff
@@ -253,6 +258,7 @@ def test_main_traceback_before_action_completion(self, monkeypatch, caplog, tmp_
         assert finish_collection_mock.call_count == 1
         assert rollback_changes_mock.call_count == 1
         assert summary_as_json_mock.call_count == 0
+        assert report_to_txt_mock.call_count == 0
         print(caplog.records)
```
r0x0d, line 262: Oh, I see we have a print left here by someone else, could you take care of removing it from here, @hosekadam?

convert2rhel/unit_tests/actions/report_test.py
hosekadam: @r0x0d I see, it's probably copy&paste mistake (I used the first two lines for mocking the path in other tests. Now it's not used since I moved the call from the `summary`). I will rather remove the two lines above your mention:
```
convert2rhel_txt_results = tmpdir.join("convert2rhel-pre-conversion.txt")
monkeypatch.setattr(report, "CONVERT2RHEL_TXT_RESULTS", str(convert2rhel_txt_results))
```
Because there is missing the write to file part and we need that for check, if the file was overwritten. I will also add `assert` to check if the `test` is not present in the file.

r0x0d: You could split the tests and make one that validates that use case... I think it you separate the cases for you testing it will make it more readable as well. 

For example: 
1. The good case where the file was written with the correct contents
2. One that there are no contents
3. File existed and it was overridden
4. etc...

hosekadam: @r0x0d I'm not sure if it makes a big difference in this case (the 2. and 3. I would say are mostly the same, it doesn't matter (from my point of view) if the file exists and have content or just exists - I would say exists and has content is stronger).

So then the only difference is if the content is written or not.
- if not, the file doesn't exists. As far as I know `tmpdir.join("convert2rhel-pre-conversion.txt")` will just create the path, not actually create the file. Covers the 1.
- if I write the content, the file is created and has content. Covers 2. and 3.

r0x0d: That's about right too. Maybe we should split the test in two then? one that covers 1 and the other that covers 2 and 3?

hosekadam: If it makes the test more readable, I'll do that!

But still, the only difference will be in:
```
if file_exists:
        convert2rhel_txt_results.write("test")
```
So I'm still wondering if it isn't good moment for using `pytest.mark.parametrize` and turn off/on this part of code using the control variable.

r0x0d: We could, that's definitely a path to follow, @hosekadam, but I think that putting conditionals inside tests is not as a good case as it is in the code.. 

Usually, the test case we want it to do only what is proposed for it to do, so in that case, parametrize this test case would be to add new variants to the same case and validate that it will work/pass for what the test is being proposed.

If you add the conditional to your test, then you are trying to assert two different behaviors instead of just one. My personal preference would be going to split the test and have them individually tested, even though, it becomes easier to debug and adapt/understand the test in the future if needed.

convert2rhel/main.py
```diff
@@ -203,6 +203,7 @@ def main_locked(process_phase):
         # parse and act upon it.
         if pre_conversion_results:
             actions.report.summary_as_json(pre_conversion_results)
+            actions.report.report_to_txt(pre_conversion_results)
```
r0x0d, line 206: Can we change the name of the function to follow the same as the above? 

```suggestion
            actions.report.summary_as_json(pre_conversion_results)
```

Mainly to keep it consistent with what we have and pass the notion that we are still talking about a summary, but this time, in textual form.

convert2rhel/unit_tests/actions/report_test.py
hosekadam: Thanks for clarification! I don't know why I had in my mind that this is good idea :D It makes totally sense what you are saying.

convert2rhel/actions/report.py
abadger: @r0x0d I think that makes perfect sense.  It probably doesn't have to be very elaborate.

When it was just summary as json and summary that goes to the user, it was awkward to design a common API because the user summary is driven through logging.  I think you'll have an easy time creating a good API that  summary_as_json and summary_as_text will share and then you'll need to decide whether to plug summary that goes to user into that same framework or keep it separate.

convert2rhel/actions/report.py
```diff
@@ -254,3 +262,33 @@ def format_report_section_heading(status_code):
...
+    combined_results_and_message = get_combined_results_and_message(results)
+
+    combined_results_and_message = find_actions_of_severity(
+        combined_results_and_message, "INFO", level_for_combined_action_data
+    )
+    combined_results_and_message = sorted(combined_results_and_message, key=lambda item: item[1]["level"], reverse=True)
+
+    for message_id, combined_result in combined_results_and_message:
+        entry = format_action_status_message(combined_result["level"], message_id[0], message_id[1], combined_result)
+        entry = colorize(entry, _STATUS_TO_COLOR[combined_result["level"]])
```
abadger, line 286: Note: IIRC, the acceptance criteria for the text file for satellite says that it should never be colorized.

r0x0d: I asked the same to Adam earlier this week and he told me that the folks asked for the ASCII codes in the text file

SpyTec:
Needs a rebase after #903 was merged

convert2rhel/actions/report.py
```diff
@@ -254,3 +262,33 @@ def format_report_section_heading(status_code):
...
+        entry += "\n"
+        txt_result += entry
```
r0x0d, line 288: Another way of doing this, without concat in two lines, would be to do:

```suggestion
        txt_result += "\n%s" % entry
```

This way, we don't override the entry and only append to the txt_result variable. Not saying that this is required, but just one tip of how to do the same thing in a different way.

hosekadam:
@SpyTec Rebased

SpyTec (approved):
LGTM

convert2rhel/unit_tests/actions/report_test.py
```diff
@@ -1103,3 +1101,371 @@ def test_summary_colors(results, expected_result, expected_message, caplog):
...
+    ),
+)
+def test_summary_as_txt_file_exists(results, text_lines, tmpdir, monkeypatch):
+    convert2rhel_txt_results = tmpdir.join("convert2rhel-pre-conversion.txt")
+    convert2rhel_txt_results.write("test")
+
+    monkeypatch.setattr(report, "CONVERT2RHEL_TXT_RESULTS", str(convert2rhel_txt_results))
+
+    # Call the summary instead of report_to_txt since report to text file is written during the
+    # summary. Also part of the summary code is used for the report to text file.
```
r0x0d, line 1344:
```suggestion
```

convert2rhel/unit_tests/actions/report_test.py
```diff
@@ -1103,3 +1101,371 @@ def test_summary_colors(results, expected_result, expected_message, caplog):
...
+        ),
+    ),
+)
+def test_summary_as_txt(results, text_lines, tmpdir, monkeypatch):
+    convert2rhel_txt_results = tmpdir.join("convert2rhel-pre-conversion.txt")
+
+    monkeypatch.setattr(report, "CONVERT2RHEL_TXT_RESULTS", str(convert2rhel_txt_results))
+
+    # Call the summary instead of report_to_txt since report to text file is written during the
+    # summary. Also part of the summary code is used for the report to text file.
```
r0x0d, line 1219:
```suggestion
```

r0x0d (commented):
Only a few minor comments

tests/integration/tier0/non-destructive/assessment-report/test_assessment_report.py
```diff
@@ -15,8 +19,17 @@ def _validate_report():
     Helper function.
     Verify the report is created in /var/log/convert2rhel/convert2rhel-pre-conversion.json,
     and it corresponds to its respective schema.
+    Additionally verify that the report is created as a .txt file.
     """
-    report_data_json = _load_json_schema(PRE_CONVERSION_REPORT)
+    # Validate the txt variant of the report exists
+    os.path.exists(PRE_CONVERSION_REPORT_TXT)
```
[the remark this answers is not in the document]
danmyway: Oh yeah, that's unfortunate :sweat_smile: 
Will fix, good catch @r0x0d !

PR #932, opened 2023-09-26
[RHELC-1166] Playbook to install convert2rhel from cdn-public.redhat.com

bocekm:
Convert2RHEL has been made available under a new domain that does not require downloading the Red Hat's self-signed SSL cert - cdn-public.redhat.com.

Our public repofiles have been updated accordingly: https://ftp.redhat.com/redhat/convert2rhel/<major ver>/convert2rhel.repo

This update makes use of the updated repofiles and stops downloading the self-signed SSL cert that is no longer needed.

Also, the source distribution validation is updated to reflect the fact that convert2rhel supports converting Alma and Rocky.

Jira Issues: [RHELC-1166](https://issues.redhat.com/browse/RHELC-1166)

Checklist

### DIFF
--- a/ansible/setup-convert2rhel.yml
+++ b/ansible/setup-convert2rhel.yml
@@ -6,8 +6,6 @@
   # to override them if needed.
   vars:
     skip_os_version_check: false
-    # added for RHELDST-7822
-    SSLCACERT: "https://ftp.redhat.com/redhat/convert2rhel/redhat-uep.pem"
     centos8_repos:
       - CentOS-Linux-AppStream.repo
       - CentOS-Linux-BaseOS.repo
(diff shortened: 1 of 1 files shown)


ansible/setup-convert2rhel.yml
```diff
@@ -21,10 +19,12 @@
 
   tasks:
 
-    - name: Validate Distribution (CentOS or Oracle Linux) and version
+    - name: Validate source distribution and major version
       fail:
-        msg: "Distribution must be CentOS or Oracle Linux"
-      when: ansible_distribution != 'CentOS' and ansible_distribution != 'OracleLinux'
+        msg: "Distribution must be CentOS/Oracle/Alma/Rocky Linux"
```
r0x0d, line 24: Should we add that they also should be in major version `7` or `8` because of the condition?

r0x0d (commented):
Only one comment, otherwise, look clean and a good change!

ansible/setup-convert2rhel.yml
bocekm: Good point. Fixed.

r0x0d (approved):
Looks great.

PR #935, opened 2023-09-26
[RHELC-1153] Port the restorable file to work with backup controller

hosekadam:
This PR ports the restorable file to work with backup controller. No functionality changed, the old restorable file is still being used and waiting for complete port. This is mainly for https://github.com/oamg/convert2rhel/pull/875 where this part of code is going to be used for now.

Jira Issues: 
- [RHELC-1153](https://issues.redhat.com/browse/RHELC-1153)
- [RHELC-855](https://issues.redhat.com/browse/RHELC-855)

Checklist

### DIFF
--- a/convert2rhel/backup.py
+++ b/convert2rhel/backup.py
@@ -348,6 +348,125 @@ class RestorableRpmKey(RestorableChange):
         super(RestorableRpmKey, self).restore()
 
 
+class NewRestorableFile(RestorableChange):
+    def __init__(self, filepath):
+        super(NewRestorableFile, self).__init__()
+        self.filepath = filepath
+
+    def enable(self):
+        """Save current version of a file"""
+        # Prevent multiple backup
+        if self.enabled:
+            return
+
+        loggerinst.info("Backing up %s." % self.filepath)
+        if os.path.isfile(self.filepath):
+            try:
+                shutil.copy2(self.filepath, BACKUP_DIR)
+                loggerinst.debug("Copied %s to %s." % (self.filepath, BACKUP_DIR))
+
+            except (OSError, IOError) as err:
+                # IOError for py2 and OSError for py3
+                loggerinst.critical("Error(%s): %s" % (err.errno, err.strerror))
+        else:
+            loggerinst.info("Can't find %s.", self.filepath)
+
+        # Set the enabled value
+        super(NewRestorableFile, self).enable()
+
+    def restore(self, rollback=True):
+        """Restore a previously backed up file"""
+        if rollback:
+            loggerinst.task("Rollback: Restore %s from backup" % self.filepath)
+        else:
+            loggerinst.info("Restoring %s from backup" % self.filepath)
+
+        # We do not have backup or not backed up by this
+        if not self.enabled:
+            loggerinst.info("%s hasn't been backed up." % self.filepath)
+            return
+
+        backup_filepath = os.path.join(BACKUP_DIR, os.path.basename(self.filepath))
+
+        if not os.path.isfile(backup_filepath):
+            loggerinst.info("%s hasn't been backed up." % self.filepath)
+            return
+
+        try:
+            shutil.copy2(backup_filepath, self.filepath)
+        except (OSError, IOError) as err:
+            # Do not call 'critical' which would halt the program. We are in
+            # a rollback phase now and we want to rollback as much as possible.
+            # IOError for py2 and OSError for py3
+            loggerinst.warning("Error(%s): %s" % (err.errno, err.strerror))
+            return
+
+        if rollback:
+            loggerinst.info("File %s restored." % self.filepath)
+            super(NewRestorableFile, self).restore()
+        else:
+            loggerinst.debug("File %s restored." % self.filepath)
+            # not setting enabled to false since this is not being rollback
+            # restoring the backed up file for conversion purposes
+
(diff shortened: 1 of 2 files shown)


convert2rhel/backup.py
```diff
@@ -339,6 +339,78 @@
...
+
+        backup_filepath = os.path.join(BACKUP_DIR, os.path.basename(self.filepath))
+        if rollback:
+            loggerinst.task("Rollback: Restore %s from backup" % self.filepath)
+        else:
+            loggerinst.info("Restoring %s from backup" % self.filepath)
+
+        # if not os.path.isfile(backup_filepath):
+
+        #    return
```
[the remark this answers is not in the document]
hosekadam: It's leftover from the legacy one, this check is moved to the beginning of the restore method - check if enabled: https://github.com/oamg/convert2rhel/pull/935/files#diff-8f4622020cde9a253bc82f69bf2b88910bcc5a84943db3c7d3a435becedfd4d5R371

I'll remove this, I left it there for not forgetting to this case. You can check if substitute with enabled is fine.

How it works now:
When we try to backup file which is not present, the enabled is set to `True` to prevent second call of this. Then while restoring, the enabled is set to `True` so the `Error(%s): %s` is printed. 

As I think about it now, it might be better for the user to read `%s hasn't been backed up.` than the error. So I might use this part of code for this

jochapma (commented):
It seems risky to me to just return if the backup_filepath doesn't exist in restore() at line 383; should the info message at least be a warning?

jochapma (commented):
In restore(), maybe you should call the superclass restore method?

hosekadam:
@jochapma about your questions:

1) "should the info message at least be a warning?"
It's this line of code (just for making sure we both are talking about the same): https://github.com/oamg/convert2rhel/pull/935/files#diff-8f4622020cde9a253bc82f69bf2b88910bcc5a84943db3c7d3a435becedfd4d5R381
I used the same part of code as we have now there: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/backup.py#L369

As I'm thinking about it now, it might be better to warn the user during the enabling (backing up) the file, since during the rollback the user can't do anything with it and the warning is useless from my point of view. I'm not sure if exists case if this message would be printed, but I didn't want to get rid of this case.

And as I'm looking into the code https://github.com/oamg/convert2rhel/pull/935/files#diff-8f4622020cde9a253bc82f69bf2b88910bcc5a84943db3c7d3a435becedfd4d5R354 it's mostly fine I would say. If the path doesn't exists we inform the user. If there we reach some exception, the critical is printed.

2) "In restore(), maybe you should call the superclass restore method?"
I'm calling it there: https://github.com/oamg/convert2rhel/pull/935/files#diff-8f4622020cde9a253bc82f69bf2b88910bcc5a84943db3c7d3a435becedfd4d5R399 or you meant something else?

pr-watson (approved):
Code looks good to me! Approved

convert2rhel/backup.py
```diff
@@ -341,6 +341,79 @@ def restore(self):
...
+        else:
+            loggerinst.info("Restoring %s from backup" % self.filepath)
+
+        if not os.path.isfile(backup_filepath):
+            loggerinst.info("%s hasn't been backed up." % self.filepath)
+            return
+
+        try:
+            shutil.copy2(backup_filepath, self.filepath)
+        except (OSError, IOError) as err:
```
SpyTec, line 389: Lots of paths not being tested here

hosekadam: @SpyTec I'm not sure if I understood. Do you mean not tested by unit tests or something else?

SpyTec: Yes that's right, L353 L381 L389 L393-394 L399

hosekadam:
@SpyTec Added the unittests - hope I haven't missed anything. Changed roder of messages in `restore` method.

hosekadam:
Changed message from comment https://github.com/oamg/convert2rhel/pull/875#discussion_r1403531575

jochapma (approved):

convert2rhel/backup.py
```diff
@@ -348,6 +348,125 @@ def restore(self):
...
+
+        loggerinst.info("Backing up %s." % self.filepath)
+        if os.path.isfile(self.filepath):
+            try:
+                shutil.copy2(self.filepath, BACKUP_DIR)
+                loggerinst.debug("Copied %s to %s." % (self.filepath, BACKUP_DIR))
+
+            except (OSError, IOError) as err:
+                # IOError for py2 and OSError for py3
+                loggerinst.critical("Error(%s): %s" % (err.errno, err.strerror))
```
SpyTec, line 370: This doesn't seem to have unit tests covering it. Unless I'm mistaken the Error we currently catch is for `restore()` function

convert2rhel/backup.py
```diff
@@ -348,6 +348,125 @@ def restore(self):
...
+    conversion so should be removed in rollback.
+    """
+
+    def __init__(self, filepath):
+        super(MissingFile, self).__init__()
+        self.filepath = filepath
+
+    def enable(self):
+        if self.enabled:
+            return
```
SpyTec, line 435: We're testing the `restore()` when enabled is off but not `enable()`

SpyTec:
> @SpyTec I would say we can close this PR as it was part of #875 which is already merged and released

Gotcha, makes sense

PR #937, opened 2023-09-26
[RHELC-1037] Adds in a dependency for  six for the auto-gen manpages action

Andrew-ang9:
The GitHub action for generating manpages was missing a dependency for six this PR adds that in so that it can be ran properly

Jira Issues: [RHELC-1037](https://issues.redhat.com/browse/RHELC-1037)

Checklist


jochapma (approved):

r0x0d (commented):
Looks good to me. Just syncing with @Andrew-ang9 to see if we can test this somehow.

.github/workflows/generate_manpage.yml
```diff
@@ -21,6 +21,24 @@ jobs:
...
+      - name: install dependencies rpm libry
+        run: pip install rpm
+
+      - name: install dependencies rpm libry
+        run: pip install rpm
+
+      - name: Install python3-rpm package with dnf
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y python3-rpm
```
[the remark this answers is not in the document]
Andrew-ang9: Yeah I was actually just about to push a new version of this code to tests the python3-rpm package and in that it changes the format so that they are grouped up! :)

.github/workflows/generate_manpage.yml
```diff
@@ -19,25 +19,36 @@ jobs:
...
-
-      - name: install dependencies rpm libry
-        run: pip install rpm
-
-      - name: install dependencies rpm libry
-        run: pip install rpm
+        run: |
+          pip install argparse-manpages
+          pip install six
+          pip install pexpect
```
r0x0d: Small tip: You can group all of them in just one line: 

```suggestion
          pip install argparse-manpages six pexpect
```

.github/workflows/generate_manpage.yml
```diff
@@ -19,7 +19,35 @@ jobs:
           python-version: 3.x
 
       - name: Install dependencies
-        run: pip install argparse-manpage
+        run: |
+          pip install argparse-manpage
+          pip install six
+          pip install pexpect
+          pip install rpm
```
r0x0d, line 26: Try combining all of them into one single command, like: 

```suggestion
          pip install argparse-manpage six pexpect rpm
```

.github/workflows/generate_manpage.yml
```diff
@@ -19,7 +19,35 @@ jobs:
...
+
+      - name: Install python3-rpm package with apt-get
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y python3-rpm
```
r0x0d, line 31:
```suggestion
      - name: Install python3-rpm and python3-dnf package with apt-get
        run: |
          sudo apt-get update
          sudo apt-get install -y python3-rpm python3-dnf
```

The problem that is happening in the CI right now might be fixed with installing python3-dnf together. This is not ideal, but it was the working solution I could find quickly here.

.github/workflows/generate_manpage.yml
```diff
@@ -19,7 +19,35 @@ jobs:
...
+
+      - name: Verify python3-rpm installation
+        run: |
+          if python3 -c "import rpm" &>/dev/null; then
+            echo "python3-rpm is installed successfully."
+          else
+            echo "ERROR: python3-rpm installation failed."
+            dpkg -l | grep python3-rpm
+            exit 1
+          fi
```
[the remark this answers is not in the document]
Andrew-ang9: Thank you for the suggestion here, I just had this in here because I was trying to debug the issue and was going to try and find a solution where it wasn't needed!

.github/workflows/generate_manpage.yml
Andrew-ang9: Thank you for this help with the format on this install command. I didn't know you could set it up like this

.github/workflows/generate_manpage.yml
```diff
@@ -19,7 +19,13 @@ jobs:
           python-version: 3.x
 
       - name: Install dependencies
-        run: pip install argparse-manpage
+        run: |
+          pip install argparse-manpage six pexpect rpm
```
r0x0d, line 23: Okay, so the `rpm` package here is not doing what we think it is doing... We are expecting to install the python3 rpm bindings for python, but in turn, we are actually installing this one here: https://pypi.org/project/rpm/

Which is a package maintained by packit (https://github.com/packit/rpm-shim).

I will say that we can remove this package from pip and rely on python3-rpm binary being installed on the next step. 

```suggestion
          pip install argparse-manpage six pexpect
```

r0x0d: Also, the python3-rpm should take care of installing the desired library we need. 

```bash
[r0x0d@fedora ~]$ rpm -ql python3-rpm
/usr/lib/.build-id
/usr/lib/.build-id/dc
/usr/lib/.build-id/dc/700247b4d4b61b9b0da932787b19546ff6e80e
/usr/lib64/python3.11/site-packages/rpm-4.18.1-py3.11.egg-info
/usr/lib64/python3.11/site-packages/rpm/__init__.py
/usr/lib64/python3.11/site-packages/rpm/__pycache__
/usr/lib64/python3.11/site-packages/rpm/__pycache__/__init__.cpython-311.opt-1.pyc
/usr/lib64/python3.11/site-packages/rpm/__pycache__/__init__.cpython-311.pyc
/usr/lib64/python3.11/site-packages/rpm/__pycache__/transaction.cpython-311.opt-1.pyc
/usr/lib64/python3.11/site-packages/rpm/__pycache__/transaction.cpython-311.pyc
/usr/lib64/python3.11/site-packages/rpm/_rpm.so
/usr/lib64/python3.11/site-packages/rpm/transaction.py
```

Andrew-ang9: Ok, cool thank you! I was just looking into python3-rpm, I will go ahead and take rpm out of the install command!

scripts/manpage_generation.sh
```diff
@@ -6,7 +6,14 @@ MANPAGE_DIR="man"
...
+echo "HERE WE ARE"
+
+/usr/bin/python3.10 -m pip install argparse-manpage six pexpect
+/usr/bin/python3.10 -m pip show --files argparse-manpage
+
+echo "THERE WE GO"
 
 # Generate the manpage using argparse-manpage
-PYTHONPATH=. argparse-manpage --pyfile man/__init__.py --function get_parser --manual-title="General Commands Manual" --description="Automates the conversion of Red Hat Enterprise Linux derivative distributions to Red Hat Enterprise Linux." --project-name "convert2rhel $VER" --prog="convert2rhel" --include man/distribution --include man/synopsis > "$MANPAGE_DIR/convert2rhel.8"
+PYTHONPATH=. /usr/bin/python3.10 /home/runner/.local/bin/argparse-manpage --pyfile man/__init__.py --function get_parser --manual-title="General Commands Manual" --description="Automates the conversion of Red Hat Enterprise Linux derivative distributions to Red Hat Enterprise Linux." --project-name "convert2rhel $VER" --prog="convert2rhel" --include man/distribution --include man/synopsis > "$MANPAGE_DIR/convert2rhel.8"
```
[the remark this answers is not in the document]
Andrew-ang9: OK, I will make sure to change that. I would like to not hardcode the python path so I am going to be looking for some solutions around that!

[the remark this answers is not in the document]
r0x0d: I don't think this is a problem as long as the version doesn't change

scripts/manpage_generation.sh
```diff
@@ -6,7 +6,11 @@ MANPAGE_DIR="man"
 VER=$(grep -oP '^Version:\s+\K\S+' packaging/convert2rhel.spec)
 
 # Generate a file with convert2rhel synopsis for argparse-manpage
-python -c 'from convert2rhel import toolopts; print("[synopsis]\n."+toolopts.CLI.usage())' > man/synopsis
+/usr/bin/python3.10 -c 'import sys;print(sys.path)'
+/usr/bin/python3.10 -c 'from convert2rhel import toolopts; print("[synopsis]\n."+toolopts.CLI.usage())' > man/synopsis
+
+/usr/bin/python3.10 -m pip install argparse-manpage six pexpect
+/usr/bin/python3.10 -m pip show --files argparse-manpage
```
r0x0d, line 13: Also, don't forget to remove those debug lines later! 

```suggestion
/usr/bin/python3.10 -c 'from convert2rhel import toolopts; print("[synopsis]\n."+toolopts.CLI.usage())' > man/synopsis

/usr/bin/python3.10 -m pip install argparse-manpage six pexpect
```

.github/workflows/generate_manpage.yml
```diff
@@ -1,7 +1,7 @@
 name: Generate Manpages
 
 on:
-  push:
+  pull_request:
```
r0x0d, line 4: @SpyTec, do you think it is worth to keep the pull_request event and the `push` one in the same workflow? 

My thought here is more like: "What if we push a new dependency and forget about it and the workflow doesn't work anymore? This way we would see right in the PR that something changed". 

But since we have it to be executed in the main branch after every push, I think it will be more a question of checking it out before releasing rather than looking at it for every PR?

SpyTec: I think we can keep whichever is best, so if pull_request is better we can keep this change

Andrew-ang9: @r0x0d Do you think I should leave it pull_request?

r0x0d: Lets leave as pull_request for now. I think it will be better to look at it constantly rather than when pushing on main-basis. I don't know if anyone keeps an eye in main pipelines.

PR #939, opened 2023-09-26
[RHELC-965,RHELC-1170] Framework to remove UNKNOWN_ERROR from the Assessment.

abadger:
We do not want to show "UNKNOWN_ERROR" in the insights WebUI.  To get away from that we need to return enough information from where the actual Action happened to where

Documentation on how to port is listed here:
  https://github.com/oamg/convert2rhel/wiki/Action-Framework#actions-and-systemexitsysexit

* This implements https://issues.redhat.com/browse/RHELC-1170
* This establishes the framework to fix: https://issues.redhat.com/browse/RHELC-965
   The ticket can be marked done once all the Actions have been ported to use this exception framework.
* This is also a small part of: https://issues.redhat.com/browse/RHELC-1168 (porting the Action framework logger.critical to logger.critical_no_exit().)

Jira Issues: [RHELC-965](https://issues.redhat.com/browse/RHELC-965) [RHELC-1170](https://issues.redhat.com/browse/RHELC-1170)
 
Checklist

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/backup_system.py
+++ b/convert2rhel/actions/pre_ponr_changes/backup_system.py
@@ -17,7 +17,7 @@ __metaclass__ = type
 
 import logging
 
-from convert2rhel import actions, repo
+from convert2rhel import actions, exceptions, repo
 from convert2rhel.redhatrelease import os_release_file, system_release_file
 
 
@@ -39,15 +39,18 @@ class BackupRedhatRelease(actions.Action):
             # https://github.com/oamg/convert2rhel/blob/v1.2/convert2rhel/subscription.py#L189-L200
             system_release_file.backup()
             os_release_file.backup()
-        except SystemExit as e:
-            # TODO(pr-watson): Places where we raise SystemExit and need to be
-            # changed to something more specific.
-            # Raised in module redhatrelease on lines 49 and 60
-            #   - If unable to find the /etc/system-release file,
-            #     SystemExit is raised
+
+        except exceptions.CriticalError as e:
             self.set_result(
-                level="ERROR", id="UNKNOWN_ERROR", title="An unknown error has occurred", description=str(e)
+                level="ERROR",
+                id=e.id,
+                title=e.title,
+                description=e.description,
+                diagnosis=e.diagnosis,
+                remediation=e.remediation,
+                variables=e.variables,
             )
+        return
 
 
 class BackupRepository(actions.Action):
--- a/convert2rhel/actions/pre_ponr_changes/backup_system.py
+++ b/convert2rhel/actions/pre_ponr_changes/backup_system.py
@@ -50,7 +50,6 @@ class BackupRedhatRelease(actions.Action):
                 remediation=e.remediation,
                 variables=e.variables,
             )
-        return
 
 
 class BackupRepository(actions.Action):
--- a/convert2rhel/backup.py
+++ b/convert2rhel/backup.py
@@ -25,7 +25,7 @@ import shutil
 
 import six
 
-from convert2rhel import utils
+from convert2rhel import exceptions, utils
 from convert2rhel.repo import get_hardcoded_repofiles_dir
 from convert2rhel.systeminfo import system_info
 from convert2rhel.utils import BACKUP_DIR, download_pkg, remove_orphan_folders, run_subprocess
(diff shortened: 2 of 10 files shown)


convert2rhel/backup.py
```diff
@@ -352,7 +352,16 @@ def backup(self):
...
-                loggerinst.critical("Error(%s): %s" % (err.errno, err.strerror))
+
+                # Replace this with loggerinst.critical_no_exit() once that is
+                # written.
+                loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
+                raise exceptions.CriticalError(
+                    id_="FAILED_TO_SAVE_FILE_TO_BACKUP_DIR",
+                    title="Failed to copy file to the backup directory.",
+                    description="While convert2rhel tests whether it is likely to succeed, spme files would be changed. To enable rollback in case we detect something that we would fail to do, we backup those files prior to changing them. In this case, we encountered a failure while performing that backup so it is unsafe to continue.",
+                    diagnosis="Failed to backup %s. Errno: %s, Error: %s" % (self.filepath, err.errno, err.strerror),
```
abadger, line 363: @mportman12 Hey Miriam! This is the first of several PRs that we will be opening to add messages that will show up in insights.  If you'd care to review it to make sure we're being clear, please do.  I've also added a note to the jira Epic tracking this work to CC you on the PRs which update the messages (If you have a different procedure you'd like to follow when you review messages destined for insights, we can remove that and teach people about that).

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -39,15 +39,18 @@ def run(self):
...
-                level="ERROR", id="UNKNOWN_ERROR", title="An unknown error has occurred", description=str(e)
+                level="ERROR",
+                id=e.id,
+                title=e.title,
+                description=e.description,
+                diagnosis=e.diagnosis,
+                remediation=e.remediation,
+                variables=e.variables,
             )
+        return
```
r0x0d, line 53: Do we need this empty return here?

convert2rhel/logger.py
```diff
@@ -37,6 +41,11 @@
 LOG_DIR = "/var/log/convert2rhel"
 
 
+class LogLevelCriticalNoExit:
+    level = 50
+    label = "CRITICAL"
```
r0x0d, line 46: Wondering if we should identify this label as something else other than just "CRITICAL" to not confuse the user with this behavior… 

I'm thinking more like: "Oh, I ran convert2rhel in the past, and whenever I saw "CRITICAL", it meant that the execution would terminate, and now it just goes on". 

But I can't think of any special label that would go well alongside the critical name... Any ideas?

convert2rhel/actions/pre_ponr_changes/backup_system.py
abadger: We don't.  I was contemplating whether it made sense to have a return there somewhere since we would need to explicitly return when we encounter the error if we add additional code after the exception block at a later time and it didn't feel right to have a return inside the block, after `set_result()` and not also at the end.

I could remove this `return` and trust that the person who does that will remember to put it in or I could add a second `return` inside of the exception block to reinforce what's happening.

What do you think?

convert2rhel/logger.py
abadger: I originally called it `CRITICAL_NO_EXIT` but later decided that the differences are probably below the level that the user will notice so it would just be confusing to the user.  Whether CRITICAL always exits or not is probably not something the user will notice for two reasons:

* We have always done a bunch of rollback stuff after the critical which obscures that we will rollback later.
* Since adding the Action framework, we perform all potential Actions before we exit due to a critical.  So it is not obvious that the exit is due to the critical rather than some other reported error.

So I think that this and `LogLevelCriticalWithExitt` (when we add it temporarily while we port the rest of the code over), should both report `CRITICAL` to the end user.

Does that make sense?

convert2rhel/actions/pre_ponr_changes/backup_system.py
r0x0d: I think we can remove it for now and leave that for future-someone who will work on that action. 

I don't think we have to take care of that right away.

convert2rhel/logger.py
r0x0d: Since it's a temporary thing, I see that it makes sense. Otherwise, I would say that this would look confusing. Even though the amount of rollback we do is quite high, and the message might get lost at some point, I can see someone looking for a critical message that does not interrupt the conversion. 

Myself would not expect that something critical happened, and we would be able to move forward...

convert2rhel/actions/pre_ponr_changes/backup_system.py
abadger: Now removed!

convert2rhel/backup.py
r0x0d: @abadger, the rest of the messages will be added in this PR or another one?

abadger: The rest of hte messages should be added in other PRs.  The idea is that this one gives us the Exception and logger methods that we'll need and a single logger.critical ported over so that we have an example.  Then other messages can be split up over multiple people and multiple PRs (using Aleksandra's proposal for large changes.)

r0x0d: Gotcha. And they are expected to land in 1.5 or 1.6?

SpyTec:
@abadger some minor rebasing required

convert2rhel/logger.py
SpyTec: Maybe we can log something like "attempting to fix"

convert2rhel/main.py
```diff
@@ -74,8 +73,6 @@ def main():
     the application lock, to do the conversion process.
     """
 
-    process_phase = ConversionPhase.INIT
-
```
SpyTec, line 78: Why do we get rid of the INIT phase?

SpyTec (approved):
Gonna merge

PR #946, opened 2023-09-27
[RHELC-1615] Add integration test coverage for EL9 systems

danmyway:
* include EL9 source systems in the integration test coverage

Jira Issues: [RHELC-1615](https://issues.redhat.com/browse/RHELC-1615)

Checklist


danmyway:
packit test --labels el9

tests/integration/tier0/non-destructive/kernel/test_custom_kernel.py
```diff
@@ -67,37 +57,55 @@ def custom_kernel(shell, hybrid_rocky_image):
...
-        custom_kernel_release = CUSTOM_KERNEL.rsplit("/", 1)[-1].replace(".rpm", "")
-        assert shell("rpm -e %s" % custom_kernel_release).returncode == 0
-
-        original_kernel = os.popen("rpm -q --last kernel | head -1 | cut -d ' ' -f1").read()
-        original_kernel_release = original_kernel.rsplit("/")[-1].replace(".rpm", "").split("-")[-1]
-
+        # This is kind of naive, but we assume the second latest installed kernel is the original one
+        # We use head to filter the first two lines of the output and tail to filter the bottom line
+        original_kernel = os.popen("rpm -q --last kernel | head -2 | tail -1 | cut -d ' ' -f1").read().strip()
+        original_kernel_title = _get_full_kernel_title(shell, kernel=original_kernel.replace("kernel-", ""))
```
kokesak, line 92: Can't we save the original kernel to a variable before the reboot? I mean somewhere before the if statement (if os.environ["TMT_REBOOT_COUNT"] == "0":)? If not, I suggest creating a tmp file on the system that would hold this information

tests/integration/tier1/destructive/proxy-conversion/test_proxy_conversion.py
```diff
@@ -91,36 +92,29 @@ def setup_rhsm(shell):
...
-            ).returncode
-            == 0
-        )
-        # On CentOS 8.5 we need to replace the $releasever in the url to 8.5,
-        # otherwise the dnf will complain with dependency issues.
-        if "centos-8" in SYSTEM_RELEASE_ENV:
-            shell(r"sed -i 's#\$releasever#8.5#' /etc/yum.repos.d/client-tools.repo")
+
+    client_tools_repo = ""
+    if re.match(r"^(centos|oracle)-7", SYSTEM_RELEASE_ENV):
```
kokesak, line 97: maybe we could use the `SystemInformationRelease.version.major == X` here instead of the `SYSTEM_RELEASE_ENV`. But that is just a thought, no need to change it

kokesak (commented):
Great work on this @danmyway , I went over it and added my suggestions to certain parts.

tests/integration/tier0/destructive/isolated-system-conversion/prepare_system.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
 
-from conftest import TEST_VARS, SystemInformationRelease
+from conftest import TEST_VARS, SystemInformationRelease, _get_full_kernel_title, grub_setup_workaround
```
SerCantus, line 6: `_get_full_kernel_title` is a private function and should not be imported.

plans/tier1.fmf
```diff
@@ -189,6 +189,13 @@ adjust+:
                     - changed-grub-file/valid_changes_to_grub_file
 
         /changed_grub_invalid:
+            adjust+:
+                - enabled: false
+                  when: distro == alma-8-latest and boot_method == uefi
+                  because: The changes to the grub file result in a unbootable system.
+                - enabled: false
+                  when: distro == alma-9, rocky-9, oracle-9
+                  because: The changes break the conversion or the bootloader at the moment.
```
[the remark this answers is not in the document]
danmyway: That is intentional, it either breaks the conversion, so it won't finish at all, or the conversion reports as successful, but the system is unbootable after the next reboot.

tests/integration/conftest.py
```diff
@@ -443,12 +449,16 @@ def remove_repositories(shell, backup_directory):
     backup_dir = os.path.join(backup_directory, "repos")
     shell(f"mkdir {backup_dir}")
     # Move all repos to other location, so it is not being used
-    assert shell(f"mv /etc/yum.repos.d/* {backup_dir}").returncode == 0
+    assert shell(f"mv -v /etc/yum.repos.d/* {backup_dir}").returncode == 0
     assert len(os.listdir("/etc/yum.repos.d/")) == 0
+    assert os.path.exists("/etc/yum.repos.d")
```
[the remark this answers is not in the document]
danmyway: yeah, that makes sense, good catch, thanks!

tests/integration/conftest.py
```diff
@@ -443,12 +449,16 @@ def remove_repositories(shell, backup_directory):
...
 
     yield
 
     if "C2R_TESTS_NONDESTRUCTIVE" in os.environ:
+        # Make sure the reposdir is present
+        # It gets removed with the *-release package on EL9 systems
+        shell("mkdir /etc/yum.repos.d")
```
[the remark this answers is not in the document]
danmyway: TBH, I don't really care if it errors out or not, but you're right, it would be nicer if it would not yield error 90 % of the cases :D

tests/integration/conftest.py
```diff
@@ -525,13 +538,10 @@ def pre_registered(shell, request, yum_conf_exclude):
...
 
     if "oracle" in SYSTEM_RELEASE_ENV:
-        # Remove the rhn-client-tools package to make way for subscription-manager installation
-        # given subman is obsoleted by the rhn-client-tools on Oracle
-        shell("yum remove -y rhn-client-tools")
         # Add the client-tools repository for Oracle linux to install subscription-manager from
         _add_client_tools_repo(shell)
 
-    assert shell("yum install -y subscription-manager").returncode == 0
+    assert shell("yum install --setopt=obsoletes=0 -y subscription-manager").returncode == 0
```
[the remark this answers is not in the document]
danmyway: Well, the obsolete flag apparently does not work on Oracle 7 for some reason, so I'll probably move back to removing the rhn-client-tools pkg.

tests/integration/conftest.py
```diff
@@ -689,41 +672,45 @@ def kernel(shell):
...
+        # an older repo. This should only happen when Alma and Rocky has just landed on
+        # a fresh minor version.
+        elif shell("yum downgrade kernel -y").returncode != 0:
+            # Assuming this can only happen with Alma and Rocky we'll try to install an older kernel
+            # from a previous minor version.
+            # For that we need to use the vault url and bump te current minor down one version.
+            major_ver = SystemInformationRelease.version.major
+            minor_ver = SystemInformationRelease.version.minor
+            minor_ver -= 1
+            if minor_ver < 0:
```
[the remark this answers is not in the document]
danmyway: I guess `if minor_ver == 0:` would work as well.
We can't have lower minor version than 0 "naturally" anyway.

tests/integration/conftest.py
```diff
@@ -689,41 +672,45 @@ def kernel(shell):
...
+            if SystemInformationRelease.distribution == "alma":
+                old_repo = f"https://vault.almalinux.org/{releasever}/BaseOS/x86_64/os/"
+            elif SystemInformationRelease.distribution == "rocky":
+                old_repo = f"https://dl.rockylinux.org/vault/rocky/{releasever}/BaseOS/x86_64/os/"
+            # Install the kernel from the url
+            shell(f"yum install kernel -y --repofromurl 'oldrepo,{old_repo}'")
+
+        # Get the oldest kernel VRA
+        # Sort the versions numerically by each item using
+        oldest_kernel = shell("rpm -q kernel | awk -F'kernel-' '{print $2}' | rpmdev-sort | head -1").output.strip()
```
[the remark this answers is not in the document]
danmyway: That's curious.
Seems like it's not that foolproof as I thought.
Anyway in my case its sorted like this
```
# rpm -q kernel | rpmdev-sort
kernel-6.9.10-200.fc40.x86_64
kernel-6.9.11-200.fc40.x86_64
kernel-6.9.12-200.fc40.x86_64
```
So I would get the latest, so I'll stick with `head -1` and hope for the best :D

tests/integration/tier0/non-destructive/basic-sanity-checks/test_basic_sanity_checks.py
```diff
@@ -153,7 +153,7 @@ def test_c2r_version_latest_override_inhibitor(convert2rhel, c2r_version, versio
         c2r.expect("Continue with the system conversion?")
         c2r.sendline("y")
 
-        assert c2r.expect("You are currently running 0.01", timeout=300) == 0
+        assert c2r.expect("You are currently running 0.01.0", timeout=300) == 0
```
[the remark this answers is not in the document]
danmyway: Yes, they have added the micro version some time ago.

tests/integration/tier1/destructive/proxy-conversion/test_proxy_conversion.py
danmyway: That's valid, I'll do that.

tests/integration/conftest.py
```diff
@@ -921,20 +901,16 @@ def backup_directory(shell, request):
...
     """
     # Add the client tools repository to install the subscription-manager from
     # This is mainly for Oracle Linux but does not hurt to do the same for CentOS as well
     _add_client_tools_repo(shell)
-    # Since we're using the yum_conf_exclude fixture, which excludes rhn-client*
-    # by default, we need to remove the pacakge to prevent issues during the yum transaction
-    if SystemInformationRelease.distribution == "oracle":
-        shell("yum remove -y rhn-client*")
-    shell(f"yum install subscription-manager -y")
+    shell(f"yum install --setopt=obsoletes=0 subscription-manager -y")
```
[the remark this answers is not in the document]
danmyway: Makes sense.

tests/integration/tier0/destructive/isolated-system-conversion/prepare_system.py
danmyway: Good catch, I was moving it around and forgot.

tests/integration/tier0/non-destructive/custom-repository/test_custom_repository.py
```diff
@@ -9,30 +9,31 @@ class AssignRepositoryVariables:
...
     """
 
-    repofile_epel7 = "rhel7"
-    repofile_epel8 = "rhel8"
-    repofile_epel8_eus = "rhel8-eus"
-    enable_repo_opt_epel7 = (
+    system_version = SystemInformationRelease.version.major
+
+    repofile_el7 = "rhel7"
+    enable_repo_opt_el7 = (
```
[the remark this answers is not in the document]
danmyway: Yes, it's evaluated only for el8 and el9, but will be used for el10 eventually as well :shrug:

tests/integration/tier0/non-destructive/duplicate-pkgs/main.fmf
```diff
@@ -9,6 +9,7 @@ enabled: false
 
 /duplicate-pkgs:
     link: https://issues.redhat.com/browse/RHELC-1070
+    enabled: false
```
SerCantus, line 12: Why is this test no longer enabled? I don't see any change in the jira link.

SerCantus (commented):
Added a few comments. I see you already checked the private import one, I will just point in the code where I saw it again. 

As far as python code and code refactoring it looks fantastic.

tests/integration/tier0/non-destructive/duplicate-pkgs/main.fmf
danmyway: TL;DR we try to work with the smallest sample when working with enablement and comparisons.
It's kind of awkward, but we have decided previously, that we will try to keep the plan/test enablement conditioning simple.
The comma separated values, can be interpreted as `a or b` whereas `and` cannot be shortened.
So we rather go with `==` instead of `!=` when comparing the distro context and (ideally) the shorter list of affected systems.
Just a line below, there is:
```
adjust+:
    - enabled: true
      when: distro == centos-7, oracle-7
```
Which is shorter and imho more readable, than:
```
adjust+:
    - enabled: true
      when: distro != centos-8 or distro != oracle-8 or distro != oracle-9 or distro != alma-8 or distro != alma-9 or distro != rocky-8 or distro != rocky-9
```
or 
```
- enabled: false
  when: distro == centos-8, oracle-8, oracle-9, alma-8, alma-9, rocky-8, rocky-9
```
Does that make sense?

tests/integration/tier0/non-destructive/kernel/test_custom_kernel.py
```diff
@@ -67,37 +57,55 @@ def custom_kernel(shell, hybrid_rocky_image):
...
+        assert shell(f"grub2-set-default '{original_kernel_title}'").returncode == 0
+        shell("grub2-mkconfig -o /boot/grub2/grub.cfg")
         # Reboot
         shell("tmt-reboot -t 600")
 
+    if os.environ["TMT_REBOOT_COUNT"] == "2":
+        custom_kernel_installed = os.popen("rpm -q --last kernel | head -1 | cut -d ' ' -f1").read().strip()
+        # After the system has the original kernel running, remove the custom kernel
+        kernel_to_remove = custom_kernel_installed.replace("kernel", "kernel*")
+        assert shell(f"yum remove -y {kernel_to_remove}").returncode == 0
```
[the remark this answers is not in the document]
danmyway: Basically, `kernel` package is just a meta-package, so removing it would not remove packages which it installed. So if I would go with just `yum remove -y kernel` it will remove just the kernel package, but I want to remove also the kernel-core, kernel-modules, etc.
```
# sudo dnf remove kernel-6.9.10-200.fc40.x86_64
Dependencies resolved.
============================================================================================================================================================================================================================
 Package                                           Architecture                                      Version                                                       Repository                                          Size
============================================================================================================================================================================================================================
Removing:
 kernel                                            x86_64                                            6.9.10-200.fc40                                               @System                                              0  

Transaction Summary
============================================================================================================================================================================================================================
Remove  1 Package

```
```
# sudo dnf remove kernel*-6.9.10-200.fc40.x86_64
Dependencies resolved.
============================================================================================================================================================================================================================
 Package                                                      Architecture                                   Version                                                  Repository                                       Size
============================================================================================================================================================================================================================
Removing:
 kernel                                                       x86_64                                         6.9.10-200.fc40                                          @System                                           0  
 kernel-core                                                  x86_64                                         6.9.10-200.fc40                                          @System                                          69 M
 kernel-modules                                               x86_64                                         6.9.10-200.fc40                                          @System                                          61 M
 kernel-modules-core                                          x86_64                                         6.9.10-200.fc40                                          @System                                          36 M
 kernel-modules-extra                                         x86_64                                         6.9.10-200.fc40                                          @System                                         2.6 M

Transaction Summary
============================================================================================================================================================================================================================
Remove  5 Packages
```
I mean working with `kernel-core` instead of just the meta-package should help as well, since removing kernel-core would remove all the packages we want to remove.
We can look into that in the future.

tests/integration/tier0/non-destructive/kernel/test_custom_kernel.py
danmyway: That makes sense.

tests/integration/tier0/non-destructive/kernel/test_custom_kernel.py
```diff
@@ -1,61 +1,51 @@
...
-
-if "alma-8" in SYSTEM_RELEASE_ENV:
-    distro = "alma-8"
-elif "rocky" in SYSTEM_RELEASE_ENV:
-    distro = "rocky-8"
-else:
-    distro = SYSTEM_RELEASE_ENV
-
-_, CUSTOM_KERNEL, GRUB_SUBSTRING = DISTRO_KERNEL_MAPPING[distro].values()
+from conftest import SYSTEM_RELEASE_ENV, SystemInformationRelease, _get_full_kernel_title, grub_setup_workaround
```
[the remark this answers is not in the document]
danmyway: I haven't pushed that in yet, but thanks.

tests/integration/tier0/non-destructive/kernel/test_custom_kernel.py
```diff
@@ -1,103 +1,114 @@
...
     if os.environ["TMT_REBOOT_COUNT"] == "1":
-        # Remove the current installed kernel and install the machine default kernel.
-        custom_kernel_release = CUSTOM_KERNEL.rsplit("/", 1)[-1].replace(".rpm", "")
-        assert shell("rpm -e %s" % custom_kernel_release).returncode == 0
-
-        original_kernel = os.popen("rpm -q --last kernel | head -1 | cut -d ' ' -f1").read()
-        original_kernel_release = original_kernel.rsplit("/")[-1].replace(".rpm", "").split("-")[-1]
-
+        # This is kind of naive, but we assume the second latest installed kernel is the original one
+        # We use head to filter the first two lines of the output and tail to filter the bottom line
```
kokesak, line 92: This comment is no longer applicable no?

tests/integration/tier1/destructive/proxy-conversion/test_proxy_conversion.py
```diff
@@ -91,36 +91,25 @@ def setup_rhsm(shell):
...
+    assert shell(ct_repo_shell_call, silent=True).returncode == 0
+
+    # On CentOS 8.5 we need to replace the $releasever in the url to 8.5,
+    # otherwise the dnf will complain with dependency issues.
+    if "centos-8" in SYSTEM_RELEASE_ENV:
+        shell(r"sed -i 's#\$releasever#8.5#' /etc/yum.repos.d/client-tools.repo")
+
+    # Install subscription-manager
+    subman = SubscriptionManager()
+    subman.install_package()
```
kokesak, line 112: Amazing usage of the classes. Love it

tests/integration/tier1/destructive/proxy-conversion/test_proxy_conversion.py
```diff
@@ -91,36 +91,25 @@ def setup_rhsm(shell):
...
+    ct_repo_shell_call = f"curl -o /etc/yum.repos.d/client-tools.repo https://cdn-public.redhat.com/content/public/repofiles/{client_tools_repo} \
+                    --proxy http://{TEST_VARS['PROXY_SERVER']}:{TEST_VARS['PROXY_PORT']}"
+
+    assert shell(ct_repo_shell_call, silent=True).returncode == 0
+
+    # On CentOS 8.5 we need to replace the $releasever in the url to 8.5,
+    # otherwise the dnf will complain with dependency issues.
+    if "centos-8" in SYSTEM_RELEASE_ENV:
+        shell(r"sed -i 's#\$releasever#8.5#' /etc/yum.repos.d/client-tools.repo")
+
```
kokesak, line 109: For this we can use the newly created class SubscriptionManager() that you introduced in this PR

tests/integration/conftest.py
```diff
@@ -109,6 +115,136 @@ def factory(command, silent=False, hide_command=False):
...
+        return self.shell(command)
+
+    def remove_package(self, package_name="subscription-manager*"):
+        """
+        Removes a package.
+        :param package_name: The package to be installed. Default is subscription-manager*.
+        :type package_name: str
+        """
+        command = f"yum remove -y {package_name}"
+        return self.shell(command)
```
kokesak, line 180: Even tough the variables are nice, I would avoid using them here. Since this class is specific to Subman then I do not expect to install or remove any other package than the subscription-manager. I would hardoce the name and do not allow using the parameter

tests/integration/conftest.py
```diff
@@ -109,6 +115,136 @@ def factory(command, silent=False, hide_command=False):
...
+
+    def prepare_full_workflow(self):
+        """
+        Usual full preparation workflow.
+        Calls, where applicable:
+            # only on Oracle Linux
+            self.remove_package(package_name="rhn-client-tools")
+            self.add_keys_and_certificates()
+            self.add_client_tools_repo()
+            self.install_package()
```
kokesak, line 218: I would change the indention or something as this comment indicate that all of the steps are applicable to ORacle instead only the first one

tests/integration/conftest.py
```diff
@@ -109,6 +115,136 @@ def factory(command, silent=False, hide_command=False):
...
+        Remove potential leftover subscription, unregister the system.
+        """
+        # Remove potential leftover subscription
+        self.shell("subscription-manager remove --all")
+        # Remove potential leftover registration
+        self.shell("subscription-manager unregister")
+
+        return
+
+    def prepare_full_workflow(self):
```
kokesak, line 210: Maybe `setup_requirements()` sounds better? And `terdo_requirements()`? WDYT?

tests/integration/conftest.py
```diff
@@ -109,6 +115,136 @@ def factory(command, silent=False, hide_command=False):
...
+            self.remove_package()
+            self.remove_client_tools_repo()
+            self.remove_keys_and_certificates()
+        """
+        self.unregister()
+        self.remove_package()
+        self.remove_client_tools_repo()
+        self.remove_keys_and_certificates()
+
+        return
```
kokesak, line 242: return is not needed here

tests/integration/conftest.py
```diff
@@ -109,6 +115,136 @@ def factory(command, silent=False, hide_command=False):
...
+            self.add_client_tools_repo()
+            self.install_package()
+        """
+        if SystemInformationRelease.distribution == "oracle":
+            self.remove_package(package_name="rhn-client-tools")
+        self.add_keys_and_certificates()
+        self.add_client_tools_repo()
+        self.install_package()
+
+        return
```
kokesak, line 226: return not needed
```suggestion
```

tests/integration/conftest.py
```diff
@@ -109,6 +115,136 @@ def factory(command, silent=False, hide_command=False):
...
+        self.unregister()
+        self.remove_package()
+        self.remove_client_tools_repo()
+        self.remove_keys_and_certificates()
+
+        return
+
+
+SUBMAN = SubscriptionManager()
+
```
kokesak, line 246: Is this needed? I would try to avoid it if possible. If some fixture or function needs it, then it can instantiate its own

tests/integration/conftest.py
```diff
@@ -631,56 +755,30 @@ def _unset_env_var(envars):
...
-            baseurl = "https://download.rockylinux.org/vault/rocky/8.6/BaseOS/$basearch/os/"
-    else:
-        pytest.fail(f"Unsupported distro ({distro}) provided.")
-    with open(f"/etc/yum.repos.d/{repo_name}.repo", "w") as f:
-        f.write(f"[{repo_name}]\n")
-        f.write(f"name={repo_name}\n")
-        f.write(f"baseurl={baseurl}\n")
-        f.write("enabled=0\n")
-        f.write("gpgcheck=0\n")
+    # Get the full name of the kernel (ignore rescue kernels)
```
kokesak, line 769: we could probably do something like:
```
if not kernel:
  return
```
?

tests/integration/conftest.py
```diff
@@ -689,41 +787,52 @@ def kernel(shell):
...
+        # so going with wildcard asterisk to generalize
+        elif SystemInformationRelease.version.major == 7:
+            older_kernel = "kernel-3.10.0-1160.118*"
+            assert shell(f"yum install -y {older_kernel}").returncode == 0
+
+        # Try to downgrade kernel version, if there is not multiple versions installed already.
+        # If the kernel downgrade fails, assume it's not possible and try to install from
+        # an older repo. This should only happen when Alma and Rocky has just landed on
+        # a fresh minor version.
+        elif shell("yum downgrade kernel -y").returncode != 0:
```
kokesak, line 808: Those `elif`s are a bit hard to read since all of them are doing completaly different thing. I propose to do use the second condition in `if` statement and then follow with `elif system==8`.
```suggestion
        if SystemInformationRelease.version.major == 7:
            older_kernel = "kernel-3.10.0-1160.118*"
            assert shell(f"yum install -y {older_kernel}").returncode == 0

        # Try to downgrade kernel version, if there is not multiple versions installed already.
        # If the kernel downgrade fails, assume it's not possible and try to install from
        # an older repo. This should only happen when Alma and Rocky has just landed on
        # a fresh minor version.
        elif SystemInformationRelease.version.major == 8:
            if shell("yum downgrade kernel -y").returncode != 0:
            ...
```

tests/integration/conftest.py
```diff
@@ -689,41 +787,52 @@ def kernel(shell):
...
+            # Assuming this can only happen with Alma and Rocky we'll try to install an older kernel
+            # from a previous minor version.
+            # For that we need to use the vault url and bump te current minor down one version.
+            major_ver = SystemInformationRelease.version.major
+            minor_ver = SystemInformationRelease.version.minor
+            previous_minor_ver = minor_ver - 1
+            if minor_ver == 0:
```
kokesak, line 815: doesn't hurt to put something like that
```suggestion
            if minor_ver <= 0:
```

tests/integration/conftest.py
```diff
@@ -777,19 +873,16 @@ def yum_conf_exclude(shell, backup_directory, request):
...
-            print(">>> Using parametrized packages requested in the fixture.")
-        yum_config = "/etc/yum.conf"
-        backup_dir = os.path.join(backup_directory, "yumconf")
-        shell(f"mkdir -v {backup_dir}")
+    exclude = [""]
+    if hasattr(request, "param"):
+        exclude = request.param
+        print(">>> Using parametrized packages requested in the fixture.")
+    yum_configs_all = ["/etc/yum.conf", "/etc/dnf/dnf.conf"]
+    # work only with existing config files
```
kokesak, line 881: I heard that the `/etc/dnf/dnf.conf` is not needed. You can write to the /etc/yum.conf and it gets redirected to dnf.conf if applicable. But I would comment this behavior if we use it.

tests/integration/conftest.py
```diff
@@ -961,11 +1035,26 @@ def workaround_remove_uek():
     Reference issue https://issues.redhat.com/browse/RHELC-1544
     """
     if SystemInformationRelease.distribution == "oracle":
-        subprocess.run(["yum", "remove", "-y", "kernel-uek"], check=False)
+        subprocess.run(
+            ["yum", "remove", "-y", "kernel-uek"], check=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
```
kokesak, line 1040: subprocess is nice, but is there a reason why we don't use `shell()` here? I mean there is inconsistency in that

tests/integration/tier0/non-destructive/custom-repository/test_custom_repository.py
```diff
@@ -1,38 +1,44 @@
...
-        enable_repo_opt = enable_repo_opt_epel7
-    elif system_version.major == 8:
-        if system_version.minor == 8:
-            repofile = repofile_epel8_eus
-            enable_repo_opt = enable_repo_opt_epel8_eus
-        else:
-            repofile = repofile_epel8
-            enable_repo_opt = enable_repo_opt_epel8
+
+REPOFILE, ENABLE_REPO_OPT = _assign_enable_repo_opt()
```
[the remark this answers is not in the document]
danmyway: Agreed, I had the same idea working with them, bud didn't want to create another out-of-scope refactor in this PR.
But we can definitely tackle it in some form as a code simplification task.

tests/integration/conftest.py
danmyway: Right, I will put the comment next to the method name

tests/integration/conftest.py
danmyway: Doesn't hurt, but doesn't make sense either.
We can't ever land on a system with `minor_ver < 0`
But yeah, I can include that, I guess :shrug:

tests/integration/conftest.py
danmyway: Yeah, on systems with dnf as package manager, `/etc/yum.conf` is just a symlink with `/etc/dnf/dnf.conf` as a target.
I'm not sure if `/etc/dnf/dnf.conf` is or is not needed, but I would rather keep it, if it exists on the system in the first place.
This was supposed to be more or less generalized implementation which would also ensure a proper restore of all configs in their vanilla settings.

tests/integration/tier0/non-destructive/kernel/test_custom_kernel.py
danmyway: Nice! 
I was sure, that there is a comment no longer applicable somewhere, but couldn't find it, thanks! :+1:

tests/integration/tier0/destructive/conversion-method/test_custom_repos.py
```diff
@@ -1,4 +1,42 @@
...
+    repos = ("rhel-7-server-rpms", "rhel-7-server-optional-rpms", "rhel-7-server-extras-rpms")
+    enable_opt_c2r = "--enablerepo"
+    enable_opt_yum = "--enable"
+
+    if system_version.major in (8, 9):
+        # We want to go for EUS repositories only where really applicable
+        #   1/ minor version matches EUS eligible version
+        #   2/ the system distribution does snapshots of said versions
+        #   3/ the current version is not the latest one available (i.e. not in the EUS phase yet)
+        if system_version.minor in (4, 6, 8) and system_distribution != "oracle" and "latest" not in SYSTEM_RELEASE_ENV:
```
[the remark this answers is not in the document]
danmyway: Out of scope for now, but I hear you.
I was undecided whether to add minor version 2 to the list, but if you say so, let's go for it.

tests/integration/tier1/destructive/proxy-conversion/test_proxy_conversion.py
danmyway: If it doesn't work through the proxy, I blame you :sweat_smile:

tests/integration/conftest.py
danmyway: Sounds better `set_up_requirements()` and `clean_up()` ? 
@kokesak

tests/integration/conftest.py
danmyway: You're right

tests/integration/conftest.py
danmyway: Makes sense

tests/integration/conftest.py
danmyway: Maybe rather `raise TypeError` or something like that?

tests/integration/conftest.py
danmyway: I feel like the `elif SystemInformationRelease.version.major == 8:` is unnecessary and would mean having 4 nested conditions.
I agree with having  `if SystemInformationRelease.version.major == 7:` as a separate condition.
With that we filter out EL7 and the others will be handled either by successful kernel downgrade or proceed inside the condition.

tests/integration/conftest.py
danmyway: We can refactor this further in some future iterations, I feel like it's out of scope of this PR which is already bigger than necessary.
We plan to rework the whole `conftest.py` anyway

tests/integration/conftest.py
danmyway: Again, having a comment here would be nice, so I'll include that.
The reasoning behind not using shell was that the fixture has a different scope than the shell fixture, which was causing issues.
Now that the context manager is refactored, it should work with calling it directly, without importing as a fixture.
From my perspective this again is out of scope of this PR, let's experiment with that in a different PR.

tests/integration/conftest.py
kokesak: yes!

tests/integration/conftest.py
kokesak: Oh yeah, you are right, I thought for some reason that you are using `previous_minor_ver` in that condition where you are subtracting -1  which could lead to negative number. The `minor_ver` is fine

tests/integration/conftest.py
kokesak: yes, sounds nice

tests/integration/conftest.py
kokesak: I think `ValueError` would be better here. https://docs.python.org/3/library/exceptions.html#TypeError - 

"Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) 
when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError)."

tests/integration/tier0/non-destructive/custom-repository/test_custom_repository.py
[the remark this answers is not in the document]
kokesak: Created a ticket for this https://issues.redhat.com/browse/RHELC-1687 and put it under my Epic to refactor the conftest.py

tests/integration/conftest.py
```diff
@@ -504,17 +638,19 @@ def _load_json_schema(path):
...
+def pre_registered(shell, request):
     """
     A fixture to install subscription manager and pre-register the system prior to the convert2rhel run.
-    For Oracle Linux we're using the _add_client_tools_repo_oracle to enable the client-tools repository
-    to install the subscription-manager package from.
-    We also exclude the rhn-client* packages for the same reason.
-    On Oracle Linux subscription-manger is obsolete and replaced by rhn-client* packages when installing subman.
+    We're using the client-tools-for-rhel-<version>-rpms repository to install the subscription-manager package from.
+        This does not apply to CentOS 8.5 since recently dependency conflicts caused inability to install the package
+        from client tools, repo. We install the package from the native repositories.
```
danmyway, line 646: Not applicable anymore. We set the releasever to 8.5 in the baseurl on CentOS 8.
https://github.com/oamg/convert2rhel/pull/946/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56R155
```suggestion
```

kokesak: 1hhv9m(1)

tests/integration/conftest.py
```diff
@@ -689,41 +783,52 @@ def kernel(shell):
...
+                # Skip the test if so
+                pytest.skip("There is no older kernel to install for this system.")
+            releasever = ".".join((str(major_ver), str(previous_minor_ver)))
+            old_repo = None
+            if SystemInformationRelease.distribution == "alma":
+                old_repo = f"https://vault.almalinux.org/{releasever}/BaseOS/x86_64/os/"
+            elif SystemInformationRelease.distribution == "rocky":
+                old_repo = f"https://dl.rockylinux.org/vault/rocky/{releasever}/BaseOS/x86_64/os/"
+            # Install the kernel from the url
+            shell(f"yum install kernel -y --repofromurl 'oldrepo,{old_repo}'")
```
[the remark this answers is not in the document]
danmyway: Oracle has somewhat rolling-release-like repositories, so there's always more than one kernel in there.
I can add else :shrug:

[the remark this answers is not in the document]
kokesak: Oh yeah I forgot that we are inside the `elif shell("yum downgrade kernel -y").returncode != 0:` branch. But yeah, please add an else branch

kokesak (approved):
Great chnages. Merging this biggie

convert2rhel/data/9/x86_64/configs/rocky-9-x86_64.cfg
```diff
@@ -10,10 +10,10 @@ gpg_fingerprints = 702d426d350d275d
 excluded_pkgs =
   rocky-obsolete-packages
   rocky-gpg-keys
+  rocky-logos
```
hosekadam, line 13: The swap doesn't work or this is a mistake?

hosekadam (approved):
Looks good

convert2rhel/data/9/x86_64/configs/rocky-9-x86_64.cfg
danmyway: It doesn't work due to some conflicting files 🥲

PR #641, opened 2022-10-26
[RHELC-772] checks: Omit exclude= options from repoquery

PicklesMcDougan:
In is_loaded_kernel_latest(), repoquery is subject to exclude= in /etc/yum.conf. If there is "exclude=kernel", the conversion can fail with the following:

	CRITICAL - Could not find any kernel from repositories to compare against the loaded kernel.

This patch resolves that condition by explicitly disabling all "exclude" options with:

	--setopt=exclude""

Signed-off-by: John Castranio <jcastran@redhat.com>

Jira Issue: [RHELC-772](https://issues.redhat.com/browse/RHELC-772)

Checklist
- [x] PR meets acceptance criteria specified in the Jira issue
- [x] PR has been tested manually in a VM (either author or reviewer)
- [ ] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [ ] Code and tests are documented properly
- [ ] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending`

PR #947, opened 2023-09-29
[RHELC-880] 8.8 as an EUS release supported for conversions

pr-watson:
This PR updates the corresponds_to_rhel_eus() function to have improved detection if a system is eus. A function get_latest_distro_release_version() is in place to query the enabled repositories on the system for their respective *- release package, the config files for each system we convert have been updated to have the hard-coded release package we need to query to extract the latest available minor version of the system. 

Then this latest released system version is compared with the running systems minor version. If they are equal, we use the standard RHEL repos for the conversion. If the running system version is older than the latest available, then we compare the running version with the hardcoded dictionary with EUS period start dates. If the date hasn't been reached yet, we again use the standard RHEL repos for the conversion. If it has been reached, we convert using EUS repos.

Integration tests related to eus have also been updated.

Jira Issues: [RHELC-880](https://issues.redhat.com/browse/RHELC-880)

Checklist

### DIFF
--- a/convert2rhel/actions/system_checks/eus.py
+++ b/convert2rhel/actions/system_checks/eus.py
@@ -1,0 +1,53 @@
+# Copyright(C) 2023 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+
+import datetime
+import logging
+
+from convert2rhel import actions
+from convert2rhel.systeminfo import EUS_MINOR_VERSIONS, system_info
+from convert2rhel.toolopts import tool_opts
+
+
+logger = logging.getLogger(__name__)
+
+
+class EusSystemCheck(actions.Action):
+    id = "EUS_SYSTEM_CHECK"
+
+    def run(self):
+        """Warn the user if their system is under EUS and past the EUS release date without using the --eus cli option."""
+        super(EusSystemCheck, self).run()
+
+        current_version = "%s.%s" % (system_info.version.major, system_info.version.minor)
+        eus_versions = list(EUS_MINOR_VERSIONS.keys())[1:]
+        if current_version in eus_versions:
+            eus_release_date = EUS_MINOR_VERSIONS.get(current_version, False)
+            # Turn eus_release_date into a datetime object
+            eus_release_date = datetime.datetime.strptime(eus_release_date, "%Y-%m-%d").date()
+            current_datetime = datetime.date.today()
+
+            # warning message if the eus release date is past and the --eus option is not set
+            if not tool_opts.eus and current_datetime > eus_release_date:
+                self.add_message(
+                    level="WARNING",
+                    id="EUS_COMMAND_LINE_OPTION_UNUSED",
+                    title="The --eus command line option is unused",
+                    description="Current system version is under Extended Lifecycle Support (EUS). You may want to consider using the --eus"
+                    " command line option to land on a system patched with the latest security errata.",
+                )
+        return
--- a/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
+++ b/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
@@ -5,6 +5,10 @@
 # Delimited by whitespace(s).
 gpg_fingerprints = 24c6a8a7f4a80eb5
 
+# Release package used to query the major and minor version
+release_pkg =
+  centos-release
+
 # List of packages to be removed before the system conversion starts.
 # Delimited by any whitespace(s).
 excluded_pkgs =
--- a/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
+++ b/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
@@ -5,10 +5,6 @@
 # Delimited by whitespace(s).
 gpg_fingerprints = 24c6a8a7f4a80eb5
 
-# Release package used to query the major and minor version
-release_pkg =
-  centos-release
-
 # List of packages to be removed before the system conversion starts.
 # Delimited by any whitespace(s).
 excluded_pkgs =
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -677,7 +677,7 @@ def _get_package_repositories(pkgs):
 
     output, retcode = utils.run_subprocess(
         ["repoquery", "--quiet", "-q"] + pkgs + ["--qf", query_format],
-        print_cmd=False,
+        print_cmd=True,
         print_output=False,
     )
     output = [line for line in output.split("\n") if line]
--- a/convert2rhel/repo.py
+++ b/convert2rhel/repo.py
@@ -39,11 +39,7 @@ def get_rhel_repoids():
     we preferably enable the RHEL EUS repoids as those provide security updates over two years, in comparison to 6 months
     in case of the standard non-EUS repoids.
     """
-
-    if system_info.corresponds_to_rhel_eus_release():
-        repos_needed = system_info.eus_rhsm_repoids
-    else:
-        repos_needed = system_info.default_rhsm_repoids
+    repos_needed = system_info.eus_rhsm_repoids if system_info.eus_system else system_info.default_rhsm_repoids
 
     loggerinst.info("RHEL repository IDs to enable: %s" % ", ".join(repos_needed))
 
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -24,9 +24,6 @@ import re
 import time
 
 from collections import namedtuple
-from datetime import datetime as dt
-
-import rpm
 
 from six.moves import configparser, urllib
 
@@ -63,11 +60,7 @@ EUS_MINOR_VERSIONS = {
 Version = namedtuple("Version", ["major", "minor"])
 
 
-class RepoqueryFailure(Exception):
-    """Custom exception for repoquery failure"""
-
-
-class SystemInfo(object):
+class SystemInfo:
     def __init__(self):
         # Operating system name (e.g. Oracle Linux)
         self.name = None
@@ -91,12 +84,14 @@ class SystemInfo(object):
             # RPM-GPG-KEY-redhat-legacy-former
             "219180cddb42a60e",
         ]
+        # Whether the system has internet access
+        self.has_internet_access = None
+        # Whether the system release corresponds to a rhel eus release
+        self.eus_system = None
         # Packages to be removed before the system conversion
         self.excluded_pkgs = []
         # Release packages to be removed before the system conversion
         self.repofile_pkgs = []
-        # Main release package used to query release version
-        self.release_pkg = None
         self.cfg_filename = None
         self.cfg_content = None
         self.system_release_file_content = None
@@ -132,11 +127,11 @@ class SystemInfo(object):
         self.fingerprints_orig_os = self._get_gpg_key_fingerprints()
         self.generate_rpm_va()
         self.releasever = self._get_releasever()
-        self.release_pkg = self._get_release_package()
         self.kmods_to_ignore = self._get_kmods_to_ignore()
         self.booted_kernel = self._get_booted_kernel()
         self.has_internet_access = self._check_internet_access()
         self.dbus_running = self._is_dbus_running()
+        self.eus_system = self.corresponds_to_rhel_eus_release()
 
     def print_system_information(self):
         """Print system related information."""
@@ -255,9 +250,6 @@ class SystemInfo(object):
 
     def _get_repofile_pkgs(self):
         return self._get_cfg_opt("repofile_pkgs").split()
-
-    def _get_release_package(self):
-        return self._get_cfg_opt("release_pkg")
 
     def _get_releasever(self):
         """
@@ -422,57 +414,15 @@ class SystemInfo(object):
         :return: Whether or not the current system has an EUS correspondent in RHEL.
         :rtype: bool
         """
+        current_version = "%s.%s" % (self.version.major, self.version.minor)
+        # This check will be dropped once 8.6 is no longer supported under EUS
+        if current_version == "8.6":
+            return True
+        if tool_opts.eus and current_version in EUS_MINOR_VERSIONS:
+            self.logger.info("EUS argument detected, automatically evaluating system as EUS")
+            return True
 
-        eus_release_date = EUS_MINOR_VERSIONS.get("%s.%s" % (self.version.major, self.version.minor), False)
-        if eus_release_date:
-            eus_release_date = dt.strptime(eus_release_date, "%Y-%m-%d")
-            current_datetime = dt.today()
-            if not self.has_internet_access:
-                return self.datetime_comparison_returns(current_datetime, eus_release_date)
-            elif current_datetime > eus_release_date:
-                try:
-                    latest_release_version = self.get_latest_distro_release_version()
-                except RepoqueryFailure:
-                    return self.datetime_comparison_returns(current_datetime, eus_release_date)
-
-                latest_major, latest_minor = latest_release_version.split(".")
-                if int(latest_minor) > int(self.version.minor):
-                    return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
-                if int(latest_minor) < int(self.version.minor):
-                    pass
-                    # some form of warning
-                    # corner case that should not happen, need to decide what to do in case we land in this condition
-                    # where the users system on a version that hasn't been released yet
         return False
-
-    def datetime_comparison_returns(self, current_datetime, eus_release_date):
-        if current_datetime > eus_release_date:
-            return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
-        return False
-
-    def get_latest_distro_release_version(self):
-        """Extract minor version from release name using yum list"""
-        cmd = [
-            "repoquery",
-            "--setopt=exclude=",
-            "--quiet",
-            "--qf",
-            "C2R\\t%{VERSION}",
-            self.release_pkg,
-        ]
-        yum_output, return_code = run_subprocess(cmd, print_output=False)
-        if return_code != 0:
-            self.logger.debug("Got the following output: %s", yum_output)
-            raise RepoqueryFailure("Repoquery failed to execute")
-
-        latest_release_version = "0.0"
-        for line in yum_output.split("\n"):
-            line = line.strip()
-            if "C2R" in line:
-                _, release_version = line.split("\t")
-                if rpm.labelCompare(("0", latest_release_version, "0"), ("0", release_version, "0")) == -1:
-                    latest_release_version = release_version
-        return latest_release_version
 
     def _is_dbus_running(self):
         """
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -24,6 +24,9 @@ import re
 import time
 
 from collections import namedtuple
+from datetime import datetime as dt
+
+import rpm
 
 from six.moves import configparser, urllib
 
@@ -51,13 +54,20 @@ RELEASE_VER_MAPPING = {
     "7.9": "7Server",
 }
 
-# List of EUS minor versions supported
-EUS_MINOR_VERSIONS = ["8.6"]
+# Dictionary of EUS minor versions supported and their EUS period start date
+EUS_MINOR_VERSIONS = {
+    "8.6": "2022-11-09",
+    "8.8": "2023-11-14",
+}
 
 Version = namedtuple("Version", ["major", "minor"])
 
 
-class SystemInfo:
+class RepoqueryFailure(Exception):
+    """Custom exception for repoquery failure"""
+
+
+class SystemInfo(object):
     def __init__(self):
         # Operating system name (e.g. Oracle Linux)
         self.name = None
@@ -85,6 +95,8 @@ class SystemInfo:
         self.excluded_pkgs = []
         # Release packages to be removed before the system conversion
         self.repofile_pkgs = []
+        # Main release package used to query release version
+        self.release_pkg = None
         self.cfg_filename = None
         self.cfg_content = None
         self.system_release_file_content = None
@@ -120,6 +132,7 @@ class SystemInfo:
         self.fingerprints_orig_os = self._get_gpg_key_fingerprints()
         self.generate_rpm_va()
         self.releasever = self._get_releasever()
+        self.release_pkg = self._get_release_package()
         self.kmods_to_ignore = self._get_kmods_to_ignore()
         self.booted_kernel = self._get_booted_kernel()
         self.has_internet_access = self._check_internet_access()
@@ -242,6 +255,9 @@ class SystemInfo:
 
     def _get_repofile_pkgs(self):
         return self._get_cfg_opt("repofile_pkgs").split()
+
+    def _get_release_package(self):
+        return self._get_cfg_opt("release_pkg")
 
     def _get_releasever(self):
         """
@@ -403,10 +419,60 @@ class SystemInfo:
         For example if we detect CentOS Linux 8.4, the corresponding RHEL 8.4 is an EUS release, but if we detect
         CentOS Linux 8.5, the corresponding RHEL 8.5 is not an EUS release.
 
-        :return: Whether or not the current system has a EUS correspondent in RHEL.
+        :return: Whether or not the current system has an EUS correspondent in RHEL.
         :rtype: bool
         """
-        return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+
+        eus_release_date = EUS_MINOR_VERSIONS.get("%s.%s" % (self.version.major, self.version.minor), False)
+        if eus_release_date:
+            eus_release_date = dt.strptime(eus_release_date, "%Y-%m-%d")
+            current_datetime = dt.today()
+            if not self.has_internet_access:
+                return self.datetime_comparison_returns(current_datetime, eus_release_date)
+            elif current_datetime > eus_release_date:
+                try:
+                    latest_release_version = self.get_latest_distro_release_version()
+                except RepoqueryFailure:
+                    return self.datetime_comparison_returns(current_datetime, eus_release_date)
+
+                latest_major, latest_minor = latest_release_version.split(".")
+                if int(latest_minor) > int(self.version.minor):
+                    return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+                if int(latest_minor) < int(self.version.minor):
+                    pass
+                    # some form of warning
+                    # corner case that should not happen, need to decide what to do in case we land in this condition
+                    # where the users system on a version that hasn't been released yet
+        return False
+
+    def datetime_comparison_returns(self, current_datetime, eus_release_date):
+        if current_datetime > eus_release_date:
+            return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+        return False
+
+    def get_latest_distro_release_version(self):
+        """Extract minor version from release name using yum list"""
+        cmd = [
+            "repoquery",
+            "--setopt=exclude=",
+            "--quiet",
+            "--qf",
+            "C2R\\t%{VERSION}",
+            self.release_pkg,
+        ]
+        yum_output, return_code = run_subprocess(cmd, print_output=False)
+        if return_code != 0:
+            self.logger.debug("Got the following output: %s", yum_output)
+            raise RepoqueryFailure("Repoquery failed to execute")
+
+        latest_release_version = "0.0"
+        for line in yum_output.split("\n"):
+            line = line.strip()
+            if "C2R" in line:
+                _, release_version = line.split("\t")
+                if rpm.labelCompare(("0", latest_release_version, "0"), ("0", release_version, "0")) == -1:
+                    latest_release_version = release_version
+        return latest_release_version
 
     def _is_dbus_running(self):
         """
--- a/convert2rhel/toolopts.py
+++ b/convert2rhel/toolopts.py
@@ -86,6 +86,7 @@ class ToolOpts:
         self.org = None
         self.arch = None
         self.no_rpm_va = False
+        self.eus = False
         self.activity = None
 
     def set_opts(self, supported_opts):
@@ -124,10 +125,10 @@ class CLI:
             "\n"
             "  convert2rhel [--version] [-h]\n"
             "  convert2rhel {subcommand} [-u username] [-p password | -c conf_file_path] [--pool pool_id | -a] [--disablerepo repoid]"
-            " [--enablerepo repoid] [--serverurl url] [--no-rpm-va] [--debug] [--restart] [-y]\n"
-            "  convert2rhel {subcommand} [--no-rhsm] [--disablerepo repoid] [--enablerepo repoid] [--no-rpm-va] [--debug] [--restart] [-y]\n"
+            " [--enablerepo repoid] [--serverurl url] [--no-rpm-va] [--eus] [--debug] [--restart] [-y]\n"
+            "  convert2rhel {subcommand} [--no-rhsm] [--disablerepo repoid] [--enablerepo repoid] [--no-rpm-va] [--eus] [--debug] [--restart] [-y]\n"
             "  convert2rhel {subcommand} [-k activation_key | -c conf_file_path] [-o organization] [--pool pool_id | -a] [--disablerepo repoid] [--enablerepo"
-            " repoid] [--serverurl url] [--no-rpm-va] [--debug] [--restart] [-y]\n"
+            " repoid] [--serverurl url] [--no-rpm-va] [--eus] [--debug] [--restart] [-y]\n"
         ).format(subcommand=subcommand_to_print)
 
         if subcommand_not_used_on_cli:
@@ -187,6 +188,13 @@ class CLI:
             " stored in log files %s and %s. At the end of the conversion, these logs are compared"
             " to show you what rpm files have been affected by the conversion."
             % (PRE_RPM_VA_LOG_FILENAME, POST_RPM_VA_LOG_FILENAME),
+        )
+        self._shared_options_parser.add_argument(
+            "--eus",
+            action="store_true",
+            help="Automatically recognize the system as eus, utilizing eus repos."
+            " 8.6 systems do not require this option as they are recognized as eus automatically."
+            " This option is meant for 8.8+ systems.",
         )
         self._shared_options_parser.add_argument(
             "--enablerepo",
@@ -408,6 +416,9 @@ class CLI:
             tool_opts.no_rhsm = True
             if not tool_opts.enablerepo:
                 loggerinst.critical("The --enablerepo option is required when --disable-submgr or --no-rhsm is used.")
+
+        if parsed_opts.eus:
+            tool_opts.eus = True
 
         if not tool_opts.disablerepo:
             # Default to disable every repo except:
--- a/convert2rhel/unit_tests/actions/system_checks/eus_test.py
+++ b/convert2rhel/unit_tests/actions/system_checks/eus_test.py
@@ -1,0 +1,73 @@
+# Copyright(C) 2023 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+
+import datetime
+
+import pytest
+
+from convert2rhel import actions, pkgmanager, systeminfo
+from convert2rhel.actions.system_checks import eus
+from convert2rhel.systeminfo import Version, system_info
+
+
+@pytest.fixture
+def eus_action():
+    return eus.EusSystemCheck()
+
+
+class DateMock(datetime.date):
+    @classmethod
+    def today(cls):
+        return cls(2023, 11, 15)
+
+
+class TestEus:
+    @pytest.mark.parametrize(
+        ("version_string", "message_reported"),
+        (
+            (Version(8, 6), False),
+            (Version(8, 8), True),
+            (Version(8, 10), False),  # Change to True after 8.10 is under eus
+            (Version(9, 2), False),  # Change to True after 9.2 is under eus
+        ),
+    )
+    @pytest.mark.skipif(pkgmanager.TYPE != "dnf", reason="el7 systems are not under eus")
+    def test_eus_warning_message(self, eus_action, monkeypatch, global_tool_opts, version_string, message_reported):
+
+        global_tool_opts.eus = False
+        monkeypatch.setattr(system_info, "version", version_string)
+        monkeypatch.setattr(systeminfo, "tool_opts", global_tool_opts)
+        monkeypatch.setattr(eus.datetime, "date", DateMock)
+
+        eus_action.run()
+        expected = set(
+            (
+                actions.ActionMessage(
+                    level="WARNING",
+                    id="EUS_COMMAND_LINE_OPTION_UNUSED",
+                    title="The --eus command line option is unused",
+                    description="Current system version is under Extended Lifecycle Support (EUS). You may want to consider using the --eus"
+                    " command line option to land on a system patched with the latest security errata.",
+                ),
+            )
+        )
+
+        if message_reported:
+            assert expected.issuperset(eus_action.messages)
+            assert expected.issubset(eus_action.messages)
+        else:
+            assert eus_action.messages == []
--- a/convert2rhel/unit_tests/systeminfo_test.py
+++ b/convert2rhel/unit_tests/systeminfo_test.py
@@ -27,12 +27,10 @@ import pytest
 import six
 
 from convert2rhel import logger, systeminfo, unit_tests, utils  # Imports unit_tests/__init__.py
-from convert2rhel.systeminfo import EUS_MINOR_VERSIONS, RELEASE_VER_MAPPING, RepoqueryFailure, Version, system_info
+from convert2rhel.systeminfo import RELEASE_VER_MAPPING, Version, system_info
 from convert2rhel.toolopts import tool_opts
-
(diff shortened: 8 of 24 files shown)


convert2rhel/systeminfo.py
SpyTec, line 1: There will be some overlap with #897

convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
```diff
@@ -5,6 +5,10 @@
 # Delimited by whitespace(s).
 gpg_fingerprints = 24c6a8a7f4a80eb5
 
+# Release package used to query the major and minor version
+release_pkg =
+  centos-release
```
r0x0d, line 10: I don't think we will have more than one package here at time, so, what do you think about inlining this package to not confuse people into thinking that this may be a list? 


```suggestion
release_pkg = centos-release
```

Its a minor suggestion, but inlining here could help thinking that this is treated as a single item

convert2rhel/systeminfo.py
```diff
@@ -397,10 +413,58 @@ def corresponds_to_rhel_eus_release(self):
         For example if we detect CentOS Linux 8.4, the corresponding RHEL 8.4 is an EUS release, but if we detect
         CentOS Linux 8.5, the corresponding RHEL 8.5 is not an EUS release.
 
-        :return: Whether or not the current system has a EUS correspondent in RHEL.
+        :return: Whether or not the current system has an EUS correspondent in RHEL.
         :rtype: bool
         """
-        return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+        eus_release_date = EUS_MINOR_VERSIONS.get("%s.%s" % (self.version.major, self.version.minor), False)
```
r0x0d, line 419: To simplify things, we could assign the default return value for this to be an empty string instead of a boolean value. 

```suggestion
        eus_release_date = EUS_MINOR_VERSIONS.get("%s.%s" % (self.version.major, self.version.minor), "")
```

convert2rhel/systeminfo.py
```diff
@@ -397,10 +413,58 @@ def corresponds_to_rhel_eus_release(self):
...
+    def datetime_comparison_returns(self, current_datetime, eus_release_date):
+        if current_datetime > eus_release_date:
+            return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+        return False
+
+    def get_latest_distro_release_version(self):
+        """Extract minor version from release name using yum list"""
+        cmd = [
+            "repoquery",
+            "--setopt=exclude=",
```
r0x0d, line 449: Do we really need to override the exclude flag with this repoquery call? It seems to me a bit odd... In your tests, @pr-watson, have you encountered a case where this was needed?

convert2rhel/systeminfo.py
```diff
@@ -397,10 +413,58 @@ def corresponds_to_rhel_eus_release(self):
...
+            "--quiet",
+            "--qf",
+            "C2R\\t%{VERSION}",
```
r0x0d, line 452: Spacing to make it easier to read that there is a query format in there.

```suggestion
            "C2R\\t %{VERSION}",
```

convert2rhel/systeminfo.py
```diff
@@ -397,10 +413,58 @@ def corresponds_to_rhel_eus_release(self):
...
+            self.release_pkg,
+        ]
+        yum_output, return_code = run_subprocess(cmd, print_output=False)
+        if return_code != 0:
+            self.logger.debug("Got the following output: %s", yum_output)
+            raise RepoqueryFailure("Repoquery failed to execute")
```
r0x0d, line 458: Since we won't be using the message here, is there any purpose in keeping it? We could transform this into a comment instead, I think... Would be more valuable to know what are the possibly failure for repoquery in this case instead of assigning a message to this exception and not using it anywhere else.

r0x0d:
@pr-watson, can you test that Ctrl+C works right after the functions that uses `rpm` are called? I don't like that we are importing rpm in such a module that is called pretty much at the beginning like this, but, if that won't break the ctrl + c support, I guess that should be fine.

convert2rhel/systeminfo.py
```diff
@@ -397,10 +413,58 @@ def corresponds_to_rhel_eus_release(self):
...
+                    return self.datetime_comparison_returns(current_datetime, eus_release_date)
+
+                latest_major, latest_minor = latest_release_version.split(".")
+                if int(latest_minor) > int(self.version.minor):
+                    return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+                if int(latest_minor) < int(self.version.minor):
+                    pass
+                    # corner case that should not happen, need to decide what to do in case we land in this condition
+                    # where the users system on a version that hasn't been released yet
+        return False
```
r0x0d, line 438: Tried to make the function a bit more cleaner to read, let me know if you think it works out the same way you intended to with the original one. 


```diff
diff --git a/convert2rhel/systeminfo.py b/convert2rhel/systeminfo.py
index 8b8b41a..7b129b4 100644
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -416,26 +416,36 @@ class SystemInfo(object):
         :return: Whether or not the current system has an EUS correspondent in RHEL.
         :rtype: bool
         """
-        eus_release_date = EUS_MINOR_VERSIONS.get("%s.%s" % (self.version.major, self.version.minor), False)
-        if eus_release_date:
-            eus_release_date = dt.strptime(eus_release_date, "%Y-%m-%d")
-            current_datetime = dt.today()
-            if not self.has_internet_access:
-                return self.datetime_comparison_returns(current_datetime, eus_release_date)
-            elif current_datetime > eus_release_date:
-                try:
-                    latest_release_version = self.get_latest_distro_release_version()
-                except RepoqueryFailure:
-                    return self.datetime_comparison_returns(current_datetime, eus_release_date)
-
-                latest_major, latest_minor = latest_release_version.split(".")
-                if int(latest_minor) > int(self.version.minor):
-                    return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
-                if int(latest_minor) < int(self.version.minor):
-                    pass
-                    # corner case that should not happen, need to decide what to do in case we land in this condition
-                    # where the users system on a version that hasn't been released yet
-        return False
+        eus_release_date = EUS_MINOR_VERSIONS.get("%s.%s" % (self.version.major, self.version.minor), "")
+        if not eus_release_date:
+            return False
+
+        # Turn eus_release_date into an datetime object
+        eus_release_date = dt.strptime(eus_release_date, "%Y-%m-%d")
+        current_datetime = dt.today()
+
+        # If our current date is not greater than eus_release_date, there is no
+        # need to continue with the function, as we will just return False at
+        # the end anyway.
+        if current_datetime < eus_release_date:
+            return False
+
+        # In case of no internet access, lets just skip the rest of the checks.
+        if not self.has_internet_access:
+            return self.datetime_comparison_returns(current_datetime, eus_release_date)
+        
+        try:
+            latest_release_version = self.get_latest_distro_release_version()
+        except RepoqueryFailure:
+            return self.datetime_comparison_returns(current_datetime, eus_release_date)
+
+        latest_major, latest_minor = latest_release_version.split(".")
+        if int(latest_minor) > int(self.version.minor):
+            return eus_release_date.strftime("%Y-%m-%d") in EUS_MINOR_VERSIONS
+        if int(latest_minor) < int(self.version.minor):
+            pass
+            # corner case that should not happen, need to decide what to do in case we land in this condition
+            # where the users system on a version that hasn't been released yet
 
     def datetime_comparison_returns(self, current_datetime, eus_release_date):
         if current_datetime > eus_release_date:
```

convert2rhel/systeminfo.py
pr-watson: I saw it used with the query in is_loaded_kernel_latest so I figured i'd stick with the same format, we can remove it

convert2rhel/systeminfo.py
pr-watson: Okay so remove the exception and keep the debug?

pr-watson:
> @pr-watson, can you test that Ctrl+C works right after the functions that uses `rpm` are called? I don't like that we are importing rpm in such a module that is called pretty much at the beginning like this, but, if that won't break the ctrl + c support, I guess that should be fine.

Yes I will test it

convert2rhel/systeminfo.py
```diff
@@ -397,10 +413,58 @@ def corresponds_to_rhel_eus_release(self):
...
+            raise RepoqueryFailure("Repoquery failed to execute")
+
+        latest_release_version = "0.0"
+        for line in yum_output.split("\n"):
+            line = line.strip()
+            if "C2R" in line:
+                _, release_version = line.split("\t")
+                if rpm.labelCompare(("0", latest_release_version, "0"), ("0", release_version, "0")) == -1:
+                    latest_release_version = release_version
+        return latest_release_version
```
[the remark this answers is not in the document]
pr-watson: This rpm.labelCompare is for instances where multiple packages are returned from the yum output so we need to determine which is the latest version

convert2rhel/systeminfo.py
pr-watson: Function looks better and accomplishes the same thing

convert2rhel/systeminfo.py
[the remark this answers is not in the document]
r0x0d: Talking privately in slack, we decided to leave this behind for now and focus on the other comments. This is not a high priority refactor for now, so if we manage to refactor this before we are ready and set to merge, we can deal with it, otherwise, ensuring that it is working correctly is most important.

convert2rhel/systeminfo.py
r0x0d: Well.. If for some reason the user specified that the `centos-release` package is excluded, then with this option we are overriding it and saying that it is not excluded anymore. 

I don't ever recall a case where I saw that it was excluded for some reason (The only one I can remember of top of my head was a kernel one). 

I would say that we can remove this part of the repoquery args, but if you want to check out with @bocekm or @abadger first, that would be good too.

convert2rhel/systeminfo.py
r0x0d: I think you can just `raise RepoqueryFailure` instead and not pass any message to it.

convert2rhel/systeminfo.py
```diff
@@ -397,10 +413,58 @@ def corresponds_to_rhel_eus_release(self):
...
+            "repoquery",
+            "--setopt=exclude=",
+            "--quiet",
+            "--qf",
+            "C2R\\t%{VERSION}",
+            self.release_pkg,
+        ]
+        yum_output, return_code = run_subprocess(cmd, print_output=False)
+        if return_code != 0:
+            self.logger.debug("Got the following output: %s", yum_output)
```
r0x0d, line 457:
```suggestion
            self.logger.debug("Repoquery execution return status codce '%s' with output %s.", return_code, yum_output)
```

what do you think about this message? I think it adds a bit more context to us with the return_code in it.

bocekm:
```suggestion
            self.logger.debug("Repoquery execution exited with return code '%s' and with output: %s.", return_code, repoquery_output)
```
We call repoquery, not yum.

convert2rhel/systeminfo.py
bocekm: It makes sense to keep the `--setopt=exclude=` here. If a user has `exclude=centos-release` in `/etc/yum.conf` this repoquery call would fail. I see no reason to adhere to the custom user exclusion here.

Related: https://github.com/oamg/convert2rhel/pull/641

convert2rhel/systeminfo.py
abadger: <nod> Note:  Both default values have use cases.  The way I decide is:

* If I just want to use the value later and my code has a fallback for when it wasn't set, then I'll use a value from within the data type that will work with that code (Empty string, empty list, and `False` are common defaults).  Example: My function gets a colon separated list of filenames from an environment variable, loops over each filename and performs an action.  I would use `for filename in os.environ.get("MYPATH", "").split(":"):` to retrieve the environment variable.  Empty string is perfect because it will split into an empty list which means the loop will not be executed.
 
* If there's a reason to differentiate between a default value and an explicitly set value, then I'll set it to `False` (or `None` if the variable holds a boolean).  Example: My function changes a number of attributes on a file, permissions, selinux context, owner, and group.  If the user has configured the permissions via a config value, then I want to use that.  If they have not and it's a new file, then I want to use a default value.  If they have not and it's an existing file, then I don't want to change the permissions at all.  So `config.get("mode", False)` is a better clue to the next reader of the code that we're dealing with a tri-state here.

This code is in the middle of those two criteria: If `eus_release_date` is not set, then it skips the rest of the code in the function and returns a hardcoded `False`, never using the value as a string.  In that circumstance, either style works here although my personal order of preference is probably:

* Use a try except:
  ``` python
  try:
      eus_release_date = 
        eus_release_date = EUS_MINOR_VERSIONS["%s.%s" % (self.version.major, self.version.minor)]
  except KeyError:
      return False
  ```
* Return False to be explicit that we're going to use this as a condition
* Use empty string.

convert2rhel/systeminfo.py
r0x0d: Then lets leave it there

r0x0d:
@pr-watson will investigate how to take care of finding a way to get the minor release number from repoquery for the different targets we support. Currently, the repoquery invocation will not give the correct information for some systems and may lead to incorrect errors. 

For example, under Oracle Linux 7, the %{version} tag will return 1.0 instead of the current version as the rest of the other systems. 
And under CentOS 7, it will just prints `7`, which is not what we expect with the repoquery output, as we want just the minor version. 

We might need to parse the result (package name) to identify and figure out how to extract that information.

convert2rhel/systeminfo.py
abadger: If you add a space here you should (perhaps must, but `labelCompare` might be stripping whitespace on its own) need to modify the code which parses it.  Right now we just do a `split('\t')` which would leave the space character intact.

Personally, I like having a single character separator rather than relying on matching any whitespace for machine parsing but I can see the readability argument.  You could use other characters as the separator which might make it more readable but still keep the unambiguous machine parsibility? For instance `|`

Side note: When making command line output robust for machine parsing, it's often advised to use the null byte as the separator (`\0`).  This is because NULL can't end up in filenames and, due to it being the terminating character for C strings, it is uncommon for it to be a valid character in the contents of the fields being separated.  However, that makes the output even more human unreadable than `\t`.

pr-watson:
After some investigation I found that the workarounds for centos7 will not be needed since it is not under EUS. Once a centos7 system gets to the corresponds_to_rhel_eus function it will automatically return False, therefore we do not require handling of centos7 systems. For oracle linux the release package that was being queried was not giving the proper minor version, the correct release package has been found and it works with the current workflow.

convert2rhel/systeminfo.py
pr-watson: Opting for no space in the query format

convert2rhel/systeminfo.py
```diff
@@ -397,10 +413,68 @@ def corresponds_to_rhel_eus_release(self):
...
+        # In case of no internet access, lets just skip the rest of the checks.
+        if not self.has_internet_access:
+            return self.datetime_comparison_returns(current_datetime, eus_release_date)
+
+        try:
+            latest_release_version = self.get_latest_distro_release_version()
+        except RepoqueryFailure:
+            return self.datetime_comparison_returns(current_datetime, eus_release_date)
+
+        latest_major, latest_minor = latest_release_version.split(".")
```
r0x0d, line 442: Since the `latest_major` is not being used, we could just repalce it with 
```suggestion
        _, latest_minor = latest_release_version.split(".")
```

pr-watson: Sounds good

SpyTec:
Needs a quick rebase

convert2rhel/systeminfo.py
r0x0d: > I would always pass a message with an Exception. The reason is that the code that catches the exception is in charge of making the decisions. So the code which raises the exception doesn't know whether a message is needed or not.

For that, then, let's make the message more meaningful and provide some context of what exactly failed with the repoquery

tests/integration/common/checks-after-conversion/test_enabled_repositories.py
```diff
@@ -28,7 +28,7 @@ def test_enabled_repositories(shell, system_release):
     enabled_repos = shell("yum repolist").output
 
     try:
-        if "redhat-8.6" in system_release:
+        if ("redhat-8.6" or "redhat-8.8") in system_release:
```
r0x0d, line 31: I don't think this will ever be true for `redhat-8.8` case. The reason for that is that Python will evaluate the first parenthesis and will check if it will use either `redhat-8.6` and `redhat-8.8`, and since both strings exists and are present, it will just default to the first one. 

For example, take a look at this example by changing `system_release` to `redhat-8.8`: 
```bash
Python 3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> system_release = "redhat-8.8"
>>> ("redhat-8.6" or "redhat-8.8") in system_release
False
>>> system_release = "redhat-8.6"
>>> ("redhat-8.6" or "redhat-8.8") in system_release
True
```

And look what happens when I executed only the parenthesis part: 

```bash
>>> ("redhat-8.6" or "redhat-8.8")
'redhat-8.6'
```

```suggestion
        if "redhat-8.6" in system_release or "redhat-8.8" in system_release:
```

r0x0d (commented):
Minor comment related to integration test. Otherwise, looks good.

tests/integration/common/checks-after-conversion/test_enabled_repositories.py
pr-watson: Good catch, I'll add this change thank you

convert2rhel/systeminfo.py
pr-watson: Okay so I can just add the debug info into the exception

hosekadam (approved):
Looks good to me.

Attaching the notes I have from testing the EUS conversion in our Vagrant box:
## Almalinux 8.6
Edit the `Vagrantfile` to get the needed version of box. For the Almalinux 8.6 it will look like the code bellow. The version of the box I got from: https://app.vagrantup.com/almalinux/boxes/8
```
when "alma8"
    BOX = "almalinux/8"
    VERSION = "8.6.20221001"
```
Then you need to edit the alma repos from latest to repos for 8.6. The URLs are from https://github.com/oamg/convert2rhel/blob/main/tests/ansible_collections/roles/hardcode-eus-version-repos/tasks/alma-86-repos.yml:
```
sed -i '/^\[appstream\]/,/^\[/ { s/^mirrorlist/#&/; s|^# baseurl=.*$|baseurl=https:\/\/repo.almalinux.org\/vault\/8.6\/AppStream\/$basearch\/os\/| }' /etc/yum.repos.d/almalinux.repo
sed -i '/^\[baseos\]/,/^\[/ { s/^mirrorlist/#&/; s|^# baseurl=.*$|baseurl=https:\/\/repo.almalinux.org\/vault\/8.6\/BaseOS\/$basearch\/os\/| }' /etc/yum.repos.d/almalinux.repo
sed -i '/^\[extras\]/,/^\[/ { s/^mirrorlist/#&/; s|^# baseurl=.*$|baseurl=https:\/\/repo.almalinux.org\/vault\/8.6\/extras\/$basearch\/os\/| }' /etc/yum.repos.d/almalinux.repo
```
Update the system, we need the latest kernel available in repos for 8.6 (4.18.0-372.26.1.el8_6 -> 4.18.0-372.32.1.el8_6) even the box is the latest for 8.6. There is few more packages which we can update too.
```
yum update -y
reboot
```
Copy the config file or use `export PR=XXX` during provisioning the machine:
```
cp /usr/lib/python3.6/site-packages/convert2rhel/data/8/x86_64/configs/almalinux-8-x86_64.cfg /usr/share/convert2rhel/configs/
```
And then run the conversion
After conversion you should see EUS repos after running `yum repolist`

## Rocky 8.6
Edit the `Vagrantfile` to get the needed version of box. For the Rocky 8.6 it will look like the code below. The version of the box I got from: https://app.vagrantup.com/eurolinux-vagrant/boxes/rocky-8 and it really the `8.6.9` for 8.6. When you get the `8.6.10` it's actually the 8.7
```
when "rocky8"
    BOX = "eurolinux-vagrant/rocky-8"
    VERSION = "8.6.9"
```
Edit the repos using https://github.com/oamg/convert2rhel/blob/main/tests/ansible_collections/roles/hardcode-eus-version-repos/tasks/rocky-86-repos.yml:
```
sed -i '/^\[appstream\]/,/^\[/ { s/^mirrorlist/#&/; s|^#baseurl=.*$|baseurl=https:\/\/dl.rockylinux.org\/vault\/rocky\/8.6\/AppStream\/$basearch\/os\/| }' /etc/yum.repos.d/Rocky-AppStream.repo
sed -i '/^\[baseos\]/,/^\[/ { s/^mirrorlist/#&/; s|^#baseurl=.*$|baseurl=https:\/\/dl.rockylinux.org\/vault\/rocky\/8.6\/BaseOS\/$basearch\/os\/| }' /etc/yum.repos.d/Rocky-BaseOS.repo
sed -i '/^\[extras\]/,/^\[/ { s/^mirrorlist/#&/; s|^#baseurl=.*$|baseurl=https:\/\/dl.rockylinux.org\/vault\/rocky\/8.6\/extras\/$basearch\/os\/| }' /etc/yum.repos.d/Rocky-Extras.repo
```
Copy the config file or use `export PR=XXX` during provisioning the machine:
```
cp /usr/lib/python3.6/site-packages/convert2rhel/data/8/x86_64/configs/rocky-8-x86_64.cfg /usr/share/convert2rhel/configs/
```
And then you can convert the system
After conversion, you should see EUS repos after running `yum repolist`

convert2rhel/unit_tests/systeminfo_test.py
```diff
@@ -311,22 +313,106 @@ def test_get_dbus_status_in_progress(monkeypatch, states, expected):
...
-def test_corresponds_to_rhel_eus_release(major, minor, expected, monkeypatch):
+def test_corresponds_to_rhel_eus_release(
+    major, minor, latest_distro_version, has_internet, repoquery_exception, expected, monkeypatch
+):
     version = Version(major, minor)
     monkeypatch.setattr(system_info, "version", version)
+    monkeypatch.setattr(system_info, "has_internet_access", has_internet)
+    if repoquery_exception:
+        return_code = 1
+        repoquery_output = "Unble to establish internet connection."
```
SpyTec, line 349:
```suggestion
        repoquery_output = "Unable to establish internet connection."
```

SpyTec (approved):
LGTM. There are a lot of unit tests though so might be a bit overkill, but not a blocker it's something we can see in the future when EUS date hits etc.

tests/integration/tier1/destructive/system-up-to-date/test_system_up_to_date.py
```diff
@@ -17,7 +17,11 @@ def test_skip_kernel_check(shell, convert2rhel):
     assert shell("mv /etc/yum.repos.d/* /tmp/s_backup/").returncode == 0
 
     # EUS version use hardcoded repos from c2r as well
-    if re.match(r"^(alma|rocky)-8\.6$", SYSTEM_RELEASE_ENV) or "centos-8-latest" in SYSTEM_RELEASE_ENV:
+    if (
+        re.match(r"^(alma|rocky)-8\.6$", SYSTEM_RELEASE_ENV)
+        or re.match(r"^(alma|rocky)-8\.8$", SYSTEM_RELEASE_ENV)
+        or "centos-8-latest" in SYSTEM_RELEASE_ENV
+    ):
```
danmyway, line 24:
```suggestion
    if (
        re.match(r"^(alma|rocky)-8\.[68]$", SYSTEM_RELEASE_ENV)
        or "centos-8-latest" in SYSTEM_RELEASE_ENV
    ):
```
This should work as well

tests/integration/common/checks-after-conversion/test_enabled_repositories.py
danmyway, line 31:
```suggestion
        if re.match(r"redhat-8\.[68]", system_release):
```
Alternatively, we could use `re` to match those.

danmyway (changes requested):
Just some minor changes.

tests/integration/tier1/destructive/system-up-to-date/test_system_up_to_date.py
pr-watson: Incorporated these changes in https://github.com/oamg/convert2rhel/pull/954/files

convert2rhel/systeminfo.py
```diff
@@ -403,10 +419,75 @@ def corresponds_to_rhel_eus_release(self):
...
+
+        if return_code != 0:
+            self.logger.debug(
+                "Repoquery execution exited with return code '%s' and with output: %s."
+                % (return_code, repoquery_output)
+            )
+            raise RepoqueryFailure(
+                "Repoquery execution exited with return code '%s' and with output: '%s'."
+                % (return_code, repoquery_output)
+            )
```
r0x0d, line 481: You can extract the text to a variable

pr-watson: Will do

tests/integration/common/checks-after-conversion/test_enabled_repositories.py
pr-watson: Incorporated these changes in https://github.com/oamg/convert2rhel/pull/954/files

SpyTec:
Ready to merge. @danmyway if #954 is fine with you lets merge this one right away

danmyway (changes requested):

convert2rhel/systeminfo.py
```diff
@@ -403,10 +419,73 @@ def corresponds_to_rhel_eus_release(self):
...
+        Compare current date and eus release date and return True if the current date is past the eus release date,
+        otherwise return False
+        """
+        if current_datetime > eus_release_date:
+            return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+        return False
+
+    def get_latest_distro_release_version(self):
+        """Extract minor version from release name using yum list"""
+        cmd = ["repoquery", "--setopt=exclude=", "--quiet", "--qf", "C2R\\t%{VERSION}", self.release_pkg]
```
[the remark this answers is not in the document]
r0x0d: Good point, @danmyway! I think that would be better suitable to be transformed into an action and make it a dependency of `RemoveRepositoryFilesPackages`

convert2rhel/systeminfo.py
```diff
@@ -243,6 +256,9 @@ def _get_excluded_pkgs(self):
     def _get_repofile_pkgs(self):
         return self._get_cfg_opt("repofile_pkgs").split()
 
+    def _get_release_package(self):
+        return self._get_cfg_opt("release_pkg")
+
```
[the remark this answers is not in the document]
pr-watson: Going to update configs so there will be no `\n`

convert2rhel/systeminfo.py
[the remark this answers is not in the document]
pr-watson: I was able to add an lru_cache to get_latest_distro_release_version to save the return once the first call is successful, is that an okay solution to this?

convert2rhel/systeminfo.py
```diff
@@ -51,12 +55,19 @@
     "7.9": "7Server",
 }
 
-# List of EUS minor versions supported
-EUS_MINOR_VERSIONS = ["8.6"]
+# Dictionary of EUS minor versions supported and their EUS period start date
+EUS_MINOR_VERSIONS = {
+    "8.6": "2022-11-09",
+    "8.8": "2022-11-14",
```
danmyway, line 61:
```suggestion
    "8.8": "2023-11-14",
```
Typo here

pr-watson: Leftover from testing, will push new version soon

convert2rhel/systeminfo.py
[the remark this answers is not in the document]
pr-watson: Ended up creating a new attribute under resolve_system_info to save the return of corresponds_to_rhel_eus_release and reference that attribute on the previous calls to corresponds_to_rhel_eus_release.

convert2rhel/repo.py
```diff
@@ -39,8 +39,7 @@ def get_rhel_repoids():
     we preferably enable the RHEL EUS repoids as those provide security updates over two years, in comparison to 6 months
     in case of the standard non-EUS repoids.
     """
-
-    if system_info.corresponds_to_rhel_eus_release():
+    if system_info.eus_system:
```
r0x0d, line 42: Not entirely needed, but since we simplified this to be a property, maybe we could simplify this line to be an inline if statement? something like: 


```suggestion
    repos_needed = system_info.eus_rhsm_repoids if system_info.eus_system else system_info.default_rhsm_repoids
```

convert2rhel/systeminfo.py
```diff
@@ -403,10 +424,73 @@ def corresponds_to_rhel_eus_release(self):
...
+            return self.datetime_comparison_returns(current_datetime, eus_release_date)
+
+        try:
+            latest_release_version = self.get_latest_distro_release_version()
+        except RepoqueryFailure:
+            return self.datetime_comparison_returns(current_datetime, eus_release_date)
+
+        _, latest_minor = latest_release_version.split(".")
+        if int(latest_minor) > int(self.version.minor):
+            return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
```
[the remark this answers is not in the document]
pr-watson: Nice suggestion

convert2rhel/repo.py
pr-watson: Sounds good to me!

convert2rhel/systeminfo.py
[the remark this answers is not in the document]
pr-watson: There is a separate function that returns the same value, should we send that as an argument to the function as well?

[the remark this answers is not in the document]
r0x0d: I think it won't hurt, yeah

convert2rhel/actions/system_checks/eus.py
```diff
@@ -0,0 +1,53 @@
...
+from convert2rhel import actions, grub, systeminfo
+from convert2rhel.systeminfo import EUS_MINOR_VERSIONS, system_info
+from convert2rhel.toolopts import tool_opts
+
+
+logger = logging.getLogger(__name__)
+
+
+class EusSystemCheck(actions.Action):
+    id = "EUS"
```
r0x0d, line 30: Lets update the id as well to match the class name

convert2rhel/actions/system_checks/eus.py
```diff
@@ -0,0 +1,53 @@
+# Copyright(C) 2016 Red Hat, Inc.
```
r0x0d, line 1: You can update the copyright to be 2023 as well

convert2rhel/actions/system_checks/eus.py
```diff
@@ -0,0 +1,55 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+
+import logging
+
+from datetime import datetime
+
+from convert2rhel import actions, grub, systeminfo
```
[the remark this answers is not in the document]
r0x0d: grub is not beign used

convert2rhel/actions/system_checks/eus.py
```diff
@@ -0,0 +1,53 @@
...
+    id = "EUS"
+
+    def run(self):
+        """Inhibit the conversion when we are not able to handle UEFI."""
+        super(EusSystemCheck, self).run()
+
+        if "%s.%s" % (system_info.version.major, system_info.version.minor) == "8.8":
+            eus_release_date = EUS_MINOR_VERSIONS.get(
+                "%s.%s" % (system_info.version.major, system_info.version.minor), False
+            )
```
r0x0d, line 39:
```suggestion
        current_version = "%s.%s" % (system_info.version.major, system_info.version.minor) 
        if current_version == "8.8":
            eus_release_date = EUS_MINOR_VERSIONS.get(current_version, False)
```

pr-watson:
Tested each scenario in an almalinux VM on 8.6 and 8.8. 8.8 systems will use non-eus repos by default, the --eus commandline option is required for it to use eus repos. If an 8.8 system runs a conversion when the 8.8 eus release date is passed without having the --eus option enabled, a warning is shown to the user in the preconversion analysis report. 8.6 systems automatically are recognized as eus systems and will not trigger the eus commandline warning message in the preconversion analysis report.

convert2rhel/actions/system_checks/eus.py
pr-watson: I'll change this, before I had current_version = "%s.%s" % (system_info.version.major, system_info.version.minor) in EUS_MINOR_VERSIONS so I would get a boolean return instead of the actual value

convert2rhel/systeminfo.py
```diff
@@ -403,10 +414,19 @@ def corresponds_to_rhel_eus_release(self):
...
         """
-        return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+        current_version = "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+        if tool_opts.eus and current_version:
+            self.logger.info("EUS argument detected, automatically evaluating system as EUS")
+            return current_version
+
+        if "%s.%s" % (self.version.major, self.version.minor) == "8.6":
+            print("WE ARE ATOMIC 8.6")
+            return current_version
```
r0x0d, line 427: I think this can be before everything else, if we detect we are in 8.6, then just let's return True and move on.

Also, remove the print

convert2rhel/systeminfo.py
```diff
@@ -403,10 +414,19 @@ def corresponds_to_rhel_eus_release(self):
         For example if we detect CentOS Linux 8.4, the corresponding RHEL 8.4 is an EUS release, but if we detect
         CentOS Linux 8.5, the corresponding RHEL 8.5 is not an EUS release.
 
-        :return: Whether or not the current system has a EUS correspondent in RHEL.
+        :return: Whether or not the current system has an EUS correspondent in RHEL.
         :rtype: bool
         """
-        return "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
+        current_version = "%s.%s" % (self.version.major, self.version.minor) in EUS_MINOR_VERSIONS
```
r0x0d, line 420: This current_version being a boolean is kinda of wrong here... I think that we can remove the `in EUS_MINOR_VERSIONS` and let the variable behave just like a string that has the composition of the versions.

convert2rhel/toolopts.py
```diff
@@ -188,6 +189,11 @@ def _register_options(self):
             " to show you what rpm files have been affected by the conversion."
             % (PRE_RPM_VA_LOG_FILENAME, POST_RPM_VA_LOG_FILENAME),
         )
+        self._shared_options_parser.add_argument(
+            "--eus",
+            action="store_true",
+            help="Automatically recognize the system as eus, utilizing eus repos",
```
r0x0d, line 195: Wondering if we should clarify in here that this doesn't affect 8.6 systems, only 8.8+

convert2rhel/unit_tests/actions/system_checks/eus_test.py
```diff
@@ -0,0 +1,63 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+
+from datetime import datetime
+from unittest.mock import MagicMock
+
+import pytest
+
+from convert2rhel import actions, subscription, systeminfo, unit_tests
```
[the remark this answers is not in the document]
r0x0d: unused imports

r0x0d (commented):
few comments

convert2rhel/toolopts.py
pr-watson: Sure I can mention it

convert2rhel/systeminfo.py
pr-watson: Ok sounds good

convert2rhel/systeminfo.py
pr-watson: Alright sounds good

convert2rhel/systeminfo.py
pr-watson: Well the `in EUS_MINOR_VERSIONS` is so we know that we are dealing with systems that are listed in that dictionary

r0x0d: You won't get rid of the `EUS_MINOR_VERSIONS` check. You will only reorganize it a bit so the `current_version` variable actually represents a version instead of a flag. 

```suggestion
        current_version = "%s.%s" % (self.version.major, self.version.minor)
```

Later in the code, 

```python
        if current_version == "8.6":
            return current_version
```

and, 

```python
        if tool_opts.eus and current_version in EUS_MINOR_VERSION:
            self.logger.info("EUS argument detected, automatically evaluating system as EUS")
            return current_version
```

convert2rhel/actions/system_checks/eus.py
```diff
@@ -0,0 +1,52 @@
...
+            current_datetime = datetime.date.today()
+
+            # warning message if the eus release date is past and the --eus option is not set
+            if not tool_opts.eus and current_datetime > eus_release_date:
+                self.add_message(
+                    level="WARNING",
+                    id="EUS_COMMAND_LINE_OPTION_UNUSED",
+                    title="The --eus command line option is unused",
+                    description="Current system version is under Extended Lifecycle Support (EUS). You may want to consider using the --eus"
+                    " command line option to land on a system pached with the latest security errata.",
```
r0x0d, line 50:
```suggestion
                    " command line option to land on a system patched with the latest security errata.",
```

convert2rhel/actions/system_checks/eus.py
```diff
@@ -0,0 +1,52 @@
...
+
+class EusSystemCheck(actions.Action):
+    id = "EUS_SYSTEM_CHECK"
+
+    def run(self):
+        """Inhibit the conversion when we are not able to handle UEFI."""
+        super(EusSystemCheck, self).run()
+
+        current_version = "%s.%s" % (system_info.version.major, system_info.version.minor)
+        if current_version == "8.8":
```
r0x0d, line 37: I don't think this will be a blocker, but I'm wondering if we should modify this a bit to actually give the warning on another EUS minor versions instead of just 8.8 (for example, 8.10, 9.2, 9.4, ...)

Maybe we could skip the first item in the dictionary in the `EUS_MINOR_VERSIONS` and check if the `current_version` is in `EUS_MINOR_VERSIONS.keys()`.

For example, that might work: 

```python
>>> current_version = "8.8"
>>> EUS_MINOR_VERSIONS.update({"8.10": "2023-11-14"})
>>> EUS_MINOR_VERSIONS
{'8.6': '2022-11-09', '8.8': '2023-11-14', '8.10': '2023-11-14'}
>>> eus_versions = list(EUS_MINOR_VERSIONS.keys())[1:]
>>> eus_versions
['8.8', '8.10']
>>> current_version in eus_versions
True
```

r0x0d: But as I mentioned, this might not be needed for now and we probably can leave this for another release

convert2rhel/systeminfo.py
```diff
@@ -24,6 +24,9 @@
 import time
 
 from collections import namedtuple
+from datetime import datetime as dt
+
+import rpm
```
r0x0d, line 29:
```suggestion
```

r0x0d: not being used

r0x0d (changes requested):
Few comments again, but this time it will be ready to be merged!

convert2rhel/actions/system_checks/eus.py
pr-watson: Good idea, would save us the trouble of having to update in the future

convert2rhel/systeminfo.py
danmyway: also the `from datetime import datetime as dt` is not being used

convert2rhel/actions/system_checks/eus.py
```diff
@@ -0,0 +1,52 @@
...
+            eus_release_date = EUS_MINOR_VERSIONS.get(current_version, False)
+            # Turn eus_release_date into a datetime object
+            eus_release_date = datetime.datetime.strptime(eus_release_date, "%Y-%m-%d")
```
danmyway, line 40:
```suggestion
            eus_release_date = datetime.datetime.strptime(eus_release_date, "%Y-%m-%d").date()
```
This needs to be converted to the `date` object, to be [comparable with the current_datetime here.](https://github.com/oamg/convert2rhel/pull/947/files#diff-1470230e5916d6ab112f1569ea65dc18608f566182e2c0ae0a33cbe0b2a0e7a4R44)

danmyway (changes requested):

convert2rhel/systeminfo.py
r0x0d: good catch

pr-watson:
Applied suggestions

convert2rhel/unit_tests/actions/system_checks/eus_test.py
```diff
@@ -0,0 +1,63 @@
...
+    @classmethod
+    def today(cls):
+        return cls(2023, 11, 15)
+
+
+class TestEus:
+    @centos8
+    def test_eus_warning_message(self, eus_action, monkeypatch, global_tool_opts, pretend_os):
+
+        version = Version(8, 8)
```
r0x0d, line 43: I think we should transform this into a pytest.parametrize to cover that 8.8, 8.10 (and 9.2?) will be covered by the test… 

Would be good to have a case where the version is set to 8.6 too, just to verify that this action is not doing something incorrectly.

pr-watson: Sounds good

convert2rhel/unit_tests/actions/system_checks/eus_test.py
```diff
@@ -0,0 +1,73 @@
...
+                    " command line option to land on a system patched with the latest security errata.",
+                ),
+            )
+        )
+
+        if message_reported:
+            assert expected.issuperset(eus_action.messages)
+            assert expected.issubset(eus_action.messages)
+        else:
+            eus_action.messages = []
```
r0x0d, line 73: why are we assigning instead of asserting here?

pr-watson: Fixing now, meant to assert

r0x0d (approved):

convert2rhel/actions/system_checks/eus.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class EusSystemCheck(actions.Action):
+    id = "EUS_SYSTEM_CHECK"
+
+    def run(self):
+        """Inhibit the conversion when we are not able to handle UEFI."""
```
hosekadam, line 33: Probably comment from elsewhere?

convert2rhel/pkghandler.py
```diff
@@ -677,7 +677,7 @@ def _get_package_repositories(pkgs):
 
     output, retcode = utils.run_subprocess(
         ["repoquery", "--quiet", "-q"] + pkgs + ["--qf", query_format],
-        print_cmd=False,
+        print_cmd=True,
```
hosekadam, line 680: Isn't the print of the command some leftover from testing?

pr-watson: Michal said we might as well always print it

convert2rhel/actions/system_checks/eus.py
pr-watson: Good catch let me fix

convert2rhel/pkghandler.py
bocekm: Yeah, it's a debug print that might help with finding problems.

PR #948, opened 2023-09-30
Scan for UNKNOWN_ERRORS that resulted from this integration test run.

abadger:
There are times that Actions return unknown errors because of a bug.  This integration test catches those errors and raises an exception to mark the test as a failure.

Checklist

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -140,6 +140,62 @@ def convert2rhel(shell):
             if unregister:
                 shell("subscription-manager unregister")
 
+            # Scan for unknown errors in the pre-conversion assessment.
+            try:
+                with open("JSON") as f:
+                    assessment = json.read(f)
+            except Exception:
+                # We only check for unknown errors when the run created the
+                # json formatted assessment.
+                pass
+            else:
+                message = []
+                for action in assessment:
+                    if action["result"].id == "UNKNOWN_ERROR":
+                        message.extend(
+                            (
+                                "== Action caught SystemExit and returned an UNKNOWN_ERROR:",
+                                "%s: %s" % (action.id, action["result"]),
+                            )
+                        )
+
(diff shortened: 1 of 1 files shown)


tests/integration/conftest.py
```diff
@@ -140,6 +140,62 @@ def factory(
...
+                            )
+                        )
+
+                    elif action["result"].id == "UNEXPECTED_ERROR":
+                        message.extend(
+                            (
+                                "== Action Framework caught SystemExit and returned an UNEXPECTED_ERROR:",
+                                "%s: %s" % (action.id, action["result"]),
+                            )
+                        )
```
r0x0d, line 168: I know this won't live here for so long, but could we do this instead?

```suggestion
                    if action["result"].id in ("UNKNOWN_ERROR", "UNEXECTED_ERROR"):
                        message.extend(
                            (
                                "== Action Framework caught SystemExit and returned an %s:",
                                "%s: %s" % (action["result"].id, action.id, action["result"]),
                            )
                        )
```

abadger: I would prefer to keep them separate since they're two different cases.

`UNKNOWN_ERROR` is for the places in `Actions` where we need to port to away from `sys.exit()/logger.critical()`.  Once we finish the port we can get rid of that.  `UNEXPECTED_ERROR` is where the Action framework is catching an exception that was unhandled by the Action plugin.  These are bugs in the Actions that we need to diagnose and fix.  This check should remain and anytime it trips, we need to look at fixing the Action that's generating it.

(Note: I see that the `UNEXPECTED_ERROR` message says it happens when SystemExit occurs.  I'll update it to say that it happens when any Exception is raised.)

r0x0d: Alright, sounds reasonable to me.

danmyway (approved):
Looks great, thanks @abadger !

PR #953, opened 2023-10-11
[RHELC-1180] Check for firewalld system unit

r0x0d:
Firewalld is having problems with the python-nftables package after we repalce the packages in the system to their rhel counterparts. This commit introduces a way to check if firewalld is available and running, and if so, we ask the user to stop it and re-enable after the conversion.

If firewalld is left running during the conversion, the user will probably see a set of errors happening in the systemctl logs for firewalld, as well, they will be unable to ssh into the machine after the conversion is done.

Jira Issues: [RHELC-1180](https://issues.redhat.com/browse/RHELC-1180)

Checklist


convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,57 @@
...
+        if system_info.version.major < 8:
+            description = "Skipping the check as it is relevant only for Oracle/Alma/Rocky Linux 8."
+            logger.info(description)
+            self.add_message(
+                level="INFO",
+                id="CHECK_FIREWALLD_AVAILABILITY_SKIP",
+                title="Skipping the check for firewalld availability.",
+                description=description,
+            )
+            return
```
SpyTec, line 44: @dkubek @r0x0d does this happen on all RHEL 8 systems or just 8.8? Can we limit it to 8.8 only?

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,57 @@
...
+
+class CheckFirewalldAvailability(actions.Action):
+    id = "CHECK_FIREWALLD_AVAILABILITY"
+
+    def run(self):
+        """Error out if the firewalld service is running on the system."""
+        super(CheckFirewalldAvailability, self).run()
+        logger.task("Prepare: Check that firewalld is running")
+
+        if system_info.version.major < 8:
```
SpyTec, line 35:
```suggestion
        if system_info.version.major != 8:
```
```suggestion
        if system_info.version.major != 8 and system_info.minor != 8:
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
r0x0d: I will take a look

r0x0d: Looks like it happens only for 8.8. 

Tried with 8.6 conversions in OL and seems that it is not happening. It is probably related to the version rather than the system itself. I will try to investigate this next week.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,57 @@
...
+
+class CheckFirewalldAvailability(actions.Action):
+    id = "CHECK_FIREWALLD_AVAILABILITY"
+
+    def run(self):
+        """Error out if the firewalld service is running on the system."""
+        super(CheckFirewalldAvailability, self).run()
+        logger.task("Prepare: Check that firewalld is running")
+
+        if system_info.version.major != 8 and system_info.minor != 8:
```
r0x0d, line 35:
```suggestion
        if system_info.id == "oracle" and (system_info.version.major == 8 and system_info.minor == 8):
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
danmyway: It's related to the packages since updates in ~ mid September '23. Not really system, nor minor version related.
There have been some updates since then, and it seems, that whatever is causing this got fixed for Alma and Rocky Linux.

SpyTec (approved):

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,58 @@
...
+        super(CheckFirewalldAvailability, self).run()
+        logger.task("Prepare: Check that firewalld is running")
+
+        if system_info.id == "oracle" and (system_info.version.major == 8 and system_info.version.minor >= 8):
+            if systeminfo.is_systemd_managed_service_running("firewalld"):
+                self.set_result(
+                    level="ERROR",
+                    id="FIREWALLD_DAEMON_RUNNING",
+                    title="Firewalld is running",
+                    description="Firewalld is running and can cause problem during the package replacement phase.",
```
mportman12, line 41:
```suggestion
                    description="Firewalld is running and can cause problems during the package replacement phase.",
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,58 @@
...
+                    diagnosis="Firewalld daemon unit is running and it may cause problems during package replacement in later phases.",
```
mportman12, line 42:
```suggestion
                    diagnosis="Firewalld daemon unit is running and might cause problems during package replacement in later phases.",
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,58 @@
...
+                    remediation="Please stop firewalld using `systemctl stop firewalld` until the conversion is done, and then, start it again with `systemctl start firewalld`",
```
mportman12, line 43:
```suggestion
                    remediation="Stop firewalld by using the `systemctl stop firewalld` command. After the conversion, restart firewalld using the `systemctl start firewalld` command.",
```

mportman12 (commented):
Left suggested edits.

tests/integration/tier0/non-destructive/firewalld-inhibitor/test_firewalld_inhibitor.py
```diff
@@ -0,0 +1,23 @@
...
+    with convert2rhel("-y --no-rpm-va --debug", unregister=True) as c2r:
+        c2r.expect(
+            "CHECK_FIREWALLD_AVAILABILITY::FIREWALLD_DAEMON_RUNNING - Firewalld is running",
+            timeout=600,
+        )
+
+    assert c2r.exitstatus == 1
+
+    # Clean up
+    assert shell("systemctl stop firewalld").returncode == 0
```
danmyway, line 23: I'd say, that cleaning up the system would be having the firewalld service running and enabled, since that is a default state of the image we test.
Considering these are non-destructive tests, we will never get past the point of no return, therefore the service won't cause the known issue.
On the other hand, if there is any yet undiscovered issue with the service before the PONR and/or any regression introduced, we should have the service running IMHO.
```suggestion

```

danmyway (changes requested):
A couple of nitpicks, otherwise looks, good, thank you!

tests/integration/tier0/non-destructive/firewalld-inhibitor/main.fmf
```diff
@@ -0,0 +1,22 @@
+summary: |
+    Check that the the c2r inhibits in certain scenario when the firewalld is running
+description: |
+    On Oracle Linux 8.8 there is an issue with the running firewalld process
+    that breaks the conversion. In such case the inhibitor has to be raised.
+
+tier: 0
```
danmyway, line 7:
```suggestion
tier: 0
```
Also unnecessary, already defined in the [root directory.](https://github.com/oamg/convert2rhel/blob/main/tests/integration/tier0/main.fmf#L8)

tests/integration/tier0/non-destructive/firewalld-inhibitor/test_firewalld_inhibitor.py
```diff
@@ -0,0 +1,23 @@
...
+@pytest.mark.test_firewalld_inhibitor
+def test_firewalld_inhibitor(shell, convert2rhel):
+    """
+    Verify that on the OL8.8 the conversion is inhibited due to
+    running firewalld on the system.
+    The reference ticket: https://issues.redhat.com/browse/RHELC-1180
+    """
+    assert shell("rpm -q firewalld").returncode == 0
+    assert shell("systemctl start firewalld").returncode == 0
+
```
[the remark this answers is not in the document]
kokesak: Added.

tests/integration/tier0/non-destructive/firewalld-inhibitor/test_firewalld_inhibitor.py
kokesak: I see your point, however, this "cleanup" should not happen in this test. What I mean is that the test should return to the state it was before the actual test, which was that the `firewalld` was stopped.

tests/integration/tier0/non-destructive/firewalld-inhibitor/main.fmf
```diff
@@ -0,0 +1,22 @@
+summary: |
+    Check that the the c2r inhibits in certain scenario when the firewalld is running
+description: |
+    On Oracle Linux 8.8 there is an issue with the running firewalld process
+    that breaks the conversion. In such case the inhibitor has to be raised.
```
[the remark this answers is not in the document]
kokesak: Changed.

tests/integration/tier0/non-destructive/firewalld-inhibitor/test_firewalld_inhibitor.py
danmyway: Since we will remove the whole case in a foreseeable future, I'd say leave it this way.

danmyway (approved):
LGTM, thanks!

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,62 @@
...
+                self.set_result(
+                    level="ERROR",
+                    id="FIREWALLD_RUNNING",
+                    title="Firewalld is running",
+                    description="Firewalld is running and can cause problems during the package replacement phase.",
+                    diagnosis="We've detected that firewalld unit is running and might cause problems related to kernel modules after the conversion is done.",
+                    remediation=(
+                        "Stop firewalld by using the `systemctl stop firewalld` command. This will prevent errors while convert2rhel replaces the system packages"
+                        " and the kernel, whoever, that might not prevent errors from appearing the firewalld logs after the conversion. After the conversion is done,"
+                        " reboot the system to load the new RHEL kernel modules and start firewalld with `systemctl start firewalld` once more."
```
bocekm, line 46: Do I recall correctly that stopping the firewall does not help and instead the user needs to change the `CleanupModulesOnExit` firewalld setting or reboot to have the firewall working fine?

I suggest providing the customer with a choice in a pre-conversion report warning message:
- either they manually change the config and then firewalld will work as expected right after convert2rhel finishes,
- or they need to reboot in order to have the firewall working after the conversion

bocekm (changes requested):

convert2rhel/actions/system_checks/check_firewalld_availability.py
kokesak: @bocekm the first bullet option is correct, the second not so much. They have to stop the firewalld before the conversion, otherwise, the conversion process breaks in the kernel replacement and the system is "dead". So it is required to stop it before. In this case, some rhel kernel modules are loaded into the still-running oracle-kernel. So for the firewall to work correctly again the user has to reboot after the conversion (which he should do anyway). Hopefully that lightness up the situation

r0x0d: +1 to @kokesak

bocekm: From @danmyway:
> Rebooting is essential, however you need to get through with the conversion first and that can be done either
1/ stop the firewalld (disable the service to make sure it wont restart during the replacement)
or
2/ change the option CleanupModulesOnExit = yes  to CleanupModulesOnExit = no and keep the firewalld running

Let's add the two options Daniel mentioned to the remediation text so the customer can choose what works better for them.

bocekm: My remediation proposal:
```
"Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf to no prior to running convert2rhel:

sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf

You can change the option back to yes after the system reboot - that is after the system boots into the RHEL kernel."
```

bocekm (changes requested):

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,62 @@
...
+        logger.task("Prepare: Check that firewalld is running")
+
+        if system_info.id == "oracle" and (system_info.version.major == 8 and system_info.version.minor >= 8):
+            if systeminfo.is_systemd_managed_service_running("firewalld"):
+                self.set_result(
+                    level="ERROR",
+                    id="FIREWALLD_RUNNING",
+                    title="Firewalld is running",
+                    description="Firewalld is running and can cause problems during the package replacement phase.",
+                    diagnosis="We've detected that firewalld unit is running and might cause problems related to kernel modules after the conversion is done.",
```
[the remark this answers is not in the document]
r0x0d: applied.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,109 @@
...
+                        title="Firewalld is running",
+                        description=(
+                            "We've detected that firewalld is running and we couldn't find check for the CleanupModulesOnExit configuration. "
+                            "This means that a reboot will be necessary after the conversion is done to reload the kernel modules and prevent firewalld from stop working."
+                        ),
+                    )
+                    return
+
+            description = "Firewalld service reported that it is not running."
+            logger.info()
```
r0x0d, line 92:
```suggestion
            logger.info(description)
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,109 @@
...
+            contents = [line.strip() for line in handler.readlines() if line]
+
+        # Contents list is empty for some reason, better to assume that there
+        # is no content in the file that was read.
+        if not contents:
+            return False
+
+        # If the config file has this option set to true/yes, then we need to
+        # return True to ask the user to change it to self.
+        if "CleanupModulesOnExit=yes" in contents or "CleanupModulesOnExit=true" in contents:
```
bocekm, line 49: Use regex here to expect whitespaces.
```
"CleanupModulesOnExit\s*=\s*yes"
```

I've just tried on RHEL 8 that:
1. `CleanupModulesOnExit = yes` works ok (firewalld doesn't complain after service restart)
2. Multiline ('\n') does not work:
```
CleanupModulesOnExit=
  yes
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,109 @@
...
+        """Error out if the firewalld service is running on the system."""
+        super(CheckFirewalldAvailability, self).run()
+        logger.task("Prepare: Check that firewalld is running")
+
+        if system_info.id == "oracle" and (system_info.version.major == 8 and system_info.version.minor >= 8):
+            if systeminfo.is_systemd_managed_service_running("firewalld"):
+                if _check_for_modules_cleanup_config():
+                    self.set_result(
+                        level="ERROR",
+                        id="FIREWALLD_MODULESS_CLEANUP_ON_EXIT_CONFIG",
```
bocekm, line 68:
```suggestion
                        id="FIREWALLD_MODULES_CLEANUP_ON_EXIT_CONFIG",
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,109 @@
...
+from convert2rhel import actions, systeminfo
+from convert2rhel.systeminfo import system_info
+
+
+logger = logging.getLogger(__name__)
+
+FIREWALLD_CONFIG_FILE = "/etc/firewalld/firewalld.conf"
+
+
+def _check_for_modules_cleanup_config():
```
bocekm, line 30: The function returns boolean so better to name it as a question.
```suggestion
def _is_modules_cleanup_enabled():
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,109 @@
...
+                    )
+                else:
+                    # Firewalld is running but the configuration for CleanupModulesOnExit was not set to true/yes
+                    self.add_message(
+                        level="WARNING",
+                        id="FIREWALLD_IS_RUNNING",
+                        title="Firewalld is running",
+                        description=(
+                            "We've detected that firewalld is running and we couldn't find check for the CleanupModulesOnExit configuration. "
+                            "This means that a reboot will be necessary after the conversion is done to reload the kernel modules and prevent firewalld from stop working."
```
bocekm, line 86: The default is yes so if the option is missing in the config, I expect that to have the same effect as if the option is set to yes.
I've just verified that removing the option does not lead to firewall complaining.
So, in the `_check_for_modules_cleanup_config` function, you should treat the missing option as if it's set to yes.

bocekm (changes requested):

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,109 @@
...
+                    return
+
+            description = "Firewalld service reported that it is not running."
+            logger.info()
+            self.add_message(
+                level="INFO",
+                id="FIREWALLD_IS_NOT_RUNNING",
+                title="Firewalld not running",
+                description=description,
+            )
```
r0x0d, line 98:
```suggestion
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,109 @@
...
+            return
+
+        description = "Skipping the check as it is relevant only for Oracle Linux 8.8 and above."
+        logger.info(description)
+        self.add_message(
+            level="INFO",
+            id="CHECK_FIREWALLD_AVAILABILITY_SKIP",
+            title="Skipping the check for firewalld availability.",
+            description=description,
+        )
```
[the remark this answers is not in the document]
r0x0d: we have this type of message in the dbus action, but I agree they don't aggregate any value to the user. I will remove.

r0x0d, line 108:
```suggestion
        logger.info(Skipping the check as it is relevant only for Oracle Linux 8.8 and above.)
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,109 @@
...
+                if _check_for_modules_cleanup_config():
+                    self.set_result(
+                        level="ERROR",
+                        id="FIREWALLD_MODULESS_CLEANUP_ON_EXIT_CONFIG",
+                        title="Firewalld is set to cleanup modules after exit.",
+                        description="Firewalld running on Oracle Linux 8 can lead to a conversion failure.",
+                        diagnosis="We've detected that firewalld unit is running and that causes iptables and nftables failures on Oracle Linux 8 and under certain conditions it can lead to a conversion failure.",
+                        remediation=(
+                            "Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf to no prior to running convert2rhel:\n"
+                            "1. sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n"
```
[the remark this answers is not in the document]
r0x0d: Thinking about it right now, it looks strange.. Will remove

convert2rhel/actions/system_checks/check_firewalld_availability.py
r0x0d: updated

r0x0d:
The unit tests are still failing because of the way we are reading the file. I will take care of this tomorrow morning. Wanted to push so folks could look at it again and suggest modifications before we are done with this.

r0x0d:
@bocekm ^

kokesak:
@r0x0d now it doesn't inhibits the conversion when the  firewalld is running and no changes to the config were made:
```
                    ========== Warning (Review and fix if needed) ==========
                    (WARNING) CHECK_FIREWALLD_AVAILABILITY::FIREWALLD_IS_RUNNING - Firewalld is
                        running
                         Description: We've detected that firewalld is running and we couldn't find
                    check for the CleanupModulesOnExit configuration. This means that a reboot will
                    be necessary after the conversion is done to reload the kernel modules and
                    prevent firewalld from stop working.
                         Diagnosis: N/A
                         Remediation: N/A
```
It is just a warning and the conversion continues and breaks

r0x0d:
> @r0x0d now it doesn't inhibits the conversion when the firewalld is running and no changes to the config were made:
> 
> ```
>                     ========== Warning (Review and fix if needed) ==========
>                     (WARNING) CHECK_FIREWALLD_AVAILABILITY::FIREWALLD_IS_RUNNING - Firewalld is
>                         running
>                          Description: We've detected that firewalld is running and we couldn't find
>                     check for the CleanupModulesOnExit configuration. This means that a reboot will
>                     be necessary after the conversion is done to reload the kernel modules and
>                     prevent firewalld from stop working.
>                          Diagnosis: N/A
>                          Remediation: N/A
> ```
> 
> It is just a warning and the conversion continues and breaks

That might be for the way that I'm reading the config file. I will fix it today.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,113 @@
...
+    id = "CHECK_FIREWALLD_AVAILABILITY"
+
+    def run(self):
+        """Error out if the firewalld service is running on the system."""
+        super(CheckFirewalldAvailability, self).run()
+        logger.task("Prepare: Check that firewalld is running")
+
+        if system_info.id == "oracle" and (system_info.version.major == 8 and system_info.version.minor >= 8):
+            if systeminfo.is_systemd_managed_service_running("firewalld"):
+                if _is_modules_cleanup_enabled():
```
bocekm, line 81:
```suggestion
        if system_info.id != "oracle" and system_info.version.major != 8 and system_info.version.minor < 8:
            logger.info("Skipping the check as it is relevant only for Oracle Linux 8.8 and above.")
            return
        if not systeminfo.is_systemd_managed_service_running("firewalld"):
            logger.info("Firewalld service reported that it is not running.")
            return
        if _is_modules_cleanup_enabled():
```
Just a style nitpick - not going neck deep into nested conditions.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,113 @@
...
+                    self.set_result(
+                        level="ERROR",
+                        id="FIREWALLD_MODULES_CLEANUP_ON_EXIT_CONFIG",
+                        title="Firewalld is set to cleanup modules after exit.",
+                        description="Firewalld running on Oracle Linux 8 can lead to a conversion failure.",
+                        diagnosis="We've detected that firewalld unit is running and that causes iptables and nftables failures on Oracle Linux 8 and under certain conditions it can lead to a conversion failure.",
+                        remediation=(
```
bocekm, line 88: Limit the line lenght of both the diagnosis and remediation to 120 chars please.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,113 @@
...
+                            "Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf to no prior to running convert2rhel: "
+                            "sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n"
```
bocekm, line 90:
```suggestion
                            "Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf to no prior to running convert2rhel:\n"
                            " sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n"
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Contents list is empty for some reason, better to assume that there
+    # is no content in the file that was read.
+    if not contents:
+        return False
+
+    # If the config file has this option set to true/yes, then we need to
+    # return True to ask the user to change it to self.
+    if list(filter(CLEANUP_MODULES_ON_EXIT_REGEX.match, contents)):
+        logger.debug("Found CleanupModulesOnExit option in firewalld.conf")
```
bocekm, line 59:
```suggestion
        logger.debug("CleanupModulesOnExit option enabled in %s" % FIREWALLD_CONFIG_FILE)
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,113 @@
...
+        return False
+
+    contents = ""
+    with open(FIREWALLD_CONFIG_FILE, mode="r") as handler:
+        contents = [line.strip() for line in handler.readlines() if line.strip()]
+
+    # Contents list is empty for some reason, better to assume that there
+    # is no content in the file that was read.
+    if not contents:
+        return False
```
bocekm, line 54:
```suggestion
    if not contents:
        logger.debug("%s is empty." % FIREWALLD_CONFIG_FILE)
        return True
```
This should be True I think. Because an empty file should mean all defaults. And the option CleanupModulesOnExit has yes as a default.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,113 @@
...
+                        remediation=(
+                            "Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf to no prior to running convert2rhel: "
+                            "sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n"
+                            "You can change the option back to yes after the system reboot - that is after the system boots into the RHEL kernel."
+                        ),
+                    )
+                else:
+                    # Firewalld is running but the configuration for CleanupModulesOnExit was not set to true/yes
+                    description = (
+                        "We've detected that firewalld is running and we couldn't find check for the CleanupModulesOnExit configuration. "
```
bocekm, line 97: In my understanding, when `_is_modules_cleanup_enabled` returns False it should means that the CleanupModulesOnExit option is set to false.

bocekm (changes requested):

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,119 @@
...
+        if _is_modules_cleanup_enabled():
+                    self.set_result(
+                        level="ERROR",
+                        id="FIREWALLD_MODULES_CLEANUP_ON_EXIT_CONFIG",
+                        title="Firewalld is set to cleanup modules after exit.",
+                        description="Firewalld running on Oracle Linux 8 can lead to a conversion failure.",
+                        diagnosis="We've detected that firewalld unit is running and that causes iptables and nftables failures on Oracle Linux 8 and under certain conditions it can lead to a conversion failure.",
+                        remediation=(
+                            "Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf to no prior to running convert2rhel:\n"
+                            " sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n"
```
kokesak, line 96: @r0x0d I remember you told me that even the `firewall-cmd --reload` command should be used. Shouldn't we mention this here as well?

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,119 @@
...
+                            "You can change the option back to yes after the system reboot - that is after the system boots into the RHEL kernel."
+                        ),
+                    )
+                else:
```
hosekadam, line 100: Bad indent here

convert2rhel/actions/system_checks/check_firewalld_availability.py
r0x0d: good catch!

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,133 @@
...
+        logger.debug("%s is empty." % FIREWALLD_CONFIG_FILE)
+        return True
+
+    # If the CleanupModulesOnExit is not present inside the contents list, we
+    # can return True since the default behavior for firewalld is to consider
+    # CleanupModulesOnExit as true. Ignoring # and ; as they are ignored in the
+    # config parser for firewalld as well:
+    # https://github.com/firewalld/firewalld/blob/46d54dcbdff94423686d67befc78ca8d601fce60/src/firewall/core/io/firewalld_conf.py#L85
+    option_present = any(
+        item.strip().startswith("CleanupModulesOnExit") for item in contents if not item.startswith(("#", ";"))
```
bocekm, line 65:
```suggestion
        item.strip().startswith("CleanupModulesOnExit") for item in contents
```

I think that that the `if not item.startswith(("#", ";"))` is not necessary because if `item.strip().startswith("CleanupModulesOnExit")` returns True, it means that there's nothing before the CleanupModulesOnExit.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # If the config file has this option set to true/yes, then we need to
+    # return True to ask the user to change it to False.
+    if list(filter(CLEANUP_MODULES_ON_EXIT_REGEX.match, contents)):
+        logger.debug("CleanupModulesOnExit option enabled in %s" % FIREWALLD_CONFIG_FILE)
+        return True
+
+    # Default to return False as it is possible that the CleanupModulesOnExit
+    # is set to no in the config already.
+    return False
```
bocekm, line 81:
```suggestion
    logger.debug("CleanupModulesOnExit option is disabled in %s" % FIREWALLD_CONFIG_FILE)
    return False
```
This is a bit of a stretch, saying that the option is disabled. There could be a typo, like `CleanupModulesOnExit=bogus`.

Thinking about it more, you could simplify the code a lot by looking for `r"(?i)CleanupModulesOnExit\s*=\s*(no|false)"`. If you can't find it, it means the option is enabled.

tests/integration/tier0/non-destructive/firewalld-inhibitor/main.fmf
```diff
@@ -0,0 +1,18 @@
+summary: |
+    Running firewalld inhibitor
+description: |
+    Verify that the inhibitor is raised, when the firewalld service is running before
+    and during the conversion.
```
bocekm, line 5: @kokesak, @danmyway, the test needs to be updated because the conversion is inhibited only when the CleanupModulesOnExit config option is enabled.

tests/ansible_collections/roles/oracle-linux-specific/tasks/stop_firewalld_workaround.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Check that the firewalld package is installed
+  shell: 'rpm -q firewalld'
+  ignore_errors: True
+  register: rpm_out
+
+- name: Stop the firewalld process
```
bocekm, line 7: @kokesak, @danmyway, it'd be better to do in the workaround what we tell the customers to do - set `CleanupModulesOnExit=no`.

tests/ansible_collections/roles/oracle-linux-specific/tasks/main.yml
```diff
@@ -1,2 +1,4 @@
 ---
 - include_tasks: boot_standard_kernel.yml
+# Workaround reference: https://issues.redhat.com/browse/RHELC-1180
+- include_tasks: stop_firewalld_workaround.yml
```
bocekm, line 4:
```suggestion
- include_tasks: stop_firewalld_workaround.yml
  when: distro == oracle-8
```
@kokesak, @danmyway, shouldn't we limit the workaround to OL8 only?

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,133 @@
...
+                "kernel modules and prevent firewalld from stop working."
+            )
+            logger.warning(description)
+            self.add_message(
+                level="WARNING",
+                id="FIREWALLD_IS_RUNNING",
+                title="Firewalld is running",
+                description=description,
+            )
+            return
```
bocekm, line 133:
```suggestion
```
I don't think there's anything to warn about. This `else:` branch is just for the case when the CleanupModulesOnExit is explicitely set to no. And in that case there should be no problem with the firewalld during or after the conversion, if I understood it correctly.
Does it match your testing, @kokesak, @danmyway?

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,133 @@
...
+
+    :returns: Whether or not the CleanupModulesOnExit is set to true in
+        firewalld config.
+    :rtype: bool
+    """
+    # Return false if the config file does not exist. Either it means that
+    # firewalld is not installed or the config file was removed.
+    if not os.path.exists(FIREWALLD_CONFIG_FILE):
+        logger.debug("%s does not exist." % FIREWALLD_CONFIG_FILE)
+        return False
```
bocekm, line 47:
```suggestion
        return True
```
This needs to be True because when the config file does not exist, firewalld will use a default configuration (that means CleanupModulesOnExit=yes).
```
WARNING: [Errno 2] No such file or directory: '/etc/firewalld/firewalld.conf'
firewalld[8585]: WARNING: Using fallback firewalld configuration settings.
```

bocekm (changes requested):

tests/ansible_collections/roles/oracle-linux-specific/tasks/main.yml
kokesak: @bocekm yeah I thought it was done on a higher level here: https://github.com/oamg/convert2rhel/blob/main/tests/ansible_collections/main.yml#L14C13-L14C13 but this also apply for the oracle-7 which we do not want. I'll change it

convert2rhel/actions/system_checks/check_firewalld_availability.py
kokesak: @bocekm that's correct, when the `firewalld` is running and the `CleanupModulesOnExit=no` then everything is fine.

kokesak: also the message `we couldn't find check for the CleanupModulesOnExit configuration.` doesn't make sense if the option is present and set to `no`

convert2rhel/actions/system_checks/check_firewalld_availability.py
r0x0d: But the `if not item.startswith(("#", ";"))` is to prevent that we are returning True when we find a line with a comment, in this case, we will only proceed on checking that if the line does not have either `#` or `;`, meaning that lines with this are skippable.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,133 @@
...
+
+class CheckFirewalldAvailability(actions.Action):
+    id = "CHECK_FIREWALLD_AVAILABILITY"
+
+    def run(self):
+        """Error out if the firewalld service is running on the system."""
+        super(CheckFirewalldAvailability, self).run()
+        logger.task("Prepare: Check that firewalld is running")
+
+        if system_info.id != "oracle" and system_info.version.major != 8 and system_info.version.minor < 8:
```
kokesak, line 92: I'm not sure this condition works. The inhibitor happens also on OL7:
http://artifacts.osci.redhat.com/testing-farm/6c246209-3b6d-4f67-82cc-17aee0fd3f5c/

I noticed this on CentOS7:
```
[2023-10-30T07:43:47+0000] TASK - [Prepare: Check that firewalld is running] *************************
[2023-10-30T07:43:47+0000] DEBUG - Calling command '/usr/bin/systemctl show -p ActiveState firewalld'
Firewalld service reported that it is not running.
CHECK_FIREWALLD_AVAILABILITY has succeeded
```
the centos VMs don't have the `firewalld` package at all I believe. Anyway, it seems that it wasn't skipped for some reason

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,133 @@
...
+                description="Firewalld running on Oracle Linux 8 can lead to a conversion failure.",
+                diagnosis=(
+                    "We've detected that firewalld unit is running and that causes iptables and nftables "
+                    "failures on Oracle Linux 8 and under certain conditions it can lead to a conversion failure."
+                ),
+                remediation=(
+                    "Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf "
+                    "to no prior to running convert2rhel:\n"
+                    " sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n && firewall-cmd --reload"
+                    "You can change the option back to yes after the system reboot "
```
kokesak, line 114:
```suggestion
                    " sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n && firewall-cmd --reload."
                    " You can change the option back to yes after the system reboot "
```

reason:
```
            out:  sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf
            out:  && firewall-cmd --reloadYou can change the option back to yes after the system reboot - that is after the system boots into the RHEL kernel.
```

convert2rhel/actions/system_checks/check_firewalld_availability.py
r0x0d: > Thinking about it more, you could simplify the code a lot by looking for `r"(?i)CleanupModulesOnExit\s*=\s*(no|false)"`. If you can't find it, it means the option is enabled.

I think that's true, but for now, I guess that I will leave it that way, and we can come back to that later if needed. Probably this will be temporary anyway if we can figure out a better solution. But if not, then we can simplify this with that suggestion.

tests/ansible_collections/roles/oracle-linux-specific/tasks/stop_firewalld_workaround.yml
kokesak: You are right, I updated the tests and the workaround so it does what we instruct customers.

tests/integration/tier0/non-destructive/firewalld-inhibitor/main.fmf
kokesak: Done!

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,122 @@
...
+                        "Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf "
+                        "to no prior to running convert2rhel:\n"
+                        " sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n && firewall-cmd --reload."
+                        " You can change the option back to yes after the system reboot "
+                        "- that is after the system boots into the RHEL kernel."
+                    ),
+                )
+                return
+
+        logger.info("Skipping the check as it is relevant only for Oracle Linux 8.8 and above.")
```
kokesak, line 122: Not a PR blocker, but: If none of the above condition is true, then this message is printed. 

Scenario:
* we are on oracle8.8
* the firewalld is running
* the config file contains the `CleanupModulesOnExit=no`
then all the ifs are skipped and no default `return` is present, so the `logger.info("Skipping the check as it is relevant only for Oracle Linux 8.8 and above.")` is printed even though we are on 8.8 :D

```
14:28:50             out: [2023-10-30T14:28:50+0000] TASK - [Prepare: Gather system information] *******************************
14:28:50             out: [2023-10-30T14:28:50+0000] DEBUG - Calling command 'uname -i'
14:28:50             out: Skipping the execution of 'rpm -Va'.
14:28:50             out: [2023-10-30T14:28:50+0000] DEBUG - Calling command 'uname -r'
14:28:50             out: [2023-10-30T14:28:50+0000] DEBUG - Booted kernel VRA (version, release, architecture): 4.18.0-477.27.1.el8_8.x86_64
14:28:50             out: Checking internet connectivity using address 'https://static.redhat.com/test/rhel-networkmanager.txt'.
14:28:50             out: Successfully connected to address 'https://static.redhat.com/test/rhel-networkmanager.txt', internet connection seems to be available.
14:28:50             out: [2023-10-30T14:28:50+0000] DEBUG - Calling command '/usr/bin/systemctl show -p ActiveState dbus'
14:28:50             out: Name:                Oracle Linux Server
14:28:50             out: OS version:          8.8
14:28:50             out: Architecture:        x86_64
14:28:50             out: Config filename:     oracle-8-x86_64.cfg
14:28:53             out: 
14:28:53             out: [2023-10-30T14:28:53+0000] TASK - [Prepare: Clear YUM/DNF version locks] *****************************
14:28:53             out: Usage of YUM/DNF versionlock plugin not detected.
14:28:53             out: 
14:28:53             out: [2023-10-30T14:28:53+0000] TASK - [Prepare: Clean yum cache metadata] ********************************
14:28:53             out: [2023-10-30T14:28:53+0000] DEBUG - Calling command 'yum clean metadata --enablerepo=* --quiet'
14:28:53             out: [2023-10-30T14:28:53+0000] DEBUG - Output of yum clean metadata:
14:28:53             out: 
14:28:53             out: Cached repositories metadata cleaned successfully.
14:28:53             out: 
14:28:53             out: [2023-10-30T14:28:53+0000] TASK - [Prepare: Check whether system is ready for conversion] ************
14:28:53             out: 
14:28:53             out: [2023-10-30T14:28:53+0000] TASK - [Prepare: Check that firewalld is running] *************************
14:28:53             out: [2023-10-30T14:28:53+0000] DEBUG - Calling command '/usr/bin/systemctl show -p ActiveState firewalld'
14:28:53             out: [2023-10-30T14:28:53+0000] DEBUG - CleanupModulesOnExit option is disabled in /etc/firewalld/firewalld.conf
14:28:53             out: Skipping the check as it is relevant only for Oracle Linux 8.8 and above.
14:28:53             out: CHECK_FIREWALLD_AVAILABILITY has succeeded
```
http://artifacts.osci.redhat.com/testing-farm/cfa41dc4-a364-4e38-9237-b41e280da9dc/work-rhsmf6bxgaaj/log.txt

convert2rhel/actions/system_checks/check_firewalld_availability.py
kokesak: Seems that it is fixed now, thanks @r0x0d!:

- http://artifacts.osci.redhat.com/testing-farm/73014ac9-4a94-4595-b8a8-21fdb72f84ce/work-activation_key8644_n6m/log.txt
- http://artifacts.osci.redhat.com/testing-farm/ed0c2fa5-0980-4160-a737-82385e5b6707/work-activation_keyenr_ngna/log.txt
- http://artifacts.osci.redhat.com/testing-farm/f7139855-6998-4b38-995c-92d38cd9361f/work-rhsm8cbya2nf/log.txt

tests/integration/tier0/non-destructive/firewalld-inhibitor/test_firewalld_inhibitor.py
```diff
@@ -0,0 +1,42 @@
...
+    running firewalld on the system and the `CleanupModulesOnExit` configuration
+    option is set to `yes` in firewalld configuration file.
+    The reference ticket: https://issues.redhat.com/browse/RHELC-1180
+    """
+    assert shell("rpm -q firewalld").returncode == 0
+    assert shell("systemctl start firewalld").returncode == 0
+    assert shell("systemctl enable firewalld").returncode == 0
+
+    assert shell(f"grep 'CleanupModulesOnExit=yes' {FIREWALLD_CONFIG_FILE}").returncode == 0
+
```
danmyway: This might be tricky, I mean we have more or less the system under our control, but if for whatever reason the Oracle changes this, we won't match and the test will fail too soon.
I would either backup the original and provide our own config with the option disabled, or try regex matching more variants.

tests/integration/tier0/non-destructive/firewalld-inhibitor/test_firewalld_inhibitor.py
```diff
@@ -0,0 +1,42 @@
...
+        c2r.expect(
+            "CHECK_FIREWALLD_AVAILABILITY::FIREWALLD_MODULES_CLEANUP_ON_EXIT_CONFIG",
+            timeout=600,
+        )
+
+    assert c2r.exitstatus == 1
+
+    assert shell(
+        "grep 'Firewalld running on Oracle Linux 8 can lead to a conversion failure' /var/log/convert2rhel/convert2rhel.log"
+    )
```
danmyway, line 42: This might be unnecessary since we check for the WARNING in the c2r runtime already.

tests/integration/tier1/destructive/firewalld-config-ol8/main.fmf
```diff
@@ -0,0 +1,16 @@
+summary: |
+    Running conversion when firewalld is running and the proper configuration option is used.
```
danmyway:
```suggestion
    Running firewalld remediation
```

tests/integration/tier1/destructive/firewalld-config-ol8/main.fmf
```diff
@@ -0,0 +1,16 @@
...
+description: |
+    Verify that on the OL8.8 the conversion does not break when
+    firewalld is running and the the `CleanupModulesOnExit` configuration
+    option is set to `no` in firewalld configuration file. If the option is not
+    set then the conversion would end up with a broken machine.
+    This only applies to Oracle Linux (>=) 8.8.
```
danmyway:
```suggestion
    This only applies to Oracle Linux (>=) 8.8.
    Verify that on the Oracle Linux the conversion does not break when
    firewalld is running and the the `CleanupModulesOnExit` configuration
    option is set to `no` in the firewalld configuration file. 
    If the option is not set as described, the conversion would end up with a 
    broken machine.
```

tests/integration/tier1/destructive/firewalld-config-ol8/main.fmf
```diff
@@ -0,0 +1,16 @@
...
+
+enabled: false
+
+tag+:
+    - firewalld
```
danmyway:
```suggestion
    - running-firewalld-remediation
```

tests/integration/tier1/destructive/firewalld-config-ol8/test_firewalld_config_ol8.py
```diff
@@ -0,0 +1,40 @@
...
+
+from convert2rhel.actions.system_checks.check_firewalld_availability import FIREWALLD_CONFIG_FILE
+
+
+@pytest.mark.test_firewalld_config
+def test_firewalld_config_ol8(shell, convert2rhel):
+    """
+    Verify that properly changing the firewalld configuration file
+    allows to convert the system without any issue.
+    The reference ticket: https://issues.redhat.com/browse/RHELC-1180
```
danmyway:
```suggestion
    Verify that changing the firewalld configuration file as recommended
    in the remediation allows to convert the system without any issue.
    The reference ticket: https://issues.redhat.com/browse/RHELC-1180
```

tests/integration/tier1/destructive/firewalld-config-ol8/test_firewalld_config_ol8.py
```diff
@@ -0,0 +1,40 @@
+import os
+
+import pytest
+
+from envparse import env
+
+from convert2rhel.actions.system_checks.check_firewalld_availability import FIREWALLD_CONFIG_FILE
```
danmyway: Same as in the other case, I would rather see this hardcoded than imported.

tests/integration/tier1/destructive/firewalld-config-ol8/test_firewalld_config_ol8.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@pytest.mark.test_firewalld_config
+def test_firewalld_config_ol8(shell, convert2rhel):
```
danmyway:
```suggestion
@pytest.mark.test_running_firewalld_remediation
def test_running_firewalld_remediation_ol8(shell, convert2rhel):
```

tests/integration/tier1/destructive/firewalld-config-ol8/test_firewalld_config_ol8.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    Verify that properly changing the firewalld configuration file
+    allows to convert the system without any issue.
+    The reference ticket: https://issues.redhat.com/browse/RHELC-1180
+    """
+    if not os.path.exists(FIREWALLD_CONFIG_FILE):
+        pytest.fail("Firewalld configuration file does not exist")
```
danmyway:
```suggestion
    if not os.path.exists(FIREWALLD_CONFIG_FILE):
        pytest.fail("Firewalld configuration file does not exist")
```
This might come and bite us in the future.
Consider this being not so probable edge case, but if for whatever reason a new image will be provided without the config, the test will fail at this point providing no valid results.
Firewalld will fall back to a default option in case no config file is provided, that being `CleanupModulesOnExit=yes`.

tests/integration/tier1/destructive/firewalld-config-ol8/test_firewalld_config_ol8.py
```diff
@@ -0,0 +1,40 @@
...
+
+    assert shell("systemctl start firewalld").returncode == 0
+    assert shell("systemctl enable firewalld").returncode == 0
+
+    assert (
+        shell(f"sed -i 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' {FIREWALLD_CONFIG_FILE}").returncode == 0
+    )
```
danmyway: This will fail if we fail to match `CleanupModulesOnExit=yes`.
We might consider the suggestion from the other (non-destructive) test case. To provide and copy over our own config file and back up the existing one.

tests/integration/tier1/destructive/firewalld-config-ol8/test_firewalld_config_ol8.py
```diff
@@ -0,0 +1,40 @@
...
+        ),
+        unregister=True,
+    ) as c2r:
+        c2r.expect("Pre-conversion analysis report")
+        c2r.expect_exact("(WARNING) CHECK_FIREWALLD_AVAILABILITY::FIREWALLD_IS_RUNNING")
+
+    assert c2r.exitstatus == 0
+
+    # There should not be any problems
+    assert shell("grep -i 'traceback' /var/log/convert2rhel/convert2rhel.log").returncode == 1
```
danmyway: Not sure why this is here.
If there would be a traceback, the c2r.exitstatus won't return 0.
On the other hand we might evaluate occurences of traceback in the logfile in one of the checks after conversion, which scans through the log file.

danmyway (approved):
Just couple of nitpicks, we can discuss and revisit later @kokesak.
I wouldn't block the merge on this.

convert2rhel/actions/system_checks/check_firewalld_availability.py
```diff
@@ -0,0 +1,122 @@
...
+                    description="Firewalld running on Oracle Linux 8 can lead to a conversion failure.",
+                    diagnosis=(
+                        "We've detected that firewalld unit is running and that causes iptables and nftables "
+                        "failures on Oracle Linux 8 and under certain conditions it can lead to a conversion failure."
+                    ),
+                    remediation=(
+                        "Set the option CleanupModulesOnExit in /etc/firewalld/firewalld.conf "
+                        "to no prior to running convert2rhel:\n"
+                        " sed -i -- 's/CleanupModulesOnExit=yes/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n && firewall-cmd --reload."
+                        " You can change the option back to yes after the system reboot "
```
bocekm, line 116:
```suggestion
                        " sed -i -- 's/^CleanupModulesOnExit.*/CleanupModulesOnExit=no/g' /etc/firewalld/firewalld.conf\n && firewall-cmd --reload\n"
                        "You can change the option back to yes after the system reboot "
```

bocekm (approved):

PR #956, opened 2023-10-12
[RHELC-965] Change `UNKNOWN_ERROR` to be more verbose with subscriptions

SpyTec:

Checklist

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/subscription.py
+++ b/convert2rhel/actions/pre_ponr_changes/subscription.py
@@ -18,7 +18,7 @@ __metaclass__ = type
 import logging
 import os.path
 
-from convert2rhel import actions, backup, cert, pkghandler, repo, subscription, toolopts, utils
+from convert2rhel import actions, backup, cert, exceptions, pkghandler, repo, subscription, toolopts, utils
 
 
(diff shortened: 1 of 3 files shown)


convert2rhel/actions/pre_ponr_changes/subscription.py
```diff
@@ -214,27 +216,28 @@ def run(self):
             logger.task("Convert: Subscription Manager - Enable RHEL repositories")
             subscription.enable_repos(rhel_repoids)
         except OSError as e:
-            # TODO(r0x0d): Places where we raise OSError and need to be
-            # changed to something more specific.
-            #  - Could fail in invoking subscription-manager to get repos     (get_avail_repos)
-            #  - Could fail in invoking subscirption-manager to disable repos (disable_repos)
-            #  - ""                                          to enable repos  (enable_repos)
+            # This should not occur anymore as all the relevant OSError has been changed to a CriticalError
+            # exception. This exception handler is just a precaution
```
r0x0d, line 220: Just to point out, when I did this exception block back then my rationale was something like: "I don't know if that will even happen, but in case it does, we have a way to catch it... Being extra paranoid, I know..." 

So I can't really confirm if that really happens.

PR #958, opened 2023-10-12
[RHELC-965] Change `UNKNOWN_ERROR` to be more verbose with pkghandler

SpyTec:

Checklist

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -27,7 +27,7 @@ from collections import namedtuple
 
 import rpm
 
-from convert2rhel import backup, pkgmanager, utils
+from convert2rhel import backup, exceptions, pkgmanager, utils
 from convert2rhel.backup import RestorableFile, remove_pkgs
 from convert2rhel.systeminfo import system_info
 from convert2rhel.toolopts import tool_opts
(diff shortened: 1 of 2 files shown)


convert2rhel/pkghandler.py
```diff
@@ -225,9 +225,16 @@ def _enable(self):
             set_releasever=False,
         )
         if ret_code:
-            loggerinst.critical(
+            loggerinst.critical_no_exit(
                 "Failed to install subscription-manager packages. See the above yum output for details."
             )
+            raise exceptions.CriticalError(
```
r0x0d, line 231: I have a feeling that we should keep using only underscores in the ids instead of mixing dashes and underscores. 

If for some reason people will try to parse our IDs to present that to some UI, then some additional logic might need to be done to avoid the dashes we have in the names.

Another reason would be to keep the pattern across the other ids since they have only underscores everywhere, even for cases where a dash would be applied. 

Let me know what you think/feel about this suggestion.

convert2rhel/pkghandler.py
```diff
@@ -225,9 +225,16 @@ def _enable(self):
...
+                id_="FAILED_TO_INSTALL_SUBSCRIPTION_MANAGER_PACKAGES",
+                title="Failed to install subscription-manager packages.",
+                description="To be able to subscribe the system to Red Hat we need subscription-manager packages to do so. Without these packages we cannot subscribe the system and we cannot install Red Hat Enterprise Linux packages.",
+                diagnosis="Failed to install packages %s. Errno: %s, Error: %s"
```
r0x0d, line 235: Maybe we should use `status` or `return code` instead of `errno`?

r0x0d: I think @jochapma is using `status` in one of his PRs too

SpyTec: We used the same in backup.py so I just kept with the consistency

SpyTec: True, though the distingtion between output and status from a command call and an exception being thrown might be good, so keeping errno and error for exceptions might be good while commands are output and status, like Joe said

Can always change later

r0x0d (approved):
Looks nice

PR #959, opened 2023-10-13
[RHELC-965] Change `UNKNOWN_ERROR` to be more verbose with backup

jochapma:
Converts two UNKNOWN_ERROR instances in package install/remove to use exceptions.CriticalError.

Checklist

### DIFF
--- a/convert2rhel/backup.py
+++ b/convert2rhel/backup.py
@@ -105,7 +105,14 @@ class ChangedRPMPackagesController:
             pkgs_as_str = " ".join(pkgs_to_install)
             loggerinst.debug(output.strip())
             if critical:
-                loggerinst.critical("Error: Couldn't install %s packages." % pkgs_as_str)
+                loggerinst.critical_no_exit("Error: Couldn't install %s packages." % pkgs_as_str)
+                raise exceptions.CriticalError(
+                    id_="COULDNT_INSTALL_PACKAGES",
+                    title="Couldn't install packages.",
+                    description="XXX DESCRIPTION.",
+                    diagnosis="Couldn't install %s packages. Command: %s Output: %s Status: %d"
+                    % (pkgs_as_str, cmd, output, ret_code),
+                )
 
             loggerinst.warning("Couldn't install %s packages." % pkgs_as_str)
             return False
@@ -355,8 +362,6 @@ class RestorableFile:
             except (OSError, IOError) as err:
                 # IOError for py2 and OSError for py3
 
-                # Replace this with loggerinst.critical_no_exit() once that is
-                # written.
                 loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
                 raise exceptions.CriticalError(
                     id_="FAILED_TO_SAVE_FILE_TO_BACKUP_DIR",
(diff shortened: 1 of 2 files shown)


convert2rhel/backup.py
```diff
@@ -105,7 +105,14 @@ def _install_local_rpms(self, pkgs_to_install, replace=False, critical=True):
...
             if critical:
-                loggerinst.critical("Error: Couldn't install %s packages." % pkgs_as_str)
+                loggerinst.critical_no_exit("Error: Couldn't install %s packages." % pkgs_as_str)
+                raise exceptions.CriticalError(
+                    id_="COULDNT_INSTALL_PACKAGES",
+                    title="Couldn't install packages.",
+                    description="XXX DESCRIPTION.",
+                    diagnosis="Couldn't install %s packages. Command: %s Output: %s Status: %d"
+                    % (pkgs_as_str, cmd, output, ret_code),
+                )
```
SpyTec, line 115: I would probably change to Failed to install packages or something, we can ask Miriam in the spreadsheet though! Same with the other messages in the PR

jochapma: Yeah, I got that feeling while editing the spreadsheet. Will fix when I fuss with the description.

r0x0d (approved):
Looks good to me.

convert2rhel/backup.py
```diff
@@ -514,10 +519,17 @@ def remove_pkgs(
             pkgs_removed.append(nevra)
 
     if pkgs_failed_to_remove:
+        pkgs_as_str = ", ".join(pkgs_failed_to_remove)
```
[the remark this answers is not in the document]
jochapma: Done.

convert2rhel/backup.py
SpyTec, line 1: Could you change L365 as well in that case to have output status

https://github.com/oamg/convert2rhel/pull/959/files#diff-8f4622020cde9a253bc82f69bf2b88910bcc5a84943db3c7d3a435becedfd4d5L365

convert2rhel/backup.py
```diff
@@ -102,10 +102,17 @@ def _install_local_rpms(self, pkgs_to_install, replace=False, critical=True):
...
             loggerinst.debug(output.strip())
             if critical:
-                loggerinst.critical("Error: Couldn't install %s packages." % pkgs_as_str)
+                loggerinst.critical_no_exit("Error: Couldn't install %s packages." % pkgs_as_str)
+                raise exceptions.CriticalError(
+                    id_="FAILED_TO_INSTALL_PACKAGES",
+                    title="Couldn't install packages.",
+                    description="While attempting to roll back changes, we encountered an unexpected failure while attempting to reinstall one or more packages that we removed as part of the conversion.",
+                    diagnosis="Couldn't install %s packages. Command: %s Output: %s Status: %d"
+                    % (pkgs_as_str, cmd, output, ret_code),
```
r0x0d, line 114: Do we need to output the command used here?

SpyTec: True, I wonder if the command is useful here at all 🤔

jochapma: I've been assuming that 'diagnosis' is something intended for a developer, which may be a misunderstanding on my part. My main rationale for including the command was that this is a "can't happen" error--we couldn't think of a way to cause it to happen for a test--so I wanted to be clear on what it was that failed, exactly.

r0x0d: But that's a good point, @jochapma... I don't think we ever discussed the intent of the messages we are producing. I, for instance, always assumed that they would be user oriented rather than developer oriented. Maybe we could bring this discussion to our Slack channel and check what others might think about that?

PR #961, opened 2023-10-16
hotfix/tests/Fix the update-ca-trust.yml

danmyway:
* update to up to date certificates


kokesak (approved):
I would wait for the tests to finish just to make sure that everything works. Otherwise looks good, nice catch @danmyway

danmyway:
For context, the certiifcates were renamed and/or will be deprecated in a near future.
There is a new bundle available, which is being updated regularly.

PR #979, opened 2023-11-02
Refactored applock to have a more robust API.

jochapma:
[RHELC-1124] Refactor ApplicationLock to have a more consistent API

Jira Issues: [RHELC-1124](https://issues.redhat.com/browse/RHELC-1124)

Checklist

- [x ] `[RHELC-]` is part of the PR title 
- [x ] Code and tests are documented properly


convert2rhel/applock.py
```diff
@@ -103,8 +101,28 @@ def _try_create(self):
...
+            return pid == self._pid
+        return False
+
+    def _read_pidfile(self):
+        """Read and return the contents of the PID file.
+
+        :returns: the file contents as an integer, or None if it doesn't exist
+        :raises: ApplicationLockedError if the contents are corrupt
+        """
+        if os.path.exists(self._pidfile):
```
abadger, line 117: Note: To avoid races, it's better to do a `try: except:` around the `open()` rather than check for file existence first.

convert2rhel/applock.py
```diff
@@ -59,15 +59,13 @@ class ApplicationLock:
     def __init__(self, name):
         # Our application name
         self._name = name
-        # Do we think we locked the pid file?
-        self._locked = False
         # Our process ID
         self._pid = os.getpid()
```
abadger, line 63:
```suggestion
        # Our process ID. We save this when the lock is created so it will be consistent even if we check from inside a fork.
        self._pid = os.getpid()
```

convert2rhel/applock.py
```diff
@@ -103,8 +101,28 @@ def _try_create(self):
```
abadger, line 101: Note: the `raise exc` on line  100/98 above would be better as `raise` without the exception object.  On Python3 this won't make a difference but on Python2, the original traceback object is used with `raise` but a new traceback object is generated with `raise exc`, losing the original.

convert2rhel/applock.py
```diff
@@ -129,31 +147,23 @@ def try_to_lock(self, _recursive=False):
...
-
+        pid = self._read_pidfile()
+        if pid == self._pid:
+            return
         if self._pid_exists(pid):
             raise ApplicationLockedError("%s locked by process %d" % (self._pidfile, pid))
         # The lock file was created by a process that has exited;
         # remove it and try again.
         loggerinst.info("Cleaning up lock held by exited process %d." % pid)
         os.unlink(self._pidfile)
```
abadger, line 162: There is a race condition here too, where something else may have already `unlinked` the file so we need to catch the exception.

Also.... I think the race could also mean that self._pidfile might not contain the same pid as we checked with `_pid_exists` (meaning between us reading the pid and `unlinking` the file, another process has unlinked the original and created their new one in its place.) Is there some way to fix that with some combination of counting links to the lockfile, adding one new link to the file, and checking whether the count of links matches how many we think it should at each step of the way?  (And if it doesn't match, having some sort of `time.sleep(random)` before retrying?)  (I'm not sure if there is a way or not.  I think there might be but a quick internet search didn't turn up a recipe so I might be overly optimisitc.  If not, it's probably better to err on the side of telling the user to check whether the lock is stale rather than overwriting the lock).

SpyTec:
@jochapma there's 3 merge commits in the PR. If you can, drop them and rebase with main instead

convert2rhel/applock.py
```diff
@@ -97,14 +108,55 @@ def _try_create(self):
...
+        pid = self._read_pidfile()
+        self._release_pidfile_flock()
+        if pid is not None:
+            return pid == self._pid
+        return False
+
+    def _release_pidfile_flock(self):
+        """Release the advisory file lock we hold on the PID file
+        and close the open file descriptor."""
+        if self._pidfile_fp is not None:
```
r0x0d, line 128:
```suggestion
        if self._pidfile_fp:
```

convert2rhel/applock.py
```diff
@@ -97,14 +108,55 @@ def _try_create(self):
...
 
     @property
     def is_locked(self):
-        """Test whether this object is locked."""
-        return self._locked
+        """Test whether this object was locked by this instance of
+        the application."""
+        pid = self._read_pidfile()
+        self._release_pidfile_flock()
+        if pid is not None:
```
r0x0d, line 121:
```suggestion
        if pid:
```

convert2rhel/applock.py
```diff
@@ -97,14 +108,55 @@ def _try_create(self):
             except OSError as exc:
                 if exc.errno == errno.EEXIST:
                     return False
-                raise exc
```
Venefilyn, line 100: Should we not keep the exception here instead of raising nothing?

convert2rhel/applock.py
```diff
@@ -97,14 +108,55 @@
...
+        If this method does not raise an exception, the
+        _release_pidfile_flock() method must be called to release the
+        advisory lock and close the file.
+
+        :returns: the file contents as an integer, or None if it doesn't exist
+        :raises: ApplicationLockedError if the contents are corrupt
+        """
+        # pylint: disable=consider-using-with
+        try:
+            self._pidfile_fp = open(self._pidfile, "r")
```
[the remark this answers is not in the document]
Venefilyn: File is never closed, we can use `with:` here or explicitly call close

convert2rhel/applock.py
```diff
@@ -97,14 +108,55 @@ def _try_create(self):
...
+            fcntl.flock(self._pidfile_fp, fcntl.LOCK_EX)
+            file_contents = self._pidfile_fp.read()
+            pid = int(file_contents.rstrip())
+            return pid
+        except IOError as exc:
+            # The open() failed because the file exists.
+            # In Python 3 this could be changed to FileNotFoundError.
+            if exc.errno == errno.ENOENT:
+                return None
+            raise
```
Venefilyn, line 156: Same here, we should be raising exception again

convert2rhel/applock.py
```diff
@@ -120,52 +172,59 @@ def _pid_exists(pid):
...
+        doesn't exist, swallow the exception; this avoids spurious
+        errors due to race conditions.
+        """
+        try:
+            os.unlink(self._pidfile)
+        except OSError as exc:
+            # In Python 3 this could be changed to FileNotFoundError.
+            if exc.errno == errno.ENOENT:
+                return
+            raise
```
Venefilyn, line 186: This isn't covered by tests

convert2rhel/applock.py
```diff
@@ -97,14 +108,55 @@ def _try_create(self):
...
+        self._release_pidfile_flock()
+        if pid:
+            return pid == self._pid
+        return False
+
+    def _release_pidfile_flock(self):
+        """Release the advisory file lock we hold on the PID file
+        and close the open file descriptor."""
+        if self._pidfile_fp:
+            fcntl.flock(self._pidfile_fp, fcntl.LOCK_UN)
```
Venefilyn, line 129: What does this do? Comments would be nice

Venefilyn (changes requested):
Overall seems fine, some exceptions being re-raised but not with the same exception. There is a file open that is never closed, which could be an issue

convert2rhel/applock.py
jochapma: We have to leave the file open and close it later explicitly because the file lock is tied to the file descriptor. The file lock is required to plug a race condition in which the pid file is overwritten by another process between the point where we read its contents and unlink it.

jochapma: The bare raise statement reraises the current exception; @abadger requested that I use this idiom.

convert2rhel/applock.py
```diff
@@ -115,57 +140,71 @@ def _pid_exists(pid):
...
+                file_contents = filep.read()
+                pid = int(file_contents.rstrip())
+                if pid == self._pid:
+                    return
+                if self._pid_exists(pid):
+                    raise ApplicationLockedError("%s locked by process %d" % (self._pidfile, pid))
+                # The lock file was created by a process that has exited;
+                # remove it and try again.
+                loggerinst.info("Cleaning up lock held by exited process %d." % pid)
+                self._safe_unlink()
```
Venefilyn, line 187: Could we not reuse the logic of `is_locked()`? Otherwise we might change it in the future on one part and miss the other etc.

jochapma: I was unable to find a way to avoid a race condition in unlink without keeping a file pointer open across multiple methods, which all my reviewers didn't like. So I removed the automatic cleanup; this also has the benefit of simplifying everything considerably, a good idea since this is my last week.

What this does is add another "can't happen" case (if the lock file somehow doesn't get cleaned up), but that can be easily remedied manually.

In the course of doing that, I addressed the `is_locked()` reuse.

PR #982, opened 2023-11-14
[RHELC-1037] Auto generate manpage Github action fix

Andrew-ang9:
The auto-gen manpage PR was believed to be working but when it was time to make the manpage for the release it wasn't correct and so it was found out that it wasn't running correctly. 

Jira Issues: [RHELC-1037](https://issues.redhat.com/browse/RHELC-1037)

Checklist

### DIFF
--- a/.github/workflows/generate_manpage.yml
+++ b/.github/workflows/generate_manpage.yml
@@ -44,6 +44,6 @@ jobs:
 
       - name: Push to Fork
         run: |
-          git push https://github.com/${{ github.repository_owner }}/${{ github.repository }}.git ${{ github.head_ref }}:${{ github.head_ref }}
+          git push origin https://github.com/${{ github.repository_owner }}/${{ github.repository }}.git ${{ github.head_ref }}:${{ github.head_ref }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/generate_manpage.yml
+++ b/.github/workflows/generate_manpage.yml
@@ -13,6 +13,9 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Create and checkout branch
+        run: git checkout -b auto_gen_man_fix
+
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
@@ -39,4 +42,4 @@ jobs:
 
       - name: Push to Fork
         run: |
-          git push https://github.com/${{ github.actor }}/${{ github.repository_name }}.git ${{ github.head_ref }}
+          git push https://github.com/${{ github.actor }}/${{ github.repository_name }}.git HEAD:auto_gen_man_fix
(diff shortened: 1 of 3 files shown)


.github/workflows/generate_manpage.yml
```diff
@@ -31,3 +31,17 @@ jobs:
         run: |
           chmod +x scripts/manpage_generation.sh
           bash scripts/manpage_generation.sh
+
+      - name: Show File Modification
+        run: |
+          FILE_PATH="man/convert2rhel.8"
+
+          # Fetch the changes from the remote repository
+          git fetch --quite
```
SpyTec, line 40:
```suggestion
          git fetch --quiet
```

.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
 
       - name: Trigger Manpage Generation
-        uses: ./.github/workflows
```
SpyTec, line 17: Did you check this in your repo? Do we know if this works?

SpyTec: Looking at the docs it seems like we're using it incorrectly still

https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow

Andrew-ang9: ohh I didn't check to see if it worked in my repo first. I do know that the last commit that I did was passing on my repo

SpyTec: To test this the easiest way would be to do it in your PR with the trigger for `release.yml` changed to 

```
on:
  push
```

.github/workflows/generate_manpage.yml
Andrew-ang9: Done!

.github/workflows/release.yml
```diff
@@ -21,13 +21,7 @@ jobs:
...
 
       - name: Trigger Manpage Generation
-        run: |
-          curl -X POST \
-            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-            -H "Accept: application/vnd.github.everest-preview+json" \
-            -H "Content-Type: application/json" \
-            https://api.github.com/repos/${{ github.repository }}/dispatches \
-            --data '{"event_type": "trigger-manpage-generation"}'
+        uses: .github/workflows/generate_manpage.yml
```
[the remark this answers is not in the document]
SpyTec: Though didn't you try `./.github/workflows/generate_manpage.yml` before and that didn't work?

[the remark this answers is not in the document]
Andrew-ang9: Yeah I did try ./.github/workflows/generate_manpage.yml and it didn't work but i'll give it a shot anyways. Thank you for the suggestion

SpyTec:
Michal mentioned something we didn't think of, that the downstream release is actually done through the source code tarball that's in the GitHub release assets page. Since we cannot and shouldn't modify this ourselves, just changing the manpage on the main branch would be a better solution as we then cover all bases with the source code tarball and release rpms

Closing for that solution and will update the Jira issue

PR #983, opened 2023-11-15
[RHELC-1193] Identify highest status in json report

pr-watson:
This PR adds a function find_highest_report_level to report.py to get the highest status level within the results and messages reported in the analysis. This highest status level is added as a new key to the report json at the top level for use in the rhc worker script. 

Jira Issues: [RHELC-1193](https://issues.redhat.com/browse/RHELC-1193)

Checklist

### DIFF
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -24,6 +24,7 @@ from convert2rhel import utils
 from convert2rhel.actions import (
     _STATUS_HEADER,
     _STATUS_NAME_FROM_CODE,
+    STATUS_CODE,
     find_actions_of_severity,
     format_action_status_message,
     level_for_combined_action_data,
@@ -75,9 +76,13 @@ def summary_as_json(results, json_file=CONVERT2RHEL_JSON_RESULTS):
     * The results are modified so that status codes use their symbolic names
       instead of the numeric values.
     """
+    # Collect the highest report level to use as status key
+    highest_level = find_highest_report_level(results)
+
     # Use an envelope so we can add other, non-result info if necessary.
     envelope = {
-        "format_version": "1.0",
+        "format_version": "1.1",
+        "status": highest_level,
         "actions": copy.deepcopy(results),
     }
 
@@ -261,6 +266,29 @@ def format_report_section_heading(status_code):
 
     heading = "{highlight} {status_header} {highlight}".format(highlight=highlight, status_header=status_header)
     return heading
+
+
+def find_highest_report_level(results):
+    """
+    Gather status codes from messages and result. We are not seeking for
+    differences between them as we want all the results, no matter where
+    they come from.
+
+    :param results: The results from the Actions which have been run.
+    :type results: dict
+    :return: The highest status code from messages and result
+    :rtype: str
+    """
+    action_level_combined = []
+    for value in results.values():
+        action_level_combined.append(value["result"]["level"])
+        for message in value["messages"]:
+            action_level_combined.append(message["level"])
+
+    valid_action_levels = [level for level in action_level_combined if level in STATUS_CODE.values()]
+    valid_action_levels.sort(reverse=True)
+    highest_action_level = _STATUS_NAME_FROM_CODE[valid_action_levels[0]]
+    return highest_action_level
 
 
 def summary_as_txt(results):
--- a/convert2rhel/unit_tests/actions/report_test.py
+++ b/convert2rhel/unit_tests/actions/report_test.py
@@ -55,7 +55,8 @@ _LONG_MESSAGE = {
                 },
             },
             {
-                "format_version": "1.0",
+                "format_version": "1.1",
+                "status": "WARNING",
                 "actions": {
                     "CONVERT2RHEL_LATEST_VERSION": {
                         "result": dict(level="SUCCESS", id="SUCCESS"),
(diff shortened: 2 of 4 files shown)


convert2rhel/actions/report.py
```diff
@@ -75,9 +76,13 @@ def summary_as_json(results, json_file=CONVERT2RHEL_JSON_RESULTS):
     * The results are modified so that status codes use their symbolic names
       instead of the numeric values.
     """
+    # Collect the highest report level to use as status key
+    highest_level = find_highest_report_level(results)
+
     # Use an envelope so we can add other, non-result info if necessary.
     envelope = {
         "format_version": "1.0",
```
hosekadam, line 84: I'm wondering if the version shouldn't be increased when there is a change in format

convert2rhel/actions/report.py
```diff
@@ -75,9 +76,13 @@ def summary_as_json(results, json_file=CONVERT2RHEL_JSON_RESULTS):
...
+        "status": highest_level,
```
hosekadam, line 85: The `schemas/assessment-schema-1.0.json` is missing the `status` item

pr-watson: Thanks i'll add that

convert2rhel/unit_tests/actions/report_test.py
```diff
@@ -1465,3 +1467,110 @@ def test_get_combined_results_and_message(results, expected):
...
+        ),
+    ),
+)
+def test_find_highest_report_level_expected(actions, expected):
+    """Should be sorted descending from the highest status to the lower one."""
+    result = report.find_highest_report_level(actions)
+    assert result == expected
+
+
+def test_find_highest_report_level_unknown_status():
```
hosekadam, line 1561: Maybe I would use the `pytest.mark.parametrize` for easier addition of new cases. But question is if that would be ever needed.

convert2rhel/actions/report.py
pr-watson: Correct, i'll add this

convert2rhel/unit_tests/actions/report_test.py
pr-watson: Likely won't have new cases. Just having FOO as an unknown status satisfies checking that unknown statuses will not be added to our output

schemas/assessment-schema-1.1.json
```diff
@@ -0,0 +1,150 @@
...
+                        "result": {
+                            "$ref": "#/$defs/action_result"
+                        }
+                    }
+                }
+            }
+        },
+        "format_version": {
+            "description": "Constant value that tells us the format of this file.",
+            "const": "1.0"
```
hosekadam, line 32:
```suggestion
            "const": "1.1"
```

convert2rhel/actions/report.py
```diff
@@ -75,9 +76,13 @@ def summary_as_json(results, json_file=CONVERT2RHEL_JSON_RESULTS):
     * The results are modified so that status codes use their symbolic names
       instead of the numeric values.
     """
+    # Collect the highest report level to use as status key
+    highest_level = find_highest_report_level(results)
+
     # Use an envelope so we can add other, non-result info if necessary.
     envelope = {
-        "format_version": "1.0",
+        "format_version": "1.1",
```
SpyTec, line 84: Why do we add the `format_version` here and not using the one in the schema file?

schemas/assessment-schema-1.1.json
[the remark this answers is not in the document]
pr-watson: The new field is a change to the format so we do need a bump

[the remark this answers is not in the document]
r0x0d: Yeah, if the format of the schema changes, a bump is needed. Especially because parsers would need to adapt their parsing mechanism based on some new field that was added to a stable version. 

I don't think bumping the version in the schema is bad just for this one field, especially because we will rarely update it.

convert2rhel/actions/report.py
```diff
@@ -264,6 +269,24 @@ def format_report_section_heading(status_code):
     return heading
 
 
+def find_highest_report_level(results):
+    """
+    Gather status codes from messages and result. We are not seeking for
+    differences between them as we want all the results, no matter from where
+    they come.
+    """
```
r0x0d, line 277: Please, if you can, update the docstring to include the param/type/return/rtype. 

```suggestion
    they come.
    
    :param results: ...
    :type results: ...
    :return: ...
    :rtype: ...
    """
```

convert2rhel/unit_tests/actions/report_test.py
r0x0d: agreed with @pr-watson. Can't see a case where we would need to parametrize this.

schemas/assessment-schema-1.1.json
```diff
@@ -0,0 +1,150 @@
...
+        },
+        "status": {
+            "description": "The highest severity of all actions.",
```
r0x0d, line 35: Maybe not necessary, but I would include _"between messages and results"_ in this description

r0x0d (commented):
It looks good to me. Just left a couple of questions and I think one of @SpyTec comments is necessary to apply. Not approving it yet, but I can see the PR is correct.

PR #984, opened 2023-11-15
[RHELC-1225] Fix an exception when listing third-party packages with Epoch 2

jochapma:
A bug processing output from the repoquery command caused packages with an Epoch equal to 2 to raise an exception in the third-party package listing task. This change fixes the processing.

Jira Issues: [RHELC-1225](https://issues.redhat.com/browse/RHELC-1225)

Checklist

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -706,6 +706,7 @@ def _get_package_repositories(pkgs):
             repositories_mapping[package] = "N/A"
     else:
         for line in output:
+            line = line.lstrip()
             if line.startswith("C2R "):
                 split_output = line[4:].split("&")
                 nevra = split_output[0]
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -706,8 +706,8 @@ def _get_package_repositories(pkgs):
             repositories_mapping[package] = "N/A"
     else:
         for line in output:
-            if "C2R" in line:
-                split_output = line.lstrip("C2R ").split("&")
+            if line.startswith("C2R "):
+                split_output = line[4:].split("&")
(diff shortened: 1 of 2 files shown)


SpyTec (approved):

SpyTec:
Unit tests are failing but code if works seems good

convert2rhel/pkghandler.py
```diff
@@ -706,8 +706,9 @@ def _get_package_repositories(pkgs):
             repositories_mapping[package] = "N/A"
     else:
         for line in output:
-            if "C2R" in line:
-                split_output = line.lstrip("C2R ").split("&")
+            line = line.lstrip()
```
SpyTec, line 709: I don't think we'll encounter this in the wild tbh, it's just something within our own tests, but doesn't hurt so let's leave it be for now

SpyTec:
Simple and effective
```
>>> pkg = "C2R 2:eog-44.1-1.fc38.x86_64&updates"
>>> pkg.lstrip("C2R ").split("&")
[':eog-44.1-1.fc38.x86_64', 'updates']
>>> pkg[4:].split("&")
['2:eog-44.1-1.fc38.x86_64', 'updates']
```

bocekm:
@jochapma, out of curiosity, why was the code failing with the epoch being specifically 2? Wouldn't other epoch cause the issue as well?

jochapma:
@bocekm, we meant to strip off "C2R" from the left side of the string. Unfortunately, we were using lstrip(), which strips off all instances of any character in the string, so it removed C, 2, and R. So epoch 2 specifically got stripped off.

bocekm:
Ugh, that was a nasty bug indeed, I would have not thought that lstrip() has this behavior. Thanks for the explanation.

PR #991, opened 2023-11-22
[RHELC-1246] Fix sub-man installation on EL8

bocekm:
Yumdownloader was downloading two different versions of the subscription-manager-rhsm-certificates package and convert2rhel failed when it wanted to install both at the same time.

subscription-manager-rhsm-certificates-1.28.36-3.el8_8.x86_64 subscription-manager-rhsm-certificates-20220623-1.el8.noarch

Pino Toscano from the subscription-manager team told us that the version 20220623 is the latest and the one we should be installing.

Since the two versions have different architecture, this quick solution is making sure that only the noarch one is downloaded and installed.

The same applies to conversions to RHEL 9. We were downloading incorrectly the x86_64 version of the package which is not the latest.

Jira Issues: [RHELC-1246](https://issues.redhat.com/browse/RHELC-1246)

Checklist

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -833,7 +833,7 @@ def _relevant_subscription_manager_pkgs():
             "python3-syspurpose",
             "python3-cloud-what",
             "json-c.x86_64",  # there's also an i686 version we don't need unless the json-c.i686 is already installed
-            "subscription-manager-rhsm-certificates",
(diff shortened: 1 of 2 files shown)


danmyway (approved):
Test fails unrelated, will be fixed in #988 
The subscription-manager package is being correctly installed.
Could be merged.

PR #998, opened 2023-11-29
[RHELC-1252] Fix sub-man-rhsm-certs installation on 8.5/8.8

bocekm:
The https://github.com/oamg/convert2rhel/pull/991 was dealing with a convert2rhel failure stemming from two different architectures of the subscription-manager-rhsm-certificates package being available in UBI repos. That fix however caused convert2rhel failures on older system minor versions where only x86_64 arch was available.

In the meantime Pino Toscano asked to have the UBI repos fixed in a way that there is now just one architecture of the package. With that we don't need to hardcode a specific arch in convert2rhel.

Jira Issues: [RHELC-1252](https://issues.redhat.com/browse/RHELC-1252)

Checklist

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -833,7 +833,7 @@ def _relevant_subscription_manager_pkgs():
             "python3-syspurpose",
             "python3-cloud-what",
             "json-c.x86_64",  # there's also an i686 version we don't need unless the json-c.i686 is already installed
-            "subscription-manager-rhsm-certificates.noarch",
+            "subscription-manager-rhsm-certificates",
         ]
 
     elif system_info.version.major >= 9:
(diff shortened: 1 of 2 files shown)


danmyway (approved):
Looks good.

convert2rhel/subscription.py
```diff
@@ -833,15 +833,15 @@ def _relevant_subscription_manager_pkgs():
...
-            "subscription-manager-rhsm-certificates.noarch",
+            "subscription-manager-rhsm-certificates",
         ]
 
     elif system_info.version.major >= 9:
         relevant_pkgs += [
             "libdnf-plugin-subscription-manager",
             "python3-subscription-manager-rhsm",
             "python3-cloud-what",
-            "subscription-manager-rhsm-certificates.noarch",
```
SpyTec, line 844: Given we had `.x86_64` before in #991 we're not going to encounter any issues?

bocekm: We won't encounter issues anymore on EL9 because https://issues.redhat.com/browse/RHELDST-21855 fixed both EL8 and EL9 UBI repos. I should have stated that in the PR desc 🫣

PR #999, opened 2023-11-30
[RHELC-1270] Update package handling check to properly report removed packages

pr-watson:
With the current setup of the conditionals in handle_packages, there are cases where a message explaining that packages were removed will be made regardless if there are any packages removed or not. This PR changes the conditionals to only display the packages removed message when there are packages contained inside the pkgs_removed variable. 

Jira Issues: [RHELC-1270](https://issues.redhat.com/browse/RHELC-1270)

image

Checklist

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -102,6 +102,9 @@ class RemoveExcludedPackages(actions.Action):
 
         logger.task("Convert: Remove excluded packages")
         logger.info("Searching for the following excluded packages:\n")
+
+        pkgs_removed = []
+
         try:
             pkgs_to_remove = sorted(pkghandler.get_packages_to_remove(system_info.excluded_pkgs))
             # this call can return None, which is not ideal to use with sorted.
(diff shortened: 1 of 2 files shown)


convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -207,13 +207,13 @@ def run(self):
                 description="Repository file packages which could not be removed",
                 diagnosis=message,
             )
-        else:
+        if pkgs_removed:
```
SpyTec, line 210: Keep in mind that `pkgs_removed` wasn't declared outside of the try-catch, so this would fail if that ever changes. I would initialize pkgs_removed before it gets sorted

SpyTec (changes requested):
If there is a Jira issue about fixing this wording, please assign the PR to that. Otherwise create one so it's documented that we address this somewhere

convert2rhel/actions/pre_ponr_changes/handle_packages.py
pr-watson: But with the current state of the code wouldn't it be fine? 
This line `pkgs_removed = sorted(pkghandler.remove_pkgs_unless_from_redhat(pkgs_to_remove) or []) ` ensures that we either get the removed packages or an empty list.

SpyTec (approved):
LGTM

convert2rhel/actions/pre_ponr_changes/handle_packages.py
pr-watson: Okay that makes sense, let me initialize pkgs_removed

PR #1004, opened 2023-12-07
Fix installation of local build

pvoborni:
The integration tests looks to have "forgotten/unused" feature to build convert2rhel at the beginning of test and install it instead of using COPR build. This is useful when developing a test along with feature/code change without a need to open a PR to have a build produced by Packit - it's faster iteration.

It is used when the following environmental args are used in tmt run:
- `ANSIBLE_BUILD_RPM=build`
- `ANSIBLE_RPM_PROVIDER=local`

It was broken in following ways:
- the previous usage was `ANSIBLE_BUILD_RPM=true` the code expected the `true` value as boolean, resp. there was a mismatch of types and the condition was not evaluated as expected. This patch is changing it to expect a "build" string to avoid this problem, i.e. to not use "truthy" values.
- `scripts/extract_version_from_rpm_spec.py` script which is used to figure out what is the version-release part of rpm name did not support a version with 3 parts, e.g. the current `1.5.0` and thus it did not return the `name` part. This patch adds support for additional "." + "number".
- the Ansible play to copy built rpms to provisioned host supported only "docker" and "virtual" providers of tmt. The "virtual" part was generalized into "non-docker" to support other types and assuming that they have SSH support.

With this, the integration tests can be executed e.g. the following way:

```
$ tmt \
   --context distro=centos-7 \
   run \
      -vvv --all \
      -e ANSIBLE_BUILD_RPM=build \
      -e ANSIBLE_RPM_PROVIDER=local \
   provision \
      --how virtual  \
      --image https://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2 \
   plan \
     --name $TEST_PLAN_NAME
```

Jira Issues: no ticket, though was done in context of [RHELC-1226](https://issues.redhat.com/browse/RHELC-1226)

Checklist

### DIFF
--- a/scripts/extract_version_from_rpm_spec.py
+++ b/scripts/extract_version_from_rpm_spec.py
@@ -29,8 +29,8 @@ def get_convert2rhel_version(spec_path: str) -> None:
                         (
                             # Any word could be here
                             (?:\w+)|
-                            # Or int or float i.e. 21 or 0.21
-                            (?:\d+[.]*\d*)
+                            # Or int or float i.e. 21 or 0.21 or 1.5.0
+                            (?:\d+[.]*\d*[.]*\d*)
                         )
                         # some special internal rpm spec var (skipping it)
                         (?:%{.+)*$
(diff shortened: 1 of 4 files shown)


hosekadam (approved):
Thanks for fixing this for us and also adding this possibility into the readme file.

PR #1005, opened 2023-12-07
[RHELC-1226] Add --payg option

pvoborni:
To configure host-metering service on RHEL 7 systems after/at the end of   conversion. 

    Thi PR contains:
    - add of the --payg option + unit tests
    - add of integration tests for --payg option
    - add of check to warn when --payg is provided on system other than RHEL 7 
      as it is not suppprted there.
    - integration test for the check
    - temporary patch which should be removed before merging to master

Jira Issues: [RHELC-1226](https://issues.redhat.com/browse/RHELC-1226)

Checklist

### DIFF
--- a/convert2rhel/actions/system_checks/payg.py
+++ b/convert2rhel/actions/system_checks/payg.py
@@ -1,0 +1,44 @@
+# Copyright(C) 2023 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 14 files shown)


pvoborni:
This is set as draft but is basically ready for review.  Draft is here only to not merge it with the `temp: use COPR repo for testing` patch.

bookwar:
As we agreed in Jira, the patch needs to be refactored so that instead of cli option it checks for the cloud system or relies on environment variable

PR #1018, opened 2023-12-13
[RHELC-1293] Introduce pre-commit hook to register custom markers

danmyway:
* Registering the custom markers is oftentimes tedious and can lead to oversight when registering a new marker or removing a deprecated one
* this should ensure, that the configuration stays up-to-date with changes in existing markers

Jira Issues: [RHELC-1293](https://issues.redhat.com/browse/RHELC-1293)

Checklist

### DIFF
--- a/scripts/regenerate_pytest_ini.py
+++ b/scripts/regenerate_pytest_ini.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+import os
+import re
+import sys
+
+
+def find_custom_markers(directories=("convert2rhel/unit_tests", "tests")):
+    """
+    Walk through the directories provided as a tuple in the function argument
+    and look for the custom pytest markers in each .py file.
+    Ignore the builtin markers by removing from the results.
+    """
+    markers = set()
+    # Regex to match the pytest markers
+    marker_pattern = re.compile(r"@pytest\.mark\.(?P<marker>\w+)\(?")
+
+    for directory in directories:
+        for root, dirs, files in os.walk(directory):
+            # Exclude hidden directories, e.g., .git and venv dirs. This doesn't really
+            # need to be here at this point, however might come in use if we decide to walk
+            # through each directory in the project not just the two currently specified.
+            dirs[:] = [d for d in dirs if not d[0] == "." and "venv" not in d]
+            for file in files:
+                if file.endswith(".py"):
+                    file_path = os.path.join(root, file)
+                    with open(file_path, "r", encoding="utf-8") as f:
+                        content = f.read()
+                        matches = marker_pattern.findall(content)
+                        markers.update(matches)
+
+    # Exclude built-in markers
+    ignored_builtin_markers = [
+        "filterwarnings",
+        "skip",
+        "skipif",
+        "xfail",
+        "parametrize",
+        "usefixtures",
+        "tryfirst",
+        "trylast",
+    ]
+    custom_markers = markers - set(ignored_builtin_markers)
+
+    return custom_markers
+
+
+def update_pytest_ini(custom_markers, ini_file="pytest.ini"):
+    """
+    Update the pytest.ini config with the custom markers.
+    """
+
+    ini_startswith = '[pytest]\ntestpaths = "convert2rhel/unit_tests"\nmarkers =\n'
+    with open(ini_file, "w", encoding="utf-8") as f:
+        if custom_markers:
+            f.write("# Generated automatically by the regenerate-pytest-ini pre-commit hook\n\n")
+            f.write(ini_startswith)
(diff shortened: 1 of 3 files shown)


scripts/regenerate_pytest_ini.py
```diff
@@ -0,0 +1,76 @@
...
+        "usefixtures",
+        "tryfirst",
+        "trylast",
+    ]
+    custom_markers = markers - set(ignored_builtin_markers)
+
+    return custom_markers
+
+
+def update_pytest_ini(custom_markers, ini_file="pytest.ini"):
```
r0x0d, line 47: Since it's a ini file, maybe you could use the configparser module to deal with it? https://docs.python.org/3/library/configparser.html

scripts/regenerate_pytest_ini.py
```diff
@@ -0,0 +1,76 @@
...
+
+def find_custom_markers(directories=("convert2rhel/unit_tests", "tests")):
+    """
+    Walk through the directories provided as a tuple in the function argument
+    and look for the custom pytest markers in each .py file.
+    Ignore the builtin markers by removing from the results.
+    """
+    markers = set()
+    # Regex to match the pytest markers
+    marker_pattern = re.compile(r"@pytest\.mark\.(?P<marker>\w+)\(?")
```
kokesak, line 15: `^@pytest\.mark\.(\w)+` should be good enough

scripts/regenerate_pytest_ini.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def update_pytest_ini(custom_markers, ini_file="pytest.ini"):
+    """
+    Update the pytest.ini config with the custom markers.
+    """
+
+    ini_startswith = '[pytest]\ntestpaths = "convert2rhel/unit_tests"\nmarkers =\n'
+    with open(ini_file, "w", encoding="utf-8") as f:
+        if custom_markers:
```
kokesak, line 54: I would move the condition to the beginning of the function:
```suggestion
    if custom_markers:
        # Nothing to do
        return

    ini_startswith = '[pytest]\ntestpaths = "convert2rhel/unit_tests"\nmarkers =\n'
    with open(ini_file, "w", encoding="utf-8") as f:
```

kokesak (changes requested):

Venefilyn:
Can we use this instead? Means we can get rid of a lot of the integration test specific markers (and just update the naming scheme of it)
https://docs.pytest.org/en/latest/example/markers.html#automatically-adding-markers-based-on-test-names

danmyway:
> Can we use this instead? Means we can get rid of a lot of the integration test specific markers (and just update the naming scheme of it)

@Venefilyn I was looking at this option as well.
IIUIC from the code example, there needs to be a common pattern in the `nodeid` to base the conditioning on.
Our markers are dynamic by default, so I don't think we would be able to implement this.

Venefilyn:
> > Can we use this instead? Means we can get rid of a lot of the integration test specific markers (and just update the naming scheme of it)
> 
> @Venefilyn I was looking at this option as well. IIUIC from the code example, there needs to be a common pattern in the `nodeid` to base the conditioning on. Our markers are dynamic by default, so I don't think we would be able to implement this.

@danmyway but if all of the markers for integration tests are with `test_` then we have a standard already that we can go by. Or we can rename it all to start with `test_int_` for example

danmyway:
@Venefilyn ah, nvm, I got it wrong.
It looks like that adding the markers work, however we still need to register those in pytest.ini, which is where we aim with this PR

Venefilyn:
> @Venefilyn ah, nvm, I got it wrong. It looks like that adding the markers work, however we still need to register those in pytest.ini, which is where we aim with this PR

@danmyway We wouldn't need to register the automated ones with a prefix, just need to do the script as the documentation says

danmyway:
> > @Venefilyn ah, nvm, I got it wrong. It looks like that adding the markers work, however we still need to register those in pytest.ini, which is where we aim with this PR
> 
> @danmyway We wouldn't need to register the automated ones with a prefix, just need to do the script as the documentation says

@Venefilyn I've copy-pasted the example from the docs, and it still yields the warning 
```
conftest.py:65
  conftest.py:65: PytestUnknownMarkWarning: Unknown pytest.mark.interface - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    item.add_marker(pytest.mark.interface)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

```
But nvm, I'm closing this, we have found a way to just ignore the warning instead, that should be easier.

PR #1029, opened 2024-01-10
[RHELC-1131, RHELC-1234] Refactor logger to not require root

SpyTec:
To do

- [x] Potentially changing integration tests
- [x] Testing locally
- [x] Fix buffer going to the log file on init

In short of the changes:

- Split the log handlers into an initialize and finalize function, where FileHandler is added later
- Logging when not being root no longer causes logs to not work or function correctly (as FileHandler isn't added)
- Also fixes the issue of clearing the log files if a duplicate process is found, we just initialize the FileHandler later
- Custom memory buffer is added to keep the last 100 logs, and once FileHandler is added we instantly log the buffer to it

Jira Issues: 
- [RHELC-1131](https://issues.redhat.com/browse/RHELC-1131)
- [RHELC-1234](https://issues.redhat.com/browse/RHELC-1234)

Checklist

### DIFF
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -77,9 +77,10 @@ class LogfileBufferHandler(BufferingHandler):
     name = "logfile_buffer_handler"
 
     def __init__(self, capacity, handler_name="file_handler"):
-        """_summary_
+        """
+        Initialize the handler with the buffer size.
 
-        :param int capacity: Initialize with a buffer size
+        :param int capacity: Buffer size for the handler
         :param str handler_name: Handler to flush buffer to, defaults to "file_handler"
         """
         super(LogfileBufferHandler, self).__init__(capacity)
@@ -88,8 +89,9 @@ class LogfileBufferHandler(BufferingHandler):
 
     @property
     def target(self):
-        """The computed Filehandler target that we are supposed to send to. This
-        is mostly copied over from logging's MemoryHandler but instead of setting
+        """The computed Filehandler target that we are supposed to send to.
+
+        This is mostly copied over from logging's MemoryHandler but instead of setting
         the target manually we find it automatically given the name of the handler
 
         :return logging.Handler: Either the found FileHandler setup or temporary NullHandler
@@ -104,8 +106,8 @@ class LogfileBufferHandler(BufferingHandler):
             self.target.handle(record)
 
     def shouldFlush(self, record):
-        """We should never flush automatically, so we set this to always return false, that way we need to flush
-        manually each time. Which is exactly what we want when it comes to keeping a buffer before we confirm we are
+        """
+        We should never flush automatically, so we set this to always return false, that way we need to flush manually each time. Which is exactly what we want when it comes to keeping a buffer before we confirm we are
         a root user.
 
         :param logging.LogRecord record: The record to log
@@ -145,9 +147,8 @@ def setup_logger_handler():
     stdout_handler.setLevel(logging.DEBUG)
     logger.addHandler(stdout_handler)
 
-    # Setup a buffer for the file handler that we will add later, that way we
     # can flush logs to the file that were logged before initializing the file handler
-    logger.addHandler(LogfileBufferHandler(100))
+    logger.addHandler(LogfileBufferHandler(capacity=100))
 
 
 def add_file_handler(log_name, log_dir):
--- a/convert2rhel/logger.py
+++ b/convert2rhel/logger.py
@@ -35,10 +35,14 @@ import os
 import shutil
 import sys
 
+from logging.handlers import BufferingHandler
 from time import gmtime, strftime
 
 
 LOG_DIR = "/var/log/convert2rhel"
+
+# get root logger
+logger = logging.getLogger("convert2rhel")
 
 
 class LogLevelCriticalNoExit:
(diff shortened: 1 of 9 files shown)


convert2rhel/logger.py
```diff
@@ -58,11 +62,52 @@ class LogLevelFile:
...
+        """
+        # We already have the filehandler, so we can
+        if self._target:
+            return self._target
+
+        for handler in logger.handlers:
+            if handler.name == "file_handler":
+                self._target = handler
+                return self._target
+        return None
```
r0x0d, line 88: To comply with the docstring, should we change this to return 

```suggestion
        return logging.NullHandler()
```

Instead of returning None?

SpyTec: Fixed!

tests/integration/tier0/non-destructive/basic-sanity-checks/test_basic_sanity_checks.py
```diff
@@ -23,9 +23,7 @@ def test_check_user_privileges(shell):
     # Check the program exits as it is required to be run by root
     assert result.returncode != 0
     # Check the program exits for the correct reason
-    assert (
-        result.output == "The tool needs to be run under the root user.\n" "\n" "No changes were made to the system.\n"
-    )
+    assert "The tool needs to be run under the root user.\n" "\n" "No changes were made to the system." in result.output
```
danmyway, line 26:
```suggestion
    assert "The tool needs to be run under the root user." "\n" "No changes were made to the system." in result.output
```
There's one extra newline which makes the validation fail.

danmyway (changes requested):

tests/integration/tier0/non-destructive/basic-sanity-checks/test_basic_sanity_checks.py
SpyTec: Thanks!

convert2rhel/logger.py
```diff
@@ -58,11 +62,63 @@ class LogLevelFile:
...
+
+    Caveat of that approach is that any logging prior to the initialization of
+    the FileHandler would be lost, to help with this we have this custom handler
+    which will keep a buffer of the logs and flush it to the FileHandler
+    """
+
+    name = "logfile_buffer_handler"
+
+    def __init__(self, capacity, handler_name="file_handler"):
+        """_summary_
```
r0x0d, line 80: Missing a summary here

convert2rhel/logger.py
```diff
@@ -58,11 +62,63 @@ class LogLevelFile:
...
+        for record in self.buffer:
+            self.target.handle(record)
+
+    def shouldFlush(self, record):
+        """We should never flush automatically, so we set this to always return false, that way we need to flush
+        manually each time. Which is exactly what we want when it comes to keeping a buffer before we confirm we are
+        a root user.
+
+        :param logging.LogRecord record: The record to log
+        :return bool: Always returns false
```
r0x0d, line 112:
```suggestion
        """
        We should never flush automatically, so we set this to always return false, that way we need to flush manually each time. Which is exactly what we want when it comes to keeping a buffer before we confirm we are
        a root user.

        :param logging.LogRecord record: The record to log
        :return bool: Always returns false
```

convert2rhel/logger.py
```diff
@@ -91,22 +145,47 @@ def setup_logger_handler(log_name, log_dir):
     stdout_handler.setLevel(logging.DEBUG)
     logger.addHandler(stdout_handler)
 
-    # create file handler
+    # Setup a buffer for the file handler that we will add later, that way we
+    # can flush logs to the file that were logged before initializing the file handler
+    logger.addHandler(LogfileBufferHandler(100))
```
r0x0d, line 150: Not something super necessary or needed, but I think adding the parameter name helps to identify what `100`means here
```suggestion
    logger.addHandler(LogfileBufferHandler(capacity=100))
```

convert2rhel/main.py
```diff
@@ -56,12 +56,16 @@ def initialize_logger(log_name, log_dir):
         log_dir.
     """
 
+    return logger_module.setup_logger_handler()
+
+
+def finalize_logger(log_name, log_dir):
```
r0x0d, line 62: Docstrings, please!

convert2rhel/main.py
```diff
@@ -98,6 +100,11 @@ def main_locked():
 
     pre_conversion_results = None
     process_phase = ConversionPhase.POST_CLI
+
+    # since we now have root, we can add the FileLogger
+    # and also archive previous logs
+    finalize_logger("convert2rhel.log", logger_module.LOG_DIR)
```
r0x0d, line 106: I'm having a bit of difficulty to wrap my head around the name here... It doesn't seem to me that it is finalizing the logger instantiated before, but rather, appending a file handler logger to it. 

Maybe I'm missing something?

convert2rhel/utils.py
```diff
@@ -287,11 +287,13 @@ def get_executable_name():
     return os.path.basename(inspect.stack()[-1][1])
 
 
+def is_root():
+    return os.geteuid() == 0
```
r0x0d, line 291: I don't think this function is adding too much value to the `require_root` function. The old way was simple enough to understand. 

It was quicker to simply look at the if statement and figure out how the condition would be met or not, here now we have to hop into a function, get a boolean result from it and negate that same result inside the `require_root`. 

I would return the check to `os.geteuid() != 0` instead.

convert2rhel/unit_tests/logger_test.py
```diff
@@ -32,7 +33,8 @@ def test_logger_handlers(monkeypatch, tmpdir, caplog, read_std, is_py2, global_t
     # initializing the logger first
     log_fname = "convert2rhel.log"
     global_tool_opts.debug = True  # debug entries > stdout if True
-    logger_module.setup_logger_handler(log_name=log_fname, log_dir=str(tmpdir))
+    logger_module.setup_logger_handler()
```
r0x0d, line 36: Can we get rid of the `caplog` and `is_py2` fixtures in this test? They seem to not be used.

convert2rhel/unit_tests/logger_test.py
```diff
@@ -125,7 +127,7 @@ def test_logger_custom_logger_insufficient_level(
     ):
         """Test CustomLogger."""
         log_fname = "convert2rhel.log"
-        logger_module.setup_logger_handler(log_name=log_fname, log_dir=str(tmpdir))
+        logger_module.setup_logger_handler()
```
r0x0d, line 130: `level_name`, `monkeypatch`, `tmpdir` and `log_fname` not used

convert2rhel/unit_tests/logger_test.py
```diff
@@ -138,7 +140,7 @@ def test_logger_custom_logger_insufficient_level(
     def test_logger_critical_insufficient_level(self, caplog, tmpdir, clear_loggers):
         """Test CustomLogger."""
         log_fname = "convert2rhel.log"
-        logger_module.setup_logger_handler(log_name=log_fname, log_dir=str(tmpdir))
+        logger_module.setup_logger_handler()
```
r0x0d, line 143: `tmpdir` and `log_fname` not used

convert2rhel/unit_tests/logger_test.py
```diff
@@ -184,3 +186,27 @@ def test_archive_old_logger_files(log_name, path_exists, tmpdir, caplog):
...
+def test_logfile_buffer_handler(read_std):
+    logbuffer_handler = logger_module.LogfileBufferHandler(2, "custom_name")
+    logger = logging.getLogger("convert2rhel")
+    logger.addHandler(logbuffer_handler)
+
+    logger.warning("message 1")
+    logger.warning("message 2")
+
+    # flushing without other handlers should work, it will just go to NullHandlers
+    logbuffer_handler.flush()
```
r0x0d, line 201: Do you know if there is a way to assert that the NullHandler actually got the messages after the flush? 

Not sure if that would be possible, probably not, right?

r0x0d (commented):
Looks good. Only a few comments & suggestions.

convert2rhel/unit_tests/logger_test.py
SpyTec: We can mock the NullHandler but I don't think it's worth the effort given it's supposed to throw it away right away. WDYT?

convert2rhel/main.py
SpyTec: It is more or less the same as logger is a global concept for the module name you put in. I'll rename it to match what it does more accurately

SpyTec:
@r0x0d addressed all review comments!

convert2rhel/unit_tests/logger_test.py
r0x0d: Makes sense. I don't think mocking it will be of any help.

r0x0d (approved):
Looks good to me! Thanks for applying the suggestions.

jochapma (commented):
Do we test what happens when we log messages beyond the buffer's capacity?

SpyTec:
> Do we test what happens when we log messages beyond the buffer's capacity?

@jochapma Yes! I do that here

https://github.com/SpyTec/convert2rhel/blob/bcfd4ea9917862ae267287b45f1c7ea3fe1f9466/convert2rhel/unit_tests/logger_test.py#L190-L205

SpyTec:
> Looks good to me! Thanks for applying the suggestions.
> 
> Have not tested it under a VM yet.

@r0x0d I have tested it in a VM and the buffer works. It outputs the command used into the logfile

PR #1030, opened 2024-01-11
[RHELC-1282] Override yum config exclude list

r0x0d:
Override the exclude list from the yum config file to avoid dependency problems during the transaction process.

This method is temporary as it will only override the config file while running the transaction, no modifications to the config file physically will be done.

Jira Issues: [RHELC-1282](https://issues.redhat.com/browse/RHELC-1282)

Checklist


r0x0d:
Suggestion from @kokesak in slack: 

>We could simply add the redhat-release-server package to the yum conf exclude list, boot into an older kernel (for example, 3.10.0-1160.95.1.el7.x86_64) and run the transaction in analysis mode. We can check explicitly for not having an occurrence of the following packages in the transaction logs: 
>
>- grubby-8.28-26.el7.x86_64 requires system-release
>- kernel-3.10.0-1160.105.1.el7.x86_64 requires system-release
>- setup-2.8.71-11.el7.noarch requires system-release
>- initscripts-9.49.53-1.el7_9.1.x86_64 requires redhat-release

r0x0d:
This is supposed to land on 1.8.

pr-watson (approved):
Looks good to me

SpyTec (approved):

r0x0d:
@danmyway, for some reason, the newly added tests are picking an "older" version instead of what is in the latest commit...

danmyway:
@r0x0d I don't think I follow, do you have an example somewhere?

r0x0d:
> @r0x0d I don't think I follow, do you have an example somewhere?

Nevermind... I think I was just seeing the wrong test.

SpyTec:
@r0x0d can you rebase and fix conflicts?

tests/integration/conftest.py
```diff
@@ -623,3 +623,87 @@ def remediation_out_of_date_packages(shell):
...
     for package in problematic_packages:
         shell(f"yum update -y {package}")
+
+
+@pytest.fixture(scope="function")
+def kernel_check_envar():
+    """
+    Fixture.
+    Set CONVERT2RHEL_UNSUPPORTED_SKIP_KERNEL_CURRENCY_CHECK environment variable
+    to skip the kernel currency check.
```
SpyTec, line 633:
```suggestion
    Fixture.
    Set CONVERT2RHEL_SKIP_KERNEL_CURRENCY_CHECK environment variable
    to skip the kernel currency check.
```

PR #1032, opened 2024-01-11
[RHELC-1067] Automatically tag every Friday

SpyTec:
Jira Issues: [RHELC-1067](https://issues.redhat.com/browse/RHELC-1067)

Checklist

### DIFF
--- a/.github/workflows/autotag.yml
+++ b/.github/workflows/autotag.yml
@@ -1,0 +1,30 @@
+name: Auto Tag
+
+on:
+  schedule:
+    - cron: "1 12 * * 5"
+
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get Latest Tag
+        id: current_tag
+        run: echo "tag=$(git describe --tags --abbrev=0)" >> $GITHUB_OUTPUT
+
+      - name: Create tag
+        id: create_tag
+        run: echo "tag=${{ steps.current_tag.outputs.tag }}-${{ github.run_number }}" >> $GITHUB_OUTPUT
+
+      - name: Push tag
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "GitHub Actions"
+          git tag ${{ steps.create_tag.outputs.tag  }}
+          git push origin ${{ steps.create_tag.outputs.tag }}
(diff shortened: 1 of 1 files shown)


.github/workflows/autotag.yml
```diff
@@ -0,0 +1,30 @@
+name: Auto Tag
+
+on:
+  schedule:
+    - cron: "1 12 * * 5"
```
r0x0d, line 5: Looks good to me, but why the `1` there instead of `0`? Any particular reason?

r0x0d (approved):

.github/workflows/autotag.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Get Latest Tag
+        id: current_tag
+        run: echo "tag=$(git describe --tags --abbrev=0)" >> $GITHUB_OUTPUT
+
+      - name: Create tag
+        id: create_tag
+        run: echo "tag=${{ steps.current_tag.outputs.tag }}-${{ github.run_number }}" >> $GITHUB_OUTPUT
```
r0x0d, line 23: Maybe we could simplify? 

```suggestion
      - name: Create tag
        id: create_tag
        run: echo "tag=$(git describe --tags --abbrev=0)-${{ github.run_number }}" >> $GITHUB_OUTPUT
```

r0x0d:
Also, probably you have come across this already, but found out this action while searching for alternatives: https://github.com/actions-ecosystem/action-push-tag. 

It seems quite simple and efficient.

.github/workflows/autotag.yml
SpyTec: I always went into the habit of not doing round numbers for scheduled stuff due to everyone doing so. So there are usually spikes of activity around the top of the hour on CI infra where you can do cronjobs

.github/workflows/autotag.yml
SpyTec: True, doesn't serve much purpose to split it

SpyTec:
> Also, probably you have come across this already, but found out this action while searching for alternatives: [actions-ecosystem/action-push-tag](https://github.com/actions-ecosystem/action-push-tag).
> 
> It seems quite simple and efficient.

@r0x0d I did see it but felt like it didn't give any benefit to doing it manually. Especially since it adds a dependency that does something super simple

.github/workflows/autotag.yml
r0x0d: Sounds good! Thanks.

PR #1035, opened 2024-01-17
[RHELC-1130] Swap base os specific packages

hosekadam:
The problem is related to how we remove the excluded packages. When “rpm -e –nodeps” is used, yum doesn’t know that the package was removed. On the removed package can be depending some other package (pkg1 depends on removed pkg). So then during processing the [yum operations](https://github.com/oamg/convert2rhel/blob/main/convert2rhel/pkgmanager/handlers/yum/__init__.py#L166) (_perform_operations), specifically update operation (reinstall is fine), the dependency (which was satisfied before conversion and then removed by us manually using rpm -e –nodeps) is not checked and results in missing dependency. 

This problem with not checking the dependency is solved in later versions, but in the one present in our converted systems. So we need to swap the packages mainly on CentOS 7 (and similar) systems.

The solution uses “yum swap” command, which is based on reinstall and install the package using yum api. The “yum swap” is mostly reimplemented on our side, using the api calls (_swap_problematic_packages).

We can use the swap command when we are removing a package from the original system vendor (like centos-logos) and this package has its replacement from rhel (in this case redhat-logos). This requires the mapping in the configuration file.

Jira Issues: [RHELC-1130](https://issues.redhat.com/browse/RHELC-1130)

Checklist

### DIFF
--- a/convert2rhel/data/7/ppc64/configs/centos-7-ppc64.cfg
+++ b/convert2rhel/data/7/ppc64/configs/centos-7-ppc64.cfg
@@ -10,11 +10,13 @@ gpg_fingerprints =
 # List of packages to be removed before the system conversion starts.
 # Delimited by any whitespace(s).
 excluded_pkgs =
-  centos-bookmarks
-  centos-logos
-  centos-indexhtml
-  libreport-centos
-  libreport-plugin-mantisbt
+
+# Mapping of packages that need to be swapped during the transaction
+swap_pkgs =
+  centos-logos | redhat-logos
+  centos-indexhtml | redhat-indexhtml
+  centos-bookmarks | redhat-bookmarks
+  libreport-centos | libreport-rhel
 
 # List of packages that either contain repofiles or affect variables in the repofiles (e.g. $releasever).
 # Delimited by any whitespace(s).
--- a/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
+++ b/convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
@@ -9,19 +9,21 @@ gpg_fingerprints = 24c6a8a7f4a80eb5
 # Delimited by any whitespace(s).
 excluded_pkgs =
   anaconda-dracut
-  centos-bookmarks
-  centos-indexhtml
-  centos-logos
   cockpit-subscriptions
   geoipupdate
   kmod-kvdo
-  libreport-centos
-  libreport-plugin-mantisbt
   mod_ldap
   mod_proxy_html
   rhn*
   yum-rhn-plugin
   gnome-documents-libs
+
+# Mapping of packages that need to be swapped during the transaction
+swap_pkgs =
+  centos-logos | redhat-logos
+  centos-indexhtml | redhat-indexhtml
+  centos-bookmarks | redhat-bookmarks
+  libreport-centos | libreport-rhel
 
 # List of packages that either contain repofiles or affect variables in the repofiles (e.g. $releasever).
 # Delimited by any whitespace(s).
--- a/convert2rhel/pkgmanager/handlers/dnf/__init__.py
+++ b/convert2rhel/pkgmanager/handlers/dnf/__init__.py
@@ -111,6 +111,26 @@ class DnfTransactionHandler(TransactionHandlerBase):
                 diagnosis="Loading repository metadata failed with error %s." % (str(e)),
             )
 
+    def _swap_base_os_specific_packages(self):
+        """Swap base os specific packages for their RHEL counterparts in the transaction.
+
+        Some packages need to be manually injected in the transaction as a
+        "swap", since those packages are not always able to be installed
+        automatically by yum if they don't exist in the system anymore, this
+        can cause problems during the transaction as missing dependencies.
+        """
+        # Related issue: https://issues.redhat.com/browse/RHELC-1130, see comments
+        # to get more proper description of solution
+        for old_package, new_package in system_info.swap_pkgs.items():
+            loggerinst.debug("Checking if %s installed for later swap." % old_package)
+            is_installed = system_info.is_rpm_installed(old_package)
+            if is_installed:
+                loggerinst.debug("Package %s will be swapped to %s during conversion." % (old_package, new_package))
+                # Order of commands based on DNF implementation of swap, different from YUM order:
+                # https://github.com/rpm-software-management/dnf/blob/master/dnf/cli/commands/swap.py#L60
+                self._base.install(pkg_spec=new_package)
+                self._base.remove(pkg_spec=old_package)
+
     def _perform_operations(self):
         """Perform the necessary operations in the transaction.
 
@@ -145,6 +165,8 @@ class DnfTransactionHandler(TransactionHandlerBase):
                     self._base.downgrade(pkg_spec=pkg)
                 except pkgmanager.exceptions.PackagesNotInstalledError:
                     loggerinst.warning("Package %s not available in RHEL repositories.", pkg)
+
+        self._swap_base_os_specific_packages()
 
     def _resolve_dependencies(self):
         """Resolve the dependencies for the transaction.
--- a/convert2rhel/pkgmanager/handlers/yum/__init__.py
+++ b/convert2rhel/pkgmanager/handlers/yum/__init__.py
@@ -163,6 +163,26 @@ class YumTransactionHandler(TransactionHandlerBase):
                 diagnosis="Loading repository metadata failed with error %s." % (str(e)),
             )
 
+    def _swap_base_os_specific_packages(self):
+        """Swap base os specific packages for their RHEL counterparts in the transaction.
+
+        Some packages need to be manually injected in the transaction as a
+        "swap", since those packages are not always able to be installed
+        automatically by yum if they don't exist in the system anymore, this
+        can cause problems during the transaction as missing dependencies.
+        """
+        # Related issue: https://issues.redhat.com/browse/RHELC-1130, see comments
+        # to get more proper description of solution
+        for old_package, new_package in system_info.swap_pkgs.items():
+            loggerinst.debug("Checking if %s installed for later swap." % old_package)
+            is_installed = system_info.is_rpm_installed(old_package)
+            if is_installed:
+                loggerinst.debug("Package %s will be swapped to %s during conversion." % (old_package, new_package))
+                # Order of operations based on YUM implementation of swap:
+                # https://github.com/rpm-software-management/yum/blob/master/yumcommands.py#L3488
+                self._base.remove(pattern=old_package)
+                self._base.install(pattern=new_package)
+
     def _perform_operations(self):
         """Perform the necessary operations in the transaction.
 
@@ -187,7 +207,6 @@ class YumTransactionHandler(TransactionHandlerBase):
                 # being outdated in the system after the conversion.
                 if can_update:
                     continue
-
                 try:
                     self._base.reinstall(pattern=pkg)
                 except (pkgmanager.Errors.ReinstallInstallError, pkgmanager.Errors.ReinstallRemoveError):
@@ -199,6 +218,10 @@ class YumTransactionHandler(TransactionHandlerBase):
                         pkgmanager.Errors.DowngradeError,
                     ):
                         loggerinst.warning("Package %s not available in RHEL repositories.", pkg)
+
+            # Swapping the packages needs to be after the operations
+            # If not, swapped packages are removed from transaction as obsolete
+            self._swap_base_os_specific_packages()
         except pkgmanager.Errors.NoMoreMirrorsRepoError as e:
             loggerinst.debug("Got the following exception message: %s", e)
             loggerinst.critical_no_exit("There are no suitable mirrors available for the loaded repositories.")
--- a/convert2rhel/systeminfo.py
+++ b/convert2rhel/systeminfo.py
@@ -93,6 +93,8 @@ class SystemInfo:
         self.eus_system = None
         # Packages to be removed before the system conversion
         self.excluded_pkgs = []
+        # Packages that need to perform a swap in the transaction
+        self.swap_pkgs = {}
         # Release packages to be removed before the system conversion
         self.repofile_pkgs = []
         self.cfg_filename = None
@@ -124,6 +126,7 @@ class SystemInfo:
         self.cfg_filename = self._get_cfg_filename()
         self.cfg_content = self._get_cfg_content()
         self.excluded_pkgs = self._get_excluded_pkgs()
+        self.swap_pkgs = self._get_swap_pkgs()
         self.repofile_pkgs = self._get_repofile_pkgs()
         self.default_rhsm_repoids = self._get_default_rhsm_repoids()
         self.eus_rhsm_repoids = self._get_eus_rhsm_repoids()
@@ -250,6 +253,36 @@ class SystemInfo:
 
     def _get_excluded_pkgs(self):
         return self._get_cfg_opt("excluded_pkgs").split()
+
+    def _get_swap_pkgs(self):
+        pkgs_to_swap = {}
+
+        try:
+            lines = self._get_cfg_opt("swap_pkgs").strip().split("\n")
+            for line in lines:
+                old_package, new_package = tuple(line.split("|"))
+
+                old_package = old_package.strip()
+                new_package = new_package.strip()
+
+                if old_package in pkgs_to_swap:
+                    self.logger.warning(
+                        "Package {old_package} redefined in swap packages list.\n"
+                        "Old package {old_package} will be swapped by {newest_package} instead of {new_package}.".format(
+                            old_package=old_package, new_package=pkgs_to_swap[old_package], newest_package=new_package
+                        )
+                    )
+                pkgs_to_swap.update({old_package: new_package})
+
+        except ValueError:
+            # Leave the swap packages dict empty, packages for swap aren't defined
+            self.logger.debug("Leaving the swap package list empty. No packages defined.")
+
+        except AttributeError:
+            # Leave the swap packages dict empty, missing swap_pkgs in config file
+            self.logger.warning("Leaving the swap package list empty. Missing swap_pkgs key in configuration file.")
+
+        return pkgs_to_swap
 
     def _get_repofile_pkgs(self):
         return self._get_cfg_opt("repofile_pkgs").split()
--- a/convert2rhel/unit_tests/pkgmanager/handlers/dnf/dnf_test.py
+++ b/convert2rhel/unit_tests/pkgmanager/handlers/dnf/dnf_test.py
@@ -125,6 +125,8 @@ class TestDnfTransactionHandler:
         monkeypatch.setattr(pkgmanager.Base, "do_transaction", value=mock.Mock())
         monkeypatch.setattr(pkgmanager.Base, "transaction", value=mock.Mock())
         monkeypatch.setattr(pkgmanager.Base, "sack", value=SackMock())
+        monkeypatch.setattr(pkgmanager.Base, "install", value=mock.Mock())
+        monkeypatch.setattr(pkgmanager.Base, "remove", value=mock.Mock())
 
     @centos8
     def test_set_up_base(self, pretend_os):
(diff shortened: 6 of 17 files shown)


convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -186,19 +225,24 @@ def _perform_operations(self):
                 # package. This is an inconsistency that could lead to packages
                 # being outdated in the system after the conversion.
                 if can_update:
+                    loggerinst.info("Package %s will be updated.", pkg)
```
r0x0d, line 228: Just a reminder that we can get rid of this later.

We don't need the info log for update/reinstall/downgrade.

convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
```diff
@@ -23,6 +21,11 @@ excluded_pkgs =
   yum-rhn-plugin
   gnome-documents-libs
 
+# Mapping of packages that need to be swapped during the transaction
+swap_pkgs =
+  centos-logos | redhat-logos
+  centos-indexhtml | redhat-indexhtml
```
r0x0d, line 27: We will need to adapt this same field to the other configs too

convert2rhel/pkgmanager/handlers/yum/__init__.py
hosekadam: I will remove it from this PR. But I'm wondering if this cannot be used for better debugging of the yum transaction. If yes, then it could be part of https://issues.redhat.com/browse/RHELC-999 What do you think? @r0x0d

hosekadam:
Finished the part with YUM swap (tests, comments). I have a few questions:
- Do you agree with `|` symbol in config file for mapping the packages?
- Do we want to implement the swap for other system than CentOS 7 (and similar, like Oracle 7)?  
The bug was occurring only on system with YUM package manager, on DNF wasn't any problem. My suggestion is to add to the config file note, that the `swap_pkgs` variable is used only on systems with YUM. And don't implement the swap for dnf.

What do you think? @r0x0d @bookwar @bocekm ? I added a short summary to the description of this PR.

I also added mapping for the packages which I found has equivalent of them in rhel repositories. Mostly, I finished up with packages centos-* mapped to redhat-* for now (so added just `centos-bookmarks`  from the previous state). I have one more candidate `libreport-centos`, but I'm not sure about this one if it couldn't cause a problem. I need to test them more (all of them). So the mapping is still in progress.

convert2rhel/pkgmanager/handlers/yum/__init__.py
r0x0d: So, I guess this is not related. The log message we used here to debug is mainly for us to identify which package is being added to the transaction and the operation related to the package. 

What the ticket https://issues.redhat.com/browse/RHELC-999 asks for is to capture error messages and have a better handling for that. We could potentially solve it by improving our callback classes and try to capture and output better messages there. 

Also, the errors about dependency solving only happen when we reach the part where try to resolve the dependencies. 

---

Said that, there might be one benefit to log those packages as a `DEBUG`. That would help us to understand the operation tied to the package, potentially saving us some time in the analysis when a report is opened. 

But I don't think we need to leave this for now, again, improving our callback classes might make more sense than placing those logs here.

convert2rhel/systeminfo.py
```diff
@@ -251,6 +254,36 @@ def _get_gpg_key_fingerprints(self):
...
+                pkgs_to_swap.update({old_package: new_package})
+
+        except ValueError:
+            # Leave the swap packages dict empty, packages for swap aren't defined
+            self.logger.debug("Leaving the swap package list empty. No packages defined.")
+
+        except AttributeError:
+            # Leave the swap packages dict empty, missing swap_pkgs in config file
+            self.logger.warning("Leaving the swap package list empty. Missing swap_pkgs key in configuration file.")
+
```
r0x0d, line 284: I liked the extra safety here! Pretty clever!

r0x0d:
> Finished the part with YUM swap (tests, comments). I have a few questions:
> 
>     * Do you agree with `|` symbol in config file for mapping the packages?
> 
>     * Do we want to implement the swap for other system than CentOS 7 (and similar, like Oracle 7)?
>       The bug was occurring only on system with YUM package manager, on DNF wasn't any problem. My suggestion is to add to the config file note, that the `swap_pkgs` variable is used only on systems with YUM. And don't implement the swap for dnf.
> 
> 
> What do you think? @r0x0d @bookwar @bocekm ? I added a short summary to the description of this PR.
> 
> I also added mapping for the packages which I found has equivalent of them in rhel repositories. Mostly, I finished up with packages centos-* mapped to redhat-* for now (so added just `centos-bookmarks` from the previous state). I have one more candidate `libreport-centos`, but I'm not sure about this one if it couldn't cause a problem. I need to test them more (all of them). So the mapping is still in progress.

> Do you agree with | symbol in config file for mapping the packages?
I don't mind having the `|`. 

> Do we want to implement the swap for other system than CentOS 7 (and similar, like Oracle 7)?
I do think that we have to propagate the configuration file changes to the other systems to keep it consistent across the variants. 

Without that, the user might not know that this option is supported by us. I'm not quite sure if we need to implement the solution to DNF as well, although, it would be nice to have, no one asked for that yet and it seems that, at least in DNF, the tool deals with it more reliably. 

But I do see that we have a bunch of packages that could benefit from this swap and be removed from the exclude list. For example, maybe this is a good candidate: https://github.com/hosekadam/convert2rhel/blob/pr/hosekadam/1035/convert2rhel/data/8/x86_64/configs/oracle-8-x86_64.cfg#L12-L13

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -163,6 +163,24 @@ def _enable_repos(self):
...
             )
 
+    def _swap_problematic_packages(self):
+        """Swap system packages for their RHEL counterparts in the transaction.
+
+        Some packages need to be manually injected in the transaction as a
+        "swap", since those packages are not always able to be installed
+        automatically by yum if they don't exist in the system anymore, this
+        can cause problems during the transaction as missing dependencies.
+        Related issue: https://issues.redhat.com/browse/RHELC-1130, see comments
```
r0x0d, line 173: I'm wondering if we should have the link to the issue in the docstring... Maybe I would move it to a comment inside the function, or try to give a better example/description inside it. I don't think that docstrings are a good place for putting that type of content in it. 

I don't have a strong opinion on that, so if others are comfortable, I'm okay with that as well.

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -163,6 +163,24 @@ def _enable_repos(self):
...
+        to get more proper description of solution
+        """
+        for old_package, new_package in system_info.swap_pkgs.items():
+            is_installed = system_info.is_rpm_installed(old_package)
+            loggerinst.debug("Checking if %s installed for later swap." % old_package)
```
r0x0d, line 178:
```suggestion
            loggerinst.debug("Checking if %s installed for later swap." % old_package)
            is_installed = system_info.is_rpm_installed(old_package)
```

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -163,6 +163,24 @@ def _enable_repos(self):
...
+            if is_installed:
+                loggerinst.debug("Package %s will be swapped during conversion." % old_package)
```
r0x0d, line 180:
```suggestion
                loggerinst.debug("Package %s will be swapped to %s during conversion." % (old_package, new_package))
```

convert2rhel/systeminfo.py
```diff
@@ -251,6 +254,36 @@ def _get_gpg_key_fingerprints(self):
...
+        return pkgs_to_swap
```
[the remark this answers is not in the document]
hosekadam: Yeah, I totally agree with having the option swap_pkgs in every config file. I just haven't done it yet.

The question is if we want to implement the swap in the code on systems using DNF. But I agree with your [comment](https://github.com/oamg/convert2rhel/pull/1035#issuecomment-1904351885) that having the consistency is great and many packages can benefit from this.

convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
```diff
@@ -9,20 +9,23 @@ gpg_fingerprints = 24c6a8a7f4a80eb5
...
-  libreport-centos
-  libreport-plugin-mantisbt
   mod_ldap
   mod_proxy_html
   rhn*
   yum-rhn-plugin
   gnome-documents-libs
 
+# Mapping of packages that need to be swapped during the transaction
+# TODO libreport might be problematic, is there just for testing
```
r0x0d, line 22: Is it problematic still?

hosekadam: I tested that, and it seems it's not any more problematic any more. The libreport-centos has dependency libreport-plugin-mantisbt (not present in rhel repos) but when the mantisbt is removed from the excluded list in config file, the mantisbt removed during the dependency resolution as problematic - so there is no need to have libreport pkgs on excluded list.

The another good thing is, that the libreport-rhel doesn't require the mantisbt package, so it can be successfully swapped and the conversion is successfull.

The log is there: [http://artifacts.osci.redhat.com/testing-farm/1c32b178-5112-4749-85bf-45cf57cf65d0/work-packages_in_system_and_config_centos7_libreport92n17s89/log.txt](http://artifacts.osci.redhat.com/testing-farm/1c32b178-5112-4749-85bf-45cf57cf65d0/work-packages_in_system_and_config_centos7_libreport92n17s89/log.txt) 

I installed libreport-centos before the conversion (also with the mantisbt as it's dependency) and after the conversion the libreport-rhel is present.

hosekadam:
The all swap pkgs lists for each system were tested on the latest version of it. I've tested by installing the packages to swap prior to the conversion, then run conversion and then checked if the target package is present.

There are the results with links to the logs. The centos7 was tested twice, first attempt for the centos* pkgs and second for the libreport.

- centos7_all_without_libreport - OK http://artifacts.osci.redhat.com/testing-farm/1c32b178-5112-4749-85bf-45cf57cf65d0/work-packages_in_system_and_config_centos7_all_without_libreportvw39w5ne/log.txt
- centos7_libreport - libreport-plugin-mantisbt removed since rhel libreport doesn't need it, OK http://artifacts.osci.redhat.com/testing-farm/1c32b178-5112-4749-85bf-45cf57cf65d0/work-packages_in_system_and_config_centos7_libreport92n17s89/log.txt
- oracle7_all - OK http://artifacts.osci.redhat.com/testing-farm/10aafd54-0587-4362-a990-1ac6c0d9ced7/work-packages_in_system_and_config_oracle7_allataibbkn/log.txt
- centos8_all - OK http://artifacts.osci.redhat.com/testing-farm/1f5d08a6-7feb-46a6-9c00-5d2b8372de0d/work-packages_in_system_and_config_centos8_allf2ltqbdw/log.txt
- oracle8_all - OK http://artifacts.osci.redhat.com/testing-farm/c84992e1-a021-457b-836a-e69b5966769f/work-packages_in_system_and_config_oracle8_allbwk1g7ry/log.txt
- alma8_all - OK http://artifacts.osci.redhat.com/testing-farm/63ff1ff9-8fa1-42f7-8cb6-91155f402283/work-packages_in_system_and_config_alma8_allggf28yse/log.txt
- rocky8_all - OK http://artifacts.osci.redhat.com/testing-farm/8caa6f2b-39dd-43e6-ae00-f461baf445a6/work-packages_in_system_and_config_rocky8_allsh_ghn_p/log.txt

So from my point of view, this is finally ready (after removing the comment from the centos7 config file, I will do it in a moment)!

convert2rhel/data/7/ppc64/configs/centos-7-ppc64.cfg
```diff
@@ -10,11 +10,13 @@ gpg_fingerprints =
 # List of packages to be removed before the system conversion starts.
 # Delimited by any whitespace(s).
 excluded_pkgs =
-  centos-bookmarks
-  centos-logos
-  centos-indexhtml
-  libreport-centos
-  libreport-plugin-mantisbt
```
r0x0d, line 17: What happened to this plugin? Do we have a reason to remove it from the exclude list?

convert2rhel/data/7/x86_64/configs/centos-7-x86_64.cfg
```diff
@@ -9,20 +9,22 @@ gpg_fingerprints = 24c6a8a7f4a80eb5
...
 excluded_pkgs =
   anaconda-dracut
-  centos-bookmarks
-  centos-indexhtml
-  centos-logos
   cockpit-subscriptions
   geoipupdate
   kmod-kvdo
-  libreport-centos
-  libreport-plugin-mantisbt
```
r0x0d, line 19: Same here

r0x0d (commented):
Looks good to me. But an extra pair of eyes might be required before approving. 

Nice work!

convert2rhel/pkgmanager/handlers/dnf/__init__.py
```diff
@@ -111,6 +111,28 @@ def _enable_repos(self):
                 diagnosis="Loading repository metadata failed with error %s." % (str(e)),
             )
 
+    def _swap_problematic_packages(self):
+        """Swap system packages for their RHEL counterparts in the transaction.
```
SpyTec, line 115: I'd change to be something like swap base os specific packages

convert2rhel/pkgmanager/handlers/dnf/__init__.py
```diff
@@ -111,6 +111,28 @@ def _enable_repos(self):
...
+
+        Some packages need to be manually injected in the transaction as a
+        "swap", since those packages are not always able to be installed
+        automatically by yum if they don't exist in the system anymore, this
+        can cause problems during the transaction as missing dependencies.
+        Related issue: https://issues.redhat.com/browse/RHELC-1130, see comments
+        to get more proper description of solution
+        """
+        # Related issue: https://issues.redhat.com/browse/RHELC-1130, see comments
+        # to get more proper description of solution
```
SpyTec, line 125:
```suggestion
        Related issue: https://issues.redhat.com/browse/RHELC-1130, see comments
        to get more proper description of solution
        """
```

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -163,6 +163,26 @@ def _enable_repos(self):
...
+    def _swap_problematic_packages(self):
+        """Swap system packages for their RHEL counterparts in the transaction.
+
+        Some packages need to be manually injected in the transaction as a
+        "swap", since those packages are not always able to be installed
+        automatically by yum if they don't exist in the system anymore, this
+        can cause problems during the transaction as missing dependencies.
+        """
+        # Related issue: https://issues.redhat.com/browse/RHELC-1130, see comments
+        # to get more proper description of solution
```
SpyTec, line 175:
```suggestion
        can cause problems during the transaction as missing dependencies.
        Related issue: https://issues.redhat.com/browse/RHELC-1130, see comments
        to get more proper description of solution
        """

**Review by SpyTec — APPROVED** [review]
Minor comments. Though one thing I'd like changed is the title. Try something that addresses the implementation and then how it fixes the bug in the description. Currently the title reads like a bug report

Implementation looks great!

**Comment by SpyTec** [comment]
Needs a rebase

File: convert2rhel/pkgmanager/handlers/dnf/__init__.py (hunk 11 above)
Comment (hosekadam) [reply convert2rhel/pkgmanager/handlers/dnf/__init__.py:115]: @SpyTec "Swap base os specific packages for their RHEL counterparts in the transaction." is fine?

File: convert2rhel/pkgmanager/handlers/yum/__init__.py (hunk 13 above)
Comment (hosekadam) [reply convert2rhel/pkgmanager/handlers/yum/__init__.py:175]: I discussed those with @r0x0d and we agreed on having the mention of the issue under the main comment, in the `#` section. https://github.com/oamg/convert2rhel/pull/1035#discussion_r1462116129

But I see I left there duplicates, I'll remove them

**Comment by hosekadam** [comment]
Rebased and applied the review

File: convert2rhel/pkgmanager/handlers/dnf/__init__.py (hunk 11 above)
Comment (SpyTec) [reply convert2rhel/pkgmanager/handlers/dnf/__init__.py]: Sure! Reflect the function name too

Comment (hosekadam) [reply convert2rhel/pkgmanager/handlers/dnf/__init__.py]: @SpyTec Do you like `_swap_base_os_specific_packages()` ?

Comment (SpyTec) [reply convert2rhel/pkgmanager/handlers/dnf/__init__.py]: @hosekadam Sure!

**Comment by hosekadam** [comment]
Rebased to have changes from #1095

File: convert2rhel/unit_tests/pkgmanager/handlers/dnf/dnf_test.py [hunk 14]
```diff
@@ -364,3 +370,35 @@ def test_run_transaction(self, pretend_os, validate_transaction, caplog, monkeyp
...
+        ("installed_pkgs", "swap_pkgs", "swaps"),
+        (
+            (["pkg0", "pkg1"], {"pkg0": "new_pkg0", "pkg1": "new_pkg1"}, 2),
+            (["pkg1"], {"pkg0": "new_pkg0", "pkg1": "new_pkg1"}, 1),
+            ([], {"pkg0": "new_pkg0", "pkg1": "new_pkg1"}, 0),
+            (["pkg0", "pkg1", "pkg2"], {}, 0),
+            ([], {}, 0),
+        ),
+    )
+    def test_swap_problematic_packages(
```
r0x0d, line 385:
```suggestion
    def test_swap_base_os_specific_packages(
```

convert2rhel/unit_tests/pkgmanager/handlers/dnf/dnf_test.py
```diff
@@ -215,17 +217,21 @@ def test_enable_repos_repo_error_exception(
 
     @centos8
     def test_perform_operations(self, pretend_os, caplog, monkeypatch):
```
r0x0d, line 219:
```suggestion
    def test_perform_operations(self, pretend_os, monkeypatch):
```

convert2rhel/unit_tests/pkgmanager/handlers/yum/yum_test.py
```diff
@@ -144,14 +146,18 @@ def test_enable_repos_repo_error(self, pretend_os, enabled_rhel_repos, caplog, m
 
     @centos7
     def test_perform_operations(self, pretend_os, caplog, monkeypatch):
```
r0x0d, line 148:
```suggestion
    def test_perform_operations(self, pretend_os, monkeypatch):
```

convert2rhel/unit_tests/pkgmanager/handlers/yum/yum_test.py
```diff
@@ -324,6 +330,38 @@ def test_package_marked_for_update(self, pretend_os, monkeypatch):
...
+        ("installed_pkgs", "swap_pkgs", "swaps"),
+        (
+            (["pkg0", "pkg1"], {"pkg0": "new_pkg0", "pkg1": "new_pkg1"}, 2),
+            (["pkg1"], {"pkg0": "new_pkg0", "pkg1": "new_pkg1"}, 1),
+            ([], {"pkg0": "new_pkg0", "pkg1": "new_pkg1"}, 0),
+            (["pkg0", "pkg1", "pkg2"], {}, 0),
+            ([], {}, 0),
+        ),
+    )
+    def test_swap_problematic_packages(
```
r0x0d, line 344:
```suggestion
    def test_swap_base_os_specific_packages(
```

r0x0d (changes requested):

r0x0d (changes requested):
Few comments, but in general, it looks pretty good!

convert2rhel/unit_tests/pkgmanager/handlers/yum/yum_test.py
hosekadam: Thanks!

r0x0d (approved):

PR #1041, opened 2024-01-19
Split `utils.py` into a module

SpyTec:
Mostly just a proof of concept with changing utils.py to a module, what does people think of the concept (or execution)?

Checklist


jochapma:
It seems a very good idea to me to break up `utils.py` into more logical pieces; having files or classes called utils is often a bad code smell.

This is a good start; it makes sense to reorganize code first and then do any refactorings. I think there are a lot of oppotunities in this code to create classes—if we have something called X, and there are a bunch of functions of the form `foo(X, ...)`, `bar(X, ...)` that suggests to me that there should be a class X with those functions as methods.

I find `format.py` a little confusing. I'd expect it to have code related to converting things into strings or formatted output like tables. Maybe stuff that manipulates collections belongs elsewhere. I could also see a refactoring in which we pull out all the code that constructs tables (and all the calculations of headings and column widths) into a general-purpose table-building class here.

In `rpm.py` we do two things: manage downloading and do some processing of rpm files. I kind of think anything beyond simple manipulation of rpm files belongs in pkgmanager, and maybe rpm stuff belongs there as well.

In `term.py` some of the functions look like they belong in `system.py`, for example `get_executable_name`.

r0x0d:
> > This is a good start; it makes sense to reorganize code first and then do any refactorings. I think there are a lot of oppotunities in this code to create classes—if we have something called X, and there are a bunch of functions of the form `foo(X, ...)`, `bar(X, ...)` that suggests to me that there should be a class X with those functions as methods.
> 
> @jochapma Definitely agree! Small steps makes things easier to manage across many PRs as the overall changes per PR is a lot less
> 
> > I find `format.py` a little confusing. I'd expect it to have code related to converting things into strings or formatted output like tables. Maybe stuff that manipulates collections belongs elsewhere. I could also see a refactoring in which we pull out all the code that constructs tables (and all the calculations of headings and column widths) into a general-purpose table-building class here.
> 
> I'm not sure of a better naming scheme. If we go with `collections` goes into the same pitfall as `utils.py`. Any suggestions?
> 
> > In `rpm.py` we do two things: manage downloading and do some processing of rpm files. I kind of think anything beyond simple manipulation of rpm files belongs in pkgmanager, and maybe rpm stuff belongs there as well.
> 
> Sounds good, we can probably move them all there and delete `rpm.py`. WDYT @r0x0d?
> 
> > In `term.py` some of the functions look like they belong in `system.py`, for example `get_executable_name`.
> 
> I was thinking in the terms of terminal specific, like current terminal environment, vs the overall system information. I feel like `system.py` would be too big if we combine it. Do you have suggestions for what we could do here?

Hi, I will get back to reading this at some point after the merge freeze.

PR #1045, opened 2024-01-24
[RHELC-1375] Port RestorableFile to BackupController

r0x0d:
First part of the port for the remaining items for the BackupController.

Changes included in this work are:

1. Remove the old RestorableFile class in order to use the NewRestorableFile
2. Rename the NewRestorableFile back to RestorableFile
3. Fix a bug where test pollution was happening in the BackupController global object

Jira Issues: [RHELC-1375](https://issues.redhat.com/browse/RHELC-1375)

Checklist


convert2rhel/backup.py
```diff
@@ -367,12 +367,19 @@ def enable(self):
...
                 # IOError for py2 and OSError for py3
-                loggerinst.critical("Error(%s): %s" % (err.errno, err.strerror))
+
+                loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
+                raise exceptions.CriticalError(
+                    id_="FAILED_TO_SAVE_FILE_TO_BACKUP_DIR",
+                    title="Failed to copy file to the backup directory.",
+                    description="During convert2rhel's tests of whether it is likely to succeed, some files on the system are changed. To enable rollback in case we detect something that we would fail to do, we backup those files prior to changing them. In the current case, we encountered a failure while performing that backup so it is unsafe to continue.",
+                    diagnosis="Failed to backup %s. Errno: %s, Error: %s" % (self.filepath, err.errno, err.strerror),
+                )
```
r0x0d, line 377: @hosekadam since you worked on making the `NewRestorableFile` class, could you take a look at this and tell me if it is correct to be this way? 

I'm not sure if we should raise loggerinst.critical() here as it was before, or, if we should go with `CriticalError`.

r0x0d:
Have not tested it in a VM yet.

convert2rhel/backup.py
hosekadam: The way you changed it looks better and more systematic to me. So I would say it's correct.

I think @SpyTec commented this part here: https://github.com/oamg/convert2rhel/pull/875/files#r1403533541 (not it's linked to different code, but in previous I think it was related to this).

r0x0d:
In this PR, I'm trying to refactor the way we organize the backup code related to files and the other modules. 

The idea is that in the next PRs we take the code from the backup/__init__.py and split them into their own modules and keep organizing the backup there. 

If people think that the complexity increases, we can rollback to the original and move from there... But I strongly think that it will help to organize the code and give the idea that the backup is an API to be used across the rest of the project.

convert2rhel/backup/files.py
```diff
@@ -0,0 +1,146 @@
...
+    conversion so should be removed in rollback.
+    """
+
+    def __init__(self, filepath):
+        super(MissingFile, self).__init__()
+        self.filepath = filepath
+
+    def enable(self):
+        if self.enabled:
+            return
```
SpyTec, line 117: Lets get this tested too

convert2rhel/backup/files.py
```diff
@@ -0,0 +1,146 @@
...
+
+        if os.path.isfile(self.filepath):
+            loggerinst.debug(
+                "Shouldn't be called, file {filepath} is present before conversion".format(filepath=self.filepath)
+            )
```
SpyTec, line 122: What does this mean?

convert2rhel/backup/files.py
```diff
@@ -0,0 +1,146 @@
...
+            return
+
+        loggerinst.info("Marking file {filepath} as missing on system.".format(filepath=self.filepath))
+        super(MissingFile, self).enable()
+
+    def restore(self):
+        if not self.enabled:
+            return
+
+        loggerinst.task("Rollback: remove file created during conversion {filepath}".format(filepath=self.filepath))
```
SpyTec, line 132: Will this always be called during a rollback?

convert2rhel/backup/files.py
```diff
@@ -0,0 +1,146 @@
...
+            loggerinst.info("File {filepath} wasn't created during conversion".format(filepath=self.filepath))
+        else:
+            try:
+                os.remove(self.filepath)
+                loggerinst.info("File {filepath} removed".format(filepath=self.filepath))
+            except OSError as err:
+                # Do not call 'critical' which would halt the program. We are in
+                # a rollback phase now and we want to rollback as much as possible.
+                loggerinst.warning("Error(%s): %s" % (err.errno, err.strerror))
+                return
```
SpyTec, line 144: Same here, do we want to use `critical_no_exit()?`

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -21,8 +21,10 @@
 import pytest
 import six
 
-from convert2rhel import backup, logger, repo, unit_tests
+from convert2rhel import backup, repo, unit_tests
```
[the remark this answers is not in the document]
SpyTec: backup can be removed

SpyTec (changes requested):
Didn't look over the unit tests yet

convert2rhel/backup/files.py
hosekadam: This can be caused in a situation, when the `rpm -Va` says that the file is missing, but when the path is checked, the file is there. So it shouldn't be marked as missing on the system. I saw this time to time.

SpyTec: I think we can reword it somehow, to me anything that says "shouldn't be called" or "shouldn't happen" and still happens from time to time feels like a contradiction. Can we change it to be more descriptive?

convert2rhel/backup/files.py
```diff
@@ -0,0 +1,146 @@
...
+            try:
+                shutil.copy2(self.filepath, BACKUP_DIR)
+                loggerinst.debug("Copied %s to %s." % (self.filepath, BACKUP_DIR))
+            except (OSError, IOError) as err:
+                # IOError for py2 and OSError for py3
+                loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
+                raise exceptions.CriticalError(
+                    id_="FAILED_TO_SAVE_FILE_TO_BACKUP_DIR",
+                    title="Failed to copy file to the backup directory.",
+                    description="During convert2rhel's tests of whether it is likely to succeed, some files on the system are changed. To enable rollback in case we detect something that we would fail to do, we backup those files prior to changing them. In the current case, we encountered a failure while performing that backup so it is unsafe to continue.",
```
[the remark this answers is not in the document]
r0x0d: I will try changing the description

convert2rhel/backup/files.py
r0x0d: For now, I think it is.

convert2rhel/backup/files.py
jochapma: Maybe it would help if we made the choices `critical` and `critical_exit`? The one that exits is the one with the side effect (and a significant side effect at that), which argues that it should be the one with the more detailed name. It would also make the code more clear to someone just browsing it.

SpyTec: Agreed @jochapma, we made it this way for making it easier to migrate later. Easier to add `critical_no_exit()` and call it than to refactor the whole codebase. Technical debt in the making but something we had to do last year

SpyTec (approved):
LGTM

convert2rhel/backup/files.py
```diff
@@ -0,0 +1,153 @@
...
+                        "In the current case, we encountered a failure while performing that backup so it is unsafe "
+                        "to continue. See the diagnosis section to identify which problem ocurred during the backup."
+                    ),
+                    diagnosis="Failed to backup %s. Errno: %s, Error: %s" % (self.filepath, err.errno, err.strerror),
+                )
+        else:
+            loggerinst.info("Can't find %s.", self.filepath)
+
+        # Set the enabled value
+        super(RestorableFile, self).enable()
```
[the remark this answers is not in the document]
hosekadam: So we should set the enabled value to `True` only when the backup was successful?

convert2rhel/backup/files.py
```diff
@@ -0,0 +1,153 @@
...
+        if rollback:
+            loggerinst.info("File %s restored." % self.filepath)
+            super(RestorableFile, self).restore()
+        else:
+            loggerinst.debug("File %s restored." % self.filepath)
+            # not setting enabled to false since this is not being rollback
+            # restoring the backed up file for conversion purposes
+
+    # Probably will be deprecated and unusable since using the BackupController
+    # Depends on specific usage of this
```
hosekadam, line 100:
```suggestion

```
I think this comment can be removed since it's still used in the right way in `register_system()`. The fix was to [RHELC-16](https://issues.redhat.com/browse/RHELC-16).

hosekadam:
I like the way how you split the backup and organized it into the modules. I agree that it help with organizing it, having it in one file was a bit messy - especially after adding more restorable things.

convert2rhel/backup/files.py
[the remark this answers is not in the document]
r0x0d: updated

pr-watson (approved):

hosekadam (approved):

jochapma (approved):
Looks good to me.

r0x0d:
The test failures doesn't seem related to the changes in the PR. Merging this now to unblock us with the next tasks.

r0x0d:
Tests failure seems to be unrelated. Merging this righ now to unblock other work.

PR #1050, opened 2024-01-26
[RHELC-1361] Fix file clash in NewRestorableFile

r0x0d:
There was a clash happening in the way we backup files, as if there is a file to backup that has the same name as one directory or another file already created, an error will be thrown to the user.

This commit fixes that clash by generating an hash directory based on the dirname of the absolute path.

Jira Issues: [RHELC-1361](https://issues.redhat.com/browse/RHELC-1361)

Checklist

### DIFF
--- a/convert2rhel/unit_tests/backup/backup_test.py
+++ b/convert2rhel/unit_tests/backup/backup_test.py
@@ -1,5 +1,6 @@
 __metaclass__ = type
 
+import hashlib
 import os
 
 import pytest
--- a/convert2rhel/backup/files.py
+++ b/convert2rhel/backup/files.py
@@ -17,6 +17,7 @@
 
 __metaclass__ = type
 
+import hashlib
 import logging
 import os
 import shutil
@@ -32,7 +33,21 @@ loggerinst = logging.getLogger(__name__)
 class RestorableFile(RestorableChange):
     def __init__(self, filepath):
         super(RestorableFile, self).__init__()
+
+        # The filepath we want to back up needs to start with at least a `/`,
+        # otherwise, let's error out and warn the developer/user that the
+        # filepath is not what we expect. This is mostly intended to be an
+        # error to catch during development, not runtime.
+        if not os.path.isabs(filepath):
+            raise TypeError("Filepath needs to be an absolute path.")
+
+        # We don't support directory globs in here *yet*, so let's prevent to
+        # pass a directory here as well.
+        if os.path.isdir(filepath):
+            raise TypeError("Path must be a file not a directory.")
+
         self.filepath = filepath
+        self._backup_path = None
 
     def enable(self):
         """Save current version of a file"""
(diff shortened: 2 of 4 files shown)


SpyTec (approved):
LGTM! @jochapma please look over the MD5 hashdir solution if you can!

SpyTec:
Integration test failures due to internet connectivity - unrelated to PR

convert2rhel/backup.py
```diff
@@ -351,7 +352,21 @@ def restore(self):
 class NewRestorableFile(RestorableChange):
     def __init__(self, filepath):
         super(NewRestorableFile, self).__init__()
+
+        # The filepath we want to back up needs to start with at least a `/`,
+        # otherwise, let's error out and warn the developer/user that the
+        # filepath is not what we expect. This is mostly intended to be an
+        # error to catch during development, not runtime.
+        if not filepath.startswith("/"):
```
jochapma, line 360:
```suggestion
        if not os.path.isabs(filepath):
```

convert2rhel/backup.py
```diff
@@ -351,7 +352,21 @@ def restore(self):
 class NewRestorableFile(RestorableChange):
     def __init__(self, filepath):
         super(NewRestorableFile, self).__init__()
+
```
jochapma, line 355: Maybe we should normalize the path? I'm not entirely convinced we should bother, but if we're trying to catch mistakes during development, it would add some bulletproofing.

convert2rhel/backup.py
```diff
@@ -374,6 +390,21 @@ def enable(self):
         # Set the enabled value
         super(NewRestorableFile, self).enable()
 
+    def _normalize_backup_path(self):
```
jochapma, line 393: I find the use of 'normalize' here confusing; in the Python world, 'normalize' refers to what `os.path.normpath` does.

jochapma:
As a general comment, I think the use of MD5 to generate a backup name in this fashion is sound.

convert2rhel/backup.py
r0x0d: +1 to that, thanks!

convert2rhel/backup.py
r0x0d: Agreed. Now reading it over again, it really seems a bit confusing.

convert2rhel/backup.py
r0x0d: I think it should be fine without it for now. The checking for an absolute path should be enough for now. Maybe if we get some issue from a user, we can investigate it more.

convert2rhel/backup.py
```diff
@@ -374,6 +390,21 @@ def enable(self):
         # Set the enabled value
         super(NewRestorableFile, self).enable()
 
+    def _hash_backup_path(self):
```
hosekadam, line 393: Maybe I would change a bit the name of the method or add to comment some example to help with understanding. Right now it feels to me, it will return the hash of the backup path. I would expect it will return the `hashed_directory = os.path.join(BACKUP_DIR, hashlib.md5(path.encode()).hexdigest())`

jochapma (approved):
Looks good.

convert2rhel/backup.py
r0x0d: I like the idea of adding an example of the usage. I will update it. 

Regarding the name, I will try to come up with something more explicit.

convert2rhel/unit_tests/backup/backup_test.py
```diff
@@ -1,5 +1,6 @@
 __metaclass__ = type
 
+import hashlib
```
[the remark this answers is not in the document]
SpyTec: Can be removed

SpyTec (approved):
Can be merged

convert2rhel/unit_tests/backup/backup_test.py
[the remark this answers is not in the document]
r0x0d: I will take care of this as part of the vulture PR.

PR #1058, opened 2024-01-30
[RHELC-1425] breaking change: Remove deprecated CLI arguments

SpyTec:
This removes a few deprecated CLI arguments, they will no longer work
or give a warning about its deprecation.

This includes
- Removal of `convert2rhel -f|--password-from-file`
- Removal of unused `convert2rhel --keep-rhsm`

Jira Issues: [RHELC-1425](https://issues.redhat.com/browse/RHELC-1425)

Checklist

### DIFF
--- a/convert2rhel/toolopts.py
+++ b/convert2rhel/toolopts.py
@@ -47,8 +47,6 @@ ARGS_WITH_VALUES = [
     "--username",
     "-p",
     "--password",
-    "-f",
-    "--password-from-file",
     "-k",
     "--activationkey",
     "-o",
@@ -69,7 +67,6 @@ class ToolOpts:
     def __init__(self):
         self.debug = False
         self.username = None
-        self.password_file = None
         self.config_file = None
         self.password = None
         self.no_rhsm = False
@@ -275,14 +272,6 @@ class CLI:
             " through a list of running processes.",
         )
         group.add_argument(
-            "-f",
-            "--password-from-file",
-            help="File containing"
-            " password for the subscription-manager in the plain"
-            " text form. It's an alternative to the --password"
-            " option. Deprecated, use --config-file instead.",
-        )
-        group.add_argument(
             "-k",
             "--activationkey",
             help="Activation key used"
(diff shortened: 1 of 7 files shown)


convert2rhel/toolopts.py
```diff
@@ -549,12 +514,6 @@ def _process_cli_options(self):
                 " configuration file. We're going to use the command line values."
             )
 
-        if (config_opts.activation_key or config_opts.password) and parsed_opts.password_from_file:
```
r0x0d, line 552: Instead of removing this completely, should we change to check to use the config file instead of the password_from_file one?

SpyTec: Don't we do this already? This was just for the warning when both were set

convert2rhel/toolopts.py
```diff
@@ -381,7 +365,7 @@ def _process_cli_options(self):
         tool_opts.config_file = parsed_opts.config_file
         # corner case: password on CLI or in password file and activation-key in the config file
         # password from CLI has precedence and activation-key must be deleted (unused)
```
hosekadam, line 367:
```suggestion
        # corner case: password on CLI  and activation-key in the config file
        # password from CLI has precedence and activation-key must be deleted (unused)
```

SpyTec: Thanks!

convert2rhel/toolopts.py
hosekadam: I agree that this was just the warning, setting the values was done in a [different place](https://github.com/oamg/convert2rhel/pull/1058/files#diff-a7331917babf0a1504893963baab7ad842640877734e5dbeb39ebf97ec584565L411).

convert2rhel/toolopts.py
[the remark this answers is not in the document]
SpyTec: Good spot, fixed!

scripts/manpage_generation.sh
hosekadam, line 1: I'm just thinking if those changes are related to this PR or those are some forgotten changes from something different.

SpyTec: Ah you're right! While generating locally I had to change it. Will fix

pr-watson (approved):
Looks good to me!

PR #1081, opened 2024-02-06
[RHELC-1269] Add diagnosis for failures in validating yum transaction

pr-watson:
This PR adds a diagnosis field for when there is a failure during the validation of a yum transaction. The error raised is added to the diagnosis field which can sometimes come out as a list which is why we have a `.join` added. The diagnosis field for the corresponding dnf validation transaction has also been updated to join the error emitted so the message to the user looks cleaner.

Jira Issues: [RHELC-1269](https://issues.redhat.com/browse/RHELC-1269)

Checklist


convert2rhel/pkgmanager/handlers/dnf/__init__.py
```diff
@@ -209,7 +209,7 @@ def _process_transaction(self, validate_transaction):
                 id_="FAILED_TO_VALIDATE_TRANSACTION",
                 title="Failed to validate dnf transaction.",
                 description="During the dnf transaction execution an error occured and convert2rhel could no longer process the transaction.",
-                diagnosis="Transaction processing failed with error %s." % (str(e)),
+                diagnosis="Transaction processing failed with error %s." % ("".join(e)),
```
r0x0d, line 212: So, we actually catch two type of exceptions here: 

```
pkgmanager.exceptions.Error,
pkgmanager.exceptions.TransactionCheckError,
```

The `pkgmanager.exceptions.TransactionCheckError` one will give you a list of packages, so your solution does apply nicely here (although, I think it could be good to separate the packages with a comma: 

```suggestion
                diagnosis="Transaction processing failed with error %s." % (",".join(e)),
```

However, I'm not 100% sure if `pkgmanager.exceptions.Error` does the same. Maybe it will only return a string to you. Need to check.

pr-watson: @r0x0d I decided to go with a space here because of this example output from the exception - `[u'Errors were encountered while downloading packages.', u'pam-1.1.8-23.el7.x86_64: [Errno 256] No more mirrors to try.']`. So that way after the join the output will look like `Transaction processing failed with error Errors were encountered while downloading packages. u'pam-1.1.8-23.el7.x86_64: [Errno 256] No more mirrors to try.

SpyTec: This isn't joining with space though, it's joining with empty string right?

SpyTec (approved):

PR #1090, opened 2024-02-08
[RHELC-1377] Port the repo backup to restorablefile

hosekadam:
Port backup of repofiles and yum/dnf var dirs to use the RestorableFile working with BackupController. 

Jira Issues: [RHELC-1377](https://issues.redhat.com/browse/RHELC-1377)

Checklist


convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -77,8 +81,51 @@ def run(self):
 
         super(BackupRepository, self).run()
 
-        repo.backup_yum_repos()
-        repo.backup_varsdir()
+        self.backup_yum_repos()
+        self.backup_varsdir()
```
r0x0d, line 85: Maybe we could create a new action class just for this guy here... It is not really a repository, but a variable that is used by the repositories.

hosekadam: Okay, makes sense! I'll change that

convert2rhel/pkghandler.py
```diff
@@ -849,6 +851,11 @@ def remove_pkgs_unless_from_redhat(pkgs_to_remove, backup=True):
     loggerinst.warning("Removing the following %s packages:\n" % str(len(pkgs_to_remove)))
     print_pkg_info(pkgs_to_remove)
     loggerinst.info("\n")
+
+    # Need to generate backed up path of repo files since hash is used to avoid
+    # conflicts during backup
+    backedup_reposdir = os.path.join(utils.BACKUP_DIR, hashlib.md5(DEFAULT_YUM_REPOFILE_DIR.encode()).hexdigest())
```
r0x0d, line 857: Not sure if I followed on why this needs to be here

r0x0d: This function is used to remove packages from other sources as well, but I don't see it being used during the backup of the variables/yum repos. 

Ideally, who should take care of that hash is the RestorableFile class. Also, this function got moved away in my last PR: https://github.com/oamg/convert2rhel/pull/1091

hosekadam: @r0x0d We need to have access to the backed up repofiles when we are trying to backup centos-release file (with this file I had issues). Before porting to the resotrablefile, just backup path was used and there were available the repofiles. But now, when we copy the .repo files to the directory named as the hash of the original path, the repofiles are inaccessible. 

So I've chosen this strategy of building the path here since it's probably just this one case.

r0x0d: Ah, I see. So we might receive this from the parameter in the function rather than hardcoding it in the function body. 

We can make this function being more generic if we receive the reposdir, and whoever needs this information, we can pass this down to the function.

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -29,6 +29,9 @@
 from convert2rhel.toolopts import PRE_RPM_VA_LOG_FILENAME
 from convert2rhel.unit_tests import CriticalErrorCallableObject
 
+# TODO test if needed
```
SpyTec, line 32:
```suggestion
```

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -77,8 +80,57 @@ def run(self):
...
+        self.backup_yum_repos()
+
+    def backup_yum_repos(self):
+        """Backup .repo files in /etc/yum.repos.d/ so the repositories can be restored on rollback."""
+        loggerinst.info("Backing up .repo files from %s." % DEFAULT_YUM_REPOFILE_DIR)
+
+        repo_files_backed_up = False
+
+        for repo in os.listdir(DEFAULT_YUM_REPOFILE_DIR):
+            if repo.endswith(".repo") and repo != "redhat.repo":
```
SpyTec, line 92: As far as I can tell we don't test when there's a redhat.repo right?

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -77,8 +80,57 @@ def run(self):
...
+                restorable_file = RestorableFile(repo_path)
+                backup.backup_control.push(restorable_file)
+                repo_files_backed_up = True
+
+        if not repo_files_backed_up:
+            loggerinst.info("No .repo files backed up.")
+
+
+class BackupVariables(actions.Action):
+    id = "BACKUP_VARIABLES"
```
r0x0d, line 103: I would change this name a bit, maybe `BackupPackageManagerVariables` (?)

Just to make it more explicit to tell what we are backing up, variables could be anything, anywhere.

r0x0d: Could also be `BackupYumVariables` as well... I don't think it does matter in that context if we say yum/dnf variables, as they are treat as one.

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -77,8 +80,57 @@ def run(self):
...
+        self._backup_variables(path=DEFAULT_YUM_VARS_DIR)
+
+        if system_info.version.major >= 8:
+            loggerinst.info("Backing up variables files from %s." % DEFAULT_DNF_VARS_DIR)
+            self._backup_variables(path=DEFAULT_DNF_VARS_DIR)
+
+    def _backup_variables(self, path):
+        """Helper internal function to backup the variables.
+
+        :parameter path: The path for the original variable.
```
r0x0d, line 121:
```suggestion
        :param path: The path for the original variable.
```

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -77,8 +80,57 @@ def run(self):
...
+        variable_files_backed_up = False
+
+        for variable in os.listdir(path):
+            variable_path = os.path.join(path, variable)
+            restorable_file = RestorableFile(variable_path)
+            backup.backup_control.push(restorable_file)
+            variable_files_backed_up = True
+
+        if not variable_files_backed_up:
+            loggerinst.info("No variables files backed up.")
```
r0x0d, line 133: Not a necessary change, but we could move this to be outside the class scope. Since this is not using the `self` reference, we tend to imagine those functions as being static (i.e: not needing any internal reference to be accessed/used, so, this could be a private function that is tied to the module, rather than the class itself. 

Of course, it has the single usage of backing up variables, but if you want to make the class less polluted without things that are not really necessary there, you could move this function to be outside the class scope.

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -65,17 +96,120 @@ def test_backup_redhat_release_calls(self, backup_redhat_release_action, monkeyp
...
-        backup_varsdir_mock = mock.Mock()
-        backup_yum_repos_mock = mock.Mock()
+    @all_systems
+    def test_backup_repository_nonexisting_paths(
+        self, tmpdir, monkeypatch, caplog, backup_repository_action, pretend_os
+    ):
+        """Test empty paths, nothing for backup."""
+        etc = tmpdir.mkdir("etc")
+
+        monkeypatch.setattr(backup_system, "DEFAULT_YUM_REPOFILE_DIR", str(etc))
```
r0x0d, line 106: We could simplify this to: 

```suggestion
        etc = tmpdir.mkdir("etc")

        monkeypatch.setattr(backup_system, "DEFAULT_YUM_REPOFILE_DIR", "/non-existing-path")
```

That would give the same effect, and much simpler as well.

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -65,17 +96,120 @@ def test_backup_redhat_release_calls(self, backup_redhat_release_action, monkeyp
...
+        # Remove the original files
+        os.remove(yum_repo)
+        assert not os.path.isfile(yum_repo)
+
+        global_backup_control.pop_all()
+
+        # Check presence of restored files
+        assert os.path.isfile(yum_repo)
+        with open(yum_repo, mode="r") as f:
+            assert f.read() == os.path.basename(yum_repo)
```
r0x0d, line 139: Let's split this to it's own class for better readability of the tests. 

```python
class TestBackupRepository:
  ....
```

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -65,17 +96,120 @@ def test_backup_redhat_release_calls(self, backup_redhat_release_action, monkeyp
...
+
+        global_backup_control.pop_all()
 
-        backup_yum_repos_mock.assert_called_once()
-        backup_varsdir_mock.assert_called_once()
+        # Check presence of restored files
+        for path in orig_path_list:
+            assert os.path.isfile(path)
+            with open(path, mode="r") as f:
+                assert f.read() == os.path.basename(path)
```
r0x0d, line 212: Same for this one: 

```python
class TestBackupVariables:
... 
```

r0x0d (commented):
Few comments, but otherwise, looks good. 

Regarding @hosekadam comment about depending on the #1091 PR, we might get this one merged first, and deal with the problems there too. 

We can apply your gist there.

convert2rhel/actions/pre_ponr_changes/backup_system.py
hosekadam: We are not, I'll add the unit test for this case

convert2rhel/actions/pre_ponr_changes/backup_system.py
hosekadam: @r0x0d If I can choose, I would rather have it there, as it's related only to this class and I have better feeling of having it together.

But your idea also makes a sense to me and I spend a bit of time thinking what I like more :D The result of having cleaner classes and having the related function in the module as private functions is attractive.

r0x0d: Sounds good to me

hosekadam:
I applied the suggestions. Fixed the `remove_pkgs_unless_from_redhat` to work with the new backup path, but in the different way as @r0x0d suggested earlier (also fixed the unit tests). Tested it in vagrant (fully working), the unit tests are fine now too.

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -80,11 +83,14 @@ def run(self):
         logger.info("Searching for the following excluded packages:\n")
 
         pkgs_removed = []
+        # Since the MD5 checksum of original path is used in backup path to avoid
+        # conflicts in backup folder, preparing the path is needed.
+        backedup_reposdir = os.path.join(BACKUP_DIR, hashlib.md5(DEFAULT_YUM_REPOFILE_DIR.encode()).hexdigest())
```
r0x0d, line 88: @hosekadam, would you mind applying the same fix for https://github.com/r0x0d/convert2rhel/blob/host-metering/convert2rhel/pkgmanager/handlers/yum/__init__.py#L78? 

I think this is the only place left out that needs the same workaround

hosekadam: @r0x0d Definitely, I missed that one.

r0x0d: Just pushed a commit that fixes that: 20437fc7a59757e0c9dcba926a73958ae57023c5. 

Not the best commit, but should do the trick.

hosekadam:
@r0x0d Thanks for fixing the yum backup part! So now this should be ready for merge.

convert2rhel/pkghandler.py
```diff
@@ -599,7 +599,7 @@
         loggerinst.info("All packages are now signed by Red Hat.")
 
 
-def remove_pkgs_unless_from_redhat(pkgs_to_remove, backup=True):
+def remove_pkgs_unless_from_redhat(pkgs_to_remove, backedup_reposdir, backup=True):
```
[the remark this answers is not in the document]
r0x0d: This was taken care in https://github.com/oamg/convert2rhel/pull/1091

hosekadam:
Rebased

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -343,3 +358,136 @@ def test_backup_package_file_complete(
                     assert f.read() == "Content for testing of file %s" % original_file_path
             elif status == "missing":
                 assert not os.path.isfile(original_file_path)
+
+
+class TestBackupRepository:
+    @all_systems
```
r0x0d, line 364: Not sure if I got why we need to use @all_systems here

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -343,3 +358,136 @@ def test_backup_package_file_complete(
...
+        global_backup_control.pop_all()
+
+        # Check presence of restored files
+        assert os.path.isfile(yum_repo)
+        with open(yum_repo, mode="r") as f:
+            assert f.read() == os.path.basename(yum_repo)
+
+    @all_systems
+    def test_backup_repository_redhat(
+        self, monkeypatch, tmpdir, backup_repository_action, global_backup_control, pretend_os, caplog
```
r0x0d, line 393:
```suggestion
        self, monkeypatch, tmpdir, backup_repository_action, pretend_os, caplog
```

r0x0d (approved):
Looks good to me.

r0x0d (approved):
Looks good to me. Only one comment for the unit_tests to see if we can remove an potentially unnecessary fixture.

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
hosekadam: It seems as a leftover from situation when backing up varsdir and repos was one action. It works without them

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
hosekadam: Yes, in this test is nothing backed up, so this is not necessary/ I'm removing it. Thanks

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
r0x0d: Cool. Let's remove all instances form that then

pr-watson (approved):
Looks good to me

danmyway:
Test fails unrelated, will be fixed by #1107

PR #1091, opened 2024-02-08
[RHELC-1382] Port RestorablePackage and ChangedRPMPackagesController

r0x0d:
Merging the two classes together and working only with RestorablePackage.

In this commit, we are introducing the refactor, as well as some other minor refactors that were required to avoid circular dependency imports.

Jira Issues: [RHELC-1382](https://issues.redhat.com/browse/RHELC-1382)

Checklist


r0x0d:
# TODO

- [ ] There is a bug with the remove excluded packages list and remove packages with repo files actions.  The bug consist mainly in rollback order.

pr-watson (approved):

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -62,143 +63,121 @@ def extract_packages(self, pkg):
...
-        installing subscription-manager to prevent package conflicts, these
-        non-RHEL packages must be present on the system during
-        subscription-manager installation so that the system can access and
-        install subscription-manager dependencies. As a result, these non-RHEL
-        packages must be manually removed after subscription-manager
-        installation.
-        """
-        super(RemoveRepositoryFilesPackages, self).run()
+def _remove_packages_unless_from_redhat(pkgs_list):
+    """Remove packages from the system that are not RHEl.
```
hosekadam, line 167:
```suggestion
    """Remove packages from the system that are not RHEL.
```

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -62,143 +63,121 @@ def extract_packages(self, pkg):
...
-        excluded packages. While other excluded packages must be removed before
-        installing subscription-manager to prevent package conflicts, these
-        non-RHEL packages must be present on the system during
-        subscription-manager installation so that the system can access and
-        install subscription-manager dependencies. As a result, these non-RHEL
-        packages must be manually removed after subscription-manager
-        installation.
-        """
-        super(RemoveRepositoryFilesPackages, self).run()
+def _remove_packages_unless_from_redhat(pkgs_list):
```
hosekadam, line 166: Why don't have this as method of `RemoveSpecialPackages`? I see it's called only from the `run` and anywhere else

convert2rhel/backup/__init__.py
```diff
@@ -20,113 +20,16 @@
...
-from convert2rhel.utils import BACKUP_DIR, download_pkg, remove_orphan_folders, run_subprocess
+from convert2rhel.utils import TMP_DIR
 
 
-loggerinst = logging.getLogger(__name__)
-
-# Note: Currently the only use case for this is package removals
-class ChangedRPMPackagesController:
-    """Keep control of installed/removed RPM pkgs for backup/restore."""
+#: Directory for temporary backing up files, packages and other relevant stuff.
```
hosekadam, line 29:
```suggestion
# Directory for temporary backing up files, packages and other relevant stuff.
```

convert2rhel/backup/packages.py
```diff
@@ -73,6 +74,140 @@
...
+        self.custom_releasever = custom_releasever
+        self.varsdir = varsdir
+
+        self._backedup_paths = []
+
+    def enable(self):
+        """Save version of RPM package.
+
+        :param reposdir: Custom repositories directory to be used in the backup.
+        :type reposdir: str
```
hosekadam, line 96: I think this is some leftover, maybe it should be placed at the `__init__`?

convert2rhel/backup/packages.py
```diff
@@ -73,6 +74,140 @@
...
+    def __init__(self, pkgs, reposdir=None, set_releasever=False, custom_releasever=None, varsdir=None):
+        super(RestorablePackage, self).__init__()
+
+        self.pkgs = pkgs
+        self.reposdir = reposdir
+        self.set_releasever = set_releasever
+        self.custom_releasever = custom_releasever
+        self.varsdir = varsdir
+
+        self._backedup_paths = []
```
hosekadam, line 90:
```suggestion
        self._backedup_pkgs_paths = []
```
For better readability, I would go with something like this

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -1235,12 +1153,12 @@ def test_install_gpg_keys_fail_create_restorable(self, monkeypatch, tmpdir, glob
     ),
 )
 @centos7
-def test_get_system_packages_for_replacement(pretend_os, pkgs, expected, monkeypatch):
+def test_get_system_packages_for_replacement(pretend_os, pkgs, expected, monkeypatch, caplog):
```
hosekadam, line 1156:
```suggestion
def test_get_system_packages_for_replacement(pretend_os, pkgs, expected, monkeypatch):
```

convert2rhel/unit_tests/pkgmanager/handlers/yum/yum_test.py
```diff
@@ -383,19 +383,21 @@ def test_resolve_yum_problematic_dependencies(
...
         assert pkgmanager.handlers.yum.remove_pkgs.called
-        pkgmanager.handlers.yum.remove_pkgs.assert_called_with(
-            pkgs_to_remove=expected_remove_pkgs,
-            backup=True,
-            critical=True,
-            reposdir=utils.BACKUP_DIR,
+        assert pkgmanager.handlers.yum.RestorablePackage.called
+        pkgmanager.handlers.yum.RestorablePackage.assert_called_with(
+            pkgs=expected_remove_pkgs,
+            reposdir=backup.BACKUP_DIR,
```
hosekadam, line 396: Maybe I would mock the backup dir to use tmpdir. It can help with avoiding troubles in unit tests

convert2rhel/actions/pre_ponr_changes/handle_packages.py
r0x0d: So, I decided to split this method because it is not using any internal reference of the class. To not pollute the class with functions that does not need any internal reference, I decided to put this outside the class scope. 

It is true that we could move this to the class as it belongs only to there, but still, if a function doesn't need any internal references, meaning that in this sense we think of them as a `static function`, I prefer to move them to be in the module level rather than in the class level.

convert2rhel/backup/packages.py
r0x0d: Yeah, thanks for catching that!

convert2rhel/unit_tests/pkgmanager/handlers/yum/yum_test.py
r0x0d: We are mocking the entire class here, it should be fine without mocking the backup dir

jochapma (approved):
Looks good to me.

convert2rhel/main.py
```diff
@@ -115,6 +115,8 @@ def main_locked():
         process_phase = ConversionPhase.PRE_PONR_CHANGES
         pre_conversion_results = actions.run_actions()
 
+        loggerinst.info("1 - before rollback: %s" % len(backup.backup_control._restorables))
```
hosekadam, line 118: Looks like forgotten debug print, also on line 157 and 333

r0x0d: Thanks!

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -345,23 +331,58 @@ def _process_transaction(self, validate_transaction):
...
 
-    @utils.run_as_child_process
     def run_transaction(self, validate_transaction=False):
         """Run the yum transaction.
 
         Perform the transaction. If the `validate_transaction` parameter set to
         true, it means the transaction will not be executed, but rather verify
         everything and do an early return.
 
+        :param vaidate_transaction: Determines if the transaction needs to be
```
hosekadam, line 341:
```suggestion
        :param validate_transaction: Determines if the transaction needs to be
```

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -345,23 +331,58 @@ def _process_transaction(self, validate_transaction):
...
+            validated or not.
+        :type validate_transaction: bool
+        :raises CriticalError: If we can't resolve the transaction dependencies.
+        """
+        resolve_deps_finished = False
+        # Do not allow this to loop until eternity.
+        attempts = 0
+        try:
+            while attempts <= MAX_NUM_OF_ATTEMPTS_TO_RESOLVE_DEPS:
+                self._set_up_base()
```
hosekadam, line 351: Why is the setup moved here from the `perform_operations`?

r0x0d: Answered in slack.

convert2rhel/actions/pre_ponr_changes/transaction.py
```diff
@@ -27,7 +27,6 @@ class ValidatePackageManagerTransaction(actions.Action):
     id = "VALIDATE_PACKAGE_MANAGER_TRANSACTION"
     dependencies = (
         "INSTALL_RED_HAT_GPG_KEY",
-        "REMOVE_EXCLUDED_PACKAGES",
```
hosekadam, line 30: Maybe I would add some comment, that there is an implicit dependency of removing the special packages (as in `pre_ponr_changes/subscription.py`). Or needs to be it removed?

convert2rhel/backup/packages.py
```diff
@@ -73,6 +74,145 @@
 )
 _UBI_9_REPO_PATH = os.path.join(_RHSM_TMP_DIR, "ubi_9.repo")
 
+# NOTE: Over time we want to replace this with pkghandler.RestorablePackageSet.
+class RestorablePackage(RestorableChange):
+    def __init__(self, pkgs, reposdir=None, set_releasever=False, custom_releasever=None, varsdir=None):
+        """Keep control of installed/removed RPM pkgs for backup/restore.
```
hosekadam, line 80: Is this comment right? This is used mainly when we are removing some packages from the system and during rollback we need to restore them.

From the comment I have a feeling I can use it also for installing some package during conversion and during rollback it will be removed.

I would go with "Keep control of removed RPM pkgs to be installed during rollback" if my understanding is right.

convert2rhel/backup/packages.py
```diff
@@ -73,6 +74,145 @@
...
+            # If we ever put Oracle Linux repofiles to ship with convert2rhel,
+            # them the second part of this condition can be dropped.
+            if system_info.eus_system and system_info.id == "centos":
+                self.reposdir = get_hardcoded_repofiles_dir()
+
+            # One of the reasons we hardcode repofiles pointing to archived
+            # repositories of older system minor versions is that we need to be
+            # able to download an older package version as a backup. Because for
+            # example the default repofiles on CentOS Linux 8.4 point only to
+            # 8.latest repositories that already don't contain 8.4 packages.
```
hosekadam, line 123: I would move this comment a bit upper, above the `if system_info.eus_system and system_info.id == "centos":` condition. I think it's related to that one.

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -237,7 +238,7 @@ def _perform_operations(self):
                 diagnosis="Repository mirrors failed with error %s." % (str(e)),
             )
 
-    def _resolve_dependencies(self, validate_transaction):
+    def _resolve_dependencies(self):
```
hosekadam, line 241: After removing the `validate_transaction` parameter there is a leftover in the comment section

convert2rhel/unit_tests/__init__.py
```diff
@@ -311,25 +314,31 @@ class SummaryAsJsonMocked(MockFunctionObject):
 #
 
 
-class RemovePkgsMocked(MockFunctionObject):
+class RestorablePackageMocked(MockFunctionObject):
     """
     Mock for the remove_pkgs function.
```
hosekadam, line 319:
```suggestion
    Mock for the RestorablePackage class.
```

convert2rhel/backup/packages.py
r0x0d: modified the docstrings

convert2rhel/backup/packages.py
r0x0d: Moved the comments to the docstring, think it keep more organized.

convert2rhel/pkgmanager/handlers/yum/__init__.py
r0x0d: applied

convert2rhel/unit_tests/__init__.py
r0x0d: updated

convert2rhel/unit_tests/backup/packages_test.py
```diff
@@ -66,6 +67,194 @@ def __call__(self, pkgs, dest, *args, **kwargs):
...
+        monkeypatch.setattr(utils, "download_pkg", DownloadPkgMocked())
+        monkeypatch.setattr(packages, "system_info", global_system_info)
+
+        packages.system_info.eus_system = True
+        packages.system_info.id = "centos"
+
+        rp = RestorablePackage(pkgs=["test.rpm"])
+        rp._backedup_pkgs_paths = ["test.rpm"]
+        rp.enable()
+
```
hosekadam, line 119: I would add here check if the `get_hardcoded_repofiles_dir()` was called

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -360,11 +360,11 @@ def test_restore_subman_uninstalled(self, caplog, monkeypatch, system_subscripti
 
 def test_install_rhel_subsription_manager(monkeypatch, global_backup_control):
     mock_backup_control = mock.Mock()
-    monkeypatch.setattr(global_backup_control, "push", mock_backup_control)
+    monkeypatch.setattr(subscription.backup.backup_control, "push", mock_backup_control)
 
     subscription.install_rhel_subscription_manager(["subscription-manager", "json-c.x86_64"])
 
-    assert mock_backup_control.push.called_once
+    assert mock_backup_control.call_count == 1
```
hosekadam, line 367: I see you changed this many times. Are there any benefits of using `call_count == 1` instead of `called_once`?

convert2rhel/unit_tests/subscription_test.py
```diff
@@ -992,7 +992,7 @@ def test_verify_rhsm_installed_failure(self, monkeypatch, caplog):
         assert "The subscription-manager package is not installed correctly." in caplog.text
 
 
-# ----
+# # ----
```
hosekadam, line 995:
```suggestion
```
Or does it have any meaning there?

r0x0d: No meaning.

convert2rhel/unit_tests/subscription_test.py
r0x0d: I think it is more clear that we are specifically saying we want to assert how many times it was called. It could be converted to called_once, but I think checking count is better to readability.

hosekadam: Ok, thanks for explanation. I agree

hosekadam (approved):

SpyTec (approved):

PR #1095, opened 2024-02-12
[RHELC-1393] Move logger initialization to initialize module

r0x0d:
Initializing the logger module earlier makes the tool work correctly when trying to find the /etc/system-release file, and if it is not there, throw a critical error to the user before anything else happens.

If we don't move the module to this point, the first /etc/system-releaes check will silently fail and only the next call will throw the exception, because at this point, we already have the logger initialized.

To prevent further executions in the system, it is more reliable to initialize the logger module earlier.

Jira Issues: [RHELC-1393](https://issues.redhat.com/browse/RHELC-1393)

Checklist

### DIFF
--- a/convert2rhel/initialize.py
+++ b/convert2rhel/initialize.py
@@ -21,6 +21,10 @@ import logging
 import os
 
 from convert2rhel import i18n
+from convert2rhel import logger as logger_module
+
+
+loggerinst = logging.getLogger(__name__)
 
 
 def disable_root_logger():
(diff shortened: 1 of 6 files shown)


convert2rhel/initialize.py
```diff
@@ -77,6 +115,13 @@ def run():
...
+    # initialize logging
+    initialize_logger()
+
+    # since we now have root, we can add the FileLogging
+    # and also archive previous logs
+    initialize_file_logging("convert2rhel.log", logger_module.LOG_DIR)
+
     from convert2rhel import main
 
     return main.main()
```
SpyTec, line 127: Given this is called before `main()` what is the safeguard in case we are not running as root?

r0x0d: Haven't tested that part... I'm not too involved in the root vs no root thread we had in the past regarding logging and etc... 

Do you think there is a way for us to be on the safe side with this, moving the logging to initialize?

SpyTec: Without splitting up main a bit more or something.. not really. We need to create the file handler where we currently do it to ensure that we have root and that we're not running another instance. We can initialize everything else though

SpyTec: That is to say, we can revert this to how it was. Just init the logger here
```suggestion

    from convert2rhel import main

    return main.main()
```

r0x0d: Ah, alright! Will apply this in a second. Thanks!

r0x0d: applied it manually because of unit_tests.

pr-watson (approved):
Changes look good to me!

pr-watson:
Only 1 test failing unrelated to code, merging

PR #1122, opened 2024-02-29
[RHELC-1432] Copy instead of move when restoring a file

r0x0d:
Moving a file back to its own directory is causing the file to lose the metadata and not letting the user enter in their system with SSH after the analysis is done.

Jira Issues: [RHELC-1432](https://issues.redhat.com/browse/RHELC-1432)

Checklist

### DIFF
--- a/convert2rhel/backup/files.py
+++ b/convert2rhel/backup/files.py
@@ -120,7 +120,8 @@ class RestorableFile(RestorableChange):
             return
 
         try:
-            shutil.move(self._backup_path, self.filepath)
+            shutil.copy2(self._backup_path, self.filepath)
+            os.remove(self._backup_path)
         except (OSError, IOError) as err:
             # Do not call 'critical' which would halt the program. We are in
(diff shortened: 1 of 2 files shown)


convert2rhel/backup/files.py
```diff
@@ -120,7 +120,8 @@ def restore(self, rollback=True):
             return
 
         try:
-            shutil.move(self._backup_path, self.filepath)
+            shutil.copy2(self._backup_path, self.filepath)
+            os.remove(self._backup_path)
```
pholica, line 124: If there's an exception raised by the `shutil.copy2` (for example the target directory doesn't exist), the `os.remove` is not called.

SpyTec: Yes, but that's how we should have it. If something didn't go well with the copy we don't want to remove the backup

hosekadam (approved):
Since the behavior of move is copy and then remove the original file, the change does in fact the same.

PR #1127, opened 2024-03-04
Improve UX for release Jira auto-linking

SpyTec:
Currently we do not add the Jira links correctly as was observed for the
v1.7.0 release as one PR title contained two RHELC Jira issues in the
title. E.g. `[RHELC-1, RHELC-2] Test title`. This change should now
be more flexible and also handles more edge-cases that could occur.

We now use python to do RegEx replace rather than sed or awk as they
were too limiting. Opted for not using perl as python shell is already
present for GitHub actions.

This is how we want it to be:
https://github.com/oamg/convert2rhel/releases/tag/v1.7.0

Tested on my repo
https://github.com/SpyTec/convert2rhel/releases/tag/v24.03.04.3

What changed
```diff
-- RHELC-1 test
-- RHEL-1, RHEL-2 testtest
-- [RHELC-1, RHEL-1234] Title test stuff here
-- [RHELC-12] Test title
+- [RHELC-1](https://issues.redhat.com/browse/RHELC-1) test
+- [RHEL-1](https://issues.redhat.com/browse/RHEL-1), [RHEL-2](https://issues.redhat.com/browse/RHEL-2) testtest
+- [[RHELC-1](https://issues.redhat.com/browse/RHELC-1), [RHEL-1234](https://issues.redhat.com/browse/RHEL-1234)] Title test stuff here
+- [[RHELC-12](https://issues.redhat.com/browse/RHELC-12)] Test title
 - [[RHELC-1234](https://issues.redhat.com/browse/RHELC-1361)] Test title
 - [RHELC-1234](https://issues.redhat.com/browse/RHELC-1361) Test title
```
 ---

This also updates the pull_request_template to be a bit nicer

Jira Issues:

-

Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,15 +1,16 @@
 <!-- Write a description of what the PR solves and how -->
 
-<!-- Link to relevant Jira issue, add multiple if necessary -->
+<!-- Link to relevant Red Hat Jira issues -->
+Jira Issues:
 
-Jira Issues: [RHELC-](https://issues.redhat.com/browse/RHELC-)
+-
 
 Checklist
 
 - [ ] PR has been tested manually in a VM (either author or reviewer)
 - [ ] Jira issue has been made public if possible
 - [ ] `[RHELC-]` is part of the PR title <!-- For a proper sync with Jira -->
-- [ ] GitHub label has been added to help with Release notes <!-- enhancement, bug-fix, no-changelog, security-hardening, breaking-change -->
+- [ ] GitHub label has been added to help with Release notes <!-- enhancement, bug-fix, no-changelog, security-hardening, breaking-change, test-coverage-enhancement -->
 - [ ] PR title explains the change from the user's point of view
 - [ ] Code and tests are documented properly
 - [ ] The commits are squashed to as few commits as possible (without losing data) <!-- The commits can be squashed to 1 commit, but then we might lose data regarding moving something to a new file and then refactoring for example. Hence squash without losing data -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,6 +3,7 @@
 <!-- Link to relevant Red Hat Jira issues -->
 Jira Issues:
 
+<!-- List below in format of [RHELC-](https://issues.redhat.com/browse/RHELC-) -->
 -
 
 Checklist
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,8 +40,26 @@ jobs:
         run: gh release view ${{github.ref_name}} --json body --jq '.body' > releaseBody.txt
 
       - name: Add Jira links
+        shell: python
         run: |
-          sed -i -e 's/\[\(RHEL.*\)\]/[[\1](https:\/\/issues.redhat.com\/browse\/\1)]/' releaseBody.txt
+          import re
+
+          #  RegEx explained
+          # (?<!\/) Negative lookbehind of / to avoid grabbing URLs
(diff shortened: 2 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -1,15 +1,16 @@
 <!-- Write a description of what the PR solves and how -->
 
-<!-- Link to relevant Jira issue, add multiple if necessary -->
+<!-- Link to relevant Red Hat Jira issues -->
+Jira Issues:
 
-Jira Issues: [RHELC-](https://issues.redhat.com/browse/RHELC-)
+-
```
bookwar, line 6: I think we still need the example here, so that I know whether I need to add a full link to a Jira issue or just mention RHEL-1234

SpyTec: Editing an existing link is a bit tedious for me at least, do you think a comment would work? Like the other comments we have

SpyTec: Added my suggestion

.github/workflows/release.yml
```diff
@@ -40,8 +40,16 @@ jobs:
...
+        shell: python
         run: |
-          sed -i -e 's/\[\(RHEL.*\)\]/[[\1](https:\/\/issues.redhat.com\/browse\/\1)]/' releaseBody.txt
+          import re
+
+          regex=re.compile(r"(?<!\/)(RHELC?-\d+)\b(?!\]\()", re.MULTILINE)
+          with open("releaseBody.txt", 'r+') as f:
+            data = f.read()
+            f.seek(0)
+            f.write(regex.sub(r"[\g<0>](https://issues.redhat.com/browse/\g<0>)", data))
```
r0x0d, line 51: This is always a nightmare to understand... Can we move this regex to its own line and only pass the result to the write function?

.github/pull_request_template.md
r0x0d: I agree with @bookwar on this one. Without the example, we might just put whatever format we would like here... Does the parser handle different types of inputs? Let's say like just using URL instead of RHELC-XXXX?

r0x0d: Particularly, I rarely pay attention to comments while opening a PR since I do that almost every day... For newcomers It would be nice, I think people just tend to pay attention to things like that when opening their first PR. 

For me, at least, I rarely see that something changed in a comment.

.github/workflows/release.yml
SpyTec: Sure! Do you want us to annotate it too? It's essentially just `\1 url \1` to grab the first group

.github/pull_request_template.md
SpyTec: > Does the parser handle different types of inputs? Let's say like just using URL instead of RHELC-XXXX?

The parser only works for the release, though I did manage to get a RegEx to match Jira ID to replace with a link, with or without a link (brace yourself): `(https.*)?(?<!\[)(RHELC?-\d+)]?`

Problem was that I couldn't find a way to edit comments in workflows so gave up on that..

.github/workflows/release.yml
r0x0d: Yeah, I think that would be a nice idea. Even though it seems very simple, having an annotation of what is the outcome might be good as well.

SpyTec: Fixed!

.github/pull_request_template.md
SpyTec: @r0x0d WDYT

r0x0d (approved):
Looks good to me

PR #1131, opened 2024-03-05
refactor: move `utils.py` to a module

SpyTec:
As we kept using `utils.py` file and adding more and more things it
became quite bloated. To solve this the best thing to do, while also
making it easy to test and reduce cyclic imports is to split it into
smaller parts.

This takes care of the initial move by putting everything on the module
level while still having backwards compatibility for the time being.

To make this easier for ourselves we should migrate the file into
more modules over time

Checklist

### DIFF
--- a/convert2rhel/backup/certs.py
+++ b/convert2rhel/backup/certs.py
@@ -24,6 +24,7 @@ import shutil
 
 from convert2rhel import exceptions, utils
 from convert2rhel.backup import RestorableChange
+from convert2rhel.utils import files
 
 
 loggerinst = logging.getLogger(__name__)
@@ -115,7 +116,7 @@ class RestorablePEMCert(RestorableChange):
             self.previously_installed = True
         else:
             try:
-                utils.mkdir_p(self._target_cert_dir)
+                files.mkdir_p(self._target_cert_dir)
                 shutil.copy2(self._source_cert_path, self._target_cert_dir)
             except OSError as err:
                 loggerinst.critical_no_exit("OSError({0}): {1}".format(err.errno, err.strerror))
--- a/convert2rhel/utils/files.py
+++ b/convert2rhel/utils/files.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright(C) 2024 Red Hat, Inc.
+#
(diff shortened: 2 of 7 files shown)


convert2rhel/backup/certs.py
```diff
@@ -115,7 +116,7 @@ def enable(self):
             self.previously_installed = True
         else:
             try:
-                utils.mkdir_p(self._target_cert_dir)
+                files.mkdir_p(self._target_cert_dir)
```
r0x0d, line 119: Can we maintain the same pattern and either import `mkdir_p` directly or use `files.mkdir_p`? I get that we are simply replacing with the new module, but maybe we could take the opportunity and decide which pattern we should use? 

This is minor, but thought to raise the question.

convert2rhel/utils/files.py
```diff
@@ -0,0 +1,35 @@
...
+    """Create all missing directories for the path and raise no exception
+    if the path exists.
+    """
+    try:
+        os.makedirs(path)
+    except OSError as err:
+        if err.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
```
r0x0d, line 35: Maybe we could rewrite the function to remove the else part like this? 

```suggestion
        if err.errno == errno.EEXIST and os.path.isdir(path):
            return
         
         raise
```

convert2rhel/utils/files.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+import errno
+import os
+
+
+def mkdir_p(path):
+    """Create all missing directories for the path and raise no exception
+    if the path exists.
+    """
```
r0x0d, line 28: If we can update docstrings with param, raises and explaining that we skip the exception with a condition, that would be nice.

convert2rhel/backup/certs.py
SpyTec: Good point, I'll change it to be just `mkdir_p`

convert2rhel/utils/files.py
SpyTec: Sure LGTM. Didn't change any code with this PR, just moved things around

r0x0d (approved):

PR #1143, opened 2024-03-13
[RHELC-1457] Remove duplicate pop_all from rollback

r0x0d:
There was a leftover in the rollback process that was triggering the pop_all function from the backup_controller twice. Calling the pop_all twice does not cause problems with the rollback, but there is no necessity to double call it.

Jira Issues: [RHELC-1457](https://issues.redhat.com/browse/RHELC-1457)

Checklist

### DIFF
--- a/convert2rhel/unit_tests/main_test.py
+++ b/convert2rhel/unit_tests/main_test.py
@@ -60,13 +60,13 @@ class TestRollbackChanges:
 
         # Note: when we remove the BackupController partition hack, the first
         # of these calls will go away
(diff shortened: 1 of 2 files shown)


convert2rhel/unit_tests/main_test.py
```diff
@@ -60,13 +60,13 @@ def test_rollback_changes(self, monkeypatch, global_backup_control):
 
         # Note: when we remove the BackupController partition hack, the first
         # of these calls will go away
```
hosekadam, line 62:
```suggestion
```

PR #1144, opened 2024-03-14
[RHELC-1342] Add post conversion stages for Action Framework

r0x0d:
In this commit, we are introducing two new stages for enabling the post conversion in the action framework.

We are also introducing an environment variable to not break the current workflow and let the conversion function as-is for now. Anyone that is willing to test and run the post conversion actions will need to use this environment var.

Jira Issues: [RHELC-1342](https://issues.redhat.com/browse/RHELC-1342)

Checklist

### DIFF
--- a/convert2rhel/actions/__init__.py
+++ b/convert2rhel/actions/__init__.py
@@ -759,8 +759,6 @@ def run_pre_actions():
 
 def run_post_actions():
     """
-    Run all of the post-ponr Actions.
-
     This function runs the Actions that occur after the Point of no Return.
     """
     # Stages are created in the opposite order that they are run in so that
@@ -768,13 +766,13 @@ def run_post_actions():
     # next_stage parameter).
     #
     # When we call check_dependencies() or run() on the first Stage
-    # (system_checks), it will operate on the first Stage and then recursively
+    # (conversion), it will operate on the first Stage and then recursively
     # call check_dependencies() or run() on the next_stage.
     post_conversion = Stage("post_conversion", "Final modifications to the system")
     conversion = Stage("conversion", "Starting Conversion", next_stage=post_conversion)
 
     try:
-        # Check dependencies are satisfied for system_checks and all subsequent
+        # Check dependencies are satisfied for conversion and all subsequent
         # Stages.
         conversion.check_dependencies()
     except DependencyError as e:
@@ -782,7 +780,7 @@ def run_post_actions():
         # way we should fail in testing before release.
         logger.critical("Some dependencies were set on Actions but not present in convert2rhel: %s" % e)
 
-    # Run the Actions in system_checks and all subsequent Stages.
+    # Run the Actions in conversion and all subsequent Stages.
     results = conversion.run()
 
     return parse_action_results(results)
--- a/convert2rhel/actions/__init__.py
+++ b/convert2rhel/actions/__init__.py
@@ -691,7 +691,42 @@ def resolve_action_order(potential_actions, previously_resolved_actions=None):
         )
 
 
-def run_actions():
+def parse_action_results(results):
+    """
+    Parse and format action results
+
+    .. note::
+        # Format results as a dictionary:
+        {
+            "$Action_id": {
+                "messages": [
+                    {
+                        "level": int,
+                        "id": "$id",
+                        "message": "" or "$message"
+                    },
+                ]
+                "result": {
+                    "level": int,
+                    "id": "$id",
+                    "message": "" or "$message"
+                },
+            },
+        }
+
+    :param results: Unformatted results given by the actions
+    :type results: dict
+    :return: Formatted dictionary with the results
+    :rtype: dict[str, dict[str, list | dict]]
+    """
+    formatted_results = {}
+    for action in itertools.chain(*results):
+        msgs = [msg.to_dict() for msg in action.messages]
+        formatted_results[action.id] = {"messages": msgs, "result": action.result.to_dict()}
+    return formatted_results
+
+
+def run_pre_actions():
     """
     Run all of the pre-ponr Actions.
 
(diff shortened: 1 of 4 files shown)


convert2rhel/main.py
```diff
@@ -183,6 +179,26 @@ def main_locked():
...
+    if failures:
+        # The report will be handled in the error handler, after rollback.
+        loggerinst.critical("Conversion failed.")
+
+    # Print the assessment just before we ask the user whether to continue past the PONR
+    report.summary(
+        results=results,
+        include_all_reports=False,
+        disable_colors=logger_module.should_disable_color_output(),
+    )
```
SpyTec, line 199: Move this out to be it's own function, doesn't have to do with analyze

jochapma (approved):
Looks reasonable to me.

convert2rhel/main.py
```diff
@@ -183,6 +179,26 @@ def main_locked():
...
+def _analyze_actions_results(results):
+    """Analyze the action results for failures
+
+    :param results: The action results from the framework
+    :type results: dict
+    """
+    failures = actions.find_actions_of_severity(results, "SKIP", level_for_raw_action_data)
+    if failures:
+        # The report will be handled in the error handler, after rollback.
+        loggerinst.critical("Conversion failed.")
```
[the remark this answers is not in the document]
r0x0d: updated

convert2rhel/main.py
r0x0d: updated

convert2rhel/actions/__init__.py
```diff
@@ -719,28 +754,38 @@ def run_actions():
...
+    Run all of the post-ponr Actions.
+
+    This function runs the Actions that occur after the Point of no Return.
+    """
+    # Stages are created in the opposite order that they are run in so that
+    # each Stage can know about the Stage that comes after it (via the
+    # next_stage parameter).
+    #
+    # When we call check_dependencies() or run() on the first Stage
+    # (system_checks), it will operate on the first Stage and then recursively
```
hosekadam, line 771:
```suggestion
    # (conversion), it will operate on the first Stage and then recursively
```

convert2rhel/actions/__init__.py
```diff
@@ -719,28 +754,38 @@ def run_actions():
...
-        formatted_results[action.id] = {"messages": msgs, "result": action.result.to_dict()}
-    return formatted_results
+    return parse_action_results(results)
+
+
+def run_post_actions():
+    """
+    Run all of the post-ponr Actions.
+
+    This function runs the Actions that occur after the Point of no Return.
```
hosekadam, line 764:
```suggestion
    This function runs the Actions that occur after the Point of no Return.
```
Maybe I would keep just one of those sentences. They say the same from my point of view, but the second is better. And if removing one, then I would remove the same under `run_pre_actions`

convert2rhel/actions/__init__.py
```diff
@@ -719,28 +754,38 @@ def run_actions():
...
+    # next_stage parameter).
+    #
+    # When we call check_dependencies() or run() on the first Stage
+    # (system_checks), it will operate on the first Stage and then recursively
+    # call check_dependencies() or run() on the next_stage.
+    post_conversion = Stage("post_conversion", "Final modifications to the system")
+    conversion = Stage("conversion", "Starting Conversion", next_stage=post_conversion)
+
+    try:
+        # Check dependencies are satisfied for system_checks and all subsequent
```
hosekadam, line 777:
```suggestion
        # Check dependencies are satisfied for conversion and all subsequent
```

convert2rhel/actions/__init__.py
```diff
@@ -719,28 +754,38 @@ def run_actions():
...
+        # Stages.
+        conversion.check_dependencies()
+    except DependencyError as e:
+        # We want to fail early if dependencies are not properly set.  This
+        # way we should fail in testing before release.
+        logger.critical("Some dependencies were set on Actions but not present in convert2rhel: %s" % e)
+
+    # Run the Actions in system_checks and all subsequent Stages.
```
hosekadam, line 785:
```suggestion
    # Run the Actions in conversion and all subsequent Stages.
```

convert2rhel/main.py
```diff
@@ -113,22 +114,17 @@ def main_locked():
         # actions.run_actions() (either from a bug or from the user hitting
```
hosekadam, line 114:
```suggestion
        # actions.run_pre_actions() (either from a bug or from the user hitting
```

convert2rhel/main.py
```diff
@@ -113,22 +114,17 @@ def main_locked():
...
-        )
-        if pre_conversion_failures:
-            # The report will be handled in the error handler, after rollback.
-            loggerinst.critical("Conversion failed.")
+        _raise_for_skipped_failures(pre_conversion_results)
 
         # Print the assessment just before we ask the user whether to continue past the PONR
         report.summary(
-            pre_conversion_results,
+            results=pre_conversion_results,
```
hosekadam, line 127: I will add this into my PR #1153 to have the call consistent across the `main.py`

SpyTec (approved):

convert2rhel/main.py
```diff
@@ -183,6 +192,20 @@ def main_locked():
     return 0
 
 
+def _raise_for_skipped_failures(results):
```
hosekadam, line 195: Maybe I would consider a different name of this function. Like `_check_for_results_skipped_failures`, `raise_when_skipped_failures` or something else. Without the function body I wouldn't understand what is it doing

SpyTec: We can change it later, merging for now

hosekadam: Agree, it's not a big problem

PR #1151, opened 2024-03-19
[RHELC-1347] Update report to include post conversion actions

r0x0d:
The report had to be updated to include the post conversion results from
the actions, since we are extending the framework and migrating the
functions to become actions, we need to make a few modifications to
adapt the reporting to work with both pre and post conversion actions.

Jira Issues: [RHELC-1347](https://issues.redhat.com/browse/RHELC-1347)

Checklist

### Depends on 

- [x] https://github.com/oamg/convert2rhel/pull/1144
- [x] https://github.com/oamg/convert2rhel/pull/1152

### DIFF
--- a/convert2rhel/actions/report.py
+++ b/convert2rhel/actions/report.py
@@ -35,8 +35,11 @@ from convert2rhel.logger import colorize
 logger = logging.getLogger(__name__)
 
 #: The filename to store the results of running preassessment
-CONVERT2RHEL_JSON_RESULTS = "/var/log/convert2rhel/convert2rhel-pre-conversion.json"
-CONVERT2RHEL_TXT_RESULTS = "/var/log/convert2rhel/convert2rhel-pre-conversion.txt"
+CONVERT2RHEL_PRE_CONVERSION_JSON_RESULTS = "/var/log/convert2rhel/convert2rhel-pre-conversion.json"
+CONVERT2RHEL_PRE_CONVERSION_TXT_RESULTS = "/var/log/convert2rhel/convert2rhel-pre-conversion.txt"
+#: The filename to store the results of the post conversion report
+CONVERT2RHEL_POST_CONVERSION_JSON_RESULTS = "/var/log/convert2rhel/convert2rhel-post-conversion.json"
+CONVERT2RHEL_POST_CONVERSION_TXT_RESULTS = "/var/log/convert2rhel/convert2rhel-post-conversion.txt"
 
 #: Map Status codes (from convert2rhel.actions.STATUS_CODE) to color name (from
 #: convert2rhel.logger.bcolor)
@@ -57,7 +60,7 @@ _STATUS_TO_COLOR = {
 }
 
 
-def summary_as_json(results, json_file=CONVERT2RHEL_JSON_RESULTS):
+def summary_as_json(results, json_file):
     """
     Output the results as a json_file.
 
@@ -149,7 +152,37 @@ def get_combined_results_and_message(results):
     return combined_results_and_message
 
 
-def summary(results, include_all_reports=False, disable_colors=False):
+def pre_conversion_report(results, include_all_reports=False, disable_colors=False):
+    """Wrapper around summary to output pre-conversion report.
+
+    :param results: Results dictionary as returned by :func:`run_actions`
+    :type results: Mapping
+    :keyword disable_colors: Whether to color the messages according to their status
+    :type disable_colors: bool
+    :keyword include_all_reports: If all reports should be logged instead of the
+        highest ones.
+    :type include_all_reports: bool
+    """
+    logger.task("Pre-conversion analysis report")
+    _summary(results, include_all_reports, disable_colors)
+
+
+def post_conversion_report(results, include_all_reports=False, disable_colors=False):
+    """Wrapper around summary to output post-conversion report.
+
+    :param results: Results dictionary as returned by :func:`run_actions`
+    :type results: Mapping
+    :keyword disable_colors: Whether to color the messages according to their status
+    :type disable_colors: bool
+    :keyword include_all_reports: If all reports should be logged instead of the
+        highest ones.
+    :type include_all_reports: bool
+    """
+    logger.task("Post-conversion report")
+    _summary(results, include_all_reports, disable_colors)
+
+
+def _summary(results, include_all_reports, disable_colors):
     """Output a summary regarding the actions execution.
 
     This summary is intended to be used to inform the user about the results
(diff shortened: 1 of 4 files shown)


convert2rhel/actions/report.py
```diff
@@ -247,7 +254,10 @@ def summary(results, include_all_reports=False, disable_colors=False):
     # If there is no other message sent to the user, then we just give a
     # happy message to them.
     if not combined_results_and_message:
-        report.append("No problems detected during the analysis!")
+        if post_conversion:
+            report.append("No problems detected uring the conversion!")
```
SpyTec, line 258: Typo in this commit

convert2rhel/actions/report.py
```diff
@@ -213,7 +216,11 @@ def summary(results, include_all_reports=False, disable_colors=False):
         highest ones.
     :type include_all_reports: bool
     """
-    logger.task("Pre-conversion analysis report")
+    if post_conversion:
+        logger.task("Post-conversion report")
+    else:
+        logger.task("Pre-conversion analysis report")
+
```
SpyTec, line 223: I feel like this is an antipattern as we have two different uses in one function. Despite the large overlap it feels like we should either make this into a class that just generates stuff easier or we just split out the functionality more and have a higher level function for pre-conversion and post-conversion with logs

SpyTec (changes requested):

convert2rhel/actions/report.py
r0x0d: I would say different purpose rather than uses here... The usability of the function is still the same despite the logging that is being shown at the beginning/end of the function. 

Having a higher level function does not seem too overkill for such a simple if statement to redirect the logging? We would essentially move this if statement to be an outside statement inside (and around) the main function to just call the high level function that will contain basically a log message that will be different from the other...

For example (I'm visualizing the higher level functions like this): 

```python
def pre_ponr_summary(results, include_all_reports=False, disable_colors=False):
    logger.task("Pre-conversion analysis report")
    summary(results, include_all_reports, disable_colors)

def post_ponr_summary(results, include_all_reports=False, disable_colors=False):
    logger.task("Post-conversion analysis report")
    summary(results, include_all_reports, disable_colors)
```

Which in turn, would move the if statement of a pre(post)-conversion report to something like: 

```python
           if post_conversion:
                report.post_ponr_summary(...)
            else:
                report.pre_ponr_summary(...)
 ```
 
Maybe I'm just overlooking this? Do you see it in a simpler way, @SpyTec?

Comment (r0x0d) [reply convert2rhel/actions/report.py:223]: I like the idea of a class, though... We could encapsulate also the `summary_as_json` as `summary_as_txt` inside the report... And with that, we could have a single interface that output different types of summary as we need...

Comment (r0x0d) [reply convert2rhel/actions/report.py:223]: Okay, I will create a Jira ticket to transform it in a class, trying to describe the best I can the idea behind it. 

I think moving all of this to a class will help make it more organized and extensible as well. I will implement the higher level functions for now.

Comment (r0x0d) [reply convert2rhel/actions/report.py:223]: Thanks for the ideas, @SpyTec

File: convert2rhel/main.py [hunk 3]
```diff
@@ -168,26 +165,33 @@ def main_locked():
 
         rollback_changes()
 
-        report.summary(
+        report.pre_conversion_report(
             pre_conversion_results,
```
SpyTec, line 169: Might as well add this
```suggestion
        report.pre_conversion_report(
            results=pre_conversion_results,
```

convert2rhel/main.py
```diff
@@ -168,26 +165,33 @@ def main_locked():
...
         loggerinst.critical_no_exit(err.diagnosis)
-        return _handle_main_exceptions(process_phase, pre_conversion_results)
-
+        results = _pick_conversion_results(process_phase, pre_conversion_results, post_conversion_results)
+        return _handle_main_exceptions(process_phase, results)
     except (Exception, SystemExit, KeyboardInterrupt) as err:
-        return _handle_main_exceptions(process_phase, pre_conversion_results)
-
+        results = _pick_conversion_results(process_phase, pre_conversion_results, post_conversion_results)
+        return _handle_main_exceptions(process_phase, results)
```
SpyTec, line 180: Would `post_conversion_result` be populated even if we're not in post-ponr phase?

Given it's only used here though we could allow some duplication if we are going to refactor it in the future, instead of a function that is difficult to define

SpyTec (commented):
Didn't review unit tests yet

convert2rhel/main.py
r0x0d: updated.

SpyTec (approved):
Implementation seems pretty good to me, just wondering about the usage of deciding pre-conversion vs post-conversion report in the try-finally. I feel like it introduces technical debt for future refactoring

r0x0d:
> Implementation seems pretty good to me, just wondering about the usage of deciding pre-conversion vs post-conversion report in the try-finally. I feel like it introduces technical debt for future refactoring

I agree, and by far, this doesn't even look like a solid implementation. 

What do you think about something like this? 

```diff
diff --git a/convert2rhel/main.py b/convert2rhel/main.py
index f59b738..f26e563 100644
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -43,6 +43,18 @@ class ConversionPhase:
     POST_PONR_CHANGES = 4
 
 
+_REPORT_MAPPING = {
+    ConversionPhase.PRE_PONR_CHANGES: (
+        report.CONVERT2RHEL_PRE_CONVERSION_JSON_RESULTS,
+        report.CONVERT2RHEL_PRE_CONVERSION_TXT_RESULTS,
+    ),
+    ConversionPhase.POST_PONR_CHANGES: (
+        report.CONVERT2RHEL_POST_CONVERSION_JSON_RESULTS,
+        report.CONVERT2RHEL_POST_CONVERSION_TXT_RESULTS,
+    ),
+}
+
+
 def initialize_file_logging(log_name, log_dir):
     """
     Archive existing file logs and setup all logging handlers that require
@@ -182,16 +194,12 @@ def main_locked():
         # Write the assessment to a file as json data so that other tools can
         # parse and act upon it.
         results = _pick_conversion_results(process_phase, pre_conversion_results, post_conversion_results)
-        json_filepath = report.CONVERT2RHEL_PRE_CONVERSION_JSON_RESULTS
-        txt_filepath = report.CONVERT2RHEL_PRE_CONVERSION_TXT_RESULTS
 
-        if process_phase == ConversionPhase.POST_PONR_CHANGES:
-            json_filepath = report.CONVERT2RHEL_POST_CONVERSION_JSON_RESULTS
-            txt_filepath = report.CONVERT2RHEL_POST_CONVERSION_TXT_RESULTS
+        if results and process_phase in _REPORT_MAPPING:
+            json_report, txt_report = _REPORT_MAPPING[process_phase]
 
-        if results:
-            report.summary_as_json(results, json_filepath)
-            report.summary_as_txt(results, txt_filepath)
+            report.summary_as_json(results, json_report)
+            report.summary_as_txt(results, txt_report)
 
     return 0
```

It looks less bad or more bad?

PR #1152, opened 2024-03-19
[RHELC-1328] Migrate post conversion transaction function

r0x0d:
First function to be converted to the actions is done in this commit.
This function is responsible for converting system packages to rhel
packages, thus, being the first function to be executed, it was the
first one to be migrated as well.

Jira Issues: [RHELC-1328](https://issues.redhat.com/browse/RHELC-1328)

Checklist

### Depends on 

- [x] https://github.com/oamg/convert2rhel/pull/1144

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -331,9 +331,6 @@ def post_ponr_changes():
 
 def post_ponr_conversion():
     """Perform main steps for system conversion."""
-    transaction_handler = pkgmanager.create_transaction_handler()
-    loggerinst.task("Convert: Replace system packages")
-    transaction_handler.run_transaction()
     loggerinst.task("Convert: Prepare kernel")
     pkghandler.preserve_only_rhel_kernel()
     loggerinst.task("Convert: List remaining non-Red Hat packages")
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -140,19 +140,18 @@ def main_locked():
         utils.ask_to_continue()
 
         process_phase = ConversionPhase.POST_PONR_CHANGES
-        # TODO(r0x0d): Just temporary, so we don't need to rush with migrating
-        # all the functions at once to the framework.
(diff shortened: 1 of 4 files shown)


r0x0d:
This will be refactored to include only the changes to actions, without the post conversion framework modifications

convert2rhel/unit_tests/actions/conversion/__init__.py
SpyTec, line 1: Do we want to use the "conversion" name? Given that we have pre_ponr_changes and post_conversion step?

convert2rhel/unit_tests/actions/conversion/transaction_test.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    convert_system_packages.run()
+
+    assert transaction_handler_instance.run_transaction.call_count == 1
+    assert transaction_handler_instance.run_transaction.call_args == mock.call()
+    assert convert_system_packages.result.level == STATUS_CODE["SUCCESS"]
+
+
+@all_systems
```
SpyTec, line 52: `@all_systems` is not needed right? We don't use `pretend_os`

r0x0d: Yeah, this is an old test... But I guess it's true, we don't need to call all_systems.

SpyTec (approved):

convert2rhel/main.py
```diff
@@ -140,19 +140,18 @@ def main_locked():
...
-            )
-        else:
-            post_ponr_changes()
+        post_conversion_results = actions.run_post_actions()
+
+        # TODO(r0x0d): Remove this after migrating all functions to Actions.
+        post_ponr_changes()
+
+        _raise_for_skipped_failures(post_conversion_results)
+        # Print the assessment just before we ask the user whether to continue past the PONR
```
r0x0d, line 149:
```suggestion
        # Print the report summary for post conversion
```

r0x0d: Not a blocker... I will update this in another PR.

PR #1153, opened 2024-03-20
[RHELC-1275, RHELC-1516] Exit with 1 when rollback fails

hosekadam:
There is a need to exit with exit code 1 when rollback fails to provide the ability to better recognize those types of failures for Insights. This situation can mainly happen when user uses analyze command or during the conversion answers "No" and user doesn't want to continue over point of no return. In those situations, rollback is processed and when it fails, the system is in an unknown state.

**Expected behavior:**
When rollback fails, c2r exits with 1, warning about unknown state and system should be restored from backup is printed. Any summaries of the report aren't printed or saved - because there is any big value of the report when the rollback failed.

**How it was solved:**
The new property `backup_failed` of BackupController was added. This property is based on instance variable `_rollback_failed` which is set when restore method of some restorable fails. 

I chose this solution because it keeps the code simple. Other solution with returning some value required a lot of code changed and added complexity to the codebase.

I also added a new function in `main.py` for printing info after rollback, which we do in two places. The goal was to reduce duplicity in this part of code.

To have captured exceptions during the rollback, I've also removed handling of exceptions in each restorable and let it up to `BackupController`. By this change I found a bug in the code, when the restorable file was removed from the backup folder when doing the restore for conversion purposes.  The file should stay in the backup folder, if we would remove it, then we won't have anything when processing the rollback.

Jira Issues: [RHELC-1275](https://issues.redhat.com/browse/RHELC-1275), [RHELC-1516](https://issues.redhat.com/browse/RHELC-1516)

Checklist

### DIFF
--- a/convert2rhel/backup/__init__.py
+++ b/convert2rhel/backup/__init__.py
@@ -59,6 +59,7 @@ class BackupController:
 
     def __init__(self):
         self._restorables = []
+        self._rollback_failures = []
 
     def push(self, restorable):
         """
@@ -97,18 +98,18 @@ class BackupController:
         """
         Restores all RestorableChanges known to the Controller and then returns them.
 
-        :returns: List of RestorableChange objects that were known to the Controller.
+        :returns list[RestorableChange]: List of RestorableChange objects that were processed by the Controller.
         :raises IndexError: If there are no RestorableChanges currently known to the Controller.
 
         After running, the Controller object will not know about any RestorableChanges.
         """
         # Only raise IndexError if there are no restorables registered.
-        # Partitions are ignored for this check as they aren't really Changes.
         if not self._restorables:
             raise IndexError("No backups to restore")
 
-        # Restore the Changes in the reverse order the changes were enabled.
         processed_restorables = []
+
+        # Restore the Changes in the reverse order the changes were enabled.
         while True:
             try:
                 restorable = self._restorables.pop()
@@ -121,11 +122,33 @@ class BackupController:
             # logger.critical in some places.
             except (Exception, SystemExit) as e:
                 # Don't let a failure in one restore influence the others
-                loggerinst.warning("Error while rolling back a %s: %s" % (restorable.__class__.__name__, str(e)))
+                message = "Error while rolling back a %s: %s" % (restorable.__class__.__name__, str(e))
+                loggerinst.warning(message)
+                # Add the rollback failures to the list
+                self._rollback_failures.append(message)
 
             processed_restorables.append(restorable)
 
         return processed_restorables
+
+    @property
+    def rollback_failed(self):
+        """
+        Return True when one or more restorables were unsuccessful.
+
+        :returns: bool True if the rollback failed, False if rollback succeeded.
+        """
+
+        return any(self._rollback_failures)
+
+    @property
+    def rollback_failures(self):
+        """
+        Errors from failed rollback. Used to provide more information to the user.
+
+        :returns: List of strings containing errors captured during rollback.
+        """
+        return self._rollback_failures
 
 
 @six.add_metaclass(abc.ABCMeta)
--- a/convert2rhel/backup/certs.py
+++ b/convert2rhel/backup/certs.py
@@ -146,6 +146,11 @@ class RestorablePEMCert(RestorableChange):
     def _restore(self):
         """The actual code to remove the certificate.  Done in a helper method so we can handle all
         the cases where we do not want to remove and still run the base class's restore().
+
+            .. warning::
+                Exceptions are not handled and left for handling by the calling code.
+
+        :raises OSError: Can be raised when the file is not found.
         """
         # Check whether any package owns the certificate file.  It is
         # possible that a new rpm package was installed after we copied the
@@ -187,8 +192,10 @@ class RestorablePEMCert(RestorableChange):
             if err.errno == errno.ENOENT:
                 # Resolves RHSM error when removing certs, as the system might not have installed any certs yet
                 loggerinst.info("No certificates found to be removed.")
-            else:
-                loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
+                return
+
+            # Will be handled in BackupController
+            raise
 
 
 def _get_cert(cert_dir):
--- a/convert2rhel/backup/files.py
+++ b/convert2rhel/backup/files.py
@@ -107,26 +107,32 @@ class RestorableFile(RestorableChange):
         return filepath
 
     def restore(self, rollback=True):
-        """Restore a previously backed up file"""
+        """Restore a previously backed up file
+
+        :arg rollback: bool value to decide if there is need print the rollback messages.
+            This argument can also be used during conversion for restoring some file needed
+            for conversion and thus won't need rollback messages.
+
+            .. warning::
+                Exceptions are not handled and left for handling by the calling code.
+
+        :raises OSError: When the backed up file is missing.
+        :raises IOError: When the backed up file is missing.
+        """
         if rollback:
             loggerinst.task("Rollback: Restore %s from backup" % self.filepath)
         else:
             loggerinst.info("Restoring %s from backup" % self.filepath)
 
-        # We do not have backup or not backed up by this
-        if (not self.enabled) or (not os.path.isfile(self._backup_path)):
+        if not self.enabled:
             loggerinst.info("%s hasn't been backed up." % self.filepath)
             return
 
-        try:
-            shutil.copy2(self._backup_path, self.filepath)
+        # Possible exceptions will be handled in the BackupController
+        shutil.copy2(self._backup_path, self.filepath)
+        if rollback:
+            # Remove the backed up file only when processing rollback
             os.remove(self._backup_path)
-        except (OSError, IOError) as err:
-            # Do not call 'critical' which would halt the program. We are in
-            # a rollback phase now and we want to rollback as much as possible.
-            # IOError for py2 and OSError for py3
-            loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
-            return
 
         if rollback:
             loggerinst.info("File %s restored." % self.filepath)
@@ -171,6 +177,14 @@ class MissingFile(RestorableChange):
         super(MissingFile, self).enable()
 
     def restore(self):
+        """Remove the file if it was created during conversion.
+
+            .. warning::
+                Exceptions are not handled and left for handling by the calling code.
+
+        :raises OSError: When the removal of the file fails.
+        :raises IOError: When the removal of the file fails.
+        """
         if not self.enabled:
             return
 
@@ -179,13 +193,8 @@ class MissingFile(RestorableChange):
         if not os.path.isfile(self.filepath):
             loggerinst.info("File {filepath} wasn't created during conversion".format(filepath=self.filepath))
         else:
-            try:
-                os.remove(self.filepath)
-                loggerinst.info("File {filepath} removed".format(filepath=self.filepath))
-            except OSError as err:
-                # Do not call 'critical' which would halt the program. We are in
-                # a rollback phase now and we want to rollback as much as possible.
-                loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
-                return
+            # Possible exceptions will be handled in the BackupController
+            os.remove(self.filepath)
+            loggerinst.info("File {filepath} removed".format(filepath=self.filepath))
 
-        super(MissingFile, self).restore()
+            super(MissingFile, self).restore()
--- a/convert2rhel/backup/files.py
+++ b/convert2rhel/backup/files.py
@@ -179,8 +179,8 @@ class MissingFile(RestorableChange):
     def restore(self):
         """Remove the file if it was created during conversion.
 
-            .. warning::
-                Exceptions are not handled and left for handling by the calling code.
+        .. warning::
+            Exceptions are not handled and left for handling by the calling code.
 
         :raises OSError: When the removal of the file fails.
         :raises IOError: When the removal of the file fails.
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -198,7 +198,7 @@ def main_locked():
         provide_status_after_rollback(pre_conversion_results, include_all_reports=True)
 
         if backup.backup_control.rollback_failed:
-            return 1
+            return ConversionExitCodes.FAILURE
 
         report.pre_conversion_report(
             results=pre_conversion_results,
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -195,6 +195,10 @@ def main_locked():
             subscription.update_rhsm_custom_facts()
 
         rollback_changes()
+        provide_status_after_rollback(pre_conversion_results, include_all_reports=True)
+
+        if backup.backup_control.rollback_failed:
+            return 1
 
         report.pre_conversion_report(
             results=pre_conversion_results,
@@ -202,7 +206,6 @@ def main_locked():
             disable_colors=logger_module.should_disable_color_output(),
         )
         return ConversionExitCodes.SUCCESSFUL
-
     except _InhibitorsFound as err:
         loggerinst.critical_no_exit(str(err))
         _handle_main_exceptions(process_phase, pre_conversion_results)
@@ -215,15 +218,16 @@ def main_locked():
         results = _pick_conversion_results(process_phase, pre_conversion_results, post_conversion_results)
         return _handle_main_exceptions(process_phase, results)
     finally:
-        # Write the assessment to a file as json data so that other tools can
-        # parse and act upon it.
-        results = _pick_conversion_results(process_phase, pre_conversion_results, post_conversion_results)
+        if not backup.backup_control.rollback_failed:
+            # Write the assessment to a file as json data so that other tools can
+            # parse and act upon it.
+            results = _pick_conversion_results(process_phase, pre_conversion_results, post_conversion_results)
 
-        if results and process_phase in _REPORT_MAPPING:
-            json_report, txt_report = _REPORT_MAPPING[process_phase]
+            if results and process_phase in _REPORT_MAPPING:
+                json_report, txt_report = _REPORT_MAPPING[process_phase]
 
-            report.summary_as_json(results, json_report)
-            report.summary_as_txt(results, txt_report)
+                report.summary_as_json(results, json_report)
+                report.summary_as_txt(results, txt_report)
 
     return ConversionExitCodes.SUCCESSFUL
 
@@ -277,14 +281,10 @@ def _handle_main_exceptions(process_phase, results=None):
             subscription.update_rhsm_custom_facts()
 
         rollback_changes()
-        if results is None:
-            loggerinst.info("\nConversion interrupted before analysis of system completed. Report not generated.\n")
-        else:
-            report.pre_conversion_report(
-                results=results,
-                include_all_reports=True,
-                disable_colors=logger_module.should_disable_color_output(),
-            )
+        provide_status_after_rollback(
+            pre_conversion_results=results,
+            include_all_reports=True,
+        )
     elif process_phase == ConversionPhase.POST_PONR_CHANGES:
         # After the process of subscription is done and the mass update of
         # packages is started convert2rhel will not be able to guarantee a
@@ -421,3 +421,32 @@ def rollback_changes():
             loggerinst.info("During rollback there were no backups to restore")
         else:
             raise
+
+    return
+
+
+def provide_status_after_rollback(pre_conversion_results, include_all_reports):
+    """Print after-rollback messages and determine if there is a report to print or
+    if report shouldn't be printed after failure in rollback."""
+    if backup.backup_control.rollback_failed:
+        loggerinst.critical_no_exit(
+            "Rollback of system wasn't completed successfully.\n"
+            "The system is left in an undetermined state that Convert2RHEL cannot fix.\n"
+            "It is strongly recommended to store the Convert2RHEL logs for later investigation, and restore"
+            " the system from a backup.\n"
+            "Following errors were captured during rollback:\n"
+            "%s" % "\n".join(backup.backup_control.rollback_failures)
+        )
+
+        return
+
+    if not pre_conversion_results:
+        loggerinst.info("\nConversion interrupted before analysis of system completed. Report not generated.\n")
+
+        return
+
+    report.pre_conversion_report(
+        results=pre_conversion_results,
+        include_all_reports=include_all_reports,
+        disable_colors=logger_module.should_disable_color_output(),
+    )
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -240,14 +240,29 @@ def register_system():
         registration_cmd = RegistrationCommand.from_tool_opts(tool_opts)
 
         try:
-            # The file /etc/os-release is needed for subscribing the system and is being removed with
-            # <system-name>-release package in one of the steps before
-            # RHELC-16
-            os_release_file.restore(rollback=False)
+            # Only in first attempt, in the next attempts the file is already restored from the first run
+            if attempt <= 0:
+                # The file /etc/os-release is needed for subscribing the system and is being removed with
+                # <system-name>-release package in one of the steps before
+                # RHELC-16
+                os_release_file.restore(rollback=False)
+        except (OSError, IOError) as e:
+            loggerinst.critical_no_exit(
+                "Failed to restore the /etc/os-release file needed for subscribing the system with message: %s" % str(e)
+            )
+            raise exceptions.CriticalError(
+                id_="FAILED_TO_SUBSCRIBE_SYSTEM",
+                title="Failed to subscribe system.",
+                description="Failed to restore the /etc/os-release file needed for subscribing the system.",
+                diagnosis="The restore failed with error %s." % (str(e)),
+            )
+
+        try:
             registration_cmd()
             # Need to remove the file, if it would stay there would be leftover /etc/os-release.rpmorig
             # after conversion
             # RHELC-16
+            # Will be removed only if the registration succeed
             os_release_file.remove()
             loggerinst.info("System registration succeeded.")
         except KeyboardInterrupt:
--- a/convert2rhel/unit_tests/main_test.py
+++ b/convert2rhel/unit_tests/main_test.py
@@ -571,7 +571,7 @@ class TestRollbackFromMain:
         assert pkghandler.clear_versionlock.call_count == 1
         assert pkgmanager.clean_yum_metadata.call_count == 1
         assert actions.run_pre_actions.call_count == 1
-        assert report._summary.call_count == 1
+        assert report._summary.call_count == 2
         assert breadcrumbs.finish_collection.call_count == 1
         assert subscription.should_subscribe.call_count == 1
         assert subscription.update_rhsm_custom_facts.call_count == 0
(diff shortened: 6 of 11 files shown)


convert2rhel/backup/__init__.py
```diff
@@ -86,18 +87,18 @@ def pop_all(self):
         """
         Restores all RestorableChanges known to the Controller and then returns them.
 
-        :returns: List of RestorableChange objects that were known to the Controller.
+        :returns processed_restorables: List of RestorableChange objects that were processed by the Controller.
```
r0x0d, line 90: In this case you don't need the variable name here, you should use the type, I believe: 

```suggestion
        :returns list[RestorableChange]: List of RestorableChange objects that were processed by the Controller.
```

convert2rhel/backup/__init__.py
```diff
@@ -48,6 +48,7 @@ class BackupController:
 
     def __init__(self):
         self._restorables = []
+        self._rollback_failed = False
```
r0x0d, line 51: @hosekadam, do you mind updating the class docstring to include those two properties? 

Just know I realized that we don't have a description even for `_restorables`, so it would be good to update it and include both of them.

convert2rhel/backup/__init__.py
```diff
@@ -111,11 +112,17 @@ def pop_all(self):
...
                 loggerinst.warning("Error while rolling back a %s: %s" % (restorable.__class__.__name__, str(e)))
+                # Set when rollback failed to know, there were issues
+                self._rollback_failed = True
 
             processed_restorables.append(restorable)
 
         return processed_restorables
 
+    @property
+    def rollback_failed(self):
```
r0x0d, line 123: A docstring here as well, please!

convert2rhel/main.py
```diff
@@ -158,13 +158,9 @@ def main_locked():
...
         rollback_changes()
+        print_info_after_rollback(pre_conversion_results, True)
 
-        report.summary(
-            pre_conversion_results,
-            include_all_reports=True,
-            disable_colors=logger_module.should_disable_color_output(),
-        )
-        return 0
+        return backup.backup_control.rollback_failed
```
r0x0d, line 163: This will return `False` instead of 1, right?

```
/tmp> python t.py                                                                                                                             03/22/24 09:27:19 AM
/tmp> $env.LAST_EXIT_CODE                                                                                                                     03/22/24 09:27:20 AM
0
/tmp> cat t.py                                                                                                                                03/22/24 09:27:23 AM
def test():
    return False

test()
```

r0x0d: And if it is 0, why not keeping it simply as 0?

convert2rhel/main.py
```diff
@@ -340,3 +329,28 @@ def rollback_changes():
...
+    if backup.backup_control.rollback_failed:
+        loggerinst.critical_no_exit(
+            "Rollback of system wasn't completed successfully.\n"
+            "The system is left in an undetermined state that Convert2RHEL cannot fix.\n"
+            "It is strongly recommended to store the Convert2RHEL logs for later investigation, and restore"
+            " the system from a backup."
+        )
+        return
+
+    elif pre_conversion_results is None:
```
r0x0d, line 348:
```suggestion
    elif not pre_conversion_results:
```

convert2rhel/main.py
r0x0d: Ah, this test is false... It will always return the exit code for the interpreter instead of the return code of the script. To make it return the exit code for the script, we would need to use sys.exit() passing the return value for the function to it

convert2rhel/backup/__init__.py
hosekadam: I will do that. Just I'll wait a bit till the PR will be close to merging

convert2rhel/main.py
```diff
@@ -363,3 +355,31 @@ def rollback_changes():
...
+
+    elif not pre_conversion_results:
+        loggerinst.info("\nConversion interrupted before analysis of system completed. Report not generated.\n")
+
+    else:
+        report.summary(
+            results=pre_conversion_results,
+            include_all_reports=include_all_reports,
+            disable_colors=logger_module.should_disable_color_output(),
+        )
```
SpyTec, line 385: Blocker. This is out of scope of the PR, please open a new one for this refactor and technical debt fixing :pray:

convert2rhel/main.py
```diff
@@ -167,12 +167,11 @@ def main_locked():
...
+
+        if backup.backup_control.rollback_failed:
+            return 1
 
-        report.summary(
-            pre_conversion_results,
-            include_all_reports=True,
-            disable_colors=logger_module.should_disable_color_output(),
-        )
         return 0
```
SpyTec, line 175: We could simply `return backup.backup_control.rollback_failed`

convert2rhel/backup/__init__.py
```diff
@@ -110,12 +111,34 @@ def pop_all(self):
...
+        """
+        Flag of fail when restoring one or more restorables.
+
+        :returns: Boolean value if the rollback failed.
+        """
+
+        return bool(self._rollback_failures)
```
SpyTec, line 131: Since this is a list I'd prefer changing it to [`any()`](https://docs.python.org/3/library/functions.html#any)
```suggestion
        return any(self._rollback_failures)
```

convert2rhel/backup/files.py
```diff
@@ -182,10 +182,5 @@ def restore(self):
...
                 loggerinst.info("File {filepath} removed".format(filepath=self.filepath))
-            except OSError as err:
-                # Do not call 'critical' which would halt the program. We are in
-                # a rollback phase now and we want to rollback as much as possible.
-                loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
-                return
-
-        super(MissingFile, self).restore()
+            finally:
+                super(MissingFile, self).restore()
```
SpyTec, line 186: `finally` is called even when an exception is raised. How will this affect this overall as this now sets `self.enabled = False`?

convert2rhel/backup/files.py
```diff
@@ -182,10 +182,5 @@ def restore(self):
             try:
                 os.remove(self.filepath)
                 loggerinst.info("File {filepath} removed".format(filepath=self.filepath))
-            except OSError as err:
-                # Do not call 'critical' which would halt the program. We are in
-                # a rollback phase now and we want to rollback as much as possible.
-                loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
-                return
```
SpyTec, line 189: Making this be handled higher up the chain is fine. LGTM!

convert2rhel/main.py
r0x0d: I asked @hosekadam to turn return explicitly here to make it easier to look at the return value and see what code we expect. 

But, regardless, with https://issues.redhat.com/browse/RHELC-1276, I'm planning to introduce an enum to match the return codes we (will) have to figure out what does 0/1/2/... means when returning.

convert2rhel/backup/__init__.py
r0x0d: I think this will have the same effect here, no? Both `any()` and `bool()` would return true in case of having at least one element in the list. Especially because the list is a list[str] and not a list[bool]. 

In any case, not against changing to use `any` here, just think the idea will be the same with `bool` and `any`. 

On the other hand, I would expect seeing an `any` check here in case we would have something like `[False, False, True]`, in this case, `any` would be applicable perfectly here, as we care about the only one `True` and not the others `False` inside the list.

SpyTec: For sure seems to be the same. I don't have to think "how does arrays get evaluated as a bool?" when I see an `any()` so I prefer that. Just personal preference for readability

convert2rhel/main.py
SpyTec: Checked again and missed the `rollback_failed` check in the initial review, thus thinking this was just a refactor that was unrelated

convert2rhel/main.py
```diff
@@ -363,3 +355,31 @@ def rollback_changes():
             loggerinst.info("During rollback there were no backups to restore")
         else:
             raise
+
+    return
+
+
+def print_info_after_rollback(pre_conversion_results, include_all_reports):
+    """Print after-rollback messages. Also determine if there is a report to print or
+    report should't be printed after failure in rollback."""
```
SpyTec, line 364: I don't like the name too much as we not only do logging in it but also create a report summary. Not sure of the best name tbh

convert2rhel/backup/files.py
hosekadam: Yes, you are right, there is no need to have the `finally`. Also, thankfully to this, I found other places where some exception in the `restore` method would be handled and haven't been sent to the `BackupController`

hosekadam:
What was done:
- renamed the `print_info_after_rollback` -> `provide_status_after_rollback`
- applied the suggestions (use the `any()` and improve comment)
- removed exception handling for `RestorablePEMCert` and `RestorableFile` (will be handled in `BackupController`)
- fixed unit tests (mainly removing the tests for exception handling)

convert2rhel/main.py
hosekadam: Renamed to `provide_status_after_rollback`

convert2rhel/backup/certs.py
```diff
@@ -187,7 +187,8 @@ def _restore(self):
                 # Resolves RHSM error when removing certs, as the system might not have installed any certs yet
                 loggerinst.info("No certificates found to be removed.")
             else:
-                loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
+                # Will be handled in BackupController
+                raise err
```
SpyTec, line 191: Can remove the else here. Codecov also claims this isn't tested

convert2rhel/backup/files.py
```diff
@@ -118,15 +118,9 @@ def restore(self, rollback=True):
...
-            os.remove(self._backup_path)
-        except (OSError, IOError) as err:
-            # Do not call 'critical' which would halt the program. We are in
-            # a rollback phase now and we want to rollback as much as possible.
-            # IOError for py2 and OSError for py3
-            loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
-            return
+        # Possible exceptions will be handled in the BackupController
+        shutil.copy2(self._backup_path, self.filepath)
+        os.remove(self._backup_path)
```
SpyTec, line 123: For good measure lets add this to the docstring. That this function throws OSError, IOError. Ask in team chat for syntax if you need it. @r0x0d probably knows if you can't find it

convert2rhel/backup/files.py
```diff
@@ -179,13 +173,8 @@ def restore(self):
...
-                os.remove(self.filepath)
-                loggerinst.info("File {filepath} removed".format(filepath=self.filepath))
-            except OSError as err:
-                # Do not call 'critical' which would halt the program. We are in
-                # a rollback phase now and we want to rollback as much as possible.
-                loggerinst.critical_no_exit("Error(%s): %s" % (err.errno, err.strerror))
-                return
+            # Possible exceptions will be handled in the BackupController
+            os.remove(self.filepath)
+            loggerinst.info("File {filepath} removed".format(filepath=self.filepath))
```
SpyTec, line 178: Same here, mention OSError, IOError in docstring

convert2rhel/main.py
```diff
@@ -167,12 +167,11 @@ def main_locked():
             subscription.update_rhsm_custom_facts()
 
         rollback_changes()
+        provide_status_after_rollback(pre_conversion_results, True)
```
SpyTec, line 170:
```suggestion
        provide_status_after_rollback(pre_conversion_results, include_all_reports=True)
```

convert2rhel/main.py
```diff
@@ -222,14 +221,7 @@ def _handle_main_exceptions(process_phase, pre_conversion_results=None):
...
         rollback_changes()
-        if pre_conversion_results is None:
-            loggerinst.info("\nConversion interrupted before analysis of system completed. Report not generated.\n")
-        else:
-            report.summary(
-                pre_conversion_results,
-                include_all_reports=(toolopts.tool_opts.activity == "analysis"),
-                disable_colors=logger_module.should_disable_color_output(),
-            )
+        provide_status_after_rollback(pre_conversion_results, (toolopts.tool_opts.activity == "analysis"))
```
SpyTec, line 224:
```suggestion
        provide_status_after_rollback(pre_conversion_results, include_all_reports=(toolopts.tool_opts.activity == "analysis"))
```
or since it's long
```suggestion
        include_all_reports = toolopts.tool_opts.activity == "analysis"
        provide_status_after_rollback(pre_conversion_results, include_all_reports)
```

convert2rhel/main.py
```diff
@@ -363,3 +355,31 @@ def rollback_changes():
             loggerinst.info("During rollback there were no backups to restore")
         else:
             raise
+
+    return
+
+
+def provide_status_after_rollback(pre_conversion_results, include_all_reports):
+    """Print after-rollback messages. Also determine if there is a report to print or
+    report should't be printed after failure in rollback."""
```
SpyTec, line 364:
```suggestion
    """Print after-rollback messages and determine if there is a report to print, or
    if report shouldn't be printed after failure in rollback."""
```

SpyTec (commented):
Seems good, minor things. Will you add integration tests?

convert2rhel/backup/files.py
r0x0d: It should be: 

```python
"""
<title>

<description>
:raises OSError: why it raises oserror
:raises IOError: Usually, same as the above ^ 
"""
```

r0x0d: Thinking more about this now, maybe we should even consider adding a `warning` in the docstring to mention that whatever exception raised here will bubble up and be handled by the caller of this plugin. 

```python
"""
<title>

<description>

     .. warning::
         Add the warning message here

:raises OSError: why it raises oserror
:raises IOError: Usually, same as the above ^ 
"""
```

SpyTec: Agreed, sounds good!

convert2rhel/backup/certs.py
hosekadam: I think I can't remove the else path there. Because above, there is a test if [`err.errno == errno.ENOENT`](https://github.com/oamg/convert2rhel/blob/main/convert2rhel/backup/certs.py#L186) and if yes, info is just printed about there is nothing to be removed and in that case is no need to raise the exception.

So I'll have to add the unit tests for this line.

convert2rhel/backup/files.py
hosekadam: Thanks! I'll add those comment improvements

convert2rhel/backup/certs.py
r0x0d: You could just: 

```python
except OSError as err:
            if err.errno == errno.ENOENT:
                # Resolves RHSM error when removing certs, as the system might not have installed any certs yet
                loggerinst.info("No certificates found to be removed.")
                return
                
          loggerinst.error("OSError({0}): {1}".format(err.errno, err.strerror))
          # Will be handled in BackupController
          raise err
```

r0x0d: Except with the crappy formatting, but you could return in that if statement.

hosekadam: Oh! Right, thanks :)

hosekadam:
What was done:
- added name of arg to calls of provide_status_after_rollback and fixed comment in it
- changed the re-raising of the exception in RestorableCert
- added the exceptions into the doc strings
- added unit test for the exception raised in cert.restore
- added exception handling of system_release file for restoring. No unit tests for this, waiting for approval if it's fine.

convert2rhel/subscription.py
```diff
@@ -190,6 +190,18 @@ def register_system():
...
+                "Failed to restore the /etc/os-release needed fo subscribing the system with message: %s" % str(e)
+            )
+            raise exceptions.CriticalError(
+                id_="FAILED_TO_SUBSCRIBE_SYSTEM",
+                title="Failed to subscribe system.",
+                description="Failed to restore the /etc/os-release needed fo subscribing the system.",
+                diagnosis="The restore failed with error %s." % (str(e)),
+            )
+
+        try:
```
SpyTec, line 204: Don't wait for approval to do unit tests, mention what you want to do in chat or elsewhere if you want to double check if it's good. Overall this try catch is fine

convert2rhel/backup/__init__.py
```diff
@@ -110,12 +111,34 @@ def pop_all(self):
...
             processed_restorables.append(restorable)
 
         return processed_restorables
 
+    @property
+    def rollback_failed(self):
+        """
+        Return True when one or more restorables were unsuccessful.
+
+        :returns: Boolean value if the rollback failed.
```
SpyTec, line 128:
```suggestion
        :returns: bool True if the rollback failed, False if rollback succeeded.
```

convert2rhel/backup/files.py
```diff
@@ -107,7 +107,18 @@ def _hash_backup_path(self):
         return filepath
 
     def restore(self, rollback=True):
-        """Restore a previously backed up file"""
+        """Restore a previously backed up file
+
+        :arg rollback: Bool value to decide if there is need print the rollback messages.
+            This method can be also used during conversion for restoring some file needed
+            for conversion.
```
SpyTec, line 114:
```suggestion
        :arg rollback: bool value to decide if there is need print the rollback messages.
            This argument can also be used during conversion for restoring some file needed
            for conversion and thus won't need rollback messages.
```

convert2rhel/backup/files.py
```diff
@@ -171,6 +176,15 @@ def enable(self):
...
     def restore(self):
+        """ "
+        Remove the file if was created during conversion.
+
+            .. warning::
+                Exceptions are not handled and left for handling by the calling code.
+
+        :raises OSError: Can be raised when error happen during removal of the file.
+        :raises IOError: Can be raised when error happen during removal of the file.
+        """
```
SpyTec, line 187:
```suggestion
        """ "
        Remove the file if it was created during conversion.

            .. warning::
                Exceptions are not handled and left for handling by the calling code.

        :raises OSError: When the removal of the file fails.
        :raises IOError: When the removal of the file fails.
        """
```

convert2rhel/backup/files.py
```diff
@@ -107,7 +107,18 @@ def _hash_backup_path(self):
...
+
+        :arg rollback: Bool value to decide if there is need print the rollback messages.
+            This method can be also used during conversion for restoring some file needed
+            for conversion.
+
+            .. warning::
+                Exceptions are not handled and left for handling by the calling code.
+
+        :raises OSError: Can be raised when there is missing the backed up file.
+        :raises IOError: Can be raised when there is missing the backed up file.
```
SpyTec, line 120:
```suggestion
        :raises OSError: When the backed up file is missing.
        :raises IOError: When the backed up file is missing.
```

convert2rhel/subscription.py
```diff
@@ -190,6 +190,18 @@ def register_system():
...
             os_release_file.restore(rollback=False)
+        except (OSError, IOError) as e:
+            loggerinst.critical_no_exit(
+                "Failed to restore the /etc/os-release needed fo subscribing the system with message: %s" % str(e)
+            )
+            raise exceptions.CriticalError(
+                id_="FAILED_TO_SUBSCRIBE_SYSTEM",
+                title="Failed to subscribe system.",
+                description="Failed to restore the /etc/os-release needed fo subscribing the system.",
+                diagnosis="The restore failed with error %s." % (str(e)),
```
SpyTec, line 201:
```suggestion
                description="Failed to restore the /etc/os-release file needed for subscribing the system.",
                diagnosis="The restore failed with error %s." % (str(e)),
```

convert2rhel/subscription.py
```diff
@@ -190,6 +190,18 @@ def register_system():
             # <system-name>-release package in one of the steps before
             # RHELC-16
             os_release_file.restore(rollback=False)
+        except (OSError, IOError) as e:
+            loggerinst.critical_no_exit(
+                "Failed to restore the /etc/os-release needed fo subscribing the system with message: %s" % str(e)
```
SpyTec, line 195:
```suggestion
            loggerinst.critical_no_exit(
                "Failed to restore the /etc/os-release file needed for subscribing the system with message: %s" % str(e)
```

convert2rhel/unit_tests/main_test.py
```diff
@@ -68,6 +70,76 @@ def test_backup_control_unknown_exception(self, monkeypatch, global_backup_contr
...
+            report_summary.assert_called_with(
+                results=pre_conversion_results,
+                include_all_reports=include_all_reports,
+                disable_colors=logger_module.should_disable_color_output(),
+            )
+        try:
+            assert message == caplog.records[-1].message
+        except IndexError:
+            # Nothing printed
+            assert message == caplog.text
```
SpyTec, line 141: This call only tests the logs atm. Since the other calls, particularly `summary()` is mocked out it is not being tested correctly, instead we're testing a mock

If you can, remove the mock for report and mock out the file handling things instead

convert2rhel/subscription.py
hosekadam: Yeah, I'll add the unit tests now :)

convert2rhel/unit_tests/main_test.py
hosekadam: @SpyTec I was thinking about it like the `summary()` is being tested elsewhere and here I just need to test if the appropriate log is printed in the right moment and also report is generated when needed (and called with right values). 

But I think I can change that, it won't be a big problem I believe.

hosekadam:
- added the unit test for handling exception in `register_system()`
- added condition to restore the os-release file only in the first attempt. In the next attempts the file is already restored
- removed check if the backup is present and let the possible exception to be raised. Without it, it won’t be handled by the BackupController/something else, and we won’t know what happened. It will be just captured in the log somewhere during the conversion
- added unit tests for restoring only in the first attempt of registering the system
- removed one case of unit test when the file for restoring wasn't present. Now this raises exception
- fixed the comments based on review

convert2rhel/unit_tests/main_test.py
SpyTec: Testing logs is fine but we're supposed to test that the function works, not just that it calls what we want. Better to mock out files IMO

convert2rhel/backup/__init__.py
SpyTec, line 1: We probably want to add  `rollback=True` to the restore() method as it's overridden by other classes

SpyTec (approved):
LGTM, one small thin

> Fixed the problem mentioned in [issues.redhat.com/browse/RHELC-1516](https://issues.redhat.com/browse/RHELC-1516)

Do you mean this PR also fixes that issue? If so please update the PR description and title

convert2rhel/backup/__init__.py
hosekadam: About the RHELC-1516 I'm not sure which of the possibilities I mentioned in the Jira is better. I added this topic to today's tech cabal.

> We probably want to add rollback=True to the restore() method as it's overridden by other classes
I'm not sure what you mean. the `rollback=True` is already present in the `RestorableFile` (I don't see any other restorable which would need that) and works like a charm.

Just about the integration tests, should I add them? Or will you @danmyway take care of it?

SpyTec: This one
image

hosekadam: Discussed with @SpyTec and we won't be changing it in this PR

SpyTec:
Has merge conflicts

hosekadam:
Solved and rebased :)

convert2rhel/main.py
```diff
@@ -394,3 +390,31 @@ def rollback_changes():
...
+
+    elif not pre_conversion_results:
+        loggerinst.info("\nConversion interrupted before analysis of system completed. Report not generated.\n")
+
+    else:
+        report.pre_conversion_report(
+            results=pre_conversion_results,
+            include_all_reports=include_all_reports,
+            disable_colors=logger_module.should_disable_color_output(),
+        )
```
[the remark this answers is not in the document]
r0x0d: The pre_conversion_report needs to deindent one level... for some reason making it throught the UI didn't worked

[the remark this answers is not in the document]
hosekadam: Agree, it looks better

hosekadam:
- rebased, solved conflicts
- fixed the `"` symbol in the one of the comments
- removed `elif` and deindented `report.pre_conversion_report` in `provide_status_after_rollback`

convert2rhel/backup/certs.py
```diff
@@ -146,6 +146,11 @@ def restore(self):
     def _restore(self):
         """The actual code to remove the certificate.  Done in a helper method so we can handle all
         the cases where we do not want to remove and still run the base class's restore().
+
+            .. warning::
+                Exceptions are not handled and left for handling by the calling code.
```
r0x0d, line 151:
```suggestion
        .. warning::
            Exceptions are not handled and left for handling by the calling code.
```

r0x0d: Deindent the waring section

convert2rhel/backup/files.py
```diff
@@ -171,6 +177,14 @@ def enable(self):
         super(MissingFile, self).enable()
 
     def restore(self):
+        """Remove the file if it was created during conversion.
+
+            .. warning::
+                Exceptions are not handled and left for handling by the calling code.
```
r0x0d, line 183:
```suggestion
        .. warning::
            Exceptions are not handled and left for handling by the calling code.
```

convert2rhel/subscription.py
```diff
@@ -240,14 +240,29 @@ def register_system():
...
+        except (OSError, IOError) as e:
+            loggerinst.critical_no_exit(
+                "Failed to restore the /etc/os-release file needed for subscribing the system with message: %s" % str(e)
+            )
+            raise exceptions.CriticalError(
+                id_="FAILED_TO_SUBSCRIBE_SYSTEM",
+                title="Failed to subscribe system.",
+                description="Failed to restore the /etc/os-release file needed for subscribing the system.",
+                diagnosis="The restore failed with error %s." % (str(e)),
+            )
```
r0x0d, line 258: Maybe we should use a different id and tittle? We are not subscribing the system at this point, only trying to restore the os_release file. 

The registration comes in the line 261

r0x0d: I mean, the procedure is for registering the system, but at this specific step we are not actually registering

r0x0d (commented):
Looks good! Only one question though.

convert2rhel/subscription.py
hosekadam: I was thinking about that, and it resulted in this. I see it that we are trying to subscribe the system and this is part of the process. The problem is then specified in the description and diagnostic. 

But if we would find a better name, I will change that, no problem. I'm not sure which, I felt like `FAILED_TO_RESTORE_OSRELEASE` is a bit misleading.... But I'm not sure there, the naming is the hardest part sometimes :smile:

SpyTec: This is a big question and I feel like we need a wiki page or something with all of these things, when to use what, and how each is triggered etc. At the moment it's just "this feels good" but it can lead to confusing things for the customer

Since this prevents the system from being registering it still fails to register, so I feel like this error is still accurate

hosekadam:
Fixed the indentation and rebased

hosekadam:
@danmyway Do we need new integration tests for this or do we have it covered?

danmyway:
Not that critical from my POV, we can add it in a separate batch @hosekadam 
I will just run the tier0, if it passes (apart from the offline conversion test on Alma 8.9) we can merge this

Venefilyn:
LGTM. @hosekadam please update the Jira issues to Dev Complete when you can!

PR #1156, opened 2024-03-20
Introduce new log levels

r0x0d:
Intorucing four new logging levels, `prepare`, `convert`, `final` and `rollback`. Essentially, this commit introduces the task headers as log levels to leave the responsability for the text of what will be printed to the logging class instead of the string in the `task` log level.

Checklist

### DIFF
--- a/convert2rhel/backup/files.py
+++ b/convert2rhel/backup/files.py
@@ -109,7 +109,7 @@ class RestorableFile(RestorableChange):
     def restore(self, rollback=True):
         """Restore a previously backed up file"""
         if rollback:
-            loggerinst.task("Rollback: Restore %s from backup" % self.filepath)
+            loggerinst.rollback("Restore %s from backup" % self.filepath)
         else:
             loggerinst.info("Restoring %s from backup" % self.filepath)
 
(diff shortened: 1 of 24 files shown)


convert2rhel/backup/files.py
```diff
@@ -109,7 +109,7 @@ def _hash_backup_path(self):
     def restore(self, rollback=True):
         """Restore a previously backed up file"""
         if rollback:
-            loggerinst.task("Rollback: Restore %s from backup" % self.filepath)
+            loggerinst.rollback("Restore %s from backup" % self.filepath)
         else:
             loggerinst.info("Restoring %s from backup" % self.filepath)
```
SpyTec, line 114: I would much prefer something that tackles this more than having to say `Rollback: ` for us. Can we address this in the same PR?

jochapma: From a design point of view, prepare, convert, final, and rollback are phases. At any point in the execution, you're in one and only one of them, and they have other effects--implicit or explicit--on how the program behaves. So while I agree that doing something other than manually labeling messages with a string like "Rollback:" is a good thing to do, I also agree with Freya that as a part of the design this seems to be at a higher level than choosing which logger method to call.

r0x0d: This PR will be closed in favor of describing a better feature as a jira ticket. This was more like a experiment to see if that would be a good fit or not. I should have done that as part of a gist, not PR. :smile:

PR #1164, opened 2024-04-01
[RHELC-601] Install subscription-manager using package manager

r0x0d:
In this commit, we are introducing the installation of subscription-manager directly through a package manager, instead of downloading the package first and then installing it with system package manager.

This is necessary as we want to do GPG verification when installing those set of packages, in which case, it was not available when we used to download the packages and then install it via local install.

Notable changes in this commit are:
* Replacing installing subscription-manager to install it directly from the client-tools repository with gpg check enabled for all supported versions
* Minor refactor to RestorablePackageSet class to allow it to be generic and use custom repository paths and content to install packages
* Small tweak to call_yum_cmd to use the reposdir option as a list instead of a string
* Move subscription-manager constants to subscription.py module instead of the packages.py backup module

Jira Issues: [RHELC-601](https://issues.redhat.com/browse/RHELC-601)

Checklist

-----

# Depends on
- [x] https://github.com/oamg/convert2rhel/pull/1194

### DIFF
--- a/convert2rhel/repo.py
+++ b/convert2rhel/repo.py
@@ -18,11 +18,18 @@
 __metaclass__ = type
 
 import logging
+import tempfile
 
-from convert2rhel.systeminfo import system_info, tool_opts
+from contextlib import closing
+
+from six.moves import urllib
+
+from convert2rhel import exceptions
+from convert2rhel.systeminfo import system_info
+from convert2rhel.toolopts import tool_opts
+from convert2rhel.utils import TMP_DIR, store_content_to_file
 
 
-DEFAULT_YUM_REPOFILE_DIR = "/etc/yum.repos.d"
 DEFAULT_YUM_VARS_DIR = "/etc/yum/vars"
 DEFAULT_DNF_VARS_DIR = "/etc/dnf/vars"
 
@@ -78,3 +85,37 @@ def get_rhel_disable_repos_command(disable_repos):
     disable_repo_command = " ".join("--disablerepo=" + repo for repo in disable_repos)
 
     return disable_repo_command
+
+
+def download_repofile(repofile_url):
+    """Download the official repofile pointing to a convert2rhel repository.
+
+    :raises exceptions.CriticalError: ...
+    :returns str: Path to a successfully downloaded repofile
+    """
+    try:
+        with closing(urllib.request.urlopen(repofile_url, timeout=15)) as response:
+            filepath = write_temporary_repofile(contents=response.read())
+            if filepath:
+                loggerinst.info("Successfully downloaded the requested repofile from %s." % repofile_url)
+
+            return filepath
+    except urllib.error.URLError as err:
+        raise exceptions.CriticalError(
+            id_="DOWNLOAD_REPOSITORY_FILE_FAILED",
+            title="Failed to download repository file",
+            description="Failed to download the requested repository file from %s.\n"
+            "Reason: %s" % (repofile_url, err.reason),
+        )
+
+
+def write_temporary_repofile(contents):
+    """ """
+    repofile_dir = tempfile.mkdtemp(prefix="convert2rhel_repo.", dir=TMP_DIR)
+
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".repo", delete=False, dir=repofile_dir) as f:
+        try:
+            store_content_to_file(filename=f.name, content=contents)
+            return f.name
+        except (OSError, IOError) as err:
+            loggerinst.warning("OSError({0}): {1}".format(err.errno, err.strerror))
--- a/convert2rhel/repo.py
+++ b/convert2rhel/repo.py
@@ -30,6 +30,7 @@ from convert2rhel.toolopts import tool_opts
 from convert2rhel.utils import TMP_DIR, store_content_to_file
 
 
+DEFAULT_YUM_REPOFILE_DIR = "/etc/yum.repos.d"
 DEFAULT_YUM_VARS_DIR = "/etc/yum/vars"
 DEFAULT_DNF_VARS_DIR = "/etc/dnf/vars"
 
@@ -90,7 +91,7 @@ def get_rhel_disable_repos_command(disable_repos):
 def download_repofile(repofile_url):
     """Download the official repofile pointing to a convert2rhel repository.
 
-    :raises exceptions.CriticalError: ...
+    :raises exceptions.CriticalError: When the repository file can't be accessed.
     :returns str: Path to a successfully downloaded repofile
     """
     try:
--- a/convert2rhel/repo.py
+++ b/convert2rhel/repo.py
@@ -89,23 +89,30 @@ def get_rhel_disable_repos_command(disable_repos):
 
 
 def download_repofile(repofile_url):
-    """Download the official repofile pointing to a convert2rhel repository.
+    """Download a repository file from a specific URL.
 
-    :raises exceptions.CriticalError: When the repository file can't be accessed.
-    :returns str: Path to a successfully downloaded repofile
+    :raises exceptions.CriticalError: When the repository file URL is inaccessible.
+    :returns str: Contents of the repofile if successfully downloaded.
     """
     try:
         with closing(urllib.request.urlopen(repofile_url, timeout=15)) as response:
             contents = response.read()
-            filepath = write_temporary_repofile(contents=contents.decode())
-            if filepath:
-                loggerinst.info("Successfully downloaded the requested repofile from %s." % repofile_url)
 
-            return filepath
+            if not contents:
+                loggerinst.critical_no_exit("")
+                raise exceptions.CriticalError(
+                    id_="REPOSITORY_FILE_EMPTY_CONTENT",
+                    title="No content was availble in requested repository file.",
+                    description="The requested repository file seems to be empty. No content received when checking for url: %s"
+                    % repofile_url,
+                )
+
+            loggerinst.info("Successfully downloaded the requested repofile from %s." % repofile_url)
+            return contents.decode()
     except urllib.error.URLError as err:
         raise exceptions.CriticalError(
             id_="DOWNLOAD_REPOSITORY_FILE_FAILED",
-            title="Failed to download repository file",
+            title="Failed to download repository file.",
             description="Failed to download the requested repository file from %s.\n"
             "Reason: %s" % (repofile_url, err.reason),
         )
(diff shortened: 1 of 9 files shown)


SpyTec (commented):
Code seems good to me, just need to get integration tests passing. Here's hoping for a flawless test run :crossed_fingers:

convert2rhel/subscription.py
```diff
@@ -583,8 +636,6 @@ def attach_subscription():
     subscription ID has been provided through command line, let the user
     interactively choose one.
     """
-    # TODO: Support attaching multiple pool IDs.
-
```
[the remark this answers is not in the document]
r0x0d: Agreed! I can revert this change and remove in another PR.

SpyTec:
Due to the nature of the change I want to wait for Satellite and offline integration tests to run successfully beforehand

SpyTec (approved):

SpyTec:
Has merge conflicts

convert2rhel/repo.py
```diff
@@ -19,9 +19,15 @@
...
+import tempfile
 
+from contextlib import closing
+
+from six.moves import urllib
+
+from convert2rhel import exceptions
 from convert2rhel.systeminfo import system_info
-from convert2rhel.utils import DATA_DIR
+from convert2rhel.utils import DATA_DIR, TMP_DIR, store_content_to_file
```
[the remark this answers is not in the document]
r0x0d:
```suggestion
from convert2rhel.utils import TMP_DIR, store_content_to_file
```

r0x0d:
Will rebase and resolve conflicts after the tests run finishes

Venefilyn:
Tests feel unrelated, same as everywhere else. Can be rebased and conflicts fixed

r0x0d:
> Tests feel unrelated, same as everywhere else. Can be rebased and conflicts fixed

Will rebase now

convert2rhel/repo.py
```diff
@@ -41,3 +48,46 @@ def get_rhel_repoids():
     loggerinst.info("RHEL repository IDs to enable: %s" % ", ".join(repos_needed))
 
     return repos_needed
+
+
+def download_repofile(repofile_url):
+    """Download the official repofile pointing to a convert2rhel repository.
+
+    :raises exceptions.CriticalError: ...
```
r0x0d, line 56:
```suggestion
    :raises exceptions.CriticalError: When the repository file can't be accessed.
```

r0x0d:
> /packit test --labels tier0

Some of the tests will keep failing until we get this merged: https://github.com/oamg/convert2rhel/pull/1194

Venefilyn:
Theres 3 tests seem off

This one might be due to the code change
- http://artifacts.osci.redhat.com/testing-farm/cd79c6d6-21ae-48de-83ec-60315d3460a0/

And these from issues unrelated to this PR
- http://artifacts.osci.redhat.com/testing-farm/7926112c-3857-4c6f-804e-402f14ed8d71/
- http://artifacts.osci.redhat.com/testing-farm/d01c8aaa-3a6d-4ba9-ad64-14f28e87e381/

r0x0d:
Tests failure seems unrelated

convert2rhel/repo.py
```diff
@@ -47,3 +54,46 @@ def get_rhel_repoids():
     loggerinst.info("RHEL repository IDs to enable: %s" % ", ".join(repos_needed))
 
     return repos_needed
+
+
+def download_repofile(repofile_url):
+    """Download the official repofile pointing to a convert2rhel repository.
+
+    :raises exceptions.CriticalError: When the repository file can't be accessed.
+    :returns str: Path to a successfully downloaded repofile
```
bocekm, line 63:
```suggestion
    """Download a repository file from a specific URL.

    :raises exceptions.CriticalError: When the repository file URL is inaccessible.
    :returns str: Path to a repofile if successfully downloaded. None if there was a problem with storing the downloaded repofile.
```

@r0x0d, I think that the function should raise an exception in both cases:
- when unable to access the URL
- when unable to store the downloaded content

Right now the former leads to raising an exception and the latter to returning None. That makes checking the result of the function difficult.

convert2rhel/subscription.py
```diff
@@ -630,19 +645,38 @@ def install_rhel_subscription_manager(pkgs_to_install, pkgs_to_upgrade=None):
...
+    # transaction.
+    if system_info.id == "oracle" and system_info.version.major == 7:
+        setopts.append("obsoletes=0")
+
+    client_tools_repofile = None
+    if system_info.id == "centos" and system_info.version.major == 8:
+        client_tools_repofile = repo.write_temporary_repofile(_CLIENT_TOOLS_RHEL_8_5_REPO)
+    else:
+        repofile_url = _CLIENT_TOOLS_REPOFILE_MAPPING[system_info.version.major]
+        client_tools_repofile = repo.download_repofile(repofile_url)
```
bocekm, line 669: The `download_repofile` function return None if unable to store the downloaded content. With that the code below would fail with a traceback. Better to change the download_repofile to raise a CriticalError instead of returning None.

convert2rhel/subscription.py
```diff
@@ -875,7 +909,7 @@ def needed_subscription_manager_pkgs():
     return to_install_pkgs
 
 
-def _dependencies_to_update(pkg_list):
+def dependencies_to_update(pkg_list):
```
bocekm, line 912:
```suggestion
```
This function is no longer needed with the direct yum installation. The reason for this function was that the UBI repos had sub-man packages in a version from the latest RHEL minor version and those required dependencies of a version not available in older Alma/Rocky/CentOS minor versions. We now have client tool repos even for older minor versions so this dependency update hack should go away.

It's mentioned in the acceptance criteria of the Jira ticket:

> The use of the new public client-tools repository brings an important improvement - we now have access to sub-man package versions corresponding to specific RHEL minor versions. That was not the case with UBI - there we had access only to the latest version for the particular RHEL major version. With this we shall remove some of the workarounds for errors that stemmed from us installing versions of sub-man intended for a different RHEL minor version:
https://github.com/oamg/convert2rhel/pull/494
https://github.com/oamg/convert2rhel/blob/v1.6.1/convert2rhel/subscription.py#L783 

Also related:
https://github.com/oamg/convert2rhel/pull/905

bocekm (changes requested):

convert2rhel/subscription.py
```diff
@@ -630,19 +645,38 @@ def install_rhel_subscription_manager(pkgs_to_install, pkgs_to_upgrade=None):
...
+    setopts = []
+    # Oracle Linux 7 needs to set the obsoletes option to avoid installing the
+    # rhc-client-tools package instead of subscription-manager, as it is marked
+    # as obsolete in the subscription-manager specfile for OL7. This option in
+    # yum will make sure to ignore the obsoletes request and install the
+    # correct packages we are requesting.
+    # WARNING: This obsoletes will be applicable to all packages in the
+    # transaction.
+    if system_info.id == "oracle" and system_info.version.major == 7:
+        setopts.append("obsoletes=0")
```
bocekm, line 662:
```suggestion
        setopts.append("exclude=rhn-client-tools")
```
Excluding a specific package is more targeted than the broad obsoletes=0 which might cause unforeseen consequences in case Red Hat starts using Obsoletes in one of the client-tools packages. I've tested it on OL 7 - the exclude makes yum to install sub-man from the client-tools repo.

bocekm (changes requested):

convert2rhel/subscription.py
r0x0d: removed

convert2rhel/subscription.py
```diff
@@ -630,19 +645,38 @@ def install_rhel_subscription_manager(pkgs_to_install, pkgs_to_upgrade=None):
...
+
+    client_tools_repofile = None
+    if system_info.id == "centos" and system_info.version.major == 8:
+        client_tools_repofile = repo.write_temporary_repofile(_CLIENT_TOOLS_RHEL_8_5_REPO)
+    else:
+        repofile_url = _CLIENT_TOOLS_REPOFILE_MAPPING[system_info.version.major]
+        client_tools_repofile = repo.download_repofile(repofile_url)
+
+    reposdir = [os.path.dirname(client_tools_repofile), backedup_reposdir]
```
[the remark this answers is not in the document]
r0x0d: I will add a comment in the code. I don't think we always rely on checking on github for past comments/references. At least, I know that I don't. So I would prefer to have this in the code, as it is the main source that I always look first.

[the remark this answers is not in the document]
r0x0d: Added it to the docstring

convert2rhel/subscription.py
r0x0d: Okay, updated the code to raise exceptions in both `download_repofile` and `write_temporary_repofile`. Please re-review, @bocekm

convert2rhel/repo.py
r0x0d: We are not raising exceptions in both cases: 
* when unable to access the URL (URLError)
* when the contents of the response is empty

Since I split the code and removed the operation of writing from this function, I thought that we should check for empty response instead. We could go further and check for status code different from 200... Do you think that would be necessary?

convert2rhel/repo.py
```diff
@@ -47,3 +54,57 @@ def get_rhel_repoids():
...
+
+    :raises exceptions.CriticalError: When the repository file URL is inaccessible.
+    :returns str: Contents of the repofile if successfully downloaded.
+    """
+    try:
+        with closing(urllib.request.urlopen(repofile_url, timeout=15)) as response:
+            contents = response.read()
+
+            if not contents:
+                loggerinst.critical_no_exit("")
```
bocekm, line 70: I suppose you wanted to pass some text here.

convert2rhel/repo.py
```diff
@@ -47,3 +54,57 @@ def get_rhel_repoids():
...
+                    description="The requested repository file seems to be empty. No content received when checking for url: %s"
+                    % repofile_url,
+                )
+
+            loggerinst.info("Successfully downloaded the requested repofile from %s." % repofile_url)
+            return contents.decode()
+    except urllib.error.URLError as err:
+        raise exceptions.CriticalError(
+            id_="DOWNLOAD_REPOSITORY_FILE_FAILED",
+            title="Failed to download repository file.",
```
bocekm, line 83:
```suggestion
            title="Failed to download repository file",
```
We don't use full stops in titles.

convert2rhel/repo.py
```diff
@@ -47,3 +54,57 @@ def get_rhel_repoids():
...
+    """
+    repofile_dir = tempfile.mkdtemp(prefix="convert2rhel_repo.", dir=TMP_DIR)
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".repo", delete=False, dir=repofile_dir) as f:
+        try:
+            store_content_to_file(filename=f.name, content=contents)
+            return f.name
+        except (OSError, IOError) as err:
+            raise exceptions.CriticalError(
+                id_="WRITE_REPOSITORY_FILE_FAILED",
+                title="Failed to write repository file.",
```
bocekm, line 107:
```suggestion
                title="Failed to store a repository file",
```
We don't use full stops in titles.

convert2rhel/repo.py
```diff
@@ -47,3 +54,57 @@ def get_rhel_repoids():
...
+    """
+    try:
+        with closing(urllib.request.urlopen(repofile_url, timeout=15)) as response:
+            contents = response.read()
+
+            if not contents:
+                loggerinst.critical_no_exit("")
+                raise exceptions.CriticalError(
+                    id_="REPOSITORY_FILE_EMPTY_CONTENT",
+                    title="No content was availble in requested repository file.",
```
bocekm, line 73:
```suggestion
                    title="No content available in a repository file",
```
Typo + we don't use full stops in titles.

convert2rhel/repo.py
```diff
@@ -47,3 +54,57 @@ def get_rhel_repoids():
...
+        raise exceptions.CriticalError(
+            id_="DOWNLOAD_REPOSITORY_FILE_FAILED",
+            title="Failed to download repository file.",
+            description="Failed to download the requested repository file from %s.\n"
+            "Reason: %s" % (repofile_url, err.reason),
+        )
+
+
+def write_temporary_repofile(contents):
+    """Write a temporary repository file inside the :py:TMP_DIR folder.
```
bocekm, line 90:
```suggestion
    """Store a temporary repository file inside the :py:TMP_DIR folder.
```

bocekm (changes requested):

convert2rhel/repo.py
r0x0d: oops, that's right. thanks for spotting that

r0x0d: Added the same message as the exception description

bocekm (approved):

PR #1174, opened 2024-04-05
[RHELC-884] Disable RHEL repos when performing checks

hosekadam:
We need to disable RHEL repos. If user registers the system to satellite which provides Centos and RHEL repos and the RHEL repos would be enabled, it would cause a fail. The recommended way is to disable the rhel repos prior the conversion, but is expected it might be forgotten. This helps avoiding the problems caused by wrong setup of the repos.

Jira Issues: https://issues.redhat.com/browse/RHELC-884

-

Checklist

### DIFF
--- a/convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
+++ b/convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
@@ -18,7 +18,7 @@ __metaclass__ = type
 import logging
 import os
 
-from convert2rhel import actions
+from convert2rhel import actions, repo
 from convert2rhel.pkghandler import compare_package_versions
 from convert2rhel.systeminfo import system_info
 from convert2rhel.utils import run_subprocess
@@ -44,10 +44,14 @@ class IsLoadedKernelLatest(actions.Action):
             )
             return
 
+        # RHELC-884 disable the RHEL repos to avoid reaching them when checking original system.
+        disable_repo_command = repo.get_rhel_disable_repos_command(repo.get_rhel_repos_to_disable())
+
         cmd = [
             "repoquery",
             "--setopt=exclude=",
             "--quiet",
+            disable_repo_command,
             "--qf",
             "C2R\\t%{BUILDTIME}\\t%{VERSION}-%{RELEASE}\\t%{REPOID}",
         ]
--- a/convert2rhel/backup/packages.py
+++ b/convert2rhel/backup/packages.py
@@ -20,7 +20,7 @@ __metaclass__ = type
 import logging
 import os
 
-from convert2rhel import exceptions, utils
+from convert2rhel import exceptions, repo, utils
 from convert2rhel.backup import BACKUP_DIR, RestorableChange
 from convert2rhel.pkgmanager import call_yum_cmd
 
@@ -83,7 +83,9 @@ _UBI_REPO_MAPPING = {
 
 # NOTE: Over time we want to replace this with pkghandler.RestorablePackageSet.
 class RestorablePackage(RestorableChange):
-    def __init__(self, pkgs, reposdir=None, set_releasever=False, custom_releasever=None, varsdir=None):
+    def __init__(
+        self, pkgs, reposdir=None, set_releasever=False, custom_releasever=None, varsdir=None, disable_repos=None
+    ):
         """
         Keep control of system packages before their removal to backup and
         restore in case of rollback.
@@ -105,6 +107,9 @@ class RestorablePackage(RestorableChange):
         self.set_releasever = set_releasever
         self.custom_releasever = custom_releasever
         self.varsdir = varsdir
+
+        # RHELC-884 disable the RHEL repos to avoid downloading pkg from them.
+        self.disable_repos = disable_repos or repo.get_rhel_repos_to_disable()
 
         self._backedup_pkgs_paths = []
 
@@ -149,6 +154,7 @@ class RestorablePackage(RestorableChange):
                 utils.download_pkg(
                     pkg=pkg,
                     dest=BACKUP_DIR,
+                    disable_repos=self.disable_repos,
                     set_releasever=self.set_releasever,
                     custom_releasever=self.custom_releasever,
                     varsdir=self.varsdir,
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -27,7 +27,7 @@ from collections import namedtuple
 
 import rpm
 
-from convert2rhel import backup, pkgmanager, utils
+from convert2rhel import backup, pkgmanager, repo, utils
 from convert2rhel.backup.certs import RestorableRpmKey
 from convert2rhel.backup.files import RestorableFile
 from convert2rhel.systeminfo import system_info
@@ -299,11 +299,13 @@ def get_installed_pkgs_w_different_fingerprint(fingerprints, name="*"):
 
 
 @utils.run_as_child_process
-def format_pkg_info(pkgs):
+def format_pkg_info(pkgs, disable_repos=None):
     """Format package information.
 
     :param pkgs: List of packages to be formatted
     :type pkgs: list[PackageInformation] | list[RPMInstalledPackage]
+    :param disable_repos: List of repo IDs to be disabled when retrieving repository information from packages.
+    :type disable_repos: List[str]
     """
     package_info = {}
     for pkg in pkgs:
@@ -344,7 +346,7 @@ def format_pkg_info(pkgs):
         + "\n"
     )
 
-    packages_with_repos = _get_package_repositories(list(package_info))
+    packages_with_repos = _get_package_repositories(list(package_info), disable_repos)
     # Update package_info reference with repoid
     for nevra, repoid in packages_with_repos.items():
         package_info[nevra]["repoid"] = repoid
@@ -368,22 +370,26 @@ def format_pkg_info(pkgs):
     return pkg_table
 
 
-def print_pkg_info(pkgs):
+def print_pkg_info(pkgs, disable_repos=None):
     """
     Print the results of format_pkg_info
 
     :param pkgs: List of packages to be printed
     :type pkgs: list[PackageInformation] | list[RPMInstalledPackage]
+    :param disable_repos: List of repo IDs to be disabled when retrieving repository information from packages.
+    :type disable_repos: List[str]
     """
 
-    loggerinst.info(format_pkg_info(pkgs))
+    loggerinst.info(format_pkg_info(pkgs, disable_repos))
 
 
-def _get_package_repositories(pkgs):
+def _get_package_repositories(pkgs, disable_repos=None):
     """Retrieve repository information from packages.
 
     :param pkgs: List of packages to get their associated repositories
     :type pkgs: list[PackageInformation]
+    :param disable_repos: List of repo IDs to be disabled when retrieving repository information from packages.
+    :type disable_repos: List[str]
     :return: Mapping of packages with their repositories names
     :rtype: dict[str, dict[str, str]
     """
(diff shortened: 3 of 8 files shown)


r0x0d:
@hosekadam, looking at how we are disabling the repos, maybe we should use `rhel*` instead of `rhel-*`? 

I'm asking this more because if RHEL introduces a new repo, or modifies the name at some point to have something like: `rhel9-something-foo-bar`, and this repository is enabled, then we will miss this repository by the time we are running our tools.

hosekadam:
@r0x0d Thanks for taking a look! I used `rhel-*` based on the names of the repos, But `rhel*` might be more generic and future-proof. I'll switch to that.

convert2rhel/pkghandler.py
```diff
@@ -918,7 +918,16 @@ def _get_packages_to_update_yum():
     :rtype: list[str] | list
     """
     all_packages = []
+    # Disable the rhel repositories when checking if the packages are up to date.
+    # When user specifies rhel repos to use during conversion, disabling them is also needed.
+    # Note: Do not depend on the --no-rhsm option, enable repos can be specified without it.
+    disable_repos = ["rhel-*"] + tool_opts.enablerepo
```
r0x0d, line 924:
```suggestion
    disable_repos = ["rhel*"] + tool_opts.enablerepo
```

convert2rhel/pkghandler.py
```diff
@@ -935,6 +944,10 @@ def _get_packages_to_update_dnf(reposdir):
     :type reposdir: str | None
     """
     packages = []
+    # Disable the rhel repositories when checking if the packages are up to date.
+    # When user specifies rhel repos to use during conversion, disabling them is also needed.
+    # Note: Do not depend on the --no-rhsm option, enable repos can be specified without it.
+    disable_repos = ["rhel-*"] + tool_opts.enablerepo
```
r0x0d, line 950:
```suggestion
    disable_repos = ["rhel*"] + tool_opts.enablerepo
```

convert2rhel/actions/system_checks/is_loaded_kernel_latest.py
```diff
@@ -48,6 +48,7 @@ def run(self):  # pylint: disable= too-many-return-statements
             "repoquery",
             "--setopt=exclude=",
             "--quiet",
+            "--disablerepo=rhel-*",
```
r0x0d, line 51:
```suggestion
            "--disablerepo=rhel*",
```

hosekadam: Thanks @r0x0d!

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -980,6 +980,105 @@ def test_get_packages_to_update_dnf(packages, reposdir, pretend_os, monkeypatch)
...
+    assert repo_for_disable not in enabled_repos
+
+
+@pytest.mark.skipif(
+    pkgmanager.TYPE != "yum",
+    reason="No yum module detected on the system, skipping it.",
+)
+def test_get_packages_to_update_yum_rhel_repos(monkeypatch):
+    # Import the YUM inside the test. If YUM is not present the test isn't run
+    import yum
```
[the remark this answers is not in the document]
hosekadam: I'll take a look, there is one more problem with mocking the `doPackageList` - the tests failed. It was just a test if that `yum.repos.Repository` works :D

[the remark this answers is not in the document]
hosekadam: Now it's working, the tests `test_get_packages_to_update_yum_rhel` and `test_get_packages_to_update_dnf_rhel` are ready @r0x0d

hosekadam:
So from the last time:
- unit tests for check if system is up-to-date were added for both yum and dnf
- changed the disable pattern from `rhel-*` to `rhel*`
- added the `--disablerepo=rhel*` into the unit tests for latest kernel
- @bocekm found there can be problem during backup, when the RHEL repos are enabled and the pkg to download has the same NEVRA in original vendor and rhel repos. There is possibility the RHEL one would be downloaded as backup of some pkg we need to backup (we are backuping just the special packages we are removing). I was looking also at the `RestorablePackageSet` but I don't think we wan't to disable RHEL repos there, I think they are needed in this place.

I'm not sure if there is requirement to add unit tests for `RestorablePackage.enable()` (right now we are just checking if `utils.download_pkg` was called) and for `IsLoadedKernelLatest.run()` (there I added the argument to the list of expected args).

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -980,6 +980,104 @@ def test_get_packages_to_update_dnf(packages, reposdir, pretend_os, monkeypatch)
...
+    # Get all enabled repos in the current system
+    base.read_all_repos()
+    enabled_repos = [repo.id for repo in base.repos.iter_enabled()]
+
+    # Check if the list is not empty and the RHEL repo is enabled
+    assert enabled_repos
+    assert rhel_repo_id in enabled_repos
+
+    # Choose one of the enabled repos on the system
+    # Avoid getting the rhel_repo_id if on 0 index, then get the repo id from 1
```
SpyTec, line 1011: Why?

convert2rhel/unit_tests/pkghandler_test.py
```diff
@@ -980,6 +980,104 @@ def test_get_packages_to_update_dnf(packages, reposdir, pretend_os, monkeypatch)
...
+    repo_for_disable = [enabled_repos[0]] if enabled_repos[0] != rhel_repo_id else [enabled_repos[1]]
+
+    # Use the previously chosen repo for disabling by the user
+    # When user adds a repo through the --enable-repo option, it should be disabled
+    monkeypatch.setattr(tool_opts, "enablerepo", repo_for_disable)
+
+    # Patch the original YUM Base to ours to have access to it and contain the added repo
+    monkeypatch.setattr(pkgmanager, "YumBase", mock.Mock(return_value=base))
```
SpyTec, line 1069: Good approach, that means we still test with the Yum/DnfBase still and not just plainly mocking it out

SpyTec (commented):
LGTM, but confused about the skip if index 0 part

convert2rhel/unit_tests/pkghandler_test.py
hosekadam: I'm trying to disable one of the repos available on the system to disabling repo provided through the tool_opts (`monkeypatch.setattr(tool_opts, "enablerepo", repo_for_disable)`). I want to have something different than the added RHEL repo but still present on the system. 

So I'm getting it in this way - I know if on the 0 index is rhel repo, on the 1 will be something different I can use

convert2rhel/unit_tests/pkghandler_test.py
hosekadam: Yes, exactly

convert2rhel/pkghandler.py
```diff
@@ -935,6 +944,10 @@ def _get_packages_to_update_dnf(reposdir):
     :type reposdir: str | None
     """
     packages = []
+    # Disable the rhel repositories when checking if the packages are up to date.
+    # When user specifies rhel repos to use during conversion, disabling them is also needed.
+    # Note: Do not depend on the --no-rhsm option, enable repos can be specified without it.
+    disable_repos = ["rhel*"] + tool_opts.enablerepo
```
r0x0d, line 950: One tip, that may not be 100% necessary here (and definitely not a blocker), but you could move the disable_repos to be a parameter in the function, this way you only have to identify the rhel repos + tool_opts.enablerepo once in the caller, and pass this down to the functions.

r0x0d (approved):
Looks good!

convert2rhel/unit_tests/pkghandler_test.py
SpyTec: I'd write that out a bit clearer in the text, avoid getting something is clear but not super clear in the comments of why you do it

hosekadam:
- rebased

convert2rhel/pkghandler.py
hosekadam: Yeah, it would help with decreasing duplicity of code. I'll do it, I like the idea. And as I'm looking on it, I see I forgot to disable the `tool_opts.enablerepo` on other places.

hosekadam:
@pr-watson needs to try something. The PR is now WIP

convert2rhel/backup/packages.py
```diff
@@ -143,11 +144,15 @@ def enable(self):
                 loggerinst.info("The repository directory %s seems to be empty or non-existent.")
                 self.reposdir = None
 
+        disable_repos = ["rhel*"] + tool_opts.enablerepo
```
r0x0d, line 147: I think we should move this to the __init__ to allow the caller to override the disable_repos if needed. This way, we will always assume that rhel*+ enablerepo will be disabled for all package backup, regardless of what we choose. 

Maybe we could improve it like this: 

```python
def __init__(self, pkgs, reposdir=None, set_releasever=False, custom_releasever=None, varsdir=None, disable_repos=None):
        """
        Keep control of system packages before their removal to backup and
        restore in case of rollback.
        :param pkgs list[str]: List of packages to backup.
        :param reposdir str: If a custom repository directory location needs to
            be used, this parameter can be set with the location for it.
        :param set_releasever bool: If there is need to set the relesever while
            downloading the package with yumdownloader.
        :param custom_releasever str: Custom releasever in case it need to be
            overwritten and it differs from the `py:system_info.releasever`.
        :param varsdir str: Location to the variables directory in case the
            repository files needs to interpolate variables from those folders.
        :param disable_repos list[str]: Repositories to disable during package download. Defaults to None.
        """
        super(RestorablePackage, self).__init__()
        self.pkgs = pkgs
        self.reposdir = reposdir
        self.set_releasever = set_releasever
        self.custom_releasever = custom_releasever
        self.varsdir = varsdir
        self.disable_repos = disable_repos or ["rhel*"] + tool_opts.enablerepo
        self._backedup_pkgs_paths = []
```

Then later while we are downloading the packages, it would use it like this: 

```python
        for pkg in self.pkgs:
            self._backedup_pkgs_paths.append(
                utils.download_pkg(
                    pkg=pkg,
                    dest=BACKUP_DIR,
                    # RHELC-884 disable the RHEL repos to avoid downloading pkg from them
                    disable_repos=self.disable_repos,
                    set_releasever=self.set_releasever,
                    custom_releasever=self.custom_releasever,
                    varsdir=self.varsdir,
                    reposdir=self.reposdir,
                )
            )
```

convert2rhel/pkghandler.py
```diff
@@ -900,25 +900,35 @@ def get_total_packages_to_update(reposdir):
...
 
 
-def _get_packages_to_update_yum():
+def _get_packages_to_update_yum(disable_repos=[]):
     """Use yum to get all the installed packages that have an update available.
 
     :return: Return a list of packages that needs to be updated.
     :rtype: list[str] | list
     """
     all_packages = []
```
r0x0d, line 924:
```suggestion
def _get_packages_to_update_yum(disable_repos=None):
    """Use yum to get all the installed packages that have an update available.

    :return: Return a list of packages that needs to be updated.
    :rtype: list[str] | list
    """
     disable_repos = disable_repos or []
    all_packages = []
```

hosekadam:
- created `repo.get_rhel_repos_to_disable()` to reduce duplicity
- moved preparation of the disablerepo list higher in the chain in many places to reduce duplicity

TBD:
- one bug waiting to be solved in `_get_package_repositories`, the repoquery should contain the `"--disablerepo=rhel*"` (pushing the rest for having backup), tested with it, needs unit tests and better implementation
- solve the conflicts :(

hosekadam:
- rebased

hosekadam:
-  _get_package_repositories, the repoquery now contains the "--disablerepo=rhel*"

convert2rhel/unit_tests/pkghandler_test.py
hosekadam: Comment about it added.

convert2rhel/pkghandler.py
```diff
@@ -393,8 +393,11 @@ def _get_package_repositories(pkgs):
     if system_info.version.major >= 8:
         query_format = "C2R %{NAME}-%{EPOCH}:%{VERSION}-%{RELEASE}.%{ARCH}&%{REPOID}\n"
 
+    # RHELC-884 disable the RHEL repos to avoid downloading pkgs from them.
+    disable_repo_command = repo.get_rhel_disable_repos_command()
```
hosekadam, line 397: This is part of discussion. WIP

hosekadam:
- refactor how the disable_repos is passed
- disabled RHEL repos when retrieving repo info from pkgs in `RemoveSpecialPkgs` and `ListThirdPartyPackages`

PR #1181, opened 2024-04-12
Drop internet connection check and its dependencies

r0x0d:
In this commit, we are dropping the internet connection check and its dependencies as well.

Some of them are not necessary, as we really want to use the system repos backed up during the conversion, and not some hardcoded repositories that we provide.

Historically, we used that for eus conversions, but right now it is not necessary anymore.

The internet connection check brought a lot of difficulties in the past and several bug reports. With this in mind, we decided to drop the check, and assume the system will be configured in the correct way.

Jira Issues:

-

Checklist

- [ ] PR has been tested manually in a VM (either author or reviewer)
- [ ] Jira issue has been made public if possible
- [ ] `[RHELC-]` is part of the PR title 
- [ ] GitHub label has been added to help with Release notes 
- [ ] PR title explains the change from the user's point of view
- [ ] Code and tests are documented properly
- [ ] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending` if relevant

PR #1182, opened 2024-04-12
[RHELC-1385] Satellite package backup failure fix

pr-watson:
This PR changes our backups to now backup the redhat.repo file when running with satellite. The reason backups were failing on satellite was due to the redhat.repo file being excluded from backups which in turn causes the .repo files not to be backed up. This change also removes the requirement for internet access to backup using hardcoded repository files and defaults to always using the the hardcoded repository files. 

A similiar failure with backing up repo files has been reported for offline conversions with satellite - [RHELC-1040](https://issues.redhat.com/browse/RHELC-1040), however this PR does not fix the issue

Jira Issues:

- [RHELC-1385](https://issues.redhat.com/browse/RHELC-1385)

Checklist

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/backup_system.py
+++ b/convert2rhel/actions/pre_ponr_changes/backup_system.py
@@ -89,11 +89,11 @@ class BackupRepository(actions.Action):
             # backing up redhat.repo so repo files are properly backed up when doing satellite conversions
 
             if not repo.endswith(".repo"):
-                loggerinst.info("No .repo files backed up.")
+                loggerinst.info("Skipping backup as file is not a repository file.")
                 return
 
             if not subscription.should_subscribe() and repo == "redhat.repo":
-                loggerinst.info("No .repo files backed up.")
+                loggerinst.info("Skipping backup of redhat.repo as it is not needed.")
                 return
 
             repo_path = os.path.join(DEFAULT_YUM_REPOFILE_DIR, repo)
--- a/convert2rhel/actions/pre_ponr_changes/backup_system.py
+++ b/convert2rhel/actions/pre_ponr_changes/backup_system.py
@@ -82,8 +82,8 @@ class BackupRepository(actions.Action):
 
         loggerinst.info("Backing up .repo files from %s." % DEFAULT_YUM_REPOFILE_DIR)
 
-        if os.listdir(DEFAULT_YUM_REPOFILE_DIR) == []:
-            loggerinst.info("No .repo files to back up.")
+        if not os.listdir(DEFAULT_YUM_REPOFILE_DIR):
+            loggerinst.info("Repository folder %s seems to be empty.", DEFAULT_YUM_REPOFILE_DIR)
 
         for repo in os.listdir(DEFAULT_YUM_REPOFILE_DIR):
             # backing up redhat.repo so repo files are properly backed up when doing satellite conversions
--- a/convert2rhel/actions/pre_ponr_changes/backup_system.py
+++ b/convert2rhel/actions/pre_ponr_changes/backup_system.py
@@ -89,10 +89,10 @@ class BackupRepository(actions.Action):
                 loggerinst.info("No .repo files backed up.")
                 return
 
-            if not subscription.should_subscribe and repo == "redhat.repo":
+            if not subscription.should_subscribe() and repo == "redhat.repo":
                 loggerinst.info("No .repo files backed up.")
-
                 return
+
             repo_path = os.path.join(DEFAULT_YUM_REPOFILE_DIR, repo)
             restorable_file = RestorableFile(repo_path)
             backup.backup_control.push(restorable_file)
--- a/convert2rhel/actions/pre_ponr_changes/backup_system.py
+++ b/convert2rhel/actions/pre_ponr_changes/backup_system.py
@@ -19,7 +19,7 @@ import logging
 import os
 import re
 
-from convert2rhel import actions, backup, exceptions, repo
+from convert2rhel import actions, backup, exceptions, repo, subscription
 from convert2rhel.backup.files import MissingFile, RestorableFile
 from convert2rhel.logger import LOG_DIR
 from convert2rhel.redhatrelease import os_release_file, system_release_file
@@ -89,11 +89,12 @@ class BackupRepository(actions.Action):
         repo_files_backed_up = False
 
         for repo in os.listdir(DEFAULT_YUM_REPOFILE_DIR):
-            if repo.endswith(".repo") and repo != "redhat.repo":
-                repo_path = os.path.join(DEFAULT_YUM_REPOFILE_DIR, repo)
-                restorable_file = RestorableFile(repo_path)
-                backup.backup_control.push(restorable_file)
-                repo_files_backed_up = True
+            if not subscription.should_subscribe:
+                if repo.endswith(".repo") and repo != "redhat.repo":
+                    repo_path = os.path.join(DEFAULT_YUM_REPOFILE_DIR, repo)
+                    restorable_file = RestorableFile(repo_path)
+                    backup.backup_control.push(restorable_file)
+                    repo_files_backed_up = True
 
         if not repo_files_backed_up:
             loggerinst.info("No .repo files backed up.")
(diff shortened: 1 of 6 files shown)


convert2rhel/actions/system_checks/duplicate_packages.py
```diff
@@ -31,7 +31,7 @@ def run(self):
         super(DuplicatePackages, self).run()
 
         logger.task("Prepare: Check if there are any duplicate installed packages on the system")
-        output, _ = utils.run_subprocess(["/usr/bin/package-cleanup", "--dupes", "--quiet"], print_output=False)
+        output, ret = utils.run_subprocess(["/usr/bin/package-cleanup", "--dupes", "--quiet"], print_output=False)
```
SpyTec, line 34: This is unrelated right? If `ret` isn't used I'd vote for `_`

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -19,7 +19,7 @@
 import os
 import re
 
-from convert2rhel import actions, backup, exceptions, repo
+from convert2rhel import actions, backup, exceptions, repo, subscription
```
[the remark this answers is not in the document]
SpyTec: Can be fixed

convert2rhel/backup/packages.py
```diff
@@ -134,37 +134,20 @@ def enable(self):
...
-                for pkg in self.pkgs:
-                    self._backedup_pkgs_paths.append(
-                        utils.download_pkg(
-                            pkg=pkg,
-                            dest=BACKUP_DIR,
-                            set_releasever=self.set_releasever,
-                            custom_releasever=self.custom_releasever,
-                            varsdir=self.varsdir,
-                        )
-                    )
```
SpyTec, line 152: Do we need to do this? In either case, it's being handled in #1181 so whichever gets merged first will have to fix

convert2rhel/actions/system_checks/duplicate_packages.py
r0x0d: So, I was discussing this with Preston last week and I asked him to check for return code in this command instead of just strings.. I think the `ret` is for him to test if there's any return code in case of failure that we can use to identify this specific problem

pr-watson: Yes that was leftover from testing, i'll revert it to _

convert2rhel/backup/packages.py
pr-watson: @bocekm said that we need to use repository files regardless if there is internet or not, so I removed the internet access dependency here

r0x0d: That's fine, either one that get merged first, we can deal with the rebase in the other PR.

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -87,9 +87,10 @@ def backup_yum_repos(self):
         loggerinst.info("Backing up .repo files from %s." % DEFAULT_YUM_REPOFILE_DIR)
 
         repo_files_backed_up = False
-
         for repo in os.listdir(DEFAULT_YUM_REPOFILE_DIR):
-            if repo.endswith(".repo") and repo != "redhat.repo":
+            if (repo.endswith(".repo") and repo != "redhat.repo") or (
+                subscription.should_subscribe and repo == "redhat.repo"
+            ):
```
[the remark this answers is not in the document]
r0x0d: What about rewriting the loop like this? 

```python
        for repo in os.listdir(DEFAULT_YUM_REPOFILE_DIR):
            # backing up redhat.repo so repo files are properly backed up when doing satellite conversions

            if not repo.endswith(".repo"):
                # Maybe log that the file is not repo?
                return

            if not subscription.should_subscribe and repo == "redhat.repo":
                return

            repo_path = os.path.join(DEFAULT_YUM_REPOFILE_DIR, repo)
            restorable_file = RestorableFile(repo_path)
            backup.backup_control.push(restorable_file)
            repo_files_backed_up = True

```

Trying to simplify this check down to two things: 
1. If it does not end with `.repo`, we can bail out as it doesn't interest to us
2. If `should_subscribe` is false, and repo is `redhat.repo`, then we return
3. Otherwise, we back it up. By default, I believe `should_subscribe` is `False`, unless we detect the system is registered before the analysis/conversion. 


Does it make sense to you, @pr-watson and @SpyTec?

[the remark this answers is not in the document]
r0x0d: Also, if possible, can we move the contents of this function (`backup_yum_repos`) to be inside the `run` method? I don't think this function is long enough to need an function like this, especially because `run` only calls this and exit. 

To me, it would be more clean if `run` performs the operations instead of calling the intermediate function.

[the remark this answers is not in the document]
pr-watson: It looks much cleaner now - I agree with this change. And I don't mind moving backup_yum_repos into the run method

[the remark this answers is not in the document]
SpyTec: Looks a lot better, thanks @r0x0d for the suggestion!

SpyTec:
Unit tests failing @pr-watson

pr-watson:
> Unit tests failing @pr-watson

Working on the fixes now

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -75,28 +75,30 @@ class BackupRepository(actions.Action):
...
         super(BackupRepository, self).run()
 
-        self.backup_yum_repos()
-
-    def backup_yum_repos(self):
-        """Backup .repo files in /etc/yum.repos.d/ so the repositories can be restored on rollback."""
         loggerinst.info("Backing up .repo files from %s." % DEFAULT_YUM_REPOFILE_DIR)
 
-        repo_files_backed_up = False
+        if os.listdir(DEFAULT_YUM_REPOFILE_DIR) == []:
```
r0x0d, line 85:
```suggestion
        if not os.listdir(DEFAULT_YUM_REPOFILE_DIR):
```

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -75,28 +75,30 @@ class BackupRepository(actions.Action):
...
+            loggerinst.info("No .repo files to back up.")
```
r0x0d, line 86: What if we used a different message? Maybe something like: 

```suggestion
            loggerinst.info("Repository folder %s seems to be empty.", DEFAULT_YUM_REPOFILE_DIR)
```

r0x0d: Just to differentiate from the other log message when a file does not end in .repo

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -75,28 +75,30 @@ class BackupRepository(actions.Action):
...
-        if not repo_files_backed_up:
-            loggerinst.info("No .repo files backed up.")
+            # backing up redhat.repo so repo files are properly backed up when doing satellite conversions
+
+            if not repo.endswith(".repo"):
+                loggerinst.info("No .repo files backed up.")
+                return
+
+            if not subscription.should_subscribe() and repo == "redhat.repo":
+                loggerinst.info("No .repo files backed up.")
```
r0x0d, line 96: Maybe improving this message here as well

r0x0d (commented):
Looks good to me. Only a couple of comments regarding the message we want to inform to the user.

r0x0d (approved):
Looks good!

danmyway:
Just tested the build with a Satellite registration and the backup works fine.

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -75,28 +75,30 @@ class BackupRepository(actions.Action):
...
-                backup.backup_control.push(restorable_file)
-                repo_files_backed_up = True
-
-        if not repo_files_backed_up:
-            loggerinst.info("No .repo files backed up.")
+            # backing up redhat.repo so repo files are properly backed up when doing satellite conversions
+
+            if not repo.endswith(".repo"):
+                loggerinst.info("Skipping backup as file is not a repository file.")
+                return
```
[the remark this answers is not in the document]
pr-watson: Good catch! Let me change this

danmyway (approved):
Looks good to me, thanks!

danmyway:
@pr-watson I'm merging this
@r0x0d this might resolve the issues in #1181, if you may pull the changes from @kokesak and rebase, so we can see, how that's going.
Thanks!

pr-watson:
@danmyway thank you!

PR #1189, opened 2024-04-23
Move common pytest arguments to config file

pholica:
Remove duplicity in pytest command line arguments.


danmyway (approved):
That is much better, thanks @pholica !

pytest.ini
```diff
@@ -1,4 +1,5 @@
 [pytest]
+addopts = --setup-show -svv
```
r0x0d, line 2: This is causing a lot of spam in our unit tests output report. I will prepare a PR to separate the configs

PR #1192, opened 2024-04-24
[RHELC-1515] Fix the enabled repository check after the conversion

kokesak:
We found several problems with the enabled repository check. This PR should fix those plus some minor code improvement changes.

Jira Issues: https://issues.redhat.com/browse/RHELC-1515

-

Checklist

- [ ] PR has been tested manually in a VM (either author or reviewer)
- [ ] Jira issue has been made public if possible
- [ ] `[RHELC-1515]` is part of the PR title 
- [ ] GitHub label has been added to help with Release notes 
- [ ] PR title explains the change from the user's point of view
- [ ] Code and tests are documented properly
- [ ] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending` if relevant

PR #1194, opened 2024-04-24
[RHELC-1314, RHELC-1524, RHELC-1387] Drop the use of the katello package from the tests

danmyway:
* do not use the katello-ca-consumer package for registration to the Satellite since it will be deprecated in the near future
* use the API call to register instead

Jira Issues:

- [RHELC-1314](https://issues.redhat.com/browse/RHELC-1314)
- [RHELC-1524](https://issues.redhat.com/browse/RHELC-1524)
- [RHELC-1387](https://issues.redhat.com/browse/RHELC-1524)

Checklist

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -727,41 +727,48 @@ def yum_conf_exclude(shell, backup_directory, request):
 
         def test_function(yum_conf_exclude):
     """
-    exclude = ["rhn-client*"]
-    if hasattr(request, "param"):
-        exclude = request.param
-        print(">>> Using parametrized packages requested in the fixture.")
-    yum_config = "/etc/yum.conf"
-    backup_dir = os.path.join(backup_directory, "yumconf")
-    shell(f"mkdir -v {backup_dir}")
-    config_bak = os.path.join(backup_dir, os.path.basename(yum_config))
-    config = configparser.ConfigParser()
-    config.read(yum_config)
+    # We need to do this only for Oracle Linux
+    # rhn* is one of the excluded packages in convert2rhel configs
+    # so if the package is present on the system, is excluded in yum.conf
+    # and convert2rhel tries to back it up by calling yumdownloader
+    # it fails to do so because the call conflicts with the exclude option
+    if "oracle" in SYSTEM_RELEASE_ENV:
+        exclude = ["rhn-client*"]
+        if hasattr(request, "param"):
+            exclude = request.param
+            print(">>> Using parametrized packages requested in the fixture.")
+        yum_config = "/etc/yum.conf"
+        backup_dir = os.path.join(backup_directory, "yumconf")
+        shell(f"mkdir -v {backup_dir}")
+        config_bak = os.path.join(backup_dir, os.path.basename(yum_config))
+        config = configparser.ConfigParser()
+        config.read(yum_config)
 
-    assert shell(f"cp -v {yum_config} {config_bak}").returncode == 0
+        assert shell(f"cp -v {yum_config} {config_bak}").returncode == 0
 
-    pkgs_to_exclude = " ".join(exclude)
-    # If there is already an `exclude` section, append to the existing value
-    if config.has_option("main", "exclude"):
-        pre_existing_value = config.get("main", "exclude")
-        config.set("main", "exclude", f"{pre_existing_value} {pkgs_to_exclude}")
-    else:
-        config.set("main", "exclude", pkgs_to_exclude)
+        pkgs_to_exclude = " ".join(exclude)
+        # If there is already an `exclude` section, append to the existing value
+        if config.has_option("main", "exclude"):
+            pre_existing_value = config.get("main", "exclude")
+            config.set("main", "exclude", f"{pre_existing_value} {pkgs_to_exclude}")
+        else:
+            config.set("main", "exclude", pkgs_to_exclude)
 
-    with open(yum_config, "w") as configfile:
-        config.write(configfile, space_around_delimiters=False)
+        with open(yum_config, "w") as configfile:
+            config.write(configfile, space_around_delimiters=False)
 
-    assert config.has_option("main", "exclude")
-    assert all(pkg in config.get("main", "exclude") for pkg in exclude)
+        assert config.has_option("main", "exclude")
+        assert all(pkg in config.get("main", "exclude") for pkg in exclude)
 
     yield
 
-    # Clean up only for non-destrucive tests
-    # We need to keep the yum.conf in the descructive tests, since restoring the original
-    # version has the discoverpkg option set to centos|oracle|alma|rocky-release trying to read
-    # the system information from the then non-existent file
-    if "C2R_TESTS_NONDESTRUCTIVE" in os.environ:
-        assert shell(f"mv {config_bak} {yum_config}").returncode == 0
+    if "oracle" in SYSTEM_RELEASE_ENV:
+        # Clean up only for non-destrucive tests
+        # We need to keep the yum.conf in the descructive tests, since restoring the original
+        # version has the discoverpkg option set to centos|oracle|alma|rocky-release trying to read
+        # the system information from the then non-existent file
+        if "C2R_TESTS_NONDESTRUCTIVE" in os.environ:
+            assert shell(f"mv {config_bak} {yum_config}").returncode == 0
 
 
 def _add_client_tools_repo(shell):
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -711,7 +711,6 @@ def kernel(shell):
 
 
 @pytest.fixture()
-@pytest.mark.yum_conf_exclude
 def yum_conf_exclude(shell, backup_directory, request):
     """
     Fixture.
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -45,17 +45,17 @@ SAT_REG_COMMAND = {
 }
 
 
-def satellite_registration_command():
+def satellite_curl_command():
     """
     Get the Satellite registration command for the respective system.
     """
-    sat_reg_command = None
+    sat_curl_command = None
     try:
-        sat_reg_command = SAT_REG_COMMAND[SYSTEM_RELEASE_ENV]
+        sat_curl_command = SAT_REG_COMMAND[SYSTEM_RELEASE_ENV]
     except KeyError:
         print(f"Key not found in satellite registration command dictionary: {SYSTEM_RELEASE_ENV}")
 
-    return sat_reg_command
+    return sat_curl_command
 
 
 SYSTEM_RELEASE_ENV = os.environ["SYSTEM_RELEASE_ENV"]
@@ -402,15 +402,14 @@ def required_packages(shell):
 
 
 @pytest.fixture(scope="function")
-def remove_repositories(shell):
+def remove_repositories(shell, backup_directory):
     """
     Fixture.
     Move all repositories to another location.
     """
-    backup_dir = "/tmp/test-backup-release_backup"
+    backup_dir = backup_directory
 
     # Move all repos to other location, so it is not being used
-    shell(f"mkdir {backup_dir}")
     assert shell(f"mv /etc/yum.repos.d/* {backup_dir}").returncode == 0
     assert len(os.listdir("/etc/yum.repos.d/")) == 0
 
@@ -418,7 +417,6 @@ def remove_repositories(shell):
 
     # Return repositories to their original location
     assert shell(f"mv {backup_dir}/* /etc/yum.repos.d/").returncode == 0
-    assert shell(f"rm -rf {backup_dir}")
 
 
 @pytest.fixture(autouse=True)
@@ -432,12 +430,12 @@ def missing_os_release_package_workaround(shell):
     yield
 
     os_to_pkg_mapping = {
-        "oracle-7": "oracle*-release-el7",
-        "oracle-8": "oraclelinux-release-el8",
-        "centos-7": "centos-release",
-        "centos-8": "centos-linux-release",
-        "alma-8": "almalinux-release",
-        "rocky-8": "rocky-release",
+        "oracle-7": ("oraclelinux-release-el7", "oraclelinux-release"),
+        "oracle-8": ("oraclelinux-release-el8", "oraclelinux-release"),
+        "centos-7": ("centos-release",),
+        "centos-8": ("centos-linux-release",),
+        "alma-8": ("almalinux-release",),
+        "rocky-8": ("rocky-release",),
     }
 
     # Run only for non-destructive tests.
@@ -447,11 +445,12 @@ def missing_os_release_package_workaround(shell):
         os_ver = SYSTEM_RELEASE_ENV.split("-")[1]
         os_key = f"{os_name}-{os_ver[0]}"
 
-        system_release_pkg = os_to_pkg_mapping.get(os_key)
+        system_release_pkgs = os_to_pkg_mapping.get(os_key)
 
-        rpm_output = shell(f"rpm -qa {system_release_pkg}").output
-        if "not installed" in rpm_output:
-            shell(f"yum install -y --releasever={os_ver} {system_release_pkg}")
+        for pkg in system_release_pkgs:
+            rpm_output = shell(f"rpm -q {pkg}").output
+            if "not installed" in rpm_output:
+                shell(f"yum install -y --releasever={os_ver} {pkg}")
 
         # Since we try to mitigate any damage caused by the incomplete rollback
         # try to update the system, in case anything got downgraded
@@ -470,22 +469,28 @@ def _load_json_schema(path):
 def pre_registered(shell, request, yum_conf_exclude):
     """
     A fixture to install subscription manager and pre-register the system prior to the convert2rhel run.
-    Requesting the ubi_repo_on_oracle fixture, to be able to install the subscription package.
+    For Oracle Linux we're using the _add_client_tools_repo_oracle to enable the client-tools repository
+    to install the subscription-manager package from.
+    We also exclude the rhn-client* packages for the same reason.
+    On Oracle Linux subscription-manger is obsolete and replaced by rhn-client* packages when installing subman.
     """
     username = TEST_VARS["RHSM_USERNAME"]
     password = TEST_VARS["RHSM_PASSWORD"]
     # Use custom parameters when the fixture is parametrized
     if hasattr(request, "param"):
         username, password = request.param
+        print(">>> Using username and password requested in the fixture parameter.")
+
+    # Add the client-tools repository for Oracle linux to install subscription-manager from
+    # Exclude the rhn-client* packages in the yum.conf
     if "oracle" in SYSTEM_RELEASE_ENV:
         yum_conf_exclude(exclude=["rhn-client*"])
-
-    _add_client_tools_repo_oracle(shell)
+        _add_client_tools_repo_oracle(shell)
 
     assert shell("yum install -y subscription-manager").returncode == 0
     # Download the SSL certificate
     shell(
-        "curl --create-dirs -o /etc/rhsm/ca/redhat-uep.pem https://cdn-public.redhat.com/content/public/repofiles/redhat-uep.pem"
+        "curl --create-dirs -ko /etc/rhsm/ca/redhat-uep.pem https://cdn-public.redhat.com/content/public/repofiles/redhat-uep.pem"
     )
     # Register the system
     assert (
@@ -508,9 +513,10 @@ def pre_registered(shell, request, yum_conf_exclude):
     # system identity: <UUID>, we need to store just the system UUID
     original_registration_uuid = uuid_raw_output.split(":")[1].strip()
 
-    _remove_client_tools_repo_oracle(shell)
-
     yield
+
+    if "oracle" in SYSTEM_RELEASE_ENV:
+        _remove_client_tools_repo_oracle(shell)
 
     # For some scenarios we do not pre-register the system, therefore we do not have the original UUID
     # and do not need to verify it stays the same
@@ -694,7 +700,8 @@ def kernel(shell):
 
 
 @pytest.fixture()
-def yum_conf_exclude(shell, request):
+@pytest.mark.yum_conf_exclude
+def yum_conf_exclude(shell, backup_directory, request):
     """
     Fixture.
     Define `exclude=kernel kernel-core` in /etc/yum.conf.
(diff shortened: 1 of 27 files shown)


tests/integration/conftest.py
```diff
@@ -29,10 +30,33 @@
...
+    elif SYSTEM_RELEASE_ENV == "oracle-8-latest":
+        sat_reg_command = SAT_REG_FILE["ORACLE8_SAT_REG"]
+    elif SYSTEM_RELEASE_ENV == "centos-8-latest":
+        sat_reg_command = SAT_REG_FILE["CENTOS8_SAT_REG"]
+    elif SYSTEM_RELEASE_ENV == "oracle-7":
+        sat_reg_command = SAT_REG_FILE["ORACLE7_SAT_REG"]
+    elif SYSTEM_RELEASE_ENV == "centos-7":
+        sat_reg_command = SAT_REG_FILE["CENTOS7_SAT_REG"]
+
+    return sat_reg_command
```
r0x0d, line 58: If I may suggest some improvements, what do you think of the following code? 

```suggestion
SAT_REG_COMMAND = {
    "alma-8-latest": SAT_REG_FILE["ALMA8_SAT_REG"],
    "alma-8.8": SAT_REG_FILE["ALMA88_SAT_REG"],
    "rocky-8-latest": SAT_REG_FILE["ROCKY8_SAT_REG"],
    "rocky-8.8":SAT_REG_FILE["ROCKY88_SAT_REG"],
    "oracle-8-latest": SAT_REG_FILE["ORACLE8_SAT_REG"],
    "centos-8-latest": SAT_REG_FILE["CENTOS8_SAT_REG"],
    "oracle-7": SAT_REG_FILE["ORACLE7_SAT_REG"],
    "centos-7": SAT_REG_FILE["CENTOS7_SAT_REG"]
}

def satellite_registration_command():
    """
    Get the Satellite registration command for the respective system.
    """
    sat_reg_command = None
    try: 
        sat_reg_command = SAT_REG_COMMAND[SYSTEM_RELEASE_ENV]
    except KeyError:
        print(f"Key not found in satellite registration command dicitonary: {SYSTEM_RELEASE_ENV}")

    return sat_reg_command
```

danmyway: That makes super sense, thank you @r0x0d !

tests/integration/conftest.py
```diff
@@ -674,3 +719,109 @@ def kernel(shell):
...
+    Helper function.
+    Runs only on Oracle Linux system
+    Create an ubi repo for its respective major version to install subscription-manager from.
+    """
+    if "oracle" in SYSTEM_RELEASE_ENV:
+        with open("/etc/system-release", "r") as sysrelease_file:
+            content = sysrelease_file.read()
+            sysver = get_system_version_number(content)
+
+        repo_url = "https://ftp.redhat.com/redhat/client-tools/client-tools-for-rhel-8-x86_64.repo"
```
r0x0d, line 782: Why not using the CDN one?

tests/integration/conftest.py
```diff
@@ -674,3 +719,109 @@ def kernel(shell):
...
+    """
+    if "oracle" in SYSTEM_RELEASE_ENV:
+        yum_conf_exclude(exclude=["rhn-client*"])
+    _add_client_tools_repo_oracle(shell)
+    # Get the registration command for the respective system
+    # from the conftest function
+    sat_reg_command = satellite_registration_command()
+
+    # Register the system to the Satellite server
+    assert shell(sat_reg_command).returncode == 0
```
r0x0d, line 819: To make it more reliable, wouldn't be better to check against the sat_reg_command not being None?

```suggestion
    sat_reg_command = satellite_registration_command()
    
    # Make sure it returned some value, otherwise it waill fail. 
    assert sat_reg_command
    
    # Register the system to the Satellite server
    assert shell(sat_reg_command).returncode == 0
```

tests/integration/conftest.py
kokesak: @danmyway https://cdn-public.redhat.com/content/public/repofiles/client-tools-for-rhel-8.repo

tests/integration/conftest.py
```diff
@@ -674,3 +719,109 @@ def kernel(shell):
...
+def _add_client_tools_repo_oracle(shell):
+    """
+    Helper function.
+    Runs only on Oracle Linux system
+    Create an ubi repo for its respective major version to install subscription-manager from.
+    """
+    if "oracle" in SYSTEM_RELEASE_ENV:
+        with open("/etc/system-release", "r") as sysrelease_file:
+            content = sysrelease_file.read()
+            sysver = get_system_version_number(content)
```
kokesak, line 780: this could also be replaced by the https://github.com/oamg/convert2rhel/pull/1192/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56R299

tests/integration/conftest.py
```diff
@@ -496,6 +539,8 @@ def pre_registered(shell, request):
     # system identity: <UUID>, we need to store just the system UUID
     original_registration_uuid = uuid_raw_output.split(":")[1].strip()
 
+    _remove_client_tools_repo_oracle(shell)
```
kokesak, line 542: shouldn't this be placed after the fixture's yield?

tests/integration/conftest.py
```diff
@@ -674,3 +719,109 @@ def kernel(shell):
...
+
+
+@pytest.fixture
+def satellite_registration(shell, yum_conf_exclude):
+    """
+    Register the system to the Satellite server
+    """
+    if "oracle" in SYSTEM_RELEASE_ENV:
+        yum_conf_exclude(exclude=["rhn-client*"])
+    _add_client_tools_repo_oracle(shell)
```
kokesak, line 813: again, this can be placed under the if statement above (it doesn't make sense to call it on other distros even though the function does nothing on other distros). The other option would be to run this function for every system - download the client tool on each distro

tests/integration/tier0/destructive/conversion-method/main.fmf
```diff
@@ -70,7 +70,7 @@ order: 49
     description: |
         Conversion method using the Satellite credentials for registration.
         The subscription-manager package is removed for this conversion method.
-        The katello-ca-consumer package is installed from the Satellite server.
+        Subscribe to the Satellite server using the curl command.
```
kokesak, line 73: Technically the curl commands only downloads some script, it is then passed to the bash through pipe `|`

tests/integration/tier0/destructive/conversion-method/test_satellite.py
```diff
@@ -1,27 +1,15 @@
...
 
 
 @pytest.mark.test_satellite_conversion
-def test_satellite_conversion(shell, convert2rhel):
+def test_satellite_conversion(shell, convert2rhel, satellite_registration):
     """
     Conversion method using the Satellite credentials for registration.
     The subscription-manager package is removed for this conversion method.
-    The katello-ca-consumer package is installed from the Satellite server.
+    Subscribe to the Satellite server using the curl command.
```
kokesak, line 11: same as above

tests/integration/tier0/destructive/offline-system-conversion/prepare_system.py
```diff
@@ -1,31 +1,36 @@
...
+from conftest import TEST_VARS
 
 
-# Replace urls in rhsm.conf file to the satellite server
-# Without doing this we get obsolete dogfood server as source of repositories
 def replace_urls_rhsm():
+    """
+    Replace urls in rhsm.conf file to the satellite server
+    Without doing this we get obsolete dogfood server as source of repositories
+    """
```
kokesak, line 12: Maybe we will not need this function anymore since we do not download the katello package

tests/integration/conftest.py
kokesak, line 819: have you tried the approach with shell `silent=true`? Because we are now leaking the whole authorization token to the logs. 
You can just silent the curl command and save the output to some file and then execute the file itself (no silent=true is needed for this one)

kokesak (changes requested):

tests/integration/conftest.py
```diff
@@ -804,16 +809,19 @@ def _remove_client_tools_repo_oracle(shell):
...
     """
-    if "oracle" in SYSTEM_RELEASE_ENV:
-        yum_conf_exclude(exclude=["rhn-client*"])
-    _add_client_tools_repo_oracle(shell)
     # Get the registration command for the respective system
     # from the conftest function
     sat_reg_command = satellite_registration_command()
+    # If the fixture is parametrized, use the parameter as the registration command
+    if hasattr(request, "param"):
+        sat_reg_command = request.param
```
r0x0d, line 821: Would be worth it to add a print statement mentioning that the sat_reg_command is now from a param instead of coming from `satellite_registration_command()`? 

I think this could help us to understand if the value was changed somehow and is different from what we expect.

tests/integration/conftest.py
danmyway: I have, and I probably will come back to it again, I have the verbosity on just for the debugging right now.
Getting the returncodes 127 from subscription-manager and 1 from curl did not help too much.

tests/integration/conftest.py
danmyway: I was using whatever convert2rhel was using, but this one is better, thanks.

tests/integration/conftest.py
danmyway: This is intentional.
I want to remove the client repo before calling convert2rhel, so it's not backed up and/or used by the utility during any yum calls.

kokesak: Backing up by c2r would be a problem? Because by removing the repos we are setting the system in a way that it contain packages without the equivalent repository. I mean, is this something even the customer would do?

tests/integration/conftest.py
```diff
@@ -346,6 +370,21 @@ def system_release(shell):
...
+    Oracle Linux Server release 7.8
+    CentOS Linux release 7.6.1810 (Core)
+    CentOS Linux release 8.1.1911 (Core)
+    """
+    match = re.search(r".+?(\d+)\.(\d+)\D?", system_release_content)
+    if not match:
+        return "not match"
+    version = namedtuple("Version", ["major", "minor"])(int(match.group(1)), int(match.group(2)))
+
+    return version
```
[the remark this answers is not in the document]
danmyway: Makes sense, will do with a rebase, when it's merged.

tests/integration/conftest.py
danmyway: That makes sense, thank you @r0x0d

tests/integration/conftest.py
danmyway: Good point, we can try if it won't yield any issues.

tests/integration/tier0/destructive/offline-system-conversion/prepare_system.py
danmyway: Good point, let's try that

tests/integration/conftest.py
```diff
@@ -407,26 +430,31 @@ def missing_os_release_package_workaround(shell):
...
-        "centos-7": "centos-release",
-        "centos-8": "centos-linux-release",
-        "alma-8": "almalinux-release",
-        "rocky-8": "rocky-release",
+        "oracle-7": ("oraclelinux-release-el7", "oraclelinux-release"),
+        "oracle-8": ("oraclelinux-release-el8", "oraclelinux-release"),
+        "centos-7": ("centos-release",),
+        "centos-8": ("centos-linux-release",),
+        "alma-8": ("almalinux-release",),
+        "rocky-8": ("rocky-release",),
```
r0x0d, line 438: Don't want to be picky, but this could be a list instead... It does not hurt to be a tuple, but a list would have the same effect, and maybe even straightforward to look at and see that we are "iterating over the results"

tests/integration/conftest.py
```diff
@@ -407,26 +430,31 @@ def missing_os_release_package_workaround(shell):
...
-        system_release_pkg = os_to_pkg_mapping.get(os_key)
+        system_release_pkgs = os_to_pkg_mapping.get(os_key)
 
-        rpm_output = shell(f"rpm -qa {system_release_pkg}").output
-        if "not installed" in rpm_output:
-            shell(f"yum install -y --releasever={os_ver} {system_release_pkg}")
+        for pkg in system_release_pkgs:
+            rpm_output = shell(f"rpm -q {pkg}").output
+            if "not installed" in rpm_output:
+                shell(f"yum install -y --releasever={os_ver} {pkg}")
```
r0x0d, line 453: If you don't want to rely on string checking, you could replace it by: 

```suggestion
            if shell(f"rpm -q {pkg}").returncode:
                shell(f"yum install -y --releasever={os_ver} {pkg}")
```

Or, to prevent undesired outcomes (if that ever happen): 

```suggestion
            rpm_output = shell(f"rpm -q {pkg}").output
            if "not installed" in rpm_output.lower():
                shell(f"yum install -y --releasever={os_ver} {pkg}")
```

tests/integration/conftest.py
```diff
@@ -438,34 +466,38 @@ def _load_json_schema(path):
...
-        )
+        yum_conf_exclude(exclude=["rhn-client*"])
+        _add_client_tools_repo_oracle(shell)
 
     assert shell("yum install -y subscription-manager").returncode == 0
     # Download the SSL certificate
     shell(
-        "curl --create-dirs -o /etc/rhsm/ca/redhat-uep.pem https://cdn-public.redhat.com/content/public/repofiles/redhat-uep.pem"
+        "curl --create-dirs -ko /etc/rhsm/ca/redhat-uep.pem https://cdn-public.redhat.com/content/public/repofiles/redhat-uep.pem"
     )
```
r0x0d, line 494: I would update the above comment to include that we are downloading the certificate allowing insecure connection to the server (because we are using `-k` option)

tests/integration/conftest.py
```diff
@@ -662,3 +697,143 @@ def kernel(shell):
...
+    Create an ubi repo for its respective major version to install subscription-manager from.
+    """
+    if "oracle" in SYSTEM_RELEASE_ENV:
+        repo_url = "https://cdn-public.redhat.com/content/public/repofiles/client-tools-for-rhel-8.repo"
+        if SystemInformationRelease.version.major == 7:
+            repo_url = "https://cdn-public.redhat.com/content/public/repofiles/client-tools-for-rhel-7-server.repo"
+
+        # Add the redhat-release GPG key
+        assert shell(f"curl -o /etc/yum.repos.d/client-tools-for-tests.repo {repo_url}")
+        shell("curl -o /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release https://www.redhat.com/security/data/fd431d51.txt")
```
r0x0d, line 761: Since this is already being called in another part of the code guarded by the condition of `"oracle" in SYSTEM_RELEASE_ENV`, I would remove the condition from here and keep it simpler for now: 

```suggestion
    repo_url = "https://cdn-public.redhat.com/content/public/repofiles/client-tools-for-rhel-8.repo"
    if SystemInformationRelease.version.major == 7:
        repo_url = "https://cdn-public.redhat.com/content/public/repofiles/client-tools-for-rhel-7-server.repo"

    # Add the redhat-release GPG key
    assert shell(f"curl -o /etc/yum.repos.d/client-tools-for-tests.repo {repo_url}")
    shell("curl -o /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release https://www.redhat.com/security/data/fd431d51.txt")
```

r0x0d: Or de-indent the function call in the places where we are calling them: 

* tests/integration/conftest.py#488
* tests/integration/conftest.py#792

tests/integration/conftest.py
danmyway: Makes sense, I'm not sure why I went for tuple to be honest :man_shrugging:

tests/integration/conftest.py
danmyway: Checking for return code makes much more sense

tests/integration/conftest.py
```diff
@@ -662,3 +700,143 @@ def kernel(shell):
 
         # Reboot after clean-up
         shell("tmt-reboot -t 600")
+
+
+@pytest.fixture()
+@pytest.mark.yum_conf_exclude
+def yum_conf_exclude(shell, backup_directory, request):
```
kokesak, line 707: The request parameter is not used anywhere

tests/integration/tier0/destructive/offline-system-conversion/prepare_system.py
```diff
@@ -1,31 +1,38 @@
...
+#     with open("/etc/rhsm/rhsm.conf", "r+") as f:
+#         file = f.read()
+#         # Replacing the urls
+#         file = re.sub("hostname = .*", "hostname = {}".format(TEST_VARS["SATELLITE_URL"]), file)
+#         file = re.sub("baseurl = .*", "baseurl = https://{}/pulp/repos".format(TEST_VARS["SATELLITE_URL"]), file)
+#
+#         # Setting the position to the top of the page to insert data
+#         f.seek(0)
+#         f.write(file)
+#         f.truncate()
```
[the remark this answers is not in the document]
danmyway: nerob

tests/integration/tier0/destructive/conversion-method/test_satellite.py
```diff
@@ -1,27 +1,16 @@
...
-from conftest import SATELLITE_PKG_DST, SATELLITE_PKG_URL, TEST_VARS
+from conftest import TEST_VARS
 
 
 @pytest.mark.test_satellite_conversion
-def test_satellite_conversion(shell, convert2rhel):
+def test_satellite_conversion(shell, convert2rhel, satellite_registration):
     """
     Conversion method using the Satellite credentials for registration.
     The subscription-manager package is removed for this conversion method.
```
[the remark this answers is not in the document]
danmyway: Good catch, thanks

tests/integration/tier0/non-destructive/eus/test_eus_support.py
```diff
@@ -6,17 +6,15 @@
...
     Backs up the convert2rhel/systeminfo.py and modifies the EUS_MINOR_VERISONS mapping.
     Restores the original version after the test is done.
     """
     eus_mapping_file = shell("find /usr -path '*/convert2rhel/systeminfo.py'").output.strip()
-    backup_dir = "/tmp/c2r_tests_backup/"
-    backup_file = f"{backup_dir}systeminfo.py.bkp"
-    if not os.path.exists(backup_dir):
-        shell(f"mkdir {backup_dir}")
+    backup_dir = backup_directory
+    backup_file = os.path.join(backup_dir, "systeminfo.py.bkp")
```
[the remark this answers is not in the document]
danmyway: I was just probably being lazy

tests/integration/tier1/destructive/kernel-check-skip/test_latest_kernel_check_skip.py
```diff
@@ -16,11 +17,17 @@ def test_skip_kernel_check(shell, convert2rhel):
         3/ Enable *just* the rhel-7-server-rpms repository prior to conversion
         4/ Run conversion verifying the conversion is not inhibited and completes successfully
     """
-    backup_dir = "/tmp/repobckp"
-    eus_backup_dir = "/tmp/eus_repobckp"
-    shell(f"mkdir {backup_dir}")
+    backup_dir = backup_directory
+    eus_backup_dir = os.path.join(backup_dir, "eus")
```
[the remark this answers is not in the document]
danmyway: Enabled for all

tests/integration/conftest.py
danmyway: fixed

tests/integration/conftest.py
```diff
@@ -41,8 +65,12 @@
 def shell(tmp_path):
     """Live shell."""
 
-    def factory(command, silent=False):
-        if not silent:
+    def factory(command, silent=False, hide_command=False):
+        if silent:
+            click.echo("The fixture is set to silent=True, therefore no output will be printed.")
+        if hide_command:
+            click.echo("The fixture is set to hide_command=False, so it won't show the called command.")
```
r0x0d, line 72:
```suggestion
            click.echo("The fixture is set to silent=True, therefore no output will be printed.")
        if hide_command:
            click.echo("The fixture is set to hide_command=True, so it won't show the called command.")
```

tests/integration/conftest.py
```diff
@@ -41,8 +65,12 @@
 def shell(tmp_path):
     """Live shell."""
 
-    def factory(command, silent=False):
-        if not silent:
+    def factory(command, silent=False, hide_command=False):
```
r0x0d, line 68: What is the difference between silent and hide_command? Seems to me that they will do the same thing in the code

tests/integration/conftest.py
```diff
@@ -662,3 +700,147 @@ def kernel(shell):
...
+
+    assert shell(f"cp -v {yum_config} {config_bak}").returncode == 0
+
+    for pkg in exclude:
+        # If there is already an `exclude` section, append to the existing value
+        if config.has_option("main", "exclude"):
+            pre_existing_value = config.get("main", "exclude")
+            config.set("main", "exclude", f"{pre_existing_value} {pkg}")
+        else:
+            config.set("main", "exclude", pkg)
```
r0x0d, line 741: We could rewrite this section as this: 


```suggestion
    pkgs_to_exclude = " ".join(exclude)
    # If there is already an `exclude` section, append to the existing value
    if config.has_option("main", "exclude"):
        pre_existing_value = config.get("main", "exclude")
        config.set("main", "exclude", f"{pre_existing_value} {pkgs_to_exclude}")
    else:
        config.set("main", "exclude", pkgs_to_exclude)
```

Main idea here is to not loop through every exclude package and perform the same operations every time (check for option, get the value, rewrite it, ...) 

We can loop through the packages once and then just append to the config once too

tests/integration/conftest.py
danmyway: `silent` is for no output from the shell fixture at all, apart from printing, that no output will be provided.
That might be used for some calls where we don't want the logs too polluted, for example when we install/uninstall packages and the output gets quite lengthy.
`hide_command` serves the purpose of trying not to leak any sensitive data. For example, if we call `subscription-manager register` directly, the shell fixture prints out the used command with the credentials in clear text first and then output of the running command.
In that case, we might care about the output for debugging purposes, but we don't want to show the command used (that you can find easily in the code if needed).

danmyway: What we would like to do in the future is to use logger and log a respective level based on the shell call returncode.
So if everything runs fine, don't print out anything, if there is a problem, fallback to debug/error/critical and print out the output.

tests/integration/conftest.py
danmyway: Good idea, thank you @r0x0d !

tests/integration/tier1/destructive/kernel-check-skip/test_latest_kernel_check_skip.py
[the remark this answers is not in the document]
kokesak: https://issues.redhat.com/browse/RHELC-1541

tests/integration/tier0/non-destructive/file-backup/test_file_backup.py
```diff
@@ -13,12 +13,11 @@
     "yum": "/etc/logrotate.d/yum",
     "dnf": "/etc/logrotate.d/dnf",
 }
-BACKUP_DIR = "/tmp/c2r_tests_backup/"
-MODIFIED_FILES_DIR = "/tmp/c2r_tests_backup/modified/"
 
 
 @pytest.fixture
-def config_files_modified(shell):
+@pytest.mark.config_files_modified
```
kokesak, line 19: Just a suggestion, maybe we should prefix those markers that are created because of the `backup_directory`. Something like:
```suggestion
@pytest.mark.backupdir_config_files_modified
```

kokesak (approved):

tests/integration/tier0/non-destructive/file-backup/test_file_backup.py
danmyway: Good idea, we can do that in a separate PR

PR #1195, opened 2024-04-24
Override pytest addopts for local development

r0x0d:
After the merge of https://github.com/oamg/convert2rhel/pull/1189, the spam of information of test setups is causing problems in github actions log rendering, as well as executing it locally/container.

This commit is intended to override the addopts option in pytest.ini when the
execution is inside a container/locally.

Jira Issues:

-

Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ PYLINT_ARGS ?=
 VENV ?= .venv3
 PRE_COMMIT ?= pre-commit
 SHOW_CAPTURE ?= no
-PYTEST_ARGS ?= -p no:cacheprovider
+PYTEST_ARGS ?= --override-ini=addopts= -p no:cacheprovider
 BUILD_IMAGES ?= 1
 
 ifdef KEEP_TEST_CONTAINER


r0x0d:
@danmyway and @kokesak, please review it once you have some time. 

The change introduced in #1189 was causing a lot of spam in the log messages in our unit tests too. To fix this problem, I'm trying to split the pytest ini file.

kokesak:
I suggest keeping the current `pytest.ini` for the integration tests and create a separate one for the unit tests. The reason is that in your current change you have to add `-c tests/pytest.ini` to every test file we have and we'll create. It seems more straightforward and less repeated code when you add the `-c convert2rhel/unit_tests/pytest.ini` just once into the `Makefile` here https://github.com/oamg/convert2rhel/blob/main/Makefile#L32 (I suspect that you run unit tests using the make)

SpyTec (changes requested):
Given that this changes the test calls everywhere, to avoid having to duplicate the config call, lets refactor this to a call in makefile instead. So we have a make command of pytest_int or something that has pytest -c tests/pytest.ini

Human factor will inevitably make it easier to make mistakes in the future otherwise

r0x0d:
Alright, sounds good.

SpyTec (approved):

PR #1196, opened 2024-04-24
Prevent workflows to execute in forks

r0x0d:
Forks might not always have the same setup as the upstream repositories, to prevent email spams from failed workflows, let's prevent them from running in the forks.

Jira Issues:

-

Checklist

### DIFF
--- a/.github/workflows/autotag.yml
+++ b/.github/workflows/autotag.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   tag:
+    if: always() && github.repository == 'oamg/convert2rhel'
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/build_images.yml
+++ b/.github/workflows/build_images.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   build_and_publish:
     name: build-${{ matrix.el.distro }}${{ matrix.el.ver }}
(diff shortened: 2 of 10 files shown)


SpyTec:
I haven't encountered any emails myself about this unless I updated the main branch once in a while maybe, or actively trying to debug workflows in something like releases 

Did something happen recently for this to occur for you?

r0x0d:
> I haven't encountered any emails myself about this unless I updated the main branch once in a while maybe, or actively trying to debug workflows in something like releases
> 
> Did something happen recently for this to occur for you?

Yes, I actually rebase & push to main in my fork quite a lot. I like to have both the local and remote updated. 

Also, the auto tagging workflow keeps running and creating tags in forks as well.

SpyTec:
Since this changes all workflows, I suppose we can use this instead. I don't want to add extra steps in the workflow if we can avoid it, and with this you can disable it for your own fork if you want

https://github.com/orgs/community/discussions/26704#discussioncomment-3252979

r0x0d:
> Since this changes all workflows, I suppose we can use this instead. I don't want to add extra steps in the workflow if we can avoid it, and with this you can disable it for your own fork if you want
> 
> https://github.com/orgs/community/discussions/26704#discussioncomment-3252979

Ah, alright, that's even better. Just disabled the workflows for this repo in my fork. 

We can close this one then.

PR #1208, opened 2024-05-06
[RHELC-1525] Fix cornercase with userpass and keyorg

Venefilyn:
If a user provided username and password CLI arguments and had set
activation key and organization in the config file, it would fail to
execute convert2rhel. Same also happens if organization is set and
activation key isn't.

This is addressed by also unsetting organization if we detect username
and password.

Jira Issues:

- [RHELC-1525](https://issues.redhat.com/browse/RHELC-1525)

Checklist

### DIFF
--- a/convert2rhel/unit_tests/toolopts_test.py
+++ b/convert2rhel/unit_tests/toolopts_test.py
@@ -694,3 +694,23 @@ def test_cli_userpass_specified(argv, message, monkeypatch, caplog, global_tool_
         # Don't care about the exception, focus on output message
         pass
     assert message in caplog.text
+
+
+@pytest.mark.parametrize(
+    ("activation_key", "organization", "argv"),
+    (
+        ("activation_key", "org", []),
+        ("activation_key", "org", ["analyze", "-u name", "-p pass"]),
+        (None, None, ["analyze", "-u name", "-p pass"]),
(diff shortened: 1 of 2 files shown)


convert2rhel/unit_tests/toolopts_test.py
```diff
@@ -694,3 +694,22 @@ def test_cli_userpass_specified(argv, message, monkeypatch, caplog, global_tool_
...
+)
+def test_cli_args_config_file_cornercase(activation_key, organization, argv, monkeypatch):
+    monkeypatch.setattr(sys, "argv", mock_cli_arguments(argv))
+    t_opts = convert2rhel.toolopts.ToolOpts()
+    t_opts.org = organization
+    t_opts.activation_key = activation_key
+    t_opts.no_rhsm = True
+    monkeypatch.setattr(convert2rhel.toolopts, "tool_opts", t_opts)
+
+    convert2rhel.toolopts.CLI()
```
r0x0d, line 715: We are not asserting anything in here. Is this expected?

Venefilyn: Supposed to be. If it doesn't pass it means it threw an exception it shouldn't. Is there a better way of indicating this?

PR #1209, opened 2024-05-06
fix: require username and password to be specified

Venefilyn:
We didn't require username and password to be specified when one was
provided but not the other. Such as username being provided but without
a password, or vice versa. Due to that convert2rhel fails much later
in the process than other warnings.

This addresses that with warning messages and aligns with other cases
like activation_key and org

Jira Issues:

-

Checklist


r0x0d (approved):

r0x0d:
Codecov is misbehaving again. I'm merging this as the sanity tests are all green and the results from the unit tests doesn't include any errors.

PR #1210, opened 2024-05-06
[RHELC-1521] Fix check for reposdir in RestorablePackage class

r0x0d:
The condition to check for the existence of the reposdir folder was set in an incorrect place. This caused the program to raise an exception as we are trying to list the contents of the directory before checking for the path availability of the folder.

Jira Issues:

- [RHELC-1521](https://issues.redhat.com/browse/RHELC-1521)

Checklist

### DIFF
--- a/convert2rhel/backup/packages.py
+++ b/convert2rhel/backup/packages.py
@@ -137,9 +137,10 @@ class RestorablePackage(RestorableChange):
         loggerinst.debug("Using repository files stored in %s." % self.reposdir)
 
         if self.reposdir:
-            # If nothing is inside the directory, or it does not exist, let's
-            # just not use it to download the packages.
-            if len(os.listdir(self.reposdir)) == 0 or not os.path.exists(self.reposdir):
+            # Check if the reposdir exists and if the directory is empty
+            if (os.path.exists(self.reposdir) and len(os.listdir(self.reposdir)) == 0) or not os.path.exists(
+                self.reposdir
+            ):
                 loggerinst.info("The repository directory %s seems to be empty or non-existent.")
                 self.reposdir = None
 


convert2rhel/backup/packages.py
```diff
@@ -136,10 +136,10 @@ def enable(self):
         loggerinst.info("Backing up the packages: %s." % ",".join(self.pkgs))
         loggerinst.debug("Using repository files stored in %s." % self.reposdir)
 
-        if self.reposdir:
+        if self.reposdir and os.path.exists(self.reposdir):
             # If nothing is inside the directory, or it does not exist, let's
             # just not use it to download the packages.
-            if len(os.listdir(self.reposdir)) == 0 or not os.path.exists(self.reposdir):
+            if len(os.listdir(self.reposdir)) == 0:
```
Venefilyn, line 142: We could combine all three but I feel like the this is the best solution atm

You could have done all three as if-statements go left to right (usually)

E.g. `if not os.path.exists(self.reposdir) or len(os.listdir(self.reposdir)) == 0:` would also have fixed the issue

However, please update the comments too! :pray:

Venefilyn: Actually scratch that.. it's intended that we go to the log no matter what, so the exception is false but implementation was (mostly) correct as it was before

However, we should still change it to be something like

```
if reposdir
	if (reposdir exists and empty) or reposdir does not exist:
		log and set reposdir empty
```

r0x0d: I don't think it will work like that... If the path doesn't exist and we pass that to os.listdir(), it will cause an exception as the path doesn't exist. Let me try that locally

r0x0d: OOoh, that's funny! 

So, I didn't know that using `and` here would not cause an exception. If we use `or`, then the exception could arise, with `and` it doesn't. That's a new learning. 

I will switch the code as you proposed.

r0x0d: Updated!

hosekadam (approved):
Works great, thanks!

convert2rhel/backup/packages.py
Venefilyn: > So, I didn't know that using `and` here would not cause an exception. If we use `or`, then the exception could arise, with `and` it doesn't. That's a new learning.

It has to do with if-statements evaluating left to right, so if you have `if os.path.exists(path) and path` it would fail. You can also do optimization with this in mind in general. Such as finding something from a database and evaluating it. Happens a lot more commonly in other OOP languages.

Venefilyn (approved):

PR #1226, opened 2024-05-16
[RHELC-1547] Fix regression in installing removed packages during rollback

r0x0d:
We were passing a set object to iterate over to compose a message of packages that got removed during yum transaction. This problem was not happening before as we were dealing with a success case of restoring packages. In this regression, we found out that a set object does not have indexing, therefore, thus bug was hidden from most of our tests and pipeline.

Jira Issues:

- [RHELC-1547](https://issues.redhat.com/browse/RHELC-1547)

Checklist

### DIFF
--- a/convert2rhel/pkgmanager/handlers/yum/__init__.py
+++ b/convert2rhel/pkgmanager/handlers/yum/__init__.py
@@ -67,7 +67,7 @@ def _resolve_yum_problematic_dependencies(output):
             packages_to_remove.append(str(resolve_error[0]).replace(" ", ""))
 
     if packages_to_remove:
-        packages_to_remove = set(packages_to_remove)
+        packages_to_remove = list(set(packages_to_remove))
         loggerinst.debug(
             "Removing problematic packages to continue with the conversion:\n%s",
             "\n".join(packages_to_remove),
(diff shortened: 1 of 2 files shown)


danmyway (approved):
That looks good to me, tests are passing, the package is getting restored correctly.
Thanks, @r0x0d !
@Venefilyn could you please take a quick look, we can merge from my POV.

convert2rhel/pkgmanager/handlers/yum/__init__.py
```diff
@@ -67,7 +67,7 @@ def _resolve_yum_problematic_dependencies(output):
             packages_to_remove.append(str(resolve_error[0]).replace(" ", ""))
 
     if packages_to_remove:
-        packages_to_remove = set(packages_to_remove)
+        packages_to_remove = list(set(packages_to_remove))
```
hosekadam, line 70: I would add a comment saying that we need to make each item in the list unique. But not a blocker, looks good
```suggestion
        # Need to make each item in the list unique, one pkg can be present more times in the list.
        packages_to_remove = list(set(packages_to_remove))
```

hosekadam (approved):

convert2rhel/pkgmanager/handlers/yum/__init__.py
r0x0d: Lovely, thanks!

PR #1229, opened 2024-05-17
[RHELC-1554] Fix ELS repositories not fallback in case of error

r0x0d:
We want to try to enable the ELS repos, and if case that fails, we want to make sure we enable the default rhel repoids.

Jira Issues:

- [RHELC-1554](https://issues.redhat.com/browse/RHELC-1554)

Checklist

### DIFF
--- a/convert2rhel/subscription.py
+++ b/convert2rhel/subscription.py
@@ -800,19 +800,19 @@ def enable_repos(rhel_repoids):
     else:
         repos_to_enable = rhel_repoids
 
-    if repos_to_enable == system_info.eus_rhsm_repoids:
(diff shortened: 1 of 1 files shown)


hosekadam (approved):
Looks good!

bookwar:
As discussed in the ticket, rather than adding fallback for ELS repositories, let's remove fallback for EUS too.

As ELS and EUS repositories are now requested explicitly via command line switch, the proper fallback is for a user to remove the incorrect cli option. We should not try to do this implicitly during the conversion.

r0x0d:
Alright, will open a PR for that in a minute.

PR #1236, opened 2024-05-22
[RHELC-1559] Update test using tmt-reboot

danmyway:
* tmt ignores any exitcode returned by pytest before the last reboot
* mitigate the issue by using try except and overriding the last pytest exitcode with reporting fail to tmt straight from the test
* minor adjustments to already existing tests - function name update, remove obsolete reboot call from a one-kernel-scenario

Jira Issues:

- [RHELC-1559](https://issues.redhat.com/browse/RHELC-1559) 

Checklist


tests/integration/tier0/destructive/isolated-system-conversion/test_offline_system_conversion.py
r0x0d, line 1: Should rename the file as well, no?

tests/integration/tier0/non-destructive/kernel/test_custom_kernel.py
```diff
@@ -107,17 +107,19 @@ def test_custom_kernel(convert2rhel, shell, custom_kernel, hybrid_rocky_image):
...
+                )
+                c2r.expect("RHEL_COMPATIBLE_KERNEL::INVALID_KERNEL_PACKAGE_SIGNATURE")
 
-            c2r.sendcontrol("c")
+                c2r.sendcontrol("c")
 
-        try:
             assert c2r.exitstatus == 1
-        except AssertionError:
+        except:  # pylint: disable=W0702
```
r0x0d, line 124: It still may be too broad, but maybe we should use Exception here, or try to see what is the base exception we could catch

danmyway: We started with an AssertionError on the c2r.exitcode, but then I realized, that anything going wrong during the contextmanager call would not be handled that way.
I felt that even though it's too broad, it might be more reliable.
Generally AssertionError, EOFError and Timeout(Error?) could be there for sure, so we might go for that, however we might lose some reliability with that 🤷

tests/integration/tier0/destructive/isolated-system-conversion/test_offline_system_conversion.py
danmyway: good catch, thank you

tests/integration/tier0/non-destructive/kernel/test_custom_kernel.py
danmyway: I have slimmed the exceptions to `AssertionError` and `pexpect.exceptions.(EOF, TIMEOUT)` in the end.
Hopefully we won't miss anything :grimacing:

tests/integration/tier0/non-destructive/single-yum-transaction-validation/main.fmf
```diff
@@ -67,6 +67,13 @@ tag+:
         pytest -m test_validation_packages_with_in_name_period
 
 /override_exclude_list_in_yum_config:
+    adjust+:
+        environment+:
+            CONVERT2RHEL_ALLOW_UNAVAILABLE_KMODS: 1
+        when: distro == oracle-8
+        because: |
+            With an older kernel, older, with RHEL incompatible kernel modules get installed as well,
```
kokesak, line 75: " With an older kernel, older, ..."
probabaly some typoo? (two "older")

tests/integration/tier1/destructive/one-kernel-scenario/test_one_kernel_scenario.py
```diff
@@ -86,5 +86,3 @@ def test_one_kernel_scenario(shell, convert2rhel, one_kernel):
         shell("yum-config-manager {}".format(enable_repo_opt))
 
         assert shell("yum install -y python3 --enablerepo=*").returncode == 0
-
```
kokesak, line 89: This `test_one_kernel_scenario` should also be in try except no? The `one_kernel` fixture also uses tmt-reboot

danmyway: Good point, but the reboot is running before the conversion, so if anything goes sideways during the conversion, tmt will be able to pick up the returncode from the pytest/convert2rhel run.
On the other hand, it probably won't pick up anything during the one_kernel() fixture :shrug:

kokesak: I see, as we talked in person - please remove the asserts from the `one_kernel` as they are pointless there anyway

tests/integration/tier0/non-destructive/single-yum-transaction-validation/main.fmf
danmyway: Fixed, hopefully

tests/integration/tier1/destructive/one-kernel-scenario/test_one_kernel_scenario.py
danmyway: I've added the try except to the fixture, so we bail out when the setup fails

kokesak (approved):

PR #1249, opened 2024-05-31
chore: build docs on PR and push to main

Venefilyn:
This builds and checks the documentation builds and for commits on main branch it will deploy to GitHub Pages

Confirmed to work on my repo
https://venefilyn.github.io/convert2rhel

Jira Issues:

-

Checklist

### DIFF
--- a/.devcontainer/alma9/alma9-development.Containerfile
+++ b/.devcontainer/alma9/alma9-development.Containerfile
@@ -5,7 +5,8 @@ ENV PYTHON python3
 ENV PIP pip
 ENV PYTHONDONTWRITEBYTECODE 1
 
-ENV APP_DEV_DEPS "alma9/requirements.txt"
+ENV APP_DEV_DEPS ".devcontainer/alma9/requirements.txt"
+ENV APP_DOCS_DEPS "docs/requirements.txt"
 ENV APP_MAIN_DEPS \
     python3 \
     python3-pip \
@@ -30,7 +31,8 @@ RUN dnf update -y && dnf install -y $APP_MAIN_DEPS && dnf clean all
 FROM install_main_deps as install_dev_deps
 
 COPY $APP_DEV_DEPS $APP_DEV_DEPS
-RUN $PIP install -r $APP_DEV_DEPS
+COPY $APP_DOCS_DEPS $APP_DOCS_DEPS
+RUN $PIP install -r $APP_DEV_DEPS -r $APP_DOCS_DEPS
 
 FROM install_dev_deps as install_application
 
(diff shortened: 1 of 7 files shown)


.github/workflows/main-docs-deploy.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Pages
+        uses: actions/configure-pages@v5
+      - uses: ammaraskar/sphinx-action@dev
+        with:
+          docs-folder: "docs/"
+          pre-build-command: "pip install sphinx_autodoc_typehints"
```
r0x0d, line 38: Can't we just use `pip install -r ...` in here? In case we add future dependencies to that, we won't need to change the workflow

Venefilyn: Fixed!

r0x0d:
@Venefilyn, the file at `.github/workflows/pr-docs-check.yml` also needs to be updated with the `pip install -r ...`.

Also, I see that you created a docs/requirements.txt. In the past, I've added the sphinx docs dependencies to each .devcontainer/*/requirements.txt, maybe to keep consistency all around we should remove that dependency from those devcontainers requirements and actually place them in a single place, like the docs/requirements.txt you created? I think that would make it more organized.

.devcontainer/alma9/alma9-development.Containerfile
```diff
@@ -30,7 +31,7 @@ RUN dnf update -y && dnf install -y $APP_MAIN_DEPS && dnf clean all
 FROM install_main_deps as install_dev_deps
 
 COPY $APP_DEV_DEPS $APP_DEV_DEPS
-RUN $PIP install -r $APP_DEV_DEPS
+RUN $PIP install -r $APP_DEV_DEPS -f $APP_DOCS_DEPS
```
Venefilyn, line 34: @r0x0d not sure if this will work with the `../docs` part. But pip install can take multiple files

.devcontainer/centos7/requirements.txt
```diff
@@ -4,5 +4,3 @@ pathlib2==2.3.7.post1
 mock==3.0.5
 pytest-cov==2.12.1
 coverage[toml]==5.5
-sphinx
-sphinx_autodoc_typehints
```
Venefilyn, line 8: Removed here as `sphinx_autodoc_typehints` requires python >= 3.9

Venefilyn:
Okay @r0x0d should work now. Changed around the syntax a bit for working directory

.github/workflows/pr-docs-check.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: ammaraskar/sphinx-action@dev
+      with:
+        docs-folder: "docs/"
+        pre-build-command: "pip install sphinx_autodoc_typehints"
```
r0x0d, line 13:
```suggestion
        pre-build-command: "pip install -r docs/requirements.txt"
```

r0x0d (approved):
Small comment, after applying, feel free to merge!

PR #1255, opened 2024-06-05
[RHELC-774] Add exclude setopt to remaining repoquery calls

r0x0d:
We need to exclude any packages set in the /etc/yum.conf set by the user so we can make sure that when we run repoquery, we are really gathering all the data we want from the command. This patch introduces the overridable --setopt=exclude= in the remaining repoquery calls.

Jira Issues:

- [RHELC-774](https://issues.redhat.com/browse/RHELC-774)

Checklist

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
+++ b/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
@@ -71,10 +71,7 @@ class EnsureKernelModulesCompatibility(actions.Action):
             "--releasever=%s" % system_info.releasever,
             "--setopt=*.skip_if_unavailable=False",
         ]
-        basecmd = [
-            "repoquery",
-            "--releasever=%s" % system_info.releasever,
-        ]
+        basecmd = ["repoquery", "--releasever=%s" % system_info.releasever, "--setopt=exclude="]
 
         if system_info.version.major >= 8:
             basecmd.append("--setopt=module_platform_id=platform:el" + str(system_info.version.major))
--- a/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
+++ b/convert2rhel/actions/pre_ponr_changes/kernel_modules.py
@@ -71,6 +71,9 @@ class EnsureKernelModulesCompatibility(actions.Action):
             "--releasever=%s" % system_info.releasever,
             "--setopt=*.skip_if_unavailable=False",
         ]
+        # Clearing the exclude field with setopt to prevent kernel being
+        # excluded in the config.
+        # https://issues.redhat.com/browse/RHELC-774
         basecmd = ["repoquery", "--releasever=%s" % system_info.releasever, "--setopt=exclude="]
 
         if system_info.version.major >= 8:
--- a/convert2rhel/pkgmanager/handlers/dnf/__init__.py
+++ b/convert2rhel/pkgmanager/handlers/dnf/__init__.py
@@ -86,6 +86,10 @@ class DnfTransactionHandler(TransactionHandlerBase):
         # going in with the packages in the transaction.
         self._base._ds_callback = DependencySolverProgressIndicatorCallback()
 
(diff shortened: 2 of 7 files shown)


r0x0d:
missing the dnf/yum cli calls. Will update the PR.

convert2rhel/actions/pre_ponr_changes/kernel_modules.py
```diff
@@ -71,10 +71,7 @@ def _get_rhel_supported_kmods(self):
             "--releasever=%s" % system_info.releasever,
             "--setopt=*.skip_if_unavailable=False",
         ]
-        basecmd = [
-            "repoquery",
-            "--releasever=%s" % system_info.releasever,
-        ]
+        basecmd = ["repoquery", "--releasever=%s" % system_info.releasever, "--setopt=exclude="]
```
Venefilyn, line 74: Could we add a comment here as well about setopt? Maybe just linking to the Jira issue. More comments the better for future maintainability IMO

Venefilyn (approved):
Looks great! Approved and can be merged when tests pass

convert2rhel/pkgmanager/handlers/dnf/__init__.py
```diff
@@ -86,6 +86,10 @@ def _set_up_base(self):
         # going in with the packages in the transaction.
         self._base._ds_callback = DependencySolverProgressIndicatorCallback()
 
+        # Override the exclude option that is loaded from the config and set it
+        # to empty.
+        self._base.conf.substitutions["exclude"] = []
```
hosekadam, line 91: I think this should be set after `self._base.read_all_repos()` is called. Otherwise, it raises an exception.

Also, I'm not sure about the name`exclude`, I couldn't find that in the docs. The only one I found and seems like what we want is `excludepkgs` https://dnf.readthedocs.io/en/latest/conf_ref.html#conf-main-options-label

r0x0d: The use of `substitutions` is wrongly used here as pointed correctly by @hosekadam. We should rely on `exclude` or `excludepkgs` as well. I'm investigating on why dnf is not loading the settings properly through the config file.

r0x0d: The `excludepkgs` is a dnf only param, but either `exclude` or `excludepkgs` can be used. `exclude` is compatible with yum, and since we are pretty much using it everywhere, I think I will keep the `exclude` in the code. 

The substitutions part is really wrong here, but I'm pushing a commit that will fix it.

hosekadam (approved):
Looks good, but cannot test because of the Candlepin outage

convert2rhel/pkgmanager/handlers/dnf/__init__.py
```diff
@@ -86,6 +86,11 @@ def _set_up_base(self):
         # going in with the packages in the transaction.
         self._base._ds_callback = DependencySolverProgressIndicatorCallback()
 
+        # Override the exclude option that is loaded from the config and set it to empty.
+        # NOTE(r0x0d): In DNF they have exclude to be `py:excludepkgs`. It is fine to use `exclude` for now as it is
```
Venefilyn, line 90: Small nitpick but some tools that consolidate NOTE/TODO etc won't find it like this, so I've started doing this instead when I can:

```suggestion
        # NOTE @r0x0d: In DNF they have exclude to be `py:excludepkgs`. It is fine to use `exclude` for now as it is
```

Venefilyn:
Failing tests relates to Oracle Linux kernel and is unrelated to the PR, merging

CC @danmyway in case you're not aware
https://dashboard.packit.dev/results/testing-farm/562322

Venefilyn:
Updated the title, WDYT @r0x0d

r0x0d:
/packit test --labels tier0

> Updated the title, WDYT @r0x0d

Sounds good! Thanks for that update.

PR #1256, opened 2024-06-05
[RHELC-1331] Port post_ponr_set_efi_configuration() to Action framework

pr-watson:
This PR ports the post_ponr_set_efi_configuration function to the action framework. Changing any relevant logger crictical, warning and info messages to Error results and Warning and Info messages respectively.

Jira Issues:

- [RHELC-1331](https://issues.redhat.com/browse/RHELC-1331)

Checklist


convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,195 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+
+import logging
+import os
+import shutil
+
+from convert2rhel import actions, grub, systeminfo, utils
```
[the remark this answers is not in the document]
r0x0d:
```suggestion
from convert2rhel import actions, grub, systeminfo
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,195 @@
...
+        files provided by Red Hat inside the RHEL_EFIDIR_CANONICAL_PATH.
+        The new UEFI bootloader entry is always created / registered and set
+        set as default.
+
+        The current (original) UEFI bootloader entry is removed under some conditions
+        (see _remove_orig_boot_entry() for more info).
+        """
+
+        try:
+            grub._replace_efi_boot_entry()
```
r0x0d, line 188: Refactor this to be public

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,195 @@
...
+    def run(self):
+        """Remove the /boot/efi/EFI/centos/ directory when no UEFI files remains.
+
+        The centos/ directory after the conversion contains usually just grubenv,
+        grub.cfg, .. files only. Which we copy into the redhat/ directory. If no
+        other UEFI files are present, we can remove this dir. However, if additional
+        UEFI files are present, we should keep the directory for now, until we
+        deal with it.
+        """
+        if systeminfo.system_info.id != "centos":
```
r0x0d, line 151:
```suggestion
        if systeminfo.system_info.id != "centos":
                    logger.debug("Skipping removing EFI files - only related to CentOS Linux.")
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,195 @@
...
+            self.set_result(
+                level="ERROR",
+                id="EFIBOOTMGR_UTILITY_NOT_INSTALLED",
+                title="Efibootmgr utility is not installed",
+                description="The /usr/sbin/efibootmgr utility is not installed.",
+                remediations="Install the efibootmgr utility via YUM/DNF.",
+            )
+
+
+class CopyGrubFiles(actions.Action):
```
r0x0d, line 82: Instead of having those two separate classes, can't we just merge them together? It seems that this is only being used inside `CopyGrubFiles`

pr-watson: We could do that but they both contain a set_result call - we would risk not having both set_results if we combine them

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,240 @@
...
+        level="ERROR",
+        id="IO_ERROR",
+        title="I/O error",
+        description=("I/O error(13): Permission denied Some GRUB files have not been copied to /boot/efi/EFI/redhat."),
+    )
+
+
+def test_remove_efi_centos_warning(monkeypatch, remove_efi_centos_instance):
+
+    monkeypatch.setattr("os.rmdir", mock.Mock())
```
r0x0d, line 211:
```suggestion
    monkeypatch.setattr(os, "rmdir", mock.Mock())
```

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,240 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+import os
+import shutil
+
+import pytest
+import six
+
+from convert2rhel import actions, exceptions, grub, unit_tests
```
r0x0d, line 23:
```suggestion
from convert2rhel import actions, grub, unit_tests
```

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,240 @@
...
+@pytest.mark.parametrize(
+    ("is_efi", "efi_file_exists", "log_msg"),
+    (
+        (False, None, "BIOS detected"),
+        (True, True, "UEFI binary found"),
+    ),
+)
+def test_new_default_efi_bin(new_default_efi_bin_instance, is_efi, efi_file_exists, log_msg, monkeypatch, caplog):
+    monkeypatch.setattr("os.path.exists", mock.Mock(return_value=efi_file_exists))
+    monkeypatch.setattr("convert2rhel.grub.is_efi", mock.Mock(return_value=is_efi))
```
r0x0d, line 66:
```suggestion
    monkeypatch.setattr(os.path, "exists", mock.Mock(return_value=efi_file_exists))
    monkeypatch.setattr(grub, "is_efi", mock.Mock(return_value=is_efi))
```

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,240 @@
...
+    assert log_msg in caplog.records[-1].message
+    unit_tests.assert_actions_result(
+        new_default_efi_bin_instance,
+        level="SUCCESS",
+    )
+
+
+def test_new_default_efi_bin_error(new_default_efi_bin_instance, monkeypatch):
+    monkeypatch.setattr("os.path.exists", mock.Mock(return_value=False))
+    monkeypatch.setattr("convert2rhel.grub.is_efi", mock.Mock(return_value=True))
```
r0x0d, line 77:
```suggestion
    monkeypatch.setattr(os.path, "exists", mock.Mock(return_value=False))
    monkeypatch.setattr(grub, "is_efi", mock.Mock(return_value=True))
```

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,240 @@
...
+            "are present in the %s directory and that\n"
+            "a new bootloader entry for Red Hat Enterprise Linux exists\n"
+            "(check `efibootmgr -v` output).\n"
+            "The entry should point to '\\EFI\\redhat\\shimx64.efi'." % grub.RHEL_EFIDIR_CANONICAL_PATH
+        ),
+    )
+
+
+def test_efi_bootmgr_utility_installed_error(efi_bootmgr_utility_installed_instance, monkeypatch):
+    monkeypatch.setattr("os.path.exists", mock.Mock(return_value=False))
```
r0x0d, line 99:
```suggestion
    monkeypatch.setattr(os.path, "exists", mock.Mock(return_value=False))
```

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,240 @@
...
+    monkeypatch.setattr("convert2rhel.systeminfo.system_info.id", sys_id)
+
+    copy_grub_files_instance.run()
+    assert any(log_msg in record.message for record in caplog.records)
+    if sys_id == "centos" and src_file_exists and not dst_file_exists:
+        assert mock_path_copy2.call_args_list == [
+            mock.call("/boot/efi/EFI/centos/grubenv", "/boot/efi/EFI/redhat/grubenv"),
+            mock.call("/boot/efi/EFI/centos/grub.cfg", "/boot/efi/EFI/redhat/grub.cfg"),
+            mock.call("/boot/efi/EFI/centos/user.cfg", "/boot/efi/EFI/redhat/user.cfg"),
+        ]
```
r0x0d, line 146:
```suggestion
@pytest.mark.parametrize(
    ("sys_id", "src_file_exists", "dst_file_exists", "log_msg"),
    (
        ("oracle", None, None, "only related to CentOS Linux"),
        ("centos", None, True, "file already exists"),
        ("centos", True, False, "Copying '"),
    ),
)
def test__copy_grub_files(
    sys_id,
    src_file_exists,
    dst_file_exists,
    log_msg,
    monkeypatch,
    caplog,
    copy_grub_files_instance,
    global_system_info,
):
    def path_exists(path):
        return src_file_exists if grub.CENTOS_EFIDIR_CANONICAL_PATH in path else dst_file_exists
    monkeypatch.setattr(os.path, "exists", path_exists)
    monkeypatch.setattr(shutil, "copy2", mock.Mock())
    global_system_info.id = sys_id

    copy_grub_files_instance.run()
    assert any(log_msg in record.message for record in caplog.records)
    if sys_id == "centos" and src_file_exists and not dst_file_exists:
        assert mock_path_copy2.call_args_list == [
            mock.call("/boot/efi/EFI/centos/grubenv", "/boot/efi/EFI/redhat/grubenv"),
            mock.call("/boot/efi/EFI/centos/grub.cfg", "/boot/efi/EFI/redhat/grub.cfg"),
            mock.call("/boot/efi/EFI/centos/user.cfg", "/boot/efi/EFI/redhat/user.cfg"),
        ]
```

We might need to check on this one if global_system_info might not need to be added as a monkeypatch in this case. I think it will probably not need.

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,240 @@
...
+    monkeypatch.setattr("convert2rhel.systeminfo.system_info.id", sys_id)
+
+    copy_grub_files_instance.run()
+    unit_tests.assert_actions_result(
+        copy_grub_files_instance,
+        level="ERROR",
+        id="UNABLE_TO_FIND_REQUIRED_FILE_FOR_GRUB_CONFIG",
+        title="Unable to find required file for GRUB config",
+        description="Unable to find the original file required for GRUB configuration: /boot/efi/EFI/centos/grubenv",
+    )
```
r0x0d, line 178:
```suggestion
@pytest.mark.parametrize(
    ("sys_id", "src_file_exists", "dst_file_exists"),
    (("centos", False, False),),
)
def test__copy_grub_files_error(
    sys_id,
    src_file_exists,
    dst_file_exists,
    monkeypatch,
    caplog,
    copy_grub_files_instance,
    global_system_info
):
    def path_exists(path):
        return src_file_exists if grub.CENTOS_EFIDIR_CANONICAL_PATH in path else dst_file_exists

    monkeypatch.setattr(os.path, "exists", path_exists)
    monkeypatch.setattr(shutil, "copy2", mock.Mock())
    global_system_info.id = sys_id

    copy_grub_files_instance.run()
    unit_tests.assert_actions_result(
        copy_grub_files_instance,
        level="ERROR",
        id="UNABLE_TO_FIND_REQUIRED_FILE_FOR_GRUB_CONFIG",
        title="Unable to find required file for GRUB config",
        description="Unable to find the original file required for GRUB configuration: /boot/efi/EFI/centos/grubenv",
    )
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,205 @@
...
+        """
+        super(ReplaceEfiBootEntry, self).run()
+
+        try:
+            grub.replace_efi_boot_entry()
+        except grub.BootloaderError as e:
+            self.set_result(
+                level="ERROR",
+                id="BOOTLOADER_ERROR",
+                title="Bootloader error",
```
r0x0d, line 203: Can we get a more descriptive title? Something like `Failed to replace EFI boot entry.`

If we can improve the `id` as well, would be good.

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,205 @@
...
+        except OSError:
+            warning_message = (
+                "The folder %s is left untouched. You may remove the folder manually"
+                " after you ensure there is no custom data you would need." % CENTOS_EFIDIR_CANONICAL_PATH
+            )
+            logger.warning(warning_message)
+            self.add_message(
+                level="WARNING",
+                id="FOLDER_NOT_REMOVED",
+                title="Folder was not removed",
```
r0x0d, line 174: Same here, a more descriptive title and id

convert2rhel/actions/conversion/set_efi_config.py
r0x0d, line 1: Line 160 and 162 are missing coverages. Let's try to add a test case for them

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,205 @@
...
+            if not os.path.exists(src_path):
+                if filename in required:
+                    # without the required files user should not reboot the system
+                    self.set_result(
+                        level="ERROR",
+                        id="UNABLE_TO_FIND_REQUIRED_FILE_FOR_GRUB_CONFIG",
+                        title="Unable to find required file for GRUB config",
+                        description="Unable to find the original file required for GRUB configuration: %s" % src_path,
+                    )
+                    return
```
r0x0d, line 127: We should break here? If more than one file does not exist, we would try to overwrite the set_result? Or it would end up in a failure because we already set an result?

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,205 @@
...
+            logger.info("Copying '%s' to '%s'" % (src_path, dst_path))
+            try:
+                shutil.copy2(src_path, dst_path)
+            except (OSError, IOError) as err:
+                # IOError for py2 and OSError for py3
+                self.set_result(
+                    level="ERROR",
+                    id="IO_ERROR",
+                    title="I/O error",
```
r0x0d, line 136: A more descriptive title and id here as well.

r0x0d (changes requested):
Some comments, otherwise, looks good.

convert2rhel/actions/conversion/set_efi_config.py
pr-watson: Good point. It probably would end in a failure if we set a result. Maybe I can use the flag from before and if it's true after the loop finishes then I can set the result

pr-watson: @r0x0d do you think this block is fine as is since it will already end in a failure because we set a result?

convert2rhel/actions/conversion/set_efi_config.py
r0x0d: I think this need some improvements still, @pr-watson

convert2rhel/actions/conversion/set_efi_config.py
r0x0d: I think this need some improvements still, @pr-watson

convert2rhel/actions/conversion/set_efi_config.py
r0x0d: I think this need some improvements still, @pr-watson

convert2rhel/actions/conversion/set_efi_config.py
pr-watson: Sorry I thought I updated this, correcting it now

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+        redhat/ directory after the conversion.
+
+        The copy of the centos/ directory should be ok. In case of the conversion
+        from Oracle Linux, the redhat/ directory is already used.
+        """
+        super(CopyGrubFiles, self).run()
+
+        if systeminfo.system_info.id != "centos":
+            logger.debug("Skipping copying GRUB files - only related to CentOS Linux.")
+            return
```
r0x0d, line 103: Let's make sure to add a test case for this

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,227 @@
...
+        remediations="Install the efibootmgr utility via YUM/DNF.",
+    )
+
+
+@pytest.mark.parametrize(
+    ("sys_id", "src_file_exists", "dst_file_exists", "log_msg"),
+    (
+        # ("oracle", None, None, "only related to CentOS Linux"),
+        ("centos", None, True, "file already exists"),
+        # ("centos", True, False, "Copying '"),
```
r0x0d, line 117:
```suggestion
        ("oracle", None, None, "only related to CentOS Linux"),
        ("centos", None, True, "file already exists"),
        ("centos", True, False, "Copying '"),
```

r0x0d:
@pr-watson, please squash the commits

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+        """Check if the Efibootmgr utility is installed"""
+        super(EfibootmgrUtilityInstalled, self).run()
+
+        if not os.path.exists("/usr/sbin/efibootmgr"):
+            self.set_result(
+                level="ERROR",
+                id="EFIBOOTMGR_UTILITY_NOT_INSTALLED",
+                title="Efibootmgr utility is not installed",
+                description="The /usr/sbin/efibootmgr utility is not installed.",
+                remediations="Install the efibootmgr utility via YUM/DNF.",
```
r0x0d, line 80: Can we provide a generic command for that? Maybe something like: 


```suggestion
                remediations="Install the efibootmgr utility using the following command:\n\n 1. yum install efibootmgr",
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+            )
+
+
+class CopyGrubFiles(actions.Action):
+    id = "COPY_GRUB_FILES"
+    dependencies = ("EFIBOOTMGR_UTILITY_INSTALLED",)
```
r0x0d, line 86: I think we don't directly depend on the efibootmgr being present on the system for this one. The action seens to be only copying files from the old centos folder to the new redhat one. We can probably drop this from here


```suggestion
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+        The centos/ directory after the conversion contains usually just grubenv,
+        grub.cfg, .. files only. Which we copy into the redhat/ directory. If no
+        other UEFI files are present, we can remove this dir. However, if additional
+        UEFI files are present, we should keep the directory for now, until we
+        deal with it.
+        """
+        super(RemoveEfiCentos, self).run()
+
+        if systeminfo.system_info.id != "centos":
+            logger.debug("Skipping removing EFI files - only related to CentOS Linux.")
```
r0x0d, line 174:
```suggestion
            logger.debug("Did not perform removal of EFI files - only related to CentOS Linux.")
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+                level="ERROR",
+                id="UNABLE_TO_FIND_REQUIRED_FILE_FOR_GRUB_CONFIG",
+                title="Unable to find required file for GRUB config",
+                description="Unable to find the original file required for GRUB configuration at: %s"
+                % ", ".join(missing_files),
+            )
+            return
+
+        for src_file in src_files:
+            # Check if the src_file already exists at the RHEL_EFIDR_CANONICAL_PATH
```
Venefilyn, line 133:
```suggestion
            # Skip already existing file in destination directory
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+                new_default_efibin = efi_path
+                break
+            logger.debug("UEFI binary %s not found. Checking next possibility..." % efi_path)
+        if not new_default_efibin:
+            self.set_result(
+                level="ERROR",
+                id="RHEL_UEFI_BINARIES_DO_NOT_EXIST",
+                title="RHEL UEFI binaries do not exist",
+                description="None of the expected RHEL UEFI binaries exist.",
+                diagnosis="The migration of the bootloader setup was not successful.",
```
Venefilyn, line 54: This is quite passive. Status is at the end instead of the beginning. Also shouldn't diagnosis be about diagnosing the issue? Atm it's written like a status
```suggestion
                diagnosis="Bootloader couldn't be migrated due to missing RHEL EFI binaries.",
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+                logger.info("UEFI binary found: %s" % efi_path)
+                new_default_efibin = efi_path
+                break
+            logger.debug("UEFI binary %s not found. Checking next possibility..." % efi_path)
+        if not new_default_efibin:
+            self.set_result(
+                level="ERROR",
+                id="RHEL_UEFI_BINARIES_DO_NOT_EXIST",
+                title="RHEL UEFI binaries do not exist",
+                description="None of the expected RHEL UEFI binaries exist.",
```
Venefilyn, line 53: Which ones? I'd say write the fullpath out for what we searched for

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+            if os.path.exists(efi_path):
+                logger.info("UEFI binary found: %s" % efi_path)
+                new_default_efibin = efi_path
+                break
+            logger.debug("UEFI binary %s not found. Checking next possibility..." % efi_path)
+        if not new_default_efibin:
+            self.set_result(
+                level="ERROR",
+                id="RHEL_UEFI_BINARIES_DO_NOT_EXIST",
+                title="RHEL UEFI binaries do not exist",
```
Venefilyn, line 52: They could exist, but we just couldn't find them
```suggestion
                title="RHEL UEFI binaries not found",
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+                diagnosis="The migration of the bootloader setup was not successful.",
+                remediations=(
+                    "Do not reboot your machine before doing a manual check of the\n"
+                    "bootloader configuration. Ensure that grubenv and grub.cfg files\n"
+                    "are present in the %s directory and that\n"
+                    "a new bootloader entry for Red Hat Enterprise Linux exists\n"
+                    "(check `efibootmgr -v` output).\n"
+                    "The entry should point to '\\EFI\\redhat\\shimx64.efi'." % grub.RHEL_EFIDIR_CANONICAL_PATH
+                ),
+            )
```
Venefilyn, line 63: We should be more colloquial and not write things out. Also why do we insert newlines?
```suggestion
                remediations=(
                    "Verify the bootloader configuration as follows and reboot the system"
                    " Ensure that `grubenv` and `grub.cfg` files"
                    " are present in the %s directory. Verify that `efibootmgr -v`"
                    " shows a bootloader entry for Red Hat Enterprise Linux"
                    " that points to to '\\EFI\\redhat\\shimx64.efi'." % grub.RHEL_EFIDIR_CANONICAL_PATH
                ),
            )
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+    def run(self):
+        """Check if the Efibootmgr utility is installed"""
+        super(EfibootmgrUtilityInstalled, self).run()
+
+        if not os.path.exists("/usr/sbin/efibootmgr"):
+            self.set_result(
+                level="ERROR",
+                id="EFIBOOTMGR_UTILITY_NOT_INSTALLED",
+                title="Efibootmgr utility is not installed",
+                description="The /usr/sbin/efibootmgr utility is not installed.",
```
Venefilyn, line 79:
```suggestion
                title="UEFI boot manager utility not found",
                description="Couldn't find the UEFI boot manager which is required for us to install and verify a RHEL boot entry.",
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,219 @@
...
+            missing_files = [
+                filename for filename in src_files if filename in required and not os.path.exists(filename)
+            ]
+            # without the required files user should not reboot the system
+            self.set_result(
+                level="ERROR",
+                id="UNABLE_TO_FIND_REQUIRED_FILE_FOR_GRUB_CONFIG",
+                title="Unable to find required file for GRUB config",
+                description="Unable to find the original file required for GRUB configuration at: %s"
+                % ", ".join(missing_files),
```
Venefilyn, line 128: I'm not entirely sure atm what we do with the copying of grub configs bit something along these lines is more friendly to the user:
```suggestion
                title="Couldn't find system GRUB config",
                description="Couldn't find any GRUB config files in the current system which is required for configuring UEFI for RHEL:{}".format("\n- ".join(missing_files)),
```

Venefilyn (changes requested):
With the Message ID guidelines I setup in the docs, can you come up with something else for the message IDs? If it's difficult or you feel like there is a better way to do message IDs just suggest it in the chat or update the wiki (and let us know)
https://github.com/oamg/convert2rhel/wiki/Action-Framework#message-ids

I also didn't go through all the `set_message()` actions as I'm just wondering now what the audience for it is. 

@oamg/convert2rhel-developers is the intention for the action messages that are displayed in Insights and others to be for readability where novice developer or sysadmins can easily know what to do without prior knowledge or something else? If that is the case we should really be improving the titles, messages, diagnosis, and remediations to be more readable

convert2rhel/actions/conversion/set_efi_config.py
pr-watson: Added a list with the ones we searched for

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,223 @@
...
+
+        try:
+            grub.replace_efi_boot_entry()
+        except grub.BootloaderError as e:
+            self.set_result(
+                level="ERROR",
+                id="FAILED_TO_REPLACE_EFI_BOOT_ENTRY",
+                title="Failed to replace EFI boot entry",
+                description=e.message,
+            )
```
Venefilyn, line 223: Needs to be rewritten

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,223 @@
...
+            ]
+            # without the required files user should not reboot the system
+            self.set_result(
+                level="ERROR",
+                id="UNABLE_TO_FIND_REQUIRED_FILE_FOR_GRUB_CONFIG",
+                title="Couldn't find system GRUB config",
+                description="Couldn't find any GRUB config files in the current system which is required for configuring UEFI for RHEL: {}".format(
+                    ", ".join(missing_files)
+                ),
+            )
```
Venefilyn, line 133: Maybe be more verbose saying any of the files need to exist. Currently it's not clear if all of them or just one needs to be there

Venefilyn (changes requested):

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+        super(RemoveEfiCentos, self).run()
+
+        if systeminfo.system_info.id != "centos":
+            logger.debug("Did not perform removal of EFI files - only related to CentOS Linux.")
+            # nothing to do
+            return
+        try:
+            os.rmdir(CENTOS_EFIDIR_CANONICAL_PATH)
+        except OSError:
```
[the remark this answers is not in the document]
pr-watson: Added IOError

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,223 @@
...
+            logger.debug("UEFI binary %s not found. Checking next possibility..." % efi_path)
+            mising_binaries.append(efi_path)
+        if not new_default_efibin:
+            self.set_result(
+                level="ERROR",
+                id="NOT_FOUND_RHEL_UEFI_BINARIES",
+                title="RHEL UEFI binaries not found",
+                description="None of the expected RHEL UEFI binaries exist.",
+                diagnosis="Bootloader couldn't be migrated due to missing RHEL EFI binaries: {} .".format(
+                    ", ".join(mising_binaries)
```
r0x0d, line 58:
```suggestion
                    ", ".join(missing_binaries)
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,223 @@
...
+        missing_binaries = []
+
+        for filename in grub.DEFAULT_INSTALLED_EFIBIN_FILENAMES:
+            efi_path = os.path.join(RHEL_EFIDIR_CANONICAL_PATH, filename)
+            if os.path.exists(efi_path):
+                logger.info("UEFI binary found: %s" % efi_path)
+                new_default_efibin = efi_path
+                break
+            logger.debug("UEFI binary %s not found. Checking next possibility..." % efi_path)
+            mising_binaries.append(efi_path)
```
r0x0d, line 50:
```suggestion
            missing_binaries.append(efi_path)
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,223 @@
...
+        other UEFI files are present, we can remove this dir. However, if additional
+        UEFI files are present, we should keep the directory for now, until we
+        deal with it.
+        """
+        super(RemoveEfiCentos, self).run()
+
+        if systeminfo.system_info.id != "centos":
+            logger.debug("Did not perform removal of EFI files - only related to CentOS Linux.")
+            # nothing to do
+            return
```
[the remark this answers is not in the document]
pr-watson: Verified with new test test_remove_efi_centos_non_centos

convert2rhel/actions/conversion/set_efi_config.py
pr-watson: Added specification for one GRUB file

Venefilyn (dismissed):

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,225 @@
...
+
+        try:
+            grub.replace_efi_boot_entry()
+        except grub.BootloaderError as e:
+            self.set_result(
+                level="ERROR",
+                id="FAILED_TO_REPLACE_EFI_BOOT_ENTRY",
+                title="Failed to replace EFI boot entry",
+                description="The EFI boot entry could not be replaced due to the following error: '%s'" % e.message,
+            )
```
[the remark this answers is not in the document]
bocekm: EFI was long ago renamed to UEFI. Per [Wikipedia](https://en.wikipedia.org/wiki/UEFI):
> In 2005, UEFI deprecated EFI 1.10 (the final release of EFI).

For that reason, use UEFI in all customer facing strings.

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,225 @@
...
+                break
+            logger.debug("UEFI binary %s not found. Checking next possibility..." % efi_path)
+            missing_binaries.append(efi_path)
+        if not new_default_efibin:
+            self.set_result(
+                level="ERROR",
+                id="NOT_FOUND_RHEL_UEFI_BINARIES",
+                title="RHEL UEFI binaries not found",
+                description="None of the expected RHEL UEFI binaries exist.",
+                diagnosis="Bootloader couldn't be migrated due to missing RHEL EFI binaries: {} .".format(
```
Venefilyn, line 59:
```suggestion
                diagnosis="Bootloader couldn't be migrated due to missing RHEL UEFI binaries: {} .".format(
```

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,237 @@
...
+    monkeypatch.setattr(os.path, "exists", mock.Mock(return_value=False))
+    monkeypatch.setattr(grub, "is_efi", mock.Mock(return_value=True))
+    new_default_efi_bin_instance.run()
+    unit_tests.assert_actions_result(
+        new_default_efi_bin_instance,
+        level="ERROR",
+        id="NOT_FOUND_RHEL_UEFI_BINARIES",
+        title="RHEL UEFI binaries not found",
+        description="None of the expected RHEL UEFI binaries exist.",
+        diagnosis="Bootloader couldn't be migrated due to missing RHEL EFI binaries: /boot/efi/EFI/redhat/shimx64.efi, /boot/efi/EFI/redhat/grubx64.efi .",
```
Venefilyn, line 85:
```suggestion
        diagnosis="Bootloader couldn't be migrated due to missing RHEL UEFI binaries: /boot/efi/EFI/redhat/shimx64.efi, /boot/efi/EFI/redhat/grubx64.efi .",
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,225 @@
...
+        The centos/ directory after the conversion contains usually just grubenv,
+        grub.cfg, .. files only. Which we copy into the redhat/ directory. If no
+        other UEFI files are present, we can remove this dir. However, if additional
+        UEFI files are present, we should keep the directory for now, until we
+        deal with it.
+        """
+        super(RemoveEfiCentos, self).run()
+
+        if systeminfo.system_info.id != "centos":
+            logger.debug("Did not perform removal of EFI files - only related to CentOS Linux.")
```
Venefilyn, line 180:
```suggestion
            logger.debug("Did not perform removal of UEFI files - only related to CentOS Linux.")
```

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,237 @@
...
+    )
+    assert expected.issuperset(remove_efi_centos_instance.messages)
+    assert expected.issubset(remove_efi_centos_instance.messages)
+
+
+def test_remove_efi_centos_non_centos(monkeypatch, remove_efi_centos_instance, global_system_info, caplog):
+    global_system_info.id = "oracle"
+    monkeypatch.setattr(systeminfo, "system_info", global_system_info)
+    remove_efi_centos_instance.run()
+    assert "Did not perform removal of EFI files" in caplog.text
```
Venefilyn, line 223:
```suggestion
    assert "Did not perform removal of UEFI files" in caplog.text
```

Venefilyn (commented):
Michal pointed out in a comment about EFI being outdated, here is suggestions to change existing logging to UEFI. Could arguably also change the action class names but I don't think it's worth it at this stage

hosekadam:
When testing in our uefi vagrant, I got this error:
```
Copying GRUB2 configuration files to the new UEFI directory /boot/efi/EFI/redhat/.
[2024-08-13T11:53:22+0000] DEBUG - The grubenv file already exists in /boot/efi/EFI/redhat/ folder. Copying skipped.
Copying '/boot/efi/EFI/centos/grub.cfg' to '/boot/efi/EFI/redhat/grub.cfg'
Copying '/boot/efi/EFI/centos/user.cfg' to '/boot/efi/EFI/redhat/user.cfg'
ERROR - (ERROR) COPY_GRUB_FILES::GRUB_FILES_NOT_COPIED_TO_BOOT_DIRECTORY - GRUB files have not been copied to boot directory
 Description: I/O error(2): No such file or directory Some GRUB files have not been copied to /boot/efi/EFI/redhat.
 Diagnosis: N/A
 Remediations: N/A
```

pr-watson:
> When testing in our uefi vagrant, I got this error:
> 
> ```
> Copying GRUB2 configuration files to the new UEFI directory /boot/efi/EFI/redhat/.
> [2024-08-13T11:53:22+0000] DEBUG - The grubenv file already exists in /boot/efi/EFI/redhat/ folder. Copying skipped.
> Copying '/boot/efi/EFI/centos/grub.cfg' to '/boot/efi/EFI/redhat/grub.cfg'
> Copying '/boot/efi/EFI/centos/user.cfg' to '/boot/efi/EFI/redhat/user.cfg'
> ERROR - (ERROR) COPY_GRUB_FILES::GRUB_FILES_NOT_COPIED_TO_BOOT_DIRECTORY - GRUB files have not been copied to boot directory
>  Description: I/O error(2): No such file or directory Some GRUB files have not been copied to /boot/efi/EFI/redhat.
>  Diagnosis: N/A
>  Remediations: N/A
> ```
> 
> EDIT: I see the PR got updated, need to check if the problem is still there

Checking on centos8 now

pr-watson:
Looks like the user.cfg file isn't present

hosekadam:
The `user.cfg` should be optional file: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/grub.py#L590

pr-watson:
The user.cfg is added to the list of files to copy even though it's not present on the system, I think that's why we see the error "No such file or directory". Looking into why it get's added to the list

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,225 @@
...
+        logger.task("Convert: Configure the bootloader")
+
+        if not grub.is_efi():
+            logger.info("BIOS detected. Nothing to do.")
+            return
+
+        new_default_efibin = None
+        missing_binaries = []
+
+        for filename in grub.DEFAULT_INSTALLED_EFIBIN_FILENAMES:
```
hosekadam, line 45:
```suggestion
        for filename in DEFAULT_INSTALLED_EFIBIN_FILENAMES:
```
I would consider importing it as `CENTOS_EFIDIR_CANONICAL_PATH, RHEL_EFIDIR_CANONICAL_PATH`

pr-watson:
Issue was we were'nt checking if the src files existed for attempting to copy them - added a condition to prevent this https://github.com/oamg/convert2rhel/pull/1256/files#diff-6ce0e50430c410250938f88c48ba475e98b8867d9be6371848c4332eef0423b9R139-R145

convert2rhel/main.py
```diff
@@ -180,7 +180,7 @@ def main_locked():
         _raise_for_skipped_failures(post_conversion_results)
         report.post_conversion_report(
             results=post_conversion_results,
-            include_all_reports=True,
```
r0x0d, line 183: This feels unrelated

r0x0d (commented):
Looks good, with the exception of my last comment.

convert2rhel/main.py
pr-watson: Must've been missed in the rebase, i'll revert

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,237 @@
...
+        deal with it.
+        """
+        super(RemoveEfiCentos, self).run()
+
+        if systeminfo.system_info.id != "centos":
+            logger.debug("Did not perform removal of EFI files - only related to CentOS Linux.")
+            # nothing to do
+            return
+        try:
+            os.rmdir(CENTOS_EFIDIR_CANONICAL_PATH)
```
hosekadam, line 196: I'm getting error when trying to remove the path. In the path are still `'grubenv', 'grub.cfg'`. When I switch `copy2` to `move` (remove the source) in `CopyGrubFiles`, the only file is `grubenv`. I think also this file can be removed, since it's mentioned in the docstring as something expected what could be removed.

pr-watson: Okay we can switch it to move then, is it worth to change the action name to MoveGrubFiles?

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,237 @@
...
+                id="NOT_INSTALLED_EFIBOOTMGR_UTILITY",
+                title="UEFI boot manager utility not found",
+                description="Couldn't find the UEFI boot manager which is required for us to install and verify a RHEL boot entry.",
+                remediations="Install the efibootmgr utility using the following command:\n\n 1. yum install efibootmgr",
+            )
+
+
+class CopyGrubFiles(actions.Action):
+    id = "COPY_GRUB_FILES"
+
```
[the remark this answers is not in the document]
pr-watson: Let's add this

convert2rhel/actions/conversion/set_efi_config.py
r0x0d: We have to make sure that both required files are not there any more before we commit to remove the directory. 

Do we have the confidence that the previous action will correctly move/copy the required files to the new folder and not leaving anything behind? If so, then calling `move` would be fine, I believe

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,237 @@
...
+        super(RemoveEfiCentos, self).run()
+
+        if systeminfo.system_info.id != "centos":
+            logger.debug("Did not perform removal of EFI files - only related to CentOS Linux.")
+            # nothing to do
+            return
+        try:
+            os.rmdir(CENTOS_EFIDIR_CANONICAL_PATH)
+        except (OSError, IOError) as err:
+            warning_message = "Failed to remove the {dir} directory as files still exist. During conversion we make sure to copy over files needed to their RHEL counterpart. However, some files we didn't expect likely exist in the directory that needs human oversight. Make sure that the files within the directory is taken care of and proceed with deleting the directory manually after conversion. We received error: '{err}'.".format(
```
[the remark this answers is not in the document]
r0x0d: Maybe we could switch to this?

```suggestion
            warning_message = (
                "Failed to remove the {dir} directory as files still exist."
                " During conversion we make sure to copy over files needed to their RHEL counterpart."
                " However, some files we didn't expect likely exist in the directory that needs human oversight."
                " Make sure that the files within the directory is taken care of and proceed with deleting the directory"
                " manually after conversion. We received error: '{err}'."
            ).format(dir=CENTOS_EFIDIR_CANONICAL_PATH, err=err)
```

In my opinion, this is easier to read without the `/` at the end of each line, and also, makes it easy to remove altogether if needed in the future.

convert2rhel/unit_tests/actions/conversion/set_efi_config_test.py
```diff
@@ -0,0 +1,241 @@
...
+    monkeypatch,
+    caplog,
+    pretend_os,
+    copy_grub_files_instance,
+):
+    monkeypatch.setattr(shutil, "copy2", mock.Mock())
+    shutil.copy2.side_effect = IOError(13, "Permission denied")
+    monkeypatch.setattr(os.path, "exists", mock.Mock(side_effect=[False, True, True, True, False, True, False]))
+    copy_grub_files_instance.run()
+    print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
```
hosekadam, line 189:
```suggestion
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,237 @@
...
+                level="WARNING",
+                id="NOT_REMOVED_CENTOS_UEFI_DIRECTORY",
+                title="CentOS UEFI directory couldn't be removed",
+                description=warning_message,
+            )
+
+
+class ReplaceEfiBootEntry(actions.Action):
+    id = "REPLACE_EFI_BOOT_ENTRY"
+    dependencies = ("REMOVE_EFI_CENTOS",)
```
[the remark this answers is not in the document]
r0x0d: The underlying code uses the efibootmgr binary, just to prevent any future problems, let's add this as a dependency as well

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,243 @@
...
+        except (OSError, IOError) as err:
+            warning_message = (
+                "Failed to remove the {dir} directory as files still exist."
+                " During conversion we make sure to copy over files needed to their RHEL counterpart."
+                " However, some files we didn't expect likely exist in the directory that needs human oversight."
+                " Make sure that the files within the directory is taken care of and proceed with deleting the directory"
+                " manually after conversion. We received error: '{err}'."
+            ).format(dir=CENTOS_EFIDIR_CANONICAL_PATH, err=err)
+                dir=CENTOS_EFIDIR_CANONICAL_PATH, err=err
+            )
```
r0x0d, line 206:
```suggestion
            warning_message = (
                "Failed to remove the {dir} directory as files still exist. During conversion we make sure to copy "
                "over files needed to their RHEL counterpart. However, some files we didn't expect likely exist in the "
                "directory that needs human oversight. Make sure that the files within the directory is taken care of "
                "and proceed with deleting the directory manually after conversion. We received error: '{err}'."
            ).format(dir=CENTOS_EFIDIR_CANONICAL_PATH, err=err)
```

convert2rhel/actions/conversion/set_efi_config.py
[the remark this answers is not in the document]
r0x0d: We don't really need the dependency there. Nothing in that action uses the efibootmgr utility, so it is fine that it runs earlier or after the copy of grub files, I think. 

We need to set this dependency on the actions that uses this in the underlying code, and I think this action is not one of those.

convert2rhel/actions/conversion/set_efi_config.py
hosekadam: I would say the `\` are needed there in the ends of the lines, when I was testing that it wasn't working correctly without them (prints just the first line without formatting)

convert2rhel/actions/conversion/set_efi_config.py
pr-watson: Just tested and I can see the move does take care of the grub.cfg file, we will need to also remove the leftover grubenv

hosekadam: The old code base had the same issue, during configuring the bootloader `WARNING - The folder /boot/efi/EFI/centos/ is left untouched. You may remove the folder manually after you ensure there is no custom data you would need.` is printed. 

But as I found by my experiments, in the folder are just `grub.cfg` and `grubenv` which I think we expect there and if in the folder are just those two files, it's fine to remove the folder. The `grub.cfg` we can `move` and the `grubenv`... we can remove manually?

>Do we have the confidence that the previous action will correctly move/copy the required files to the new folder and not leaving anything behind?

Since the `RemoveEfiCentos` has dependency `COPY_GRUB_FILES` I believe this won't run if the copy wouldn't be successfull.

convert2rhel/actions/conversion/set_efi_config.py
r0x0d: Could you share your test? Reproducing this in a python shell works correctly

r0x0d:
```python
>>> CENTOS_EFIDIR_CANONICAL_PATH="test"
>>> err="test2"
>>> warning_message = (
                "Failed to remove the {dir} directory as files still exist. During conversion we make sure to copy "
                "over files needed to their RHEL counterpart. However, some files we didn't expect likely exist in the "
                "directory that needs human oversight. Make sure that the files within the directory is taken care of "
                "and proceed with deleting the directory manually after conversion. We received error: '{err}'."
            ).format(dir=CENTOS_EFIDIR_CANONICAL_PATH, err=err)
>>> logger.info(warning_message)
INFO:__main__:Failed to remove the test directory as files still exist. During conversion we make sure to copy over files needed to their RHEL counterpart. However, some files we didn't expect likely exist in the directory that needs human oversight. Make sure that the files within the directory is taken care of and proceed with deleting the directory manually after conversion. We received error: 'test2'.
>>> print(warning_message)
Failed to remove the test directory as files still exist. During conversion we make sure to copy over files needed to their RHEL counterpart. However, some files we didn't expect likely exist in the directory that needs human oversight. Make sure that the files within the directory is taken care of and proceed with deleting the directory manually after conversion. We received error: 'test2'.
```

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,243 @@
...
+    RHEL_EFIDIR_CANONICAL_PATH,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+class NewDefaultEfiBin(actions.Action):
+    id = "NEW_DEFAULT_EFI_BIN"
+
```
hosekadam, line 35: I wonder if there shouldn't be some dependency as `FIX_DEFAULT_KERNEL` (which is the last action from `preserve_only_rhel_kernel.py` if I'm right) and the EFI actions seems related to this

convert2rhel/actions/conversion/set_efi_config.py
hosekadam: I see, the difference is the `()`, so `()` or the `\` is needed. Sorry

convert2rhel/actions/conversion/set_efi_config.py
r0x0d: Ah, cool! Alright, makes sense then

convert2rhel/actions/conversion/set_efi_config.py
pr-watson: We can add it

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,243 @@
...
+        super(ReplaceEfiBootEntry, self).run()
+
+        try:
+            grub.replace_efi_boot_entry()
+        except grub.BootloaderError as e:
+            self.set_result(
+                level="ERROR",
+                id="FAILED_TO_REPLACE_UEFI_BOOT_ENTRY",
+                title="Failed to replace UEFI boot entry to RHEL",
+                description="As the current UEFI bootloader entry could be invalid or missing we need to ensure that a RHEL UEFI entry exists. The UEFI boot entry could not be replaced due to the following error: '%s'"
```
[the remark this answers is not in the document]
r0x0d: please apply this one.

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,243 @@
...
+        """Replace the current UEFI bootloader entry with the RHEL one.
+
+        The current UEFI bootloader entry could be invalid or misleading. It's
+        expected that the new bootloader entry will refer to one of the standard UEFI binary
+        files provided by Red Hat inside the RHEL_EFIDIR_CANONICAL_PATH.
+        The new UEFI bootloader entry is always created / registered and set
+        set as default.
+
+        The current (original) UEFI bootloader entry is removed under some conditions
+        (see _remove_orig_boot_entry() for more info).
```
r0x0d, line 230:
```suggestion
        (see `py:grub._remove_orig_boot_entry()` for more info).
```

pr-watson:
Should we stick with the original intended approach of only removing the centos boot folder if the folder is empty @r0x0d @hosekadam ?

convert2rhel/actions/conversion/set_efi_config.py
hosekadam: @pr-watson @r0x0d Or... it depends. As I'm looking at the [original order](https://github.com/oamg/convert2rhel/blob/v1.7.1/convert2rhel/main.py#L299), it should be `transaction.py` > `preserve_only_rhel_kernel.py` > `list_non_red_hat_pgs_left.py` > `set_efi_config.py`. But the `list_non_red_hat_pgs_left.py` is missing its dependencies too, so we would have to add it (and it makes sense, or the `list_non_red_hat_pgs_left.py` could be possibly ran before the transaction which would be useless)

hosekadam:
> Should we stick with the original intended approach of only removing the centos boot folder if the folder is empty @r0x0d @hosekadam ?

I think when we are just copying the files, the action `RemoveEfiCentos` will always fail as we have it now - in the folder would be at least the `grub.cfg` and grubenv we expect to be in the folder.

I just tried it in vagrant with dependencies mentioned in https://github.com/oamg/convert2rhel/pull/1256#discussion_r1718100982 (the order from previous versions of codebase), removed the folder `/boot/efi/EFI/centos` (simulated it's empty with `shutil.rmtree`) and it works great, no warnings and system rebooted correctly.

WDYT @r0x0d ?

convert2rhel/actions/conversion/set_efi_config.py
hosekadam: If we decide to add the dependencies to `list_non_red_hat_pgs_left.py`, I can include it into the https://github.com/oamg/convert2rhel/pull/1341 where I'm touching this action

r0x0d:
> > Should we stick with the original intended approach of only removing the centos boot folder if the folder is empty @r0x0d @hosekadam ?
> 
> I think when we are just copying the files, the action `RemoveEfiCentos` will always fail as we have it now - in the folder would be at least the `grub.cfg` and `grubenv` we expect to be in the folder.
> 
> I just tried it in vagrant with dependencies mentioned in [#1256 (comment)](https://github.com/oamg/convert2rhel/pull/1256#discussion_r1718100982) (the order from previous versions of codebase), removed the folder `/boot/efi/EFI/centos` (simulated it's empty with `shutil.rmtree`) and it works great, no warnings and system rebooted correctly.
> 
> WDYT @r0x0d ?
> 
> EDIT: I see you already changed the copy to move, forgot to pull the changes 🤦 But still, in the folder is the `grubenv`

I don't think this is necessary to change right now. I would prefer to be closer to the original implementation for the time being, and of course, in the future improve it when we have more time. 

The reason for that is that we are on a tight schedule and I don't think this change will improve it too much right now. Definitely it won't be worse, thus, I would prefer to have a ticket to have a closer look at those actions after the release and try to improve them somehow. 

For beyond this tight release schedule, @hosekadam, I think this will cause no harm to the system as we always left this folder there. I also keep thinking if the users might have custom settings/configs in those files, and if we deliberately remove the files and the folder altogether, they might lose it.

hosekadam:
> I also keep thinking if the users might have custom settings/configs in those files, and if we deliberately remove the files and the folder altogether, they might lose it.

I meant to remove the files we expect (grub.cfg and grubenv) and try to remove the folder. If anything is in it, it won't be removed.

But I agree, it's not a big deal to leave the folder on the system and the schedule is tight.

So this can stay as it is, the only thing I would change are the dependencies to have the right order of actions. I wrongly suggested the `preserve_only_rhel_kernel`, more info: https://github.com/oamg/convert2rhel/pull/1256#discussion_r1718100982

convert2rhel/actions/conversion/set_efi_config.py
r0x0d: Thanks for spotting the missing dependency, @hosekadam. Here's the PR that fixes the list_non_red_hat_pkgs_left.py: https://github.com/oamg/convert2rhel/pull/1342

pr-watson: Okay so I should remove the new dependency here?

r0x0d: So, as we discussed in the demo presentation about action framework a few weeks (months) ago, we don't strictly need every dependency to be in correct order. 

For example, the actions here doesn't need to have a dependency for `list_non_red_hat_pkgs_left`, as this is something that only makes sense for `transaction.py` actions. 

But regarding the original order, and you point, I think you're right, we want to add the dependencies at least for the preserve_only_rhel_kernel, and the preserve_only_rhel_kernel should have a dependency on transaction. 

I would vote for having this fixed in this PR, in that case, make the `preserve_only_rhel_kernel` a dependency of `set_efi_configuration`

pr-watson: Alright so the dependency we have now will satisfy that

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def run(self):
+        """Check that the expected RHEL UEFI binaries exist."""
+        super(NewDefaultEfiBin, self).run()
+
+        logger.task("Convert: Configure the bootloader")
+
+        if not grub.is_efi():
+            logger.info(
+                "Unable to collect data about UEFI on a BIOS system, did not perform check of" " RHEL UEFI binaries."
```
hosekadam, line 45:
```suggestion
                "Unable to collect data about UEFI on a BIOS system, did not perform check of RHEL UEFI binaries."
```

danmyway (approved):
Tests look good from QE POV

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,273 @@
...
+                logger.debug(
+                    "The %s file already exists in %s folder. Moving skipped."
+                    % (os.path.basename(src_file), RHEL_EFIDIR_CANONICAL_PATH)
+                )
+                continue
+
+            logger.info("Moving '%s' to '%s'" % (src_file, dst_file))
+
+            try:
+                shutil.move(src_file, dst_file)
```
hosekadam, line 177: I'm thinking if we shouldn't go back to the `copy2`, since it doesn't make a big difference (the `/boot/efi/EFI/centos` won't be removed) and as we discussed here https://github.com/oamg/convert2rhel/pull/1256#issuecomment-2291248107 we can refactor this later and for now we should be closer to the original code.

If we decide to move, sorry Preston for an extra work you had with my ideas

pr-watson: Yes we're sticking with `move`, it's a good addition

r0x0d (approved):
Let's merge after the required actions passes

convert2rhel/actions/conversion/set_efi_config.py
```diff
@@ -0,0 +1,269 @@
...
+            self.set_result(
+                level="ERROR",
+                id="NOT_INSTALLED_EFIBOOTMGR_UTILITY",
+                title="UEFI boot manager utility not found",
+                description="Couldn't find the UEFI boot manager which is required for us to install and verify a RHEL boot entry.",
+                remediations="Install the efibootmgr utility using the following command:\n\n 1. yum install efibootmgr",
+            )
+
+
+class MoveGrubFiles(actions.Action):
```
hosekadam, line 100: Rename this back to copy, since we use `copy2`

r0x0d:
rhel 7 rpm not being built is not a PR problem. It is happening in other PRs as well.

PR #1266, opened 2024-06-18
[RHELC-1334] Port ModifiedRPMFilesDiff to Action framework

hosekadam:
This new action provides information about difference between 'rpm -Va' from before and after the conversion.

Durign this port was needed to change the logging in the SystemInfo class to use global provided logging instead of the one inside of the class. Which was overall the same, just created dependency for logging to be inside of the class, or in the systeminfo.py would be duplicity for logging.

Jira Issues:

-[RHELC-1334](https://issues.redhat.com/browse/RHELC-1334)

Checklist

### DIFF
--- a/convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
+++ b/convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
@@ -1,0 +1,82 @@
+# Copyright(C) 2024 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+
+import difflib
+import logging
+import os
+
+from convert2rhel import actions, utils
+from convert2rhel.logger import LOG_DIR
+from convert2rhel.systeminfo import system_info
+from convert2rhel.toolopts import POST_RPM_VA_LOG_FILENAME, PRE_RPM_VA_LOG_FILENAME
+
+
+logger = logging.getLogger(__name__)
+
+
+class ModifiedRPMFilesDiff(actions.Action):
+    id = "MODIFIED_RPM_FILES_DIFF"
+
+    def run(self):
+        """
+        Get a list of modified rpm files after the conversion and
+        compare it to the one from before the conversion.
+        """
+        super(ModifiedRPMFilesDiff, self).run()
+
+        logger.task("Final: Show RPM files modified by the conversion")
+
+        system_info.generate_rpm_va(log_filename=POST_RPM_VA_LOG_FILENAME)
+
+        pre_rpm_va_log_path = os.path.join(LOG_DIR, PRE_RPM_VA_LOG_FILENAME)
+        if not os.path.exists(pre_rpm_va_log_path):
+            logger.info("Skipping comparison of the 'rpm -Va' output from before and after the conversion.")
+            self.add_message(
+                level="INFO",
+                id="SKIPPED_MODIFIED_RPM_FILES_DIFF",
+                title="Skipped comparison of 'rpm -Va' output from before and after the conversion.",
+                description="Comparison of 'rpm -Va' output was skipped due missing output "
+                "of the 'rpm -Va' run before the conversion.",
+                diagnosis="This is caused mainly by using '--no-rpm-va' argument for convert2rhel.",
+            )
+            return
+
+        pre_rpm_va = utils.get_file_content(pre_rpm_va_log_path, True)
+        post_rpm_va_log_path = os.path.join(LOG_DIR, POST_RPM_VA_LOG_FILENAME)
+        post_rpm_va = utils.get_file_content(post_rpm_va_log_path, True)
+        modified_rpm_files_diff = "\n".join(
+            difflib.unified_diff(
+                pre_rpm_va,
+                post_rpm_va,
+                fromfile=pre_rpm_va_log_path,
+                tofile=post_rpm_va_log_path,
+                n=0,
+                lineterm="",
+            )
+        )
+
+        if modified_rpm_files_diff:
+            logger.info(
+                "Comparison of modified rpm files from before and after the conversion:\n%s" % modified_rpm_files_diff
+            )
+            self.add_message(
+                level="INFO",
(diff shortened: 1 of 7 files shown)


convert2rhel/systeminfo.py
```diff
@@ -549,5 +500,26 @@ def is_systemd_managed_service_running(service):
...
+    logger.info(
+        "Running the 'rpm -Va' command which can take several"
+        " minutes. It can be disabled by using the"
+        " --no-rpm-va option."
+    )
+    rpm_va, _ = utils.run_subprocess(["rpm", "-Va"], print_output=False)
+    output_file = os.path.join(LOG_DIR, log_filename)
+    utils.store_content_to_file(output_file, rpm_va)
+    logger.info("The 'rpm -Va' output has been stored in the %s file." % output_file)
+
```
r0x0d, line 522: To minimize the changes to the systeminfo that doesn't seem related to the code, why don't we keep this inside the class and only call system_info.generate_rpm_va() inside the action classs? That would minimize all of those logger changes

r0x0d: To expand a bit on this, I think this is a good change, moving the logger instance to be outside the class is definitely a way to go. But it also feels unrelated to the scope of the ticket. We could open a new PR that patches this as well.

convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
```diff
@@ -0,0 +1,82 @@
...
+            logger.info("Skipping comparison of the 'rpm -Va' output from before and after the conversion.")
+            self.set_result(
+                level="SKIP",
+                id="SKIPPED_MODIFIED_RPM_FILES_DIFF",
+                title="Skipped comparison of 'rmp -Va' output from before and after the conversion.",
+                description="Comparison of 'rpm -Va' output was skipped due missing output "
+                "of the 'rpm -Va' run before the conversion.",
+                diagnosis="This is caused mainly by using '--no-rpm-va' argument for convert2rhel.",
+            )
+            return
```
r0x0d, line 56: I'm not sure if the use of the `skip` level here is the following the same pattern. We usually set `SKIP` automatically if any action dependant on this one fails. I think I will to double check.

r0x0d (commented):
Looks good. Some minors coments here and there. Also, will need to double check the usage of level = `SKIP`.

convert2rhel/systeminfo.py
hosekadam: Okay. I wanted to take the `generate_rpm_va` outside the class. But you are right, this is unrelated to the ticket. I'll open a new PR with the changes.

convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
hosekadam: Soo, it should be `INFO`?

hosekadam: I see the info cannot be the result. From the possible values (success, skip, fail) I like the most skip. But not sure about the ussage you are mentioning

hosekadam:
Applied the changes - no big changes to the systeminfo, porting (better say writing new) unit tests. Still WIP, will continue after PTO

convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
r0x0d: I commented in our slack channel about that. For now, it is totally fine to have this as skip

convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
```diff
@@ -0,0 +1,82 @@
...
+
+        system_info.generate_rpm_va(log_filename=POST_RPM_VA_LOG_FILENAME)
+
+        pre_rpm_va_log_path = os.path.join(LOG_DIR, PRE_RPM_VA_LOG_FILENAME)
+        if not os.path.exists(pre_rpm_va_log_path):
+            logger.info("Skipping comparison of the 'rpm -Va' output from before and after the conversion.")
+            self.set_result(
+                level="SKIP",
+                id="SKIPPED_MODIFIED_RPM_FILES_DIFF",
+                title="Skipped comparison of 'rmp -Va' output from before and after the conversion.",
```
r0x0d, line 51:
```suggestion
                title="Skipped comparison of 'rpm -Va' output from before and after the conversion.",
```

convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
```diff
@@ -0,0 +1,82 @@
...
+            logger.info("Skipping comparison of the 'rpm -Va' output from before and after the conversion.")
+            self.add_message(
+                level="INFO",
+                id="SKIPPED_MODIFIED_RPM_FILES_DIFF",
+                title="Skipped comparison of 'rpm -Va' output from before and after the conversion.",
+                description="Comparison of 'rpm -Va' output was skipped due missing output "
+                "of the 'rpm -Va' run before the conversion.",
+                diagnosis="This is caused mainly by using '--no-rpm-va' argument for convert2rhel.",
+            )
+            return
```
r0x0d, line 56: Just to avoid having different messages in the waring and the message.

```suggestion
            title = "Skipping comparison of the 'rpm -Va' output from before and after the conversion."
            logger.info(title)
            self.add_message(
                level="INFO",
                id="SKIPPED_MODIFIED_RPM_FILES_DIFF",
                title=title,
                description="Comparison of 'rpm -Va' output was skipped due missing output "
                "of the 'rpm -Va' run before the conversion.",
                diagnosis="This is caused mainly by using '--no-rpm-va' argument for convert2rhel.",
            )
            return
```

convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
```diff
@@ -0,0 +1,82 @@
...
+            logger.info(
+                "Comparison of modified rpm files from before and after the conversion:\n%s" % modified_rpm_files_diff
+            )
+            self.add_message(
+                level="INFO",
+                id="FOUND_MODIFIED_RPM_FILES",
+                title="Modified rpm files from before and after the conversion were found.",
+                description="Comparison of modified rpm files from before and after "
+                "the conversion: \n%s" % modified_rpm_files_diff,
+            )
```
r0x0d, line 82:
```suggestion
            description = "Comparison of modified rpm files from before and after the conversion:\n%s" % modified_rpm_files_diff
            logger.info(description)
            self.add_message(
                level="INFO",
                id="FOUND_MODIFIED_RPM_FILES",
                title="Modified rpm files from before and after the conversion were found.",
                description=description,
            )
```

r0x0d (approved):
Looks good to me

r0x0d (approved):

convert2rhel/actions/post_conversion/modified_rpm_files_diff.py
hosekadam: Yeah, makes sense, I'll fix that in different PR, since this is already merged

hosekadam: https://github.com/oamg/convert2rhel/pull/1333

PR #1267, opened 2024-06-18
[RHELC-1606] Fix of multiple backup of one file

hosekadam:
File can be present multiple times in the output of 'rpm -Va' command. The file is backed up only once.

Also, the unit test was updated to handle this situation.

Jira Issues:

- [RHELC-1606](https://issues.redhat.com/browse/RHELC-1606)

Checklist

### DIFF
--- a/convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
+++ b/convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
@@ -286,6 +286,13 @@ class TestBackupSystem:
                 ],
                 (False, True, False, False, True),
             ),
+            (
+                [
+                    "S.5.?..T.     c   {path}/filename",
+                    "S.5.?..T.     c   {path}/filename",
+                ],
+                (True, True),
+            ),
         ),
     )
     def test_backup_package_file_complete(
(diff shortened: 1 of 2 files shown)


convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -297,37 +304,45 @@ def test_backup_package_file_complete(
         backup_package_files_action,
         global_backup_control,
     ):
-        # Prepare the rpm -va ouput to the PRE_RPM_VA_LOG_FILENAME file
+        ### Prepare the 'rpm -Va' ouput to the PRE_RPM_VA_LOG_FILENAME file
```
Venefilyn, line 307: Since I'm a bit unclear about the triple hashes. What is the significance of `#` vs `###`?

Venefilyn (approved):
Seems mostly fine!

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
r0x0d: Probably just some typo or re-commenting the same comment two more times

Venefilyn: I thought so too but then saw a comment else where in the PR with

```
### Prepare the files from the 'rpm -Va' output by creating them
# Use only unique paths, since one path can be present multiple times in the output
# https://issues.redhat.com/browse/RHELC-1606
```

hosekadam: The plan was to create some headings to each part of the test. But I think it wasn't good plan :D I'll remove that, since it's unclear

Venefilyn: No it's fine but was curious of the reason why. Wonder if we can make it more clear somehow..

I think when it's like this of a heading and comments underneath I'd go with this or something like it
```
# Heading
# 
# Text here
# Text here
```

Venefilyn:
@hosekadam I had forgotten that I even reviewed this, but I created a solution too. What is stopping us from doing something like this?

https://github.com/Venefilyn/convert2rhel/commit/1e480493e74b7f2c672bd876348b26baf1c88bf8

hosekadam:
@Venefilyn looking at it, I would say they are really similar. Maybe the mine solution is easier to read (but it's probably because I did that :D) - like just adding the backed up path to the list of already backed up paths (which we already use, just extending it).

And yours where we have in the paths to back up just unique paths. Maybe the memory usage will be lower in this case.

Just one case came to my mind - what if the status or type of file would differ, but the path would be the same? In that case it would be added to the list, right? But maybe this is a too much corner case...

But I would definitely use the unit tests from this PR, which checks the complete process of the backup and rollback, I think the unit test is good because it tests the real usage (and I spent some time on it :D). But I understand, with your solution it's probably not needed so much.

Venefilyn:
> what if the status or type of file would differ, but the path would be the same?

Good question. Feels like that would never occur though :thinking: 

@bookwar @r0x0d  WDYT

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -342,17 +357,27 @@ def test_backup_package_file_complete(
...
+                # Remove the original file
+                try:
+                    os.remove(original_file_path)
+                    removed_paths.append(original_file_path)
+                # FileNotFound on Python 3+, due compatibility with Python 2.7 using OSError
+                except OSError:
+                    # If the path is present multiple times in the 'rpm -Va' output
+                    # it's possible, the path was already removed. If not, that's fail.
+                    if original_file_path not in removed_paths:
+                        assert False
```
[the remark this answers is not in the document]
hosekadam: @Venefilyn I checked this and the comments `###` to have this prepared. So both of the commits are ready and we can discuss which we will merge.

Comments are updated, but I don't think this can be changed. I'm also checking if the path was really removed and the exception doesn' have to be raised each time.

r0x0d:
> > what if the status or type of file would differ, but the path would be the same?
> 
> Good question. Feels like that would never occur though 🤔
> 
> @bookwar @r0x0d WDYT



> > what if the status or type of file would differ, but the path would be the same?
> 
> Good question. Feels like that would never occur though 🤔
> 
> @bookwar @r0x0d WDYT

I would merge the two PRs in this one. I prefer @Venefilyn solution as it is simpler and does not require always updating a list after each new backed up file. Ideally, we should receive the data in that function already prepared to parse and backup. 

But I agree with @hosekadam to keep his unit_tests as it is testing the whole situation. I would try to look for a way to split the test or simplify if possible. The test itself is very difficult to read :thinking:. 

My vote here is then to pick @Venefilyn solution and @hosekadam unit_tests.

Venefilyn:
We decided to merge this and my PR so closing this one

PR #1272, opened 2024-06-19
[RHELC-1507] Port environment variables to config file

r0x0d:
This patch introduces the initial work to port the environment variables
to be read by a configuration file, parsed, and put into the ToolOpts
class.

With that, we are aiming to remove the environments variables completely
in the next release, managing all the configurations/inhibitors/settings
through that config file.

Jira Issues:

- [RHELC-1507](https://issues.redhat.com/browse/RHELC-1507)

Checklist


convert2rhel/toolopts.py
```diff
@@ -515,70 +539,118 @@ def _process_cli_options(self):
...
-    # Otherwise it won't be used
-    supported_opts = {"username": None, "password": None, "activation_key": None, "org": None}
+    # Paths for the configuration files. In case we have cfg_path defined
+    # (meaning that the user entered something through the `-c` option), we
+    # will use only that, as it has a higher priority over the rest
+    config_paths = [cfg_path] if cfg_path else CONFIG_PATHS
+    paths = [os.path.expanduser(path) for path in config_paths if os.path.exists(os.path.expanduser(path))]
+
+    if cfg_path and not paths:
+        raise FileNotFoundError("No such file or directory: %s" % ", ".join(paths))
```
hosekadam, line 576: This would be problem on systems with python 2.7 - this exception is not defined. I would go with
`raise OSError(2, "No such file or directory: '%s'" % cfg_path)`

convert2rhel/toolopts.py
```diff
@@ -515,70 +539,118 @@ def _process_cli_options(self):
...
+    found_opts = _parse_options_from_config(paths)
+    return found_opts
 
+
+def _parse_options_from_config(paths):
+    """Parse the options from the given config files.
+
+    .. note::
+        If no configuration file is provided through the command line option
+        (`-c`), we will use the default paths and follow their priority.
```
hosekadam, line 587: I think this comment is not related to the function - this just parse what is given to it from `options_from_config_files`

convert2rhel/toolopts.py
```diff
@@ -515,70 +539,118 @@ def _process_cli_options(self):
...
-    headers = ["subscription_manager"]  # supported sections in config file
-    # Create dict with all supported options, all of them set to None
-    # needed for avoiding problems with files priority
-    # The name of supported option MUST correspond with the name in ToolOpts()
-    # Otherwise it won't be used
-    supported_opts = {"username": None, "password": None, "activation_key": None, "org": None}
+    # Paths for the configuration files. In case we have cfg_path defined
+    # (meaning that the user entered something through the `-c` option), we
+    # will use only that, as it has a higher priority over the rest
+    config_paths = [cfg_path] if cfg_path else CONFIG_PATHS
```
hosekadam, line 572: I'm not sure about this. I think if the config path is specified, it should have the highest priority, but the lower ones should be also loaded.

convert2rhel/toolopts.py
```diff
@@ -515,70 +539,118 @@ def _process_cli_options(self):
...
+def _parse_options_from_config(paths):
+    """Parse the options from the given config files.
+
+    .. note::
+        If no configuration file is provided through the command line option
+        (`-c`), we will use the default paths and follow their priority.
+
+    :param paths: List of paths to iterate through and gather the options from
+        them.
+    :type paths: list[str]
```
hosekadam, line 591:
```suggestion
    :return: Return a dict of loaded values under all headers from all the config files with solved priority
    :rtype: dict[str, str]
```
I would add this comment about return value

convert2rhel/toolopts.py
```diff
@@ -515,70 +539,118 @@ def _process_cli_options(self):
...
+    """Helper function to iterate through the options in a config file.
+
+    :param config_file: An instance of `py:ConfigParser` after reading the file
+        to iterate through the options.
+    :type config_file: configparser.ConfigParser
+    :param header: The header name to get options from.
+    :type header: str
+    :param supported_opts: List of supported options that can be parsed from
+        the config file.
+    :type supported_opts: list[str]
```
hosekadam, line 627:
```suggestion
    :type supported_opts: list[str]
    :return: Dict of keys and values loaded from the config file under the provided header
    :rtype: dict[str, str]
```

convert2rhel/toolopts.py
r0x0d: good point! thanks for suggestion

convert2rhel/toolopts.py
r0x0d: It makes sense. My idea here was that if the user has defined their values inside the config file they will provide, then it would be unnecessary for us to go through all of the possibilities just to override all in the end with their choice in the custom config file. 

But before I left for PTO, we discussed the possibility of having multiple config files in a folder, this way the user would be able to separate subman secrets from overridable options, so reverting this line and making it go from bottom-top in the for loop actually makes more sense now.

convert2rhel/toolopts.py
r0x0d: Removed the comment

convert2rhel/toolopts.py
```diff
@@ -515,70 +539,122 @@ def _process_cli_options(self):
...
+        loggerinst.debug("No options found for %s. It seems to be empty or commented." % header)
+        return options
+
+    for option in conf_options:
+        if option.lower() not in supported_opts:
+            loggerinst.warning("Unsupported option '%s' in '%s'" % (option, header))
+            continue
+
+        options[option] = config_file.get(header, option).strip('"')
+        loggerinst.debug("Found %s in %s" % (option, header))
```
r0x0d, line 646: @Venefilyn, I'm thinking about the way we save the value associated with the option here... I was about to use a `strip('"')` to remove any trailing `"` we found in the option, but I'm afraid that this could cause problems when dealing with subman secrets. 

Take a look at the following example: 

```
>>> a = '"aushs'
>>> a.strip('"')
'aushs'
>>> a = '"aushs"'
>>> a.strip('"')
'aushs'
>>> a = '"aus"hs"'
>>> a.strip('"')
'aus"hs'
>>> a = '"aus"h"s"'
>>> a.strip('"')
'aus"h"s'
```

When we have a `"` in the middle of the string, we don't accidently remove it, so if the password or any other important secret has a `"` in the middle of it, we would parse normally. Whoever, if the secret starts with only a `"` (as the first example there) we would remove it and make it incorrect. 

Currently, when we parse the value we would store them with the `"` in our dictionary, so the following config:

```config
[subscription_manager]
password = "test12345"
```

would become something like: 

```python
{"subscription_manager": {"password": '"test12345"'}}
```

For that reason, I think it is better to keep the current parser the way it is... What do you think? In case that they have those type of special characters, we would not mess with it, making it more realible maybe? At this point, if something fails, it would be easier to tell the customer that they need to remove the trailing `"` from their string rather than trying to be smart with this, right?

r0x0d: @hosekadam what is your input on this as well?

hosekadam: I as the customer would be happy if I don't have to care about the formatting much - so we should accept both for having the best user experience.

But I as developer would be also fine with the approach having it as it is and do not make any changes to the values. Just use what is in the config file. And also not adding the complexity to the code.

When I put those feelings together, I think the compromise would be removing the `"` symbol only if it's present in the start and in the end of the string. In that case, it would be a big coincidence if the `"` would be the real part of the string. We can use some regex to check if the is in the start and end and `if`, so I believe it won't increase the complexity of the code.

Anyway, let's wait for another inputs for this.

r0x0d: Yeah, after thinking more about that, I really think that we should go with the way it was. It's a bit scary coming up with a sophisticated parser without need, just to guarantee that we are removing correctly `"`. I think that I prefer to leave as-is and parse the user data from that file as they added it. Less room for errors.

jochapma: Maybe you could use `ast.literal_eval()`?

jochapma (approved):
Looks good to me.

convert2rhel/toolopts.py
Venefilyn: Why do we store it as `password = "test12345"` in our config file to begin with? Can't we just get rid of the quotes there. It doesn't make sense to parse anything out, especially not passwords. You could set your own password anywhere that would both start and end with double-quote (")

convert2rhel/toolopts.py
```diff
@@ -87,13 +102,22 @@ def __init__(self):
...
         self.activity = None
 
+        # Settings
+        self.incomplete_rollback = None
+        self.tainted_kernel_module_check_skip = None
+        self.outdated_package_check_skip = None
+        self.allow_older_version = None
+        self.allow_unavailable_kmods = None
+        self.configure_host_metering = None
+        self.skip_kernel_currency_check = None
```
Venefilyn, line 112: I'd say apply at class level instead of inside init. Then add stuff like this (haven't verified what it should be)
```suggestion
        # Settings
        self.incomplete_rollback = None  # type: List[str]|None
        self.tainted_kernel_module_check_skip = None  # type: str|None
        self.outdated_package_check_skip = None  # type: str|None
        self.allow_older_version = None  # type: str|None
        self.allow_unavailable_kmods = None  # type: str|None
        self.configure_host_metering = None#   type: str|None
        self.skip_kernel_currency_check = None  # type: str|None
```

r0x0d: We can't initialize them at that moment. Our code does not handle that very well right now, so it is better to have then as `None` instead of `0`. 

https://github.com/oamg/convert2rhel/pull/1285/files this would be the next step to get them initialized and properly checked

r0x0d: Regarding moving them to be a class level instead of under init, what would be the advantage of that? I'm personally not much a fan of that approach, but maybe there is a better reason to that and I'm not aware of.

r0x0d: Added type comments

Venefilyn:
Tests failing with two different errors depending on the run

```
[2024-08-12T16:58:38+0000] DEBUG - Found password in subscription_manager

WARNING - Passing the RHSM password or activation key through the --activationkey or --password options is insecure as it leaks the values through the list of running processes. We recommend using the safer --config-file option instead.

WARNING - Either a password or an activation key can be used for system registration. We're going to use the activation key.

CRITICAL - Either the --org or the --activationkey option is missing. You can't use one without the other.

Either the --org or the --activationkey option is missing. You can't use one without the other.

FAILED
        TEARDOWN F convert2rhel
```

```
System registration failed with error: com.redhat.RHSM1.Error: {"exception": "RestlibException", "severity": "error", "message": "Organization config_org does not exist."}
CRITICAL - Unable to register the system through subscription-manager.
ERROR - (ERROR) SUBSCRIBE_SYSTEM::FAILED_TO_SUBSCRIBE_SYSTEM - Failed to subscribe system.
 Description: After several attempts, convert2rhel was unable to subscribe the system using subscription-manager. This issue might occur because of but not limited to DBus, file permission-related issues, bad credentials, or network issues.
 Diagnosis: System registration failed with error com.redhat.RHSM1.Error: {"exception": "RestlibException", "severity": "error", "message": "Organization config_org does not exist."}.
 Remediations: N/A

ERROR - Skipped ENSURE_KERNEL_MODULES_COMPATIBILITY. Skipped because SUBSCRIBE_SYSTEM was not successful
ERROR - Skipped VALIDATE_PACKAGE_MANAGER_TRANSACTION. Skipped because ENSURE_KERNEL_MODULES_COMPATIBILITY and SUBSCRIBE_SYSTEM were not successful
CRITICAL - The conversion process failed.
```

tests/integration/tier0/non-destructive/config-file/test_config_file.py
```diff
@@ -2,64 +2,107 @@
...
+
+    _remove_files(config_files)
+
+
+@pytest.mark.parametrize(
+    "set_up_config_files",
+    [
+        [
+            Config(
+                "~/.convert2rhel_custom.ini",
```
kokesak, line 39: It is a small thing, but I think we can try to have a custom config file name that douesn't contain the convert2rhel in its name. Let's try something like

```suggestion
                "~/.custom_foo_config.ini",
```

tests/integration/tier0/non-destructive/config-file/test_config_file.py
```diff
@@ -70,47 +113,75 @@ def test_config_cli_priority(convert2rhel):
...
-        Config("/etc/convert2rhel.ini", "[subscription_manager]\nactivation_key = config2_activationkey"),
-    ]
-    create_files(config)
 
+@pytest.mark.parametrize(
+    "set_up_config_files",
+    [
+        [
+            Config(
+                "~/.convert2rhel.ini",
```
kokesak, line 122: Just to be sure, I would put an exact path
```suggestion
                "/root/.convert2rhel.ini",
```

tests/integration/tier0/non-destructive/config-file/test_config_file.py
```diff
@@ -70,47 +113,75 @@ def test_config_cli_priority(convert2rhel):
...
+def test_config_standard_paths_priority_diff_methods(convert2rhel, set_up_config_files):
     with convert2rhel("--debug") as c2r:
-        c2r.expect("DEBUG - Found password in /root/.convert2rhel.ini")
-        c2r.expect("DEBUG - Found activation_key in /etc/convert2rhel.ini")
+        c2r.expect("DEBUG - Checking configuration file at /root/.convert2rhel.ini", timeout=120)
+        c2r.expect("DEBUG - Found password in subscription_manager")
+        c2r.expect("DEBUG - Found activation_key in subscription_manager")
+
         c2r.expect(
             "WARNING - Passing the RHSM password or activation key through the --activationkey or --password options"
```
kokesak, line 146: It doesn't seem correct to me. The c2r reports "Passing the RHSM password or activation key through the --activationkey or --password options is insecure..." but we aer calling convrt2rhel only with the `--debug` option, @danmyway, @r0x0d?

kokesak (commented):
Good changes on how we handle config inside the integration tests. Thanks!

tests/integration/tier0/non-destructive/config-file/test_config_file.py
r0x0d: Yeah, it feels off as well but this debug log has been there since forever. 

@kokesak, the ticket got delayed for next version, so I will have more time to revisit it and rework some of those things.

r0x0d:
Delayed to v2.2.0. Need more time to work on a couple of bugs and things that are off in the original implementation.

r0x0d:
There is one failing test that will depend on https://github.com/oamg/convert2rhel/pull/1373 and https://github.com/oamg/convert2rhel/pull/1327

config/convert2rhel.ini
```diff
@@ -11,3 +11,14 @@
 # password       = <insert_password>
 # activation_key = <insert_activation_key>
 # org            = <insert_org>
+
+[host_metering]
+# configure_host_metering          = "0"
```
kokesak, line 16: Just a question, but shouldn't the host metering value be either "auto" or "force"?

config/convert2rhel.ini
```diff
@@ -11,3 +11,14 @@
...
+
+[inhibitor_overrides]
+# incomplete_rollback              = "0"
```
kokesak, line 19: Also, I think that c2r does not care what value is present in the env var (it is enough just to be set to any value), but I think we should use the value "1" in the examples, as it can be more intuitive. WDYT?
```suggestion
# incomplete_rollback              = "1"
```

kokesak (commented):
Tests are passing. But I think we can extend our test suite. In integration tests that are verifying a certain inhibitor and ability to override it, we could also introduce a variant that puts the override value to the config file instead of the environment. I wouldn't block this PR on this - we can add this in the integration test ticket - https://issues.redhat.com/browse/RHELC-1535

config/convert2rhel.ini
Venefilyn: We could always change it to need a truthy value, i.e. `1`, `true`, `yes`

Venefilyn (commented):
Overall looks great!

config/convert2rhel.ini
r0x0d: I'm inclined to change this to a boolean value to have it forced in convert2rhel to operate correctly from now on that this is set. 

I'm doing a minimal conversion from "0" and "1" to true/false, but I don't like that solution that much... I will see about converting them to boolean values when reading the config

config/convert2rhel.ini
r0x0d: Hmmm. That's true, not sure why I changed to 0 or 1... Let me fix that

kokesak: Thanks!

config/convert2rhel.ini
r0x0d: Okay, changed the code to parse the options from inhibitor_overrides as booleans directly. 

From the configparse docs it seems that the accepted values are `on/off, yes/no, 1/0, true/false`. We should be safer as any of those get converted to bool right away and we will be able to enforce it properly in the codebase later.

kokesak: If the tests pass now, I'm ok with merging it

kokesak (approved):

Venefilyn (approved):

PR #1274, opened 2024-06-20
chore: improve sphinx workflow with composite action

Venefilyn:
This changes the Sphinx GitHub action to use a composite action that does not install Sphinx on its own, but rather takes care of comments so that we can more easily run everything as we want

We also add packages using apt instead of pip, as we otherwise ran into some weird issues. But I've tested on my branch and works fine

See
https://venefilyn.github.io/convert2rhel

Jira Issues:

-

Checklist

### DIFF
--- a/.github/workflows/main-docs-deploy.yml
+++ b/.github/workflows/main-docs-deploy.yml
@@ -32,10 +32,17 @@ jobs:
         uses: actions/checkout@v4
       - name: Setup Pages
         uses: actions/configure-pages@v5
-      - uses: ammaraskar/sphinx-action@dev
+      - name: Install system dependencies
+        uses: awalsh128/cache-apt-pkgs-action@latest
+        with:
+          packages: python3-pytest python3-sphinx python3-sphinx-autodoc-typehints python3-pexpect python3-six python3-rpm python3-hawkey python3-dnf python3-dbus python3-pytest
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -20,6 +20,7 @@
     "python": ("https://docs.python.org/3/", None),
 }
 
+autodoc_mock_imports = ["dnf", "yum", "rpm", "hawkey", "dbus", "pexpect"]
```
r0x0d, line 23: Can we have a link to some docs that explain what the autodoc_mock do?

Venefilyn: @r0x0d Added!

.github/workflows/main-docs-deploy.yml
```diff
@@ -32,10 +32,17 @@ jobs:
         uses: actions/checkout@v4
       - name: Setup Pages
         uses: actions/configure-pages@v5
-      - uses: ammaraskar/sphinx-action@dev
+      - name: Install system dependencies
+        uses: awalsh128/cache-apt-pkgs-action@latest
```
r0x0d, line 36: What is the difference between using this thirdy-party action and the official actions/cache from github?

r0x0d (approved):

PR #1275, opened 2024-06-21
[RHELC-1293] Automatically assign pytest markers, ignore warnings

danmyway:
* assign pytest markers automatically
  * https://docs.pytest.org/en/latest/example/markers.html#automatically-adding-markers-based-on-test-names
* ignore PytestUnknownWarning on custom markers starting with test_
  * https://docs.python.org/3/library/warnings.html#warnings.filterwarnings
* remove unused imports
* remove custom markers from both function decorators and pytest.ini
* change some test function names 
* remove a kernel module test verifying, that unloading kernel module will not yield previously appearing inhibitor, as it has litttle to no value

Jira Issues:

- [RHELC-1293](https://issues.redhat.com/browse/RHELC-1293)

Checklist


tests/integration/tier1/destructive/firewalld-disabled-ol8/test_firewalld_disabled_ol8.py
```diff
@@ -8,7 +8,6 @@
 FIREWALLD_CONFIG_FILE = "/etc/firewalld/firewalld.conf"
 
 
-@pytest.mark.test_firewalld_disabled
 def test_firewalld_disabled_ol8(shell, convert2rhel):
```
Venefilyn, line 11: What's the reason for removing all the markers? Are they not needed?

r0x0d: Yeah, as per documentation it is supposed to be done automatically by pytest via the hook they have: https://docs.pytest.org/en/latest/example/markers.html#automatically-adding-markers-based-on-test-names

danmyway: Yep, it is indeed, as Rodolfo says.
I'm applying the suggestion from #1018 
There is also a way to automatically register the marker for the sake of the run during the `pytest_sessionstart()` test collection, but that I see as unnecessary.
Simply ignoring the markers based on the custom prefix and registering those without one is good enough.

tests/integration/conftest.py
```diff
@@ -29,6 +31,23 @@
 logging.basicConfig(level=os.environ.get("DEBUG", "INFO"), stream=sys.stderr)
 logger = logging.getLogger(__name__)
 
+
+def pytest_collection_modifyitems(items):
```
kokesak, line 35: great description, thanks!

kokesak:
rebase and rerun please

kokesak (approved):
LGTM!

PR #1281, opened 2024-06-27
[RHELC-1622] Require yum on RHEL 8+ based systems

bocekm:
The package manager on RHEL 8+ is dnf but for a backwards compatibility reason a yum rpm is available to create symlinks pointing to dnf.

If the yum package is missing, convert2rhel fails because it calls yum.

Having yum as a dependency of convert2rhel on RHEL 8+ based systems will ensure that yum will be available (unless someone intentionally removes yum after the convert2rhel installation but then ¯\_(ツ)_/¯).

Jira Issues:

- [RHELC-1622](https://issues.redhat.com/browse/RHELC-1622)

Checklist

### DIFF
--- a/packaging/convert2rhel.spec
+++ b/packaging/convert2rhel.spec
@@ -44,6 +44,8 @@ Requires:       python%{python_pkgversion}-setuptools
 Requires:       python%{python_pkgversion}-six
 %if 0%{?rhel} && 0%{?rhel} >= 8
 Requires:       dnf
+# The yum rpm mainly takes care of creating symlinks to dnf. We need it since we call yum in convert2rhel.
+Requires:       yum
 # dnf-utils includes yumdownloader and package-cleanup we use
 Requires:       dnf-utils
 Requires:       grubby


r0x0d:
Should we add a test case to this to replicate the description of the ticket? It would be a simple test, I believe

Venefilyn:
> Should we add a test case to this to replicate the description of the ticket? It would be a simple test, I believe... @danmyway

@SerCantus WDYT?

danmyway:
The integration tests task is triaged for this sprint.
I don't see an issue of merging this and including the test later in the cycle.
Feel free to merge when you feel like (and the sanity suite is passing) @r0x0d @Venefilyn

r0x0d:
/packit test --labels sanity

retrigger in case of infra problems

Venefilyn:
@SerCantus @danmyway @kokesak Are these flaky test results?

kokesak:
Three of the fails are for the ` /tests/integration/tier0/non-destructive/cve-2022-1662/passing_password_to_submgr` (it only passed for Rocky8). Do we know if there is some issue with the test @danmyway ?

danmyway:
@kokesak @Venefilyn I mean the branch is 34 commits behind.
The issue with the tests failing got fixed two weeks ago in https://github.com/oamg/convert2rhel/pull/1275
So a rebase would help.

danmyway (approved):
Anyway, the failure is unrelated, fixed in main already and I don't see it as a blocker for merging.

PR #1287, opened 2024-07-03
[RHELC-1606] Fix duplicating backups

Venefilyn:
Jira Issues:

-[RHELC-1606](https://issues.redhat.com/browse/RHELC-1606)

Checklist

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/backup_system.py
+++ b/convert2rhel/actions/pre_ponr_changes/backup_system.py
@@ -170,7 +170,10 @@ class BackupPackageFiles(actions.Action):
     def _get_changed_package_files(self):
         """Get the output from rpm -Va command from during resolving system info
         to get changes made to package files.
-        Return them as a list of dict, for example:
+
+        The dict itself is unique and does not have duplicate entries
+
+        :return dict: Return them as a list of dict, for example:
         [{"status":"S5T", "file_type":"c", "path":"/etc/yum.repos.d/CentOS-Linux-AppStream.repo"}]
         """
         data = []
@@ -194,7 +197,12 @@ class BackupPackageFiles(actions.Action):
         lines = output.strip().split("\n")
         for line in lines:
             parsed_line = self._parse_line(line.strip())
-            if parsed_line["path"] and parsed_line["status"]:
+            # We first check that it has a path and status, otherwise nothing to backup
+            #
+            # For checking it already exists, it is possible two applications that
+            # use the same config file, such as packages with different archs
+            # For more info see https://issues.redhat.com/browse/RHELC-1606
+            if parsed_line["path"] and parsed_line["status"] and parsed_line not in data:
                 data.append(parsed_line)
 
         return data
--- a/convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
+++ b/convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
@@ -141,6 +141,7 @@ class TestBackupSystem:
             (
                 """     missing   d /usr/share/info/ed.info.gz
                 S.5.?..T. c     /etc/yum.repos.d/CentOS-Linux-AppStream.repo
+                S.5.?..T. c     /etc/yum.repos.d/CentOS-Linux-AppStream.repo
                 S.5.?..T.        /etc/yum.repos.d/CentOS-Linux-AppStream.repo
 
 


pr-watson (approved):
Looks good to me

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
```diff
@@ -141,6 +141,7 @@ def test_backup_redhat_release_error_os_release_file(self, backup_redhat_release
             (
                 """     missing   d /usr/share/info/ed.info.gz
                 S.5.?..T. c     /etc/yum.repos.d/CentOS-Linux-AppStream.repo
+                S.5.?..T. c     /etc/yum.repos.d/CentOS-Linux-AppStream.repo
                 S.5.?..T.        /etc/yum.repos.d/CentOS-Linux-AppStream.repo
```
Venefilyn, line 145: @r0x0d I don't know what the different statuses mean really. But what does c indicate here? We have it listed 3 times and we do AFAIK expect to have this correct:

```
                    {
                        "status": "S5T",
                        "file_type": "c",
                        "path": "/etc/yum.repos.d/CentOS-Linux-AppStream.repo",
                    },
                    {
                        "status": "S5T",
                        "file_type": None,
                        "path": "/etc/yum.repos.d/CentOS-Linux-AppStream.repo",
                    },
```

bocekm: http://ftp.rpm.org/max-rpm/s1-rpm-verify-output.html

> c appears only if the file is a configuration file.

bocekm: Better resource: https://www.man7.org/linux/man-pages/man8/rpm.8.html

>        The format of the output is a string of 9 characters, a possible
>        attribute marker:
> 
>               a %artifact a build side-effect file (such as buildid links)
>               c %config configuration file.
>               d %doc documentation file.
>               g %ghost file (i.e. the file contents are not included in the package payload).
>               l %license license file.
>               m %missingok file missing is not a verify failure.
>               r %readme readme file.
> 
>        from the package header, followed by the file name.  Each of the
>        9 characters denotes the result of a comparison of attribute(s)
>        of the file to the value of those attribute(s) recorded in the
>        database.  A single "." (period) means the test passed, while a
>        single "?" (question mark) indicates the test could not be per‐
>        formed (e.g. file permissions prevent reading).  Otherwise, the
>        (mnemonically emBoldened) character denotes failure of the corre‐
>        sponding --verify test:
> 
>               S file Size differs
>               M Mode differs (includes permissions and file type)
>               5 digest (formerly MD5 sum) differs
>               D Device major/minor number mismatch
>               L readLink(2) path mismatch
>               U User ownership differs
>               G Group ownership differs
>               T mTime differs
>               P caPabilities differ

bocekm: Because you're adding another entry to the `rpm_va_output` test parameter, you need to add also an entry to the `expected` parameter.

```
                    {
                        "status": "S5T",
                        "file_type": "c",
                        "path": "/etc/yum.repos.d/CentOS-Linux-AppStream.repo",
                    },
                    {
                        "status": "S5T",
                        "file_type": "c",
                        "path": "/etc/yum.repos.d/CentOS-Linux-AppStream.repo",
                    },
                    {
                        "status": "S5T",
                        "file_type": None,
                        "path": "/etc/yum.repos.d/CentOS-Linux-AppStream.repo",
                    },
```

bocekm (dismissed):

convert2rhel/unit_tests/actions/pre_ponr_changes/backup_system_test.py
Venefilyn: Right, but since we back it up only once we just want it listed once right?

bocekm: Ah, right, indeed.

bocekm, line 145: I'd then remove the non-c entry from the `rpm_va_output` because I'd expect that two entries for the same file would be of the same type.
```suggestion
                S.5.?..T. c     /etc/yum.repos.d/CentOS-Linux-AppStream.repo
```
And then remove the following from the `expected`:
```
                    {
                        "status": "S5T",
                        "file_type": None,
                        "path": "/etc/yum.repos.d/CentOS-Linux-AppStream.repo",
                    },
```

convert2rhel/backup/__init__.py
```diff
@@ -70,6 +70,12 @@ def push(self, restorable):
         if not isinstance(restorable, RestorableChange):
             raise TypeError("`%s` is not a RestorableChange object" % restorable)
 
+        # check if the file is already added
```
hosekadam, line 73: I think in this place we cannot expect only file, but there can be any other restorable. What if we would try to back up e.g. package - there isn't any `restorable.backup_path`.

Maybe it's a corner case, but I won't "lock" it just to the files.

Venefilyn: Ah right, yeah wording can probably be changed then. This is just checking if the two items are identical. By default two different class instances will be different unless `__eq__` is implemented like in RestorableFile

hosekadam: Hmm, so maybe I would also add a comment saying that we have `__eq__` implemented only in the restorablefile, so working with the `backup_path` is safe

convert2rhel/backup/__init__.py
```diff
@@ -70,6 +70,12 @@ def push(self, restorable):
...
+        for r in self._restorables:
+            if r == restorable:
+                loggerinst.debug("{} is already backed up".format(restorable.backup_path))
```
hosekadam, line 76:
```suggestion
                loggerinst.debug("{} is already backed up".format(r.backup_path))
```
or `loggerinst.debug("{} is already backed up".format(restorable.filepath))`

convert2rhel/backup/__init__.py
```diff
@@ -70,6 +70,13 @@ def push(self, restorable):
         if not isinstance(restorable, RestorableChange):
             raise TypeError("`%s` is not a RestorableChange object" % restorable)
 
+        # check if the restorable is already backed up
+        # if it is, we skip it
+        for r in self._restorables:
+            if r == restorable:
+                loggerinst.debug("Skipping: has already been backed up")
```
hosekadam, line 77:
```suggestion
                loggerinst.debug("Skipping: %s has already been backed up" % r.filepath)
```
I tested it like this and it works great, the message looks like `DEBUG - Skipping: /etc/security/limits.conf has already been backed up` and as you said, we only have the `__eq__` for restorablefile, so it's safe to use.

Venefilyn: We can't use `r.filepath` as not all restorables have that. I think RFE for the future is to implement a `__str__()`

hosekadam (approved):

convert2rhel/backup/__init__.py
```diff
@@ -70,6 +70,13 @@ def push(self, restorable):
         if not isinstance(restorable, RestorableChange):
             raise TypeError("`%s` is not a RestorableChange object" % restorable)
 
+        # check if the restorable is already backed up
```
hosekadam, line 73:
```suggestion
        # Check if the restorable is already backed up
```

convert2rhel/backup/files.py
```diff
@@ -150,6 +150,14 @@ def remove(self):
         except (OSError, IOError):
             loggerinst.debug("Couldn't remove restored file %s" % self.filepath)
 
+    def __eq__(self, value):
+        if hash(self) == hash(value):
+            return True
+        return False
```
r0x0d, line 156:
```suggestion
        return hash(self) == hash(value)
```

What about make it shorter?

Venefilyn: Was following the same style we have elsewhere

convert2rhel/backup/__init__.py
Venefilyn: Me and Adam came up with getting the class name which has now been added

convert2rhel/actions/pre_ponr_changes/backup_system.py
```diff
@@ -170,7 +170,10 @@ def run(self):
     def _get_changed_package_files(self):
         """Get the output from rpm -Va command from during resolving system info
         to get changes made to package files.
-        Return them as a list of dict, for example:
+
+        The dict itself is unique and does not have duplicate entries
```
hosekadam, line 174:
```suggestion
```

hosekadam (approved):
Looks good, just minor comment about comment

PR #1288, opened 2024-07-03
[RHELC-1747] Use enablerepo instead of no_rhsm to check custom repos

r0x0d:
We were using the no_rhsm option to skip the check for custom repositories, and since the system can be unregistered or pre-registered, there is no need anymore to use no_rhsm in this action, instead, we can use enablerepo directly to validate if the repositories are valid or not.

Jira Issues:

- [RHELC-1747](https://issues.redhat.com/browse/RHELC-1747)

Checklist

---
# TODO

- [ ] Create a Jira issue for this

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/custom_repos_are_valid.py
+++ b/convert2rhel/actions/pre_ponr_changes/custom_repos_are_valid.py
@@ -27,6 +27,7 @@ logger = root_logger.getChild(__name__)
 
 class CustomReposAreValid(actions.Action):
     id = "CUSTOM_REPOSITORIES_ARE_VALID"
+    dependencies = ("SUBSCRIBE_SYSTEM",)
 
     def run(self):
         """To prevent failures past the PONR, make sure that the enabled custom repositories are valid.
@@ -38,7 +39,7 @@ class CustomReposAreValid(actions.Action):
         logger.task("Prepare: Check if --enablerepo repositories are accessible")
 
         if not tool_opts.enablerepo:
-            logger.info("Skipping the check as there was no --enablerepo option detected in the command-line.")
+            logger.info("Did not perform the check of repositories due to the use of RHSM for the conversion.")
             return
 
         output, ret_code = call_yum_cmd(
--- a/convert2rhel/actions/system_checks/custom_repos_are_valid.py
+++ b/convert2rhel/actions/system_checks/custom_repos_are_valid.py
@@ -37,8 +37,8 @@ class CustomReposAreValid(actions.Action):
         super(CustomReposAreValid, self).run()
         logger.task("Prepare: Check if --enablerepo repositories are accessible")
 
-        if not tool_opts.no_rhsm:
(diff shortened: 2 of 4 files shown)


convert2rhel/actions/system_checks/custom_repos_are_valid.py
```diff
@@ -37,8 +37,8 @@ def run(self):
         super(CustomReposAreValid, self).run()
         logger.task("Prepare: Check if --enablerepo repositories are accessible")
 
-        if not tool_opts.no_rhsm:
-            logger.info("Did not perform the check of repositories due to the use of RHSM for the conversion.")
+        if not tool_opts.enablerepo:
```
bocekm, line 40: This way the custom repo accessibility check would fail when using RHSM **and** without the system being pre-registered. At that point the repo the user wants convert2rhel to enable during the conversion is not yet available. One possible solution is to make this check run much later, after sub-man enables the repos needed.

bocekm (changes requested):

convert2rhel/actions/pre_ponr_changes/custom_repos_are_valid.py
```diff
@@ -37,8 +38,8 @@ def run(self):
         super(CustomReposAreValid, self).run()
         logger.task("Prepare: Check if --enablerepo repositories are accessible")
 
-        if not tool_opts.no_rhsm:
-            logger.info("Did not perform the check of repositories due to the use of RHSM for the conversion.")
+        if not tool_opts.enablerepo:
+            logger.info("Skipping the check as there was no --enablerepo option detected in the command-line.")
```
pr-watson, line 42: We changed this from orignally saying "Skipping" so we don't confuse customers with our SKIP result level. Reference - https://issues.redhat.com/browse/RHELC-1173

r0x0d:
```suggestion
            logger.info("Did not perform the check of repositories due to the use of RHSM for the conversion.")
```

pr-watson (commented):
Looks good to me!

hosekadam (approved):
The problem Michal mentioned is fixed, the changes looks good.

Venefilyn:
@r0x0d can you create a Jira issue, update the PR text, and merge?

convert2rhel/actions/pre_ponr_changes/custom_repos_are_valid.py
```diff
@@ -37,7 +38,7 @@ def run(self):
         super(CustomReposAreValid, self).run()
         logger.task("Prepare: Check if --enablerepo repositories are accessible")
 
-        if not tool_opts.no_rhsm:
+        if not tool_opts.enablerepo:
             logger.info("Did not perform the check of repositories due to the use of RHSM for the conversion.")
```
bocekm, line 42: The log message does not match the condition anymore. I'll create a PR for that.

PR #1297, opened 2024-07-10
[RHELC-1635] State action in special package removal report messages

bocekm:
Customers have been confused when seeing the messages, not knowing what should they do.

Jira Issues:

- [RHELC-1635](https://issues.redhat.com/browse/RHELC-1635)

Checklist

[RHELC-1635]: https://redhat.atlassian.net/browse/RHELC-1635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/convert2rhel/actions/pre_ponr_changes/handle_packages.py
+++ b/convert2rhel/actions/pre_ponr_changes/handle_packages.py
@@ -155,14 +155,22 @@ class RemoveSpecialPackages(actions.Action):
         # shows which packages were not removed, if false, all packages were removed
         pkgs_not_removed = sorted(frozenset(pkghandler.get_pkg_nevras(all_pkgs)).difference(pkgs_removed))
         if pkgs_not_removed:
-            message = "The following packages were not removed: {}".format(", ".join(pkgs_not_removed))
+            message = "The following packages cannot be removed: {}".format(", ".join(pkgs_not_removed))
             logger.warning(message)
             self.add_message(
                 level="WARNING",
                 id="SPECIAL_PACKAGES_NOT_REMOVED",
-                title="Special packages not removed",
-                description="Special packages which could not be removed",
+                title="Some packages cannot be removed",
                 diagnosis=message,
+                description=(
+                    "The packages in diagnosis match a pre-defined list of packages that are to be removed during the"
+                    " conversion. This list includes packages that are known to cause a conversion failure."
+                ),
+                remediations=(
+                    "Remove the packages manually before running convert2rhel again:\n" "yum remove -y {}".format(
+                        " ".join(pkgs_not_removed)
+                    )
+                ),
             )
 
         if pkgs_removed:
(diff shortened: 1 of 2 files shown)


pr-watson (approved):
Looks good to me!

bocekm:
@pr-watson, I've reworded the messages a but. Please take a look.

pr-watson (approved):

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -162,12 +169,13 @@ def run(self):
...
-                description=(
+                title="Packages to be removed",
+                description=message,
+                diagnosis=(
                     "We have identified installed packages that match a pre-defined list of packages that are"
-                    " to be removed during the conversion"
+                    " known to cause a conversion failure."
                 ),
-                diagnosis=message,
+                remediations=("Check that the system runs correctly without the packages after the conversion."),
```
bocekm, line 178: @bookwar, do you see a better action the user can take?

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -146,14 +146,21 @@ def run(self):
...
-                diagnosis=message,
+                title="Some packages cannot be removed",
+                description=message,
+                diagnosis=(
+                    "The packages match a pre-defined list of packages that are to be removed during the"
+                    " conversion. This list includes packages that are known to cause a conversion failure."
+                ),
+                remediations=(
+                    "We do a conversion dry-run (yum transaction validation) later on as part of the pre-conversion"
+                    " analysis. If the dry-run fails, try removing the packages listed above manually."
```
hosekadam, line 162: Do we want to add there also command customer can use?

convert2rhel/actions/pre_ponr_changes/handle_packages.py
r0x0d:
```suggestion
                remediations="Check that the system runs correctly without the packages after the conversion.",
```

convert2rhel/actions/pre_ponr_changes/handle_packages.py
```diff
@@ -146,14 +146,21 @@ def run(self):
...
-            message = "The following packages were not removed: %s" % ", ".join(pkgs_not_removed)
+            message = "The following packages cannot be removed: %s" % ", ".join(pkgs_not_removed)
             logger.warning(message)
             self.add_message(
                 level="WARNING",
                 id="SPECIAL_PACKAGES_NOT_REMOVED",
-                title="Special packages not removed",
-                description="Special packages which could not be removed",
-                diagnosis=message,
+                title="Some packages cannot be removed",
```
r0x0d, line 154: Maybe we should also update the id? It feels strange that in the id we call them as "special" and in the title is just "some packages"

convert2rhel/actions/pre_ponr_changes/handle_packages.py
r0x0d: The first sentence in the remediations seems to be they would be a better fit in the diagnosis, I belive... Right now, it doesn't seem that they are fitting in the remediations correctly. 

I tend to think that remediation would be a step-to-step on how to solve the problem, not really an explanation of why it failed.

> Do we want to add there also command customer can use?

If we know what command can be used, I don't see why not. Would be useful

r0x0d (commented):
Looks good, only some small comments

PR #1299, opened 2024-07-11
[RHELC-1339] Port of update_rhsm_custom_facts

Andrew-ang9:
This PR is a Port for update_rhsm_custom_facts function to the action framwork

Jira Issues: 

- [RHELC-1339](https://issues.redhat.com/browse/RHELC-1339) 

Checklist

- [x] PR has been tested manually in a VM (either author or reviewer)
- [ ] Jira issue has been made public if possible
- [x] `[RHELC-]` or `[HMS-]` is part of the PR title 
- [x] GitHub label has been added to help with Release notes 
- [x] PR title explains the change from the user's point of view
- [x] Code and tests are documented properly
- [ ] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending` if relevant

PR #1305, opened 2024-07-15
[RHELC-1338] Port breadcrumbs.finish_collection to action framework

hosekadam:
Port one of the final steps of conversion to be called by action framework.

Jira Issues:

- [RHELC-1338](https://issues.redhat.com/browse/RHELC-1338)

Checklist

### DIFF
--- a/convert2rhel/actions/post_conversion/breadcrumbs_finish_collection.py
+++ b/convert2rhel/actions/post_conversion/breadcrumbs_finish_collection.py
@@ -1,0 +1,37 @@
+# Copyright(C) 2024 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


convert2rhel/actions/post_conversion/breadcrumbs_finish_collection.py
```diff
@@ -0,0 +1,33 @@
...
+import logging
+
+from convert2rhel import actions, breadcrumbs
+
+
+logger = logging.getLogger(__name__)
+
+
+class BreadcumbsFinishCollection(actions.Action):
+    id = "BREADCRUMBS_FINISH_COLLECTION"
```
r0x0d, line 27: This should have a dependency in the end and be the last thing to run, we don't want to give a false report to the user about breadcrumbs if something fails before it

r0x0d: Actually, we need to define the `dependencies` property in the `BreadcumbsFinishCollection` class itself. We will just need to make sure that it is being executed after the "last action". 

If we defined `BreadcumbsFinishCollection` as a dependency for every action, it means that all of them will depend on `BreadcumbsFinishCollection` executing first, which is not what we want. We want that `BreadcumbsFinishCollection` is executed as last thing possible

r0x0d: Yeah, all or at least one that contains all. We want to make sure that this will be kinda of the latest thing to run (alongside with the https://github.com/oamg/convert2rhel/pull/1299)

r0x0d (approved):
It looks good, we just want to make sure that this will be merged after https://github.com/oamg/convert2rhel/pull/1299, so we can update the dependencies and make it run as last thing possible

PR #990, opened 2023-11-22
[RHELC-1173] Replace the word `skip` in report messages

pr-watson:
This PR replaces the use of 'skip' and 'skipping' with 'Did not perform' in the preconversion analysis report messages so as to not create confusion with the skip status put in place for when we cannot run an action because another one failed. 

Jira Issues: [RHELC-1173](https://issues.redhat.com/browse/RHELC-1173)

Checklist

- [ ] PR has been tested manually in a VM (either author or reviewer)
- [x] Jira issue has been made public if possible
- [x] `[RHELC-]` is part of the PR title 
- [x] GitHub label has been added to help with Release notes 
- [x] PR title explains the change from the user's point of view
- [x] Code and tests are documented properly
- [x] The commits are squashed to as few commits as possible (without losing data) 
- [ ] When merged: Jira issue has been updated to `Release Pending` if relevant

PR #1311, opened 2024-07-18
[RHELC-1390] Port hostmetering to Action framework

hosekadam:
Port functions related to hostmetering service to be part of post conversion action.

Refactor:
- removed variable `should_configure_metering` as it is not needed
- conditions of configuring hostmetering are easier to read and less
  complex
- part about checking content of environment variable is moved to
  function to make the code less complex and reduce count of return
  statements
- get the rhsm facts is now inside the method instead of passing it as
  argument

Jira Issues:

- [RHELC-1390](https://issues.redhat.com/browse/RHELC-1390)

Checklist


convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,235 @@
...
+
+        self.env_var = os.environ.get("CONVERT2RHEL_CONFIGURE_HOST_METERING", None)
+        if not self._check_env_var_correct():
+            return False
+
+        if system_info.version.major != 7 and self.env_var != "force":
+            logger.info("Skipping host metering configuration. Only supported for RHEL 7.")
+            self.add_message(
+                level="INFO",
+                id="SKIPPED_CONFIGURE_HOST_METERING",
```
hosekadam, line 61: For reviewers:

When I was testing I found out that when there is used level SKIP, then the action fails, which I think is not what we want here. So I set it to info message and added to the ID `CONFIGURE_HOST_METERING_IF_NEEDED` the `IF_NEEDED` part. I believe this will help to avoid misunderstandings like "Why the host metering was set? I don't need that".

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,235 @@
...
+                title="Host metering service is not running.",
+                description="host-metering.service is not running.",
+                remediations="You can try to start the service manually"
+                " by running following command:\n"
+                "   systemctl start host-metering.service",
+            )
+
+        return service_running
+
+    def _check_env_var_correct(self):
```
hosekadam, line 138: For reviewers:

I had to split the code a bit since there were a lot of `return` statements.

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,235 @@
...
+                title="The `force' option has been used for the CONVERT2RHEL_CONFIGURE_HOST_METERING environment variable.",
+                description="Please note that this option is mainly used for testing and"
+                " will configure host-metering unconditionally."
+                " For generic usage please use the 'auto' option.",
+            )
+        elif self.env_var == "auto":
+            self.should_configure_metering = True
+            logger.debug("Automatic detection of host hyperscaler and configuration.")
+        else:
+            self.should_configure_metering = False
```
Venefilyn, line 183: We never reach this as we always verify `env_var`. If it fails verification we return right away. Do we even need to set `self.should_configure_metering` as it will always be `True` if we execute?
```suggestion
```

Venefilyn:
Do update the commit message as well to contain if there are any changes with the refactor, not just that it was moved/refactored

convert2rhel/actions/post_conversion/hostmetering.py
pr-watson: @hosekadam so the "SKIP" level isn't supposed to be used in set_result. It's a level for when an action fails and we skip the other ones that depended on it. Reference line - https://github.com/oamg/convert2rhel/blob/81c59f667ac09d9bb1bc148d458db188961effb4/convert2rhel/actions/__init__.py#L53

pr-watson: The INFO level is what we want here. For cases where we want to allow the user to skip over a result we use "OVERRIDABLE"

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,235 @@
...
+        self._set_should_configure_metering()
+
+        rhsm_facts = get_rhsm_facts()
+        is_hyperscaler = self.is_running_on_hyperscaler(rhsm_facts)
+        if (not is_hyperscaler or not self.should_configure_metering) and self.env_var != "force":
+            logger.info("Skipping host-metering configuration.")
+            self.add_message(
+                level="INFO",
+                id="SKIPPED_CONFIGURE_HOST_METERING",
+                title="Skipping host metering configuration as not needed.",
```
pr-watson, line 76:
```suggestion
                title="Did not perform host metering configuration as not needed.",
```
We want to avoid the use of skip in our messages to not confuse customers with our "SKIP" level. Reference -https://github.com/oamg/convert2rhel/pull/990/files

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,235 @@
...
+                description="Host metering is supportted only for RHEL 7.",
+            )
+            return False
+
+        self._set_should_configure_metering()
+
+        rhsm_facts = get_rhsm_facts()
+        is_hyperscaler = self.is_running_on_hyperscaler(rhsm_facts)
+        if (not is_hyperscaler or not self.should_configure_metering) and self.env_var != "force":
+            logger.info("Skipping host-metering configuration.")
```
pr-watson, line 72:
```suggestion
            logger.info("Did not perform host-metering configuration.")
```

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,235 @@
...
+        logger.info("Installing host-metering packages.")
+        output, ret_install = call_yum_cmd("install", ["host-metering"])
+        if ret_install:
+            logger.warning("Failed to install host-metering rpms.")
+            # TODO Consider this when setting this action as dependency of other actions
+            # Fail in this step can result in incomplete conversion - even if it would be
+            # enough to re-run the install and config steps.
+            self.set_result(
+                level="ERROR",
+                id="FAILED_INSTALL_HOST_METERING",
```
pr-watson, line 90:
```suggestion
                id="INSTALL_HOST_METERING_FAILURE",
```

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,235 @@
...
+                "   systemctl start host-metering.service",
+            )
+            return False
+
+        command, error_message = self._enable_host_metering_service()
+        # If there is any failure, the failed command would be present
+        if any(command):
+            self.set_result(
+                level="ERROR",
+                id="FAILED_CONFIGURE_HOST_METERING",
```
pr-watson, line 109:
```suggestion
                id="CONFIGURE_HOST_METERING_FAILURE",
```

convert2rhel/actions/post_conversion/hostmetering.py
hosekadam: Good point, it was already present in the code and I didn't want to touch the logic much. But it seems it can be removed or reworked

hosekadam:
Applied the changes, removed `should_configure_metering` variable as not needed. Made conditions easier to read. Replaced `Skipping` by `Did not performed`, moved `FAILURE` from begging of the IDs to the end (did the same for the `SKIP` in ID to be consitent)

hosekadam:
Changed the results\messages to correspond with the logging levels.

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        logger.task("Final: Configure host-metering")
+
+        super(ConfigureHostMetering, self).run()
+
+        self.env_var = os.environ.get("CONVERT2RHEL_CONFIGURE_HOST_METERING", None)
+        if not self._check_env_var():
+            return False
+
+        if system_info.version.major != 7 and self.env_var == "auto":
```
r0x0d, line 64: This is a bit different from the original code, is it intended? https://github.com/oamg/convert2rhel/blob/main/convert2rhel/hostmetering.py#L73

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,238 @@
...
+                description="Please note that this option is mainly used for testing and"
+                " will configure host-metering unconditionally."
+                " For generic usage please use the 'auto' option.",
+            )
+        elif self.env_var == "auto":
+            logger.debug("Automatic detection of host hyperscaler and configuration.")
+
+        return True
+
+    def is_running_on_hyperscaler(self, rhsm_facts):
```
r0x0d, line 188: This can be moved to the module level instead of the class level since it does not depend on anything from the class (self). 

We can make it private as well to indicate that this is intended to be used only inside this module

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,238 @@
...
+        logger.info("Starting host-metering service.")
+        command = ["systemctl", "start", "host-metering.service"]
+        output, ret_start = run_subprocess(command)
+        if output:
+            logger.debug("Output of systemctl call: %s" % output)
+        if ret_start:
+            logger.warning("Failed to start host-metering service.")
+            return " ".join(command), output
+
+        # Any command failed, no error output
```
r0x0d, line 237:
```suggestion
        # All commands succeeded, no error output found
```

convert2rhel/actions/post_conversion/hostmetering.py
hosekadam: After good point from @Venefilyn here: https://github.com/oamg/convert2rhel/pull/1311#discussion_r1688132019 I spent some time on refactoring the conditions a bit. I think having the non negation state of the `env_var` is easier to read. And since it can have only `auto` or `force` (we check it in the code before), I believe it's fine, and the functionality is the same.

convert2rhel/actions/post_conversion/hostmetering.py
hosekadam: Yeah, that sound's better :D I removed the negation there https://github.com/oamg/convert2rhel/pull/1311#discussion_r1706921105 but added here... :D

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,239 @@
...
+class ConfigureHostMetering(actions.Action):
+    """Configure host metering on a machine if it's needed.
+
+    env_var: str
+        Content of CONVERT2RHEL_CONFIGURE_HOST_METERING environment variable.
+    """
+
+    id = "CONFIGURE_HOST_METERING_IF_NEEDED"
+    env_var = None
+
```
Venefilyn, line 41:
```suggestion
    env_var = None  # type: str|None
```

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,239 @@
...
+                description="Failed to install host-metering rpm.",
+                diagnosis="`yum install host-metering` command returned {ret_install} with message {output}".format(
+                    ret_install=ret_install, output=output
+                ),
+                remediations="You can try install and set up the host metering"
+                " manually using following commands:\n"
+                "   yum install host-metering\n"
+                "   systemctl enable host-metering.service\n"
+                "   systemctl start host-metering.service",
+            )
```
Venefilyn, line 103:
```suggestion
            self.add_message(
                level="WARNING",
                id="INSTALL_HOST_METERING_FAILURE",
                title="Failed to install host metering package.",
                description="When installing host metering package an error occured meaning we can " \
                    " not enable host metering on the system." \
                    " \n\nAs the conversion is set to install host metering this must be resolved before continuing.",
                diagnosis="`yum install host-metering` command returned {ret_install} with message {output}".format(
                    ret_install=ret_install, output=output
                ),
                remediations="You can try install and set up the host metering"
                " manually using following commands:\n"
                " - `yum install host-metering`\n"
                " - `systemctl enable host-metering.service`\n"
                " - `systemctl start host-metering.service`",
            )
```

convert2rhel/unit_tests/actions/post_conversion/hostmetering_test.py
```diff
@@ -0,0 +1,440 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+
+import pytest
+import six
+
+from convert2rhel import actions, systeminfo, unit_tests
```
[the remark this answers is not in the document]
Venefilyn:
```suggestion
from convert2rhel import actions, systeminfo
```

convert2rhel/actions/post_conversion/hostmetering.py
```diff
@@ -0,0 +1,239 @@
...
+                title="Failed to configure host metering.",
+                description="Failed to enable and start host metering service.",
+                diagnosis="Command {command} failed with {error_message}".format(
+                    command=command, error_message=error_message
+                ),
+                remediations="You can try set up the host metering"
+                " manually using following commands:\n"
+                "   systemctl enable host-metering.service\n"
+                "   systemctl start host-metering.service",
+            )
```
Venefilyn, line 121:
```suggestion
            self.add_message(
                level="WARNING",
                id="CONFIGURE_HOST_METERING_FAILURE",
                title="Failed to enable and start host metering service.",
                description="The host metering service failed to start" \
                    " successfully and won't be able to keep track.",
                diagnosis="Command {command} failed with {error_message}".format(
                    command=command, error_message=error_message
                ),
                remediations="You can try set up the host metering"
                " service manually using following commands:\n"
                " - `systemctl enable host-metering.service`\n"
                " - `systemctl start host-metering.service`",
            )
```

hosekadam:
Applied the changes, rebased.

Note about the unit_test import in the `hostmetering_test.py`: it's needed for the `assert_actions_result`, but I changed it to import this excplicitly.

pr-watson (approved):
Looks good to me!

PR #1313, opened 2024-07-18
[RHELC-1600] Update the bufsize value of run_subprocess

pr-watson:
This PR updates the bufsize value for our run_subprocess function from 1 to -1 to avoid the RuntimeWarning introduced in python 3.8. The bufsize value of -1 is the new default value because it is compatible with most code according to the python docs - https://docs.python.org/3/library/subprocess.html .

Jira Issues:

- [RHELC-1600](https://issues.redhat.com/browse/RHELC-1600)

Checklist

### DIFF
--- a/convert2rhel/utils/__init__.py
+++ b/convert2rhel/utils/__init__.py
@@ -350,7 +350,7 @@ def run_subprocess(cmd, print_cmd=True, print_output=True):
         cmd,
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
-        bufsize=1,
+        bufsize=-1,
     )
     output = ""
     for line in iter(process.stdout.readline, b""):


Venefilyn:
I'd like a longer commit message explaining why we had it in the first place, why -1 is the solution, and why we can't use 0 for example. Also, we probably do want buffered text. Can we use `universal_newlines=True`?

https://docs.python.org/3/library/subprocess.html

pr-watson:
> I'd like a longer commit message explaining why we had it in the first place, why -1 is the solution, and why we can't use 0 for example. Also, we probably do want buffered text as that is what is in the code atm, is there a git history explaining why? Can we use `universal_newlines=True`?
> 
> https://docs.python.org/3/library/subprocess.html

Updated the commit message. I choose -1 because it is the new default value. Since the code has buffered text I can use `universal_newlines=True` and that should get rid of the RuntimeWarning and maintain the same behaviour as before.

pr-watson:
@Venefilyn Just tested the code with `bufsize=1` and `universal_newlines=True` and the conversion stopped at the TASK - [Prepare: Gather system information] so we will need to choose a different value for bufsize

Venefilyn:
> @Venefilyn Just tested the code with `bufsize=1` and `universal_newlines=True` and the conversion stopped at the TASK - [Prepare: Gather system information] so we will need to choose a different value for bufsize

@pr-watson Ack, and it works with `-1`?

hosekadam (approved):
Looks good, even better when it's recommended value from docs :)

Venefilyn (approved):
LGTM, but push up the commit message change you did :)

pr-watson:
> LGTM, but push up the commit message change you did :)

I added the update to the PR description

Venefilyn:
> > LGTM, but push up the commit message change you did :)
> 
> I added the update to the PR description

@pr-watson That's not the commit message though, it's ideal to have the commit history in the branches more descriptive to make it easier to figure out what happened in the commit later down the line. For example, if someone of us or newly onboarded people find a bug that relates to using `-1` and is unsure why it is there to begin with, they will search through git blame to find the commit

At the moment when they find the commit https://github.com/oamg/convert2rhel/commit/f65b8c14d00f20218048ef465ff9ba67338bd9f6 they will see `[RHELC-1652] Update the bufsize value of run_subprocess (#1313)` and no description

It's ideal to have the commit descriptive of why the change has been done and why this solution was chosen

For more info see https://cbea.ms/git-commit/

pr-watson:
@Venefilyn My apologies I will include it next time - since it's already merged I don't think I can edit the commit message

PR #1321, opened 2024-07-24
[RHELC-1332] Port PkgManagerConf() to the Action framework

pr-watson:
This PR calls the PkgManagerConf class inside a new action called pkg_manager_config so it can be executed alongside the other actions in the conversion section. Relevant tests have been written.

Jira Issues:

- [RHELC-1332](https://issues.redhat.com/browse/RHELC-1332) 

Checklist

### DIFF
--- a/convert2rhel/actions/conversion/pkg_manager_config.py
+++ b/convert2rhel/actions/conversion/pkg_manager_config.py
@@ -23,15 +23,15 @@ from convert2rhel import actions, redhatrelease
 logger = logging.getLogger(__name__)
 
 
-class PkgManagerConfig(actions.Action):
-    id = "PKG_MANAGER_CONFIG"
+class ConfigurePkgManager(actions.Action):
+    id = "CONFIGURE_PKG_MANAGER"
 
     def run(self):
         """
         Check if the distroverpkg tag inside the package manager config has been modified before the conversion and if so
         comment it out and write to the file.
         """
-        super(PkgManagerConfig, self).run()
+        super(ConfigurePkgManager, self).run()
 
         logger.task("Convert: Patch package manager configuration file")
         pmc = redhatrelease.PkgManagerConf()
--- a/convert2rhel/actions/conversion/pkg_manager_config.py
+++ b/convert2rhel/actions/conversion/pkg_manager_config.py
@@ -1,0 +1,37 @@
+# Copyright(C) 2024 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__metaclass__ = type
+
+import logging
+
+from convert2rhel import actions, redhatrelease
+
+
+logger = logging.getLogger(__name__)
+
+
+class PkgManagerConfig(actions.Action):
+    id = "PKG_MANAGER_CONFIG"
+
+    def run(self):
+        """
+        Check if the distoverpkg tag inside the package manager config has been modified before the conversion and if so
+        comment it out and write to the file.
+        """
+        super(PkgManagerConfig, self).run()
+
+        logger.task("Convert: Patch package manager configuration file")
+        redhatrelease.PkgManagerConf.patch()
--- a/convert2rhel/actions/conversion/pkg_manager_config.py
+++ b/convert2rhel/actions/conversion/pkg_manager_config.py
@@ -28,7 +28,7 @@ class PkgManagerConfig(actions.Action):
 
     def run(self):
         """
-        Check if the distoverpkg tag inside the package manager config has been modified before the conversion and if so
+        Check if the distroverpkg tag inside the package manager config has been modified before the conversion and if so
         comment it out and write to the file.
         """
         super(PkgManagerConfig, self).run()
--- a/convert2rhel/actions/conversion/pkg_manager_config.py
+++ b/convert2rhel/actions/conversion/pkg_manager_config.py
@@ -34,4 +34,5 @@ class PkgManagerConfig(actions.Action):
         super(PkgManagerConfig, self).run()
 
         logger.task("Convert: Patch package manager configuration file")
-        redhatrelease.PkgManagerConf.patch()
+        pmc = redhatrelease.PkgManagerConf()
+        pmc.patch()
--- a/convert2rhel/actions/conversion/pkg_manager_config.py
+++ b/convert2rhel/actions/conversion/pkg_manager_config.py
@@ -25,6 +25,7 @@ logger = logging.getLogger(__name__)
 
 class ConfigurePkgManager(actions.Action):
     id = "CONFIGURE_PKG_MANAGER"
+    dependencies = ("CONVERT_SYSTEM_PACKAGES",)
 
     def run(self):
         """
--- a/convert2rhel/redhatrelease.py
+++ b/convert2rhel/redhatrelease.py
@@ -53,7 +53,7 @@ class PkgManagerConf:
     """
     Check if the config file of the systems package manager has been modified and if it has then
     remove those changes before the conversion completes.
-    .. note::  
+    .. note::
         The pkg manager config file path only needs to be set to yum.conf as there is a symlink between yum and dnf.
     This means that on dnf systems the dnf.conf will be modified even the path is for the yum.conf.
     """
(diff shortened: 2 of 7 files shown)


convert2rhel/redhatrelease.py
```diff
@@ -49,35 +49,42 @@ def get_system_release_content():
...
 
 
-class YumConf:
-    _yum_conf_path = "/etc/yum.conf"
+class PkgManagerConf:
+    """
+    Check if the config file of the systems package manager has been modified and if it has then
+    remove those changes before the conversion completes.
+    NOTE: The pkg manager config file path only needs to be set to yum.conf as there is a symlink between yum and dnf.
+    This means that on dnf systems the dnf.conf will be modified.
```
hosekadam, line 57:
```suggestion
    This means that on dnf systems the dnf.conf will be modified even the path is for the yum.conf.
```
If my understanding is good, I would extend the comment a bit

convert2rhel/redhatrelease.py
```diff
@@ -49,35 +49,42 @@ def get_system_release_content():
         loggerinst.critical("%s\n%s file is essential for running this tool." % (err, filepath))
 
 
-class YumConf:
-    _yum_conf_path = "/etc/yum.conf"
+class PkgManagerConf:
+    """
+    Check if the config file of the systems package manager has been modified and if it has then
+    remove those changes before the conversion completes.
+    NOTE: The pkg manager config file path only needs to be set to yum.conf as there is a symlink between yum and dnf.
```
hosekadam, line 56:
```suggestion
    .. note::  
        The pkg manager config file path only needs to be set to yum.conf as there is a symlink between yum and dnf.
```
I would go with this format: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/initialize.py#L62

convert2rhel/redhatrelease.py
```diff
@@ -49,35 +49,42 @@ def get_system_release_content():
...
+    This means that on dnf systems the dnf.conf will be modified.
+    """
+
+    _pkg_manager_conf_path = "/etc/yum.conf" if pkgmanager.TYPE == "yum" else "/etc/dnf/dnf.conf"
```
hosekadam, line 60: I think this could be also used in the `is_modified()` method. I know, now it's static one, but since I see it's called only from `patch()` I think it's not needed anymore.

pr-watson: Yes I was also thinking that but the static part made me reconsider. I'll remove the static and use attirbute there

r0x0d (commented):
Looks good to me. 

The only thing is that the action looks confusing right now, as the name of the class is almost the name of what we call underneath it. The only main difference in the naming is the `ig` at the name of the action class :thinking: 

Is there a way we can improve it somehow?

pr-watson:
> Looks good to me.
> 
> The only thing is that the action looks confusing right now, as the name of the class is almost the name of what we call underneath it. The only main difference in the naming is the `ig` at the name of the action class 🤔
> 
> Is there a way we can improve it somehow?



> Looks good to me.
> 
> The only thing is that the action looks confusing right now, as the name of the class is almost the name of what we call underneath it. The only main difference in the naming is the `ig` at the name of the action class 🤔
> 
> Is there a way we can improve it somehow?

I could extend the name of the action class to PackageManagerConfiguration or maybe ConfigurePkgManager

Venefilyn (changes requested):

Venefilyn (changes requested):
Some changes necessary

convert2rhel/redhatrelease.py
```diff
@@ -49,45 +49,52 @@ def get_system_release_content():
...
+    This means that on dnf systems the dnf.conf will be modified even the path is for the yum.conf.
+    """
 
-    def __init__(self):
-        self._yum_conf_content = utils.get_file_content(self._yum_conf_path)
+    def __init__(self, config_path=None):
+        self._pkg_manager_conf_path = config_path
+        if config_path is None:
+            self._pkg_manager_conf_path = "/etc/yum.conf" if pkgmanager.TYPE == "yum" else "/etc/dnf/dnf.conf"
+        self._pkg_manager_conf_content = utils.get_file_content(self._pkg_manager_conf_path)
```
Venefilyn, line 65:
```suggestion
    _pkg_manager_conf_path = ""  # type: str
    _pkg_manager_conf_content = ""  # type: str

    def __init__(self, config_path=None):  # type: (str|None) -> None

        if config_path is None:
            self._pkg_manager_conf_path = "/etc/yum.conf" if pkgmanager.TYPE == "yum" else "/etc/dnf/dnf.conf"
        else:
            self._pkg_manager_conf_path = config_path
        self._pkg_manager_conf_content = utils.get_file_content(self._pkg_manager_conf_path)
```

Venefilyn (changes requested):

convert2rhel/unit_tests/redhatrelease_test.py
```diff
@@ -96,36 +98,37 @@ def test_patch_yum_conf_existing_distroverpkg(version, monkeypatch):
...
     _comment_out_distroverpkg_tag = mock.Mock()
     monkeypatch.setattr(
-        YumConf,
+        PkgManagerConf,
         "_comment_out_distroverpkg_tag",
         value=_comment_out_distroverpkg_tag,
     )
-    _write_altered_yum_conf = mock.Mock()
-    monkeypatch.setattr(YumConf, "_write_altered_yum_conf", value=_write_altered_yum_conf)
+    _write_altered_pkg_manager_conf = mock.Mock()
```
Venefilyn, line 121: Change this mock to instead mock out `open()` somehow. Ideally mocking out the os part so it doesn't write into a file (faster that way)

convert2rhel/actions/conversion/pkg_manager_config.py
```diff
@@ -0,0 +1,38 @@
...
+    def run(self):
+        """
+        Check if the distroverpkg tag inside the package manager config has been modified before the conversion and if so
+        comment it out and write to the file.
+        """
+        super(ConfigurePkgManager, self).run()
+
+        logger.task("Convert: Patch package manager configuration file")
+        pmc = redhatrelease.PkgManagerConf()
+        pmc.patch()
```
Venefilyn, line 38: Do we not want to set results? Even for successful runs?

pr-watson: The super call will set the result to SUCCESS if no other result is set

convert2rhel/actions/conversion/pkg_manager_config.py
```diff
@@ -0,0 +1,38 @@
...
+
+from convert2rhel import actions, redhatrelease
+
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigurePkgManager(actions.Action):
+    id = "CONFIGURE_PKG_MANAGER"
+
```
hosekadam, line 28: In the past times, the patching was executed after the [main transaction](https://github.com/oamg/convert2rhel/blob/v1.6.1/convert2rhel/main.py#L297). I've tried that, and it seems it's not a problem, having the `yum.conf` patched before the transaction.

But maybe just for staying safe - shouldn't there be the transaction as dependency?

hosekadam: Otherwise, the PR looks good!

convert2rhel/unit_tests/redhatrelease_test.py
pr-watson: The test completes in .03 seconds so we likely won't save any noticable time by changing this

convert2rhel/actions/conversion/pkg_manager_config.py
pr-watson: Sure let me add that

convert2rhel/unit_tests/redhatrelease_test.py
Venefilyn: @pr-watson point still stands. Right now it's mocking out functionality, meaning you're not testing anything. There is a function that is completely skipped now because of it. The idea behind mocking out the os part is to not create temporary files, as that will decrease the speed of the test. But the test should still make sure to execute through `_write_altered_pkg_manager_conf()`

pr-watson: Wasn't able to avoid creating temp files but the test executes through _write_altered_pkg_manager_conf now without adding any noticable execution time

r0x0d (approved):
Looks good!

pr-watson:
Failing int tests are because we have updated the patch file that gets asserted to be yum/dnf based on the pkg manager. Before the patch file that got output was always yum.conf

hosekadam (approved):
Just needs to be rebased

danmyway (approved):

tests/integration/tier1/destructive/changed-yum-conf/test_patch_yum_conf.py
```diff
@@ -9,6 +9,9 @@ def test_yum_conf_patch(convert2rhel, shell):
     expanding the $releasever variable properly.
     """
     shell("echo '#random text' >> /etc/yum.conf")
```
kokesak, line 11: This shell command should be placed little down, once we know which config we want to edit (either the dnf or yum conf)

danmyway: It doesn't need to since, `/etc/yum.conf` is either a config file or a symlink to `/etc/dnf/dnf.conf`
A comment would be nice though

kokesak: I see, thanks for the info. I haven't realized this. And +1 for the comment

PR #1323, opened 2024-07-25
[RHELC-1672, RHELC-1707, RHELC-1708] Detect a newer version of RHEL kernel in the main transaction

danmyway:
In this PR we try to detect, if a RHEL kernel was installed during the main transaction.
We compare which of the installed kernel packages has a higher version. 
* If the installed RHEL kernel has a higher version than the original vendor kernel, we do not perform any further action and set such kernel as default.
* If there is no non-RHEL kernel detected, we assume that the main transaction handled the whole kernel replacement by itself.
* If the latest available RHEL kernel package is the same version as the currently running non-RHEL kernel, we fall back to the original procedure of trying to install an older version and subsequently updating it.

Jira Issues:

- [RHELC-1672](https://issues.redhat.com/browse/RHELC-1672)
- [RHELC-1708](https://issues.redhat.com/browse/RHELC-1708)
- [RHELC-1707](https://issues.redhat.com/browse/RHELC-1707)

Checklist

### DIFF
--- a/convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
+++ b/convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
@@ -24,6 +24,8 @@ from convert2rhel import actions, pkghandler, pkgmanager, utils
 from convert2rhel.systeminfo import system_info
 
 
+_kernel_update_needed = None
+
 loggerinst = logging.getLogger(__name__)
 
 
@@ -36,9 +38,13 @@ class InstallRhelKernel(actions.Action):
         super(InstallRhelKernel, self).run()
         loggerinst.task("Convert: Prepare kernel")
 
+        # Solution for RHELC-1707
+        # Update is needed in the UpdateKernel action
+        global _kernel_update_needed  # pylint: disable=global-statement
+
         loggerinst.info("Installing RHEL kernel ...")
         output, ret_code = pkgmanager.call_yum_cmd(command="install", args=["kernel"])
-        kernel_update_needed = False
+        _kernel_update_needed = False
 
         if ret_code != 0:
             self.set_result(
@@ -83,7 +89,7 @@ class InstallRhelKernel(actions.Action):
                 description=info_message,
             )
             pkghandler.handle_no_newer_rhel_kernel_available()
-            kernel_update_needed = True
+            _kernel_update_needed = True
 
         # In this case all kernel packages were already replaced during the main transaction.
         # Having elif here to prevent breaking the action. Otherwise, when the first condition applies,
@@ -116,10 +122,7 @@ class InstallRhelKernel(actions.Action):
             # That might happen and happened before, when the original vendor patches the package
             # with a higher release number.
             pkghandler.handle_no_newer_rhel_kernel_available()
-            kernel_update_needed = True
-
-        if kernel_update_needed:
-            pkghandler.update_rhel_kernel()
+            _kernel_update_needed = True
 
 
 class VerifyRhelKernelInstalled(actions.Action):
@@ -295,3 +298,20 @@ class KernelPkgsInstall(actions.Action):
             if name != "kernel":
                 loggerinst.info("Installing RHEL %s" % name)
                 pkgmanager.call_yum_cmd("install", args=[name])
+
+
+class UpdateKernel(actions.Action):
+    id = "UPDATE_KERNEL"
+    dependencies = ("FIX_DEFAULT_KERNEL",)
+
+    def run(self):
+        super(UpdateKernel, self).run()
+        # Solution for RHELC-1707
+        # This variable is set in the InstallRhelKernel action
+        global _kernel_update_needed
+
+        if _kernel_update_needed:
+            # Note: Info message is in the function
+            pkghandler.update_rhel_kernel()
+        else:
+            loggerinst.info("RHEL kernel already present in latest version. Update not needed.\n")
--- a/convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
+++ b/convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
@@ -50,34 +50,63 @@ class InstallRhelKernel(actions.Action):
             )
             return
 
-        # Check if kernel with same version is already installed.
+        # Check which of the kernel versions are already installed.
         # Example output from yum and dnf:
         #  "Package kernel-4.18.0-193.el8.x86_64 is already installed."
-        already_installed = re.search(r" (.*?)(?: is)? already installed", output, re.MULTILINE)
-        if already_installed:
-            rhel_kernel_nevra = already_installed.group(1)
-            non_rhel_kernels = pkghandler.get_installed_pkgs_w_different_fingerprint(
-                system_info.fingerprints_rhel, "kernel"
+        # When calling install, yum/dnf essentially reports all the already installed versions.
+        already_installed = re.findall(r" (.*?)(?: is)? already installed", output, re.MULTILINE)
+        # Get list of kernel pkgs not signed by Red Hat
+        non_rhel_kernels_pkg_info = pkghandler.get_installed_pkgs_w_different_fingerprint(
+            system_info.fingerprints_rhel, "kernel"
+        )
+        # Extract the NEVRA from the package object to a list
+        non_rhel_kernels = [pkghandler.get_pkg_nevra(kernel) for kernel in non_rhel_kernels_pkg_info]
+        rhel_kernels = [kernel for kernel in already_installed if kernel not in non_rhel_kernels]
+
+        # There is no RHEL kernel installed on the system at this point.
+        # Generally that would mean, that there is either only one kernel
+        # package installed on the system by the time of the conversion.
+        # Or none of the kernel packages installed is possible to be handled
+        # during the main transaction.
+        if not rhel_kernels:
+            info_message = "Conflict of kernels: The running kernel has the same version as the latest RHEL kernel."
+            loggerinst.info("\n%s" % info_message)
+            self.add_message(
+                level="INFO",
+                id="CONFLICT_OF_KERNELS",
+                title="Conflict of installed kernel versions",
+                description=info_message,
             )
-            for non_rhel_kernel in non_rhel_kernels:
-                # We're comparing to NEVRA since that's what yum/dnf prints out
-                if rhel_kernel_nevra == pkghandler.get_pkg_nevra(non_rhel_kernel):
-                    # If the installed kernel is from a third party (non-RHEL) and has the same NEVRA as the one available
-                    # from RHEL repos, it's necessary to install an older version RHEL kernel and the third party one will
-                    # be removed later in the conversion process. It's because yum/dnf is unable to reinstall a kernel.
-                    info_message = (
-                        "Conflict of kernels: One of the installed kernels"
-                        " has the same version as the latest RHEL kernel."
-                    )
-                    loggerinst.info("\n%s" % info_message)
-                    self.add_message(
-                        level="INFO",
-                        id="CONFLICT_OF_KERNELS",
-                        title="Conflict of installed kernel versions",
-                        description=info_message,
-                    )
-                    pkghandler.handle_no_newer_rhel_kernel_available()
-                    kernel_update_needed = True
+            pkghandler.handle_no_newer_rhel_kernel_available()
+            kernel_update_needed = True
+
+        # In this case all kernel packages were already replaced during the main transaction
+        elif not non_rhel_kernels:
+            return
+
+        # At this point we need to decide if the highest package version in the rhel_kernels list
+        # is higher than the highest package version in the non_rhel_kernels list
+        else:
+            latest_installed_non_rhel_kernel = pkghandler.get_highest_package_version(
+                ("non-RHEL kernel", non_rhel_kernels)
+            )
+            latest_installed_rhel_kernel = pkghandler.get_highest_package_version(("RHEL kernel", rhel_kernels))
+            is_rhel_kernel_higher = pkghandler.compare_package_versions(
+                latest_installed_rhel_kernel, latest_installed_non_rhel_kernel
+            )
+
+            # If the highest version of the RHEL kernel package installed at this point is indeed
+            # higher than any non-RHEL package, we don't need to do anything else.
+            if is_rhel_kernel_higher == 1:
+                return
+
+            # This also contains a scenario, where the running non-RHEL kernel is of a higher version
+            # than the latest one available in the RHEL repositories.
+            # That might happen and happened before, when the original vendor patches the package
+            # with a higher release number.
+            else:
+                pkghandler.handle_no_newer_rhel_kernel_available()
+                kernel_update_needed = True
 
         if kernel_update_needed:
             pkghandler.update_rhel_kernel()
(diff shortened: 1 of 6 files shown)


bocekm:
@danmyway, the `install_rhel_kernel()` function used to be called before the main transaction. Nowadays it's called after it. For that reason it should IMHO be refactored a bit more. Here's a proposal of what should it be doing (https://miro.com/app/board/uXjVKvQvEEY=/):
Convert2RHEL_ Install RHEL kernel
The key difference is not trying to install a RHEL kernel when it might very well be already installed from the main transaction.

hosekadam:
@danmyway Do you want to implement the refactor Michal mentioned, or should I take care of it? It seems it will be a bit more work :D

danmyway:
@hosekadam we can discuss the approach, and I'll see if I'll keep it as a "side quest" for myself :smile:

convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
```diff
@@ -50,34 +50,63 @@ def run(self):
             )
             return
 
-        # Check if kernel with same version is already installed.
+        # Check which of the kernel versions are already installed.
         # Example output from yum and dnf:
         #  "Package kernel-4.18.0-193.el8.x86_64 is already installed."
-        already_installed = re.search(r" (.*?)(?: is)? already installed", output, re.MULTILINE)
-        if already_installed:
```
danmyway, line 57: Dropping this condition, since it will always match something.
I can't imagine a case, where `yum install kernel` it won't return, that some kernel version is already installed at that point.

convert2rhel/pkghandler.py
```diff
@@ -1057,3 +1057,21 @@ def _parse_pkg_with_dnf(pkg):
...
+
+def get_highest_package_version(pkgs):
+    """
+    Get the highest version from the provided list of packages.
+    :param pkgs: A tuple containing the name of the package list (as a string) and the list of package versions (as a list of strings).
+    :type pkgs: tuple(str,list[str])
+    :return: Return a single package with the highest version.
+    :rtype: str
+
+    :raises ValueError: If the list is empty, raise a ValueError.
```
Venefilyn, line 1070: Atm this is untrue as it doesn't reraise the exception. If the list is empty it returns None atm

convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
```diff
@@ -50,34 +50,63 @@ def run(self):
...
+        # during the main transaction.
+        if not rhel_kernels:
+            info_message = "Conflict of kernels: The running kernel has the same version as the latest RHEL kernel."
+            loggerinst.info("\n%s" % info_message)
+            self.add_message(
+                level="INFO",
+                id="CONFLICT_OF_KERNELS",
+                title="Conflict of installed kernel versions",
+                description=info_message,
             )
```
Venefilyn, line 79: Could you also reword this to be easier to understand what this means, why it matters, and how to try to fix it. Atm the comment above gives more detail than the message

convert2rhel/pkghandler.py
danmyway: Right, I forgot to put it back :sweat_smile: 
Thanks!

convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
danmyway: I'll try.

convert2rhel/pkghandler.py
```diff
@@ -1057,3 +1057,21 @@ def _parse_pkg_with_dnf(pkg):
...
+    """
+    try:
+        return max(pkgs[1])
+    except ValueError:
+        loggerinst.debug("The list of %s packages is empty." % pkgs[0])
+
+    return
```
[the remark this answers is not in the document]
r0x0d: I would vote to continue without the return statement. It does not add anything to the current implementation. 

Also, not needed to explicitly say that it returns `None`. An empty return will always be `None`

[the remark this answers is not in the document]
Venefilyn: Going by this, and it makes sense IMO. When returns have some explicit return then we should be explicit everywhere

https://github.com/oamg/convert2rhel/security/code-scanning/926

convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
```diff
@@ -50,34 +50,68 @@ def run(self):
...
+                "We will try to install a lower version of the package, "
+                "remove the conflicting kernel and then update to the latest security patched version."
+            )
+            loggerinst.info("\n%s" % info_message)
+            self.add_message(
+                level="INFO",
+                id="CONFLICT_OF_KERNELS",
+                title="Conflict of installed kernel versions",
+                description=info_message,
+            )
```
Venefilyn, line 84: Given that the info_message talks about doing something but the action message won't be displayed until it's done means it's a bit confusing. We could reword it a bit as a refactor (not a blocker on this PR as it's out of scope, only change if you feel like it)

danmyway: Agreed, it's kind of mind-bending to figure out the wording :exploding_head:

convert2rhel/pkghandler.py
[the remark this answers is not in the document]
danmyway: I think I'll go with the exception.
Given the context where it's used, if the list is indeed empty, there is something very wrong, and it should raise the exception.
If we ever decide to use the function elsewhere, it can be refactored

Venefilyn:
Unit test failing with this, which seems like something that shouldn't happen. Is there something with the implementation?
```
version1 = 'kernel-2.6.32-754.33.1.el7.x86_64'
version2 = 'kernel-uek-headers-0.1-1.x86_64'

    def compare_package_versions(version1, version2):
        """Compare two package versions against each other, including name and arch.
        This function will receive packages in any of the following formats:
            * ENVR
            * ENVRA
            * NVR
            * NEVR
            * NVRA
            * NEVRA
        :param version1: The version to be compared.
        :type version1: str
        :param version2: The version to compare against.
        :type version2: str
    
        .. example::
            >>> match = compare_package_versions("kernel-core-5.14.10-300.fc35", "kernel-core-5.14.15-300.fc35")
            >>> print(match) # -1
    
        .. note::
            Since the return type is a int, this could be difficult to understand
            the meaning of each number, so here is a list that represents
            every possible number:
                * -1 if the version1 is less then version2 version
                * 0 if the version1 is equal version2 version
                * 1 if the version1 is greater than version2 version
    
        :raises ValueError: In case of packages name being different.
        :raises ValueError: In case of architectures being different.
    
        :return: Return a number indicating if the versions match, are less or greater then.
        :rtype: int
        """
        # call parse_pkg_string to obtain a list containing name, epoch, release, version and arch fields
        # order of fields returned: name, epoch, version, release, arch
        version1_components = parse_pkg_string(version1)
        version2_components = parse_pkg_string(version2)
    
        # ensure package names match, error if not
        if version1_components[0] != version2_components[0]:
            raise ValueError(
                "The package names ('%s' and '%s') do not match. Can only compare versions for the same packages."
>               % (version1_components[0], version2_components[0])
            )
E           ValueError: The package names ('kernel' and 'kernel-uek-headers') do not match. Can only compare versions for the same packages.
```

convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
```diff
@@ -50,34 +50,68 @@ def run(self):
...
+            pkghandler.handle_no_newer_rhel_kernel_available()
+            kernel_update_needed = True
+
+        # In this case all kernel packages were already replaced during the main transaction
+        elif not non_rhel_kernels:
+            return
```
r0x0d, line 90: Not at all required for this PR, but I think we could move this outside and turn into a if statement to skip early the rest of the code. 

```suggestion
        if not non_rhel_kernels:
            return
```

r0x0d: Would look better for the workflow and easier to debug in future versions

convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
```diff
@@ -50,34 +50,68 @@ def run(self):
...
+            if is_rhel_kernel_higher == 1:
+                return
+
+            # This also contains a scenario, where the running non-RHEL kernel is of a higher version
+            # than the latest one available in the RHEL repositories.
+            # That might happen and happened before, when the original vendor patches the package
+            # with a higher release number.
+            else:
+                pkghandler.handle_no_newer_rhel_kernel_available()
+                kernel_update_needed = True
```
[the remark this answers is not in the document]
danmyway: Agreed

convert2rhel/actions/conversion/preserve_only_rhel_kernel.py
danmyway: Reverted to `elif` in the end with the current implementation.
I've left an explanation in the comment.

tests/integration/common/checks-after-conversion/test_yum_check_update.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+
+@pytest.mark.parametrize("package", ["kernel"])
```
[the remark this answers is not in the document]
danmyway: We could go with asterisk, I'll include it in #946, so the running tests are not dismissed

kokesak (approved):
qe ack

danmyway:
Sanity suit passed:
```
+--------------------------------------+-----------------------------------------+------------------------------------------------------------------------------------------+--------+
| UUID                                 | Target                                  | Test Plan                                                                                | Result |
+--------------------------------------+-----------------------------------------+------------------------------------------------------------------------------------------+--------+
| dd148f6d-f68c-48a0-bcca-be847b6c88b0 | AlmaLinux OS 8.10.20240530 x86_64       |                                                                                          |        |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/activation_key_conversion        | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/pre_registered_system_conversion | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/rhsm_conversion                  | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/satellite_conversion             | PASSED |
|                                      |                                         | plans/tier0/sanity/non-destructive                                                       | PASSED |
| cd7d8d4f-8e98-425c-93f9-148bf07610ac | AlmaLinux OS 8.8.20230524 x86_64        |                                                                                          |        |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/activation_key_conversion        | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/pre_registered_system_conversion | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/rhsm_conversion                  | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/satellite_conversion             | PASSED |
|                                      |                                         | plans/tier0/sanity/non-destructive                                                       | PASSED |
| 824313ff-ef5f-4920-a46f-671745b933e3 | OL7.9-x86_64-HVM-2023-01-05             |                                                                                          |        |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/activation_key_conversion        | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/pre_registered_system_conversion | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/rhsm_conversion                  | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/satellite_conversion             | PASSED |
|                                      |                                         | plans/tier0/sanity/non-destructive                                                       | PASSED |
| 180bb850-0d16-457e-aa33-0ab07d35696e | OL8.9-x86_64-HVM-2024-02-02             |                                                                                          |        |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/activation_key_conversion        | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/pre_registered_system_conversion | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/rhsm_conversion                  | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/satellite_conversion             | PASSED |
|                                      |                                         | plans/tier0/sanity/non-destructive                                                       | PASSED |
| fa109008-bb49-4fba-84a1-af5c3a850d85 | Rocky-8-EC2-Base-8.10-20240528.0.x86_64 |                                                                                          |        |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/activation_key_conversion        | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/pre_registered_system_conversion | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/rhsm_conversion                  | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/satellite_conversion             | PASSED |
|                                      |                                         | plans/tier0/sanity/non-destructive                                                       | PASSED |
| 678b80a6-584d-4d96-836c-8ade09d3ffca | Rocky-8-EC2-Base-8.8-20230518.0.x86_64  |                                                                                          |        |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/activation_key_conversion        | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/pre_registered_system_conversion | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/rhsm_conversion                  | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/satellite_conversion             | PASSED |
|                                      |                                         | plans/tier0/sanity/non-destructive                                                       | PASSED |
| de0a6860-8b17-4edd-ac35-0e1039996805 | CentOS-7-latest                         |                                                                                          |        |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/activation_key_conversion        | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/one_key_satellite_conversion     | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/pre_registered_system_conversion | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/rhsm_conversion                  | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/satellite_conversion             | PASSED |
|                                      |                                         | plans/tier0/sanity/non-destructive                                                       | PASSED |
| a832297e-500b-43d0-a29c-fe5f22a74d7e | CentOS-8-latest                         |                                                                                          |        |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/activation_key_conversion        | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/pre_registered_system_conversion | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/rhsm_conversion                  | PASSED |
|                                      |                                         | plans/tier0/sanity/destructive/basic_conversion_methods/satellite_conversion             | PASSED |
|                                      |                                         | plans/tier0/sanity/non-destructive                                                       | PASSED |
+--------------------------------------+-----------------------------------------+------------------------------------------------------------------------------------------+--------+
```

r0x0d (approved):
Code looks good. 

We are okay with having the `global` keyword for now, but we will open a ticket to refactor this in the future. No need to block the PR because of it.

tests/integration/common/checks-after-conversion/test_yum_check_update.py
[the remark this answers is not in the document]
danmyway: Included here.

PR #1336, opened 2024-08-08
[RHELC-1673] Include all reports in post-conversion report

hosekadam:
Include all reports in post-conversion report to be consistent with behavior of the analysis report, where all the reports are included.

Jira Issues:

- [RHELC-1673](https://issues.redhat.com/browse/RHELC-1673)

Checklist

### DIFF
--- a/convert2rhel/main.py
+++ b/convert2rhel/main.py
@@ -180,7 +180,7 @@ def main_locked():
         _raise_for_skipped_failures(post_conversion_results)
         report.post_conversion_report(
             results=post_conversion_results,
-            include_all_reports=False,
+            include_all_reports=True,
             disable_colors=logger_module.should_disable_color_output(),
         )
 


Venefilyn (dismissed):
IMO underlying `_report()` should be refactored completely so that we are using a dict with levels and name, just like how python logging works. I.e. assigning a number to each level

On that note too, `include_all_reports` isn't very telling of what it means. I'd go for refactoring it completely with a key-value pair instead like mentioned above

pr-watson (approved):
Looks good to me!

r0x0d:
> IMO underlying `_report()` should be refactored completely so that we are using a dict with levels and name, just like how python logging works. I.e. assigning a number to each level
> 
> On that note too, `include_all_reports` isn't very telling of what it means. I'd go for refactoring it completely with a key-value pair instead like mentioned above

We have a specific ticket for that refactor. It is not viable to handle this as part of this PR I would say.

Venefilyn:
> We have a specific ticket for that refactor. It is not viable to handle this as part of this PR I would say.

That's fine. We don't even know what action statuses we have atm (well we do but like, there is nothing preventing us from typing WARMING instead of WARNING somewhere AFAIK) which isn't great when saying yeah lets include everything

@pr-watson @hosekadam could either of you create a Jira issue about this though? Making actions type safe and having a key-value dict between levels and their value

hosekadam:
> could either of you create a Jira issue about this though? Making actions type safe and having a key-value dict between levels and their value
@Venefilyn I think there is some checking, when I tried to create a wrong level, it crashed on the check. Maybe it's this https://github.com/oamg/convert2rhel/blob/main/convert2rhel/actions/__init__.py#L366 ? But I'm not sure if that's what you meant. Also I see some mapping there: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/actions/__init__.py#L68

pr-watson:
> > could either of you create a Jira issue about this though? Making actions type safe and having a key-value dict between levels and their value
> 
> @Venefilyn I think there is some checking, when I tried to create a wrong level, it crashed on some check. Maybe it's this https://github.com/oamg/convert2rhel/blob/main/convert2rhel/actions/__init__.py#L366 ? But I'm not sure if that's what you meant. Also I see some mapping there: https://github.com/oamg/convert2rhel/blob/main/convert2rhel/actions/__init__.py#L68

Yes those are the correct places @hosekadam

PR #1342, opened 2024-08-15
Add dependency for non redhat pkgs listing

r0x0d:
Without this dependency, the action can run before the package conversion in the system. Ideally, we want this to run after the `KERNEL_PACKAGES_INSTALLATION`, as it is no use to have this run before the package replacement.

Pointed out by @hosekadam in https://github.com/oamg/convert2rhel/pull/1256#discussion_r1718100982

Jira Issues:

-

Checklist

### DIFF
--- a/convert2rhel/actions/conversion/list_non_red_hat_pkgs_left.py
+++ b/convert2rhel/actions/conversion/list_non_red_hat_pkgs_left.py
@@ -27,6 +27,7 @@ loggerinst = logging.getLogger(__name__)
 
 class ListNonRedHatPkgsLeft(actions.Action):
     id = "LIST_NON_RED_HAT_PKGS_LEFT"
+    dependencies = ("KERNEL_PACKAGES_INSTALLATION",)
 
     def run(self):
         """List all the packages that have not been replaced by the


convert2rhel/actions/conversion/list_non_red_hat_pkgs_left.py
```diff
@@ -27,6 +27,7 @@
 
 class ListNonRedHatPkgsLeft(actions.Action):
     id = "LIST_NON_RED_HAT_PKGS_LEFT"
+    dependencies = ("CONVERT_SYSTEM_PACKAGES",)
```
hosekadam, line 30: Shouldn't it be the `FIX_DEFAULT_KERNEL`? (which I believe is the last from `preserve_only_rhel_kernel.py`)

The original order I use as source: https://github.com/oamg/convert2rhel/blob/v1.7.1/convert2rhel/main.py#L299

Maybe in the nonredhatpkgs could be listed some non-RHEL kernel which would be removed during `KernelPkgsInstall`

r0x0d: That's true. Good point. 

In that case, I believe it would be `KERNEL_PACKAGES_INSTALLATION`, as we would have the installed kernel ready and all the dependencies from it as well. Will update the code

r0x0d:
```suggestion
    dependencies = ("KERNEL_PACKAGES_INSTALLATION",)
```

hosekadam (approved):
Yeah, this should work too, you are right. LGTM!

hosekadam:
I think the failed tests are not related to this PR, can be merged

PR #1358, opened 2024-08-27
[RHELC-1715] - conftest.py refactor - rearange functions

kokesak:
Rearange functions in conftest.py (part1)

Do NOT merge before https://github.com/oamg/convert2rhel/pull/1349 (DONE)

Jira Issues:

- [RHELC-1715](https://issues.redhat.com/browse/RHELC-1715) 
- [RHELC-1620]( https://issues.redhat.com/browse/RHELC-1620) 

Checklist

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -171,7 +171,7 @@ class SubscriptionManager:
 
         return self.shell(command)
 
-    def set_up_requirements(self):
+    def set_up_requirements(self, use_staging_cdn=False):
         """
         Usual full preparation workflow.
         Calls, where applicable:
@@ -179,12 +179,26 @@ class SubscriptionManager:
             self.add_keys_and_certificates()
             self.add_client_tools_repo()
             self.install_package()
+            self.set_up_to_stagecdn() # only when use_staging_cdn is True
         """
         if SystemInformationRelease.distribution == "oracle":
             self.remove_package(package_name="rhn-client-tools")
         self.add_keys_and_certificates()
         self.add_client_tools_repo()
         self.install_package()
+        if use_staging_cdn:
+            self.set_up_to_stagecdn()
+
+    def set_up_to_stagecdn(self):
+        # Point the server hostname to the staging environment,
+        # so we don't need to pass it to convert2rhel explicitly
+        # RHSM baseurl gets pointed to a stage cdn
+        self.shell(
+            "subscription-manager config --rhsm.baseurl=https://{0} --server.hostname={1}".format(
+                TEST_VARS["RHSM_STAGECDN"], TEST_VARS["RHSM_SERVER_URL"]
+            ),
+            silent=True,
+        )
 
     def clean_up(self):
         """
@@ -199,6 +213,24 @@ class SubscriptionManager:
         self.remove_package()
         self.remove_client_tools_repo()
         self.remove_keys_and_certificates()
+
+
+@pytest.fixture()
+def fixture_subman():
+    """
+    Fixture.
+    Set up the subscription manager on the system. Wrapper around SubscriptionManager class and its methods.
+    By default sets the subscription manager to the stagecdn (needed for SCA Enabled accounts). If you want
+    to disable it, please edit this fixture to utilize parametrization.
+    """
+    subman = SubscriptionManager()
+
+    subman.set_up_requirements(use_staging_cdn=True)
+
+    yield
+
+    if "C2R_TESTS_NONDESTRUCTIVE" in os.environ:
+        subman.clean_up()
 
 
 @pytest.fixture()
@@ -505,27 +537,6 @@ def log_file_data():
 
 
 @pytest.fixture(scope="function")
-def required_packages(shell):
-    """
-    Installs packages based on values under TEST_REQUIRES envar in tmt metadata, when called.
-    """
-    try:
-        required_packages = os.environ.get("TEST_REQUIRES").split(" ")
-        for package in required_packages:
-            print(f"\nPREPARE: Installing required {package}")
-            assert shell(f"yum install -y {package}")
-
-        yield
-
-        for package in required_packages:
-            print(f"\nCLEANUP: Removing previously installed required {package}")
-            assert shell(f"yum remove -y *{package}*")
-
-    except KeyError:
-        raise
-
-
-@pytest.fixture(scope="function")
 def remove_repositories(shell, backup_directory):
     """
     Fixture.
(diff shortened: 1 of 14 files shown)


tests/integration/conftest.py
```diff
@@ -201,6 +212,32 @@ def clean_up(self):
...
+    """
+    subman = SubscriptionManager()
+
+    subman.set_up_requirements()
+
+    if hasattr(request, "param"):
+        set_to_stage = request.param
+        print(f">>> Using parametrized option requested in the fixture - {set_to_stage}.")
+        if set_to_stage == "set_to_stage=true":
+            subman.set_up_to_stagecdn()
```
danmyway, line 231: The fixture is nice, but feels kind of overengineered.
Here's a suggestion, what about just having `set_cdn_to_stage=False` parameter for the `SubscriptionManager.set_up_requirements()` method and condition that inside?
I understand that we would lose some visibility around the setup in tests using the fixture, but that can be tackled in the docstring of the test function.
With that, the fixture usability would narrow down just to subman installation and setting the baseurl to the stage.
But we don't really need to wrap the subman class for any other scenario than the stage setup.
Its methods are either called directly in tests, if needed, or inside another fixtures (pre_register, satellite)
WDYT @kokesak ?
```suggestion
@pytest.fixture()
def fixture_subman():
    """
    Fixture.
    Set up the subscription manager on the system. Wrapper around SubscriptionManager class and its methods.
    """
    subman = SubscriptionManager()

    subman.set_up_requirements(set_cdn_to_stage=True)

```

tests/integration/conftest.py
```diff
@@ -1087,3 +980,122 @@ def fixture_satellite(request):
...
+        subprocess.run(
+            ["yum", "remove", "-y", "kernel-uek"], check=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+
+    yield
+
+
+def workaround_grub_setup(shell):
+    """
+    Workaround fixture.
```
SerCantus, line 1078: Is this still a fixture?

tests/integration/conftest.py
```diff
@@ -201,6 +212,32 @@ def clean_up(self):
         self.remove_keys_and_certificates()
 
 
+@pytest.fixture()
+def fixture_subman(request):
```
SerCantus, line 216: Maybe the fixture name could be improved, what was wrong with the old name?

SerCantus (changes requested):

tests/integration/conftest.py
```diff
@@ -512,18 +549,18 @@ def required_packages(shell):
...
-        for package in required_packages:
-            print(f"\nCLEANUP: Removing previously installed required {package}")
-            assert shell(f"yum remove -y *{package}*")
+    yield
 
-    except KeyError:
-        raise
+    for package in required_packages:
+        print(f"\nCLEANUP: Removing previously installed required {package}")
+        assert shell(f"yum remove -y *{package}*")
```
danmyway, line 563: This might be a good opportunity to drop this (`required_packages()`) in favor of https://tmt.readthedocs.io/en/stable/spec/tests.html#require
Honestly, I'm not sure if tmt will handle the teardown part and remove the package after each test, or if it will be able to handle different sets of packages for different distributions.
Looking in the docs we might need to use prepare to be able to run with different sets of packages, which might not be viable for our current set up.
I'll leave that up to you if you want to pursue this in either this or another PR.

tests/integration/conftest.py
```diff
@@ -201,6 +212,32 @@ def clean_up(self):
...
+    subman.set_up_requirements()
+
+    if hasattr(request, "param"):
+        set_to_stage = request.param
+        print(f">>> Using parametrized option requested in the fixture - {set_to_stage}.")
+        if set_to_stage == "set_to_stage=true":
+            subman.set_up_to_stagecdn()
+
+    # TODO: should we add register here?
+    # Resolve under the: https://issues.redhat.com/browse/RHELC-1134
```
danmyway, line 234: We don't need to pre-register here.
The intention behind this is to just set the baseurl to stage.
That cannot be done without having the subscription-manager installed prior.
The registration is handled by convert2rhel itself.

danmyway (commented):
I like the unification overall.
Just a couple of suggestions/discussion points.

tests/integration/conftest.py
```diff
@@ -512,18 +549,18 @@ def required_packages(shell):
     """
     try:
         required_packages = os.environ.get("TEST_REQUIRES").split(" ")
-        for package in required_packages:
-            print(f"\nPREPARE: Installing required {package}")
-            assert shell(f"yum install -y {package}")
+    except KeyError:
+        raise
```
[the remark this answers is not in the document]
kokesak: I actually realized that `get()` return None if nothing is found so this `KeyError` exception can never happen.
```
>>> os.environ.get("TEST_REQUIRES").split(" ")
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'split'
```
I'll have to change it to something else

tests/integration/conftest.py
kokesak: Right, good catch. It is not a fixture.

tests/integration/conftest.py
kokesak: Ack, I remove it

tests/integration/conftest.py
kokesak: We did not have this fixture before, so the name is new :D. However, I understand that this doesn't look the best, but I wanted to keep the prefix `fixture` in the name as it correspond to what it actually is and is consistent with the fixture_satellite [here](https://github.com/oamg/convert2rhel/blob/main/tests/integration/conftest.py#L1071). If you have a better name for this, feel free to suggest it

tests/integration/conftest.py
```diff
@@ -185,6 +185,19 @@ def set_up_requirements(self):
...
+        if set_to_stage:
+            self.set_up_to_stagecdn()
+
+    def set_up_to_stagecdn(self):
+        # Point the server hostname to the staging environment,
+        # so we don't need to pass it to convert2rhel explicitly
+        # RHSM baseurl gets pointed to a stage cdn
+        self.shell(
+            "subscription-manager config --rhsm.baseurl=https://{} --server.hostname={}".format(
+                TEST_VARS["RHSM_STAGECDN"], TEST_VARS["RHSM_SERVER_URL"]
```
Venefilyn, line 197: Just for clarity
```suggestion
            "subscription-manager config --rhsm.baseurl=https://{0} --server.hostname={1}".format(
                TEST_VARS["RHSM_STAGECDN"], TEST_VARS["RHSM_SERVER_URL"]
```

tests/integration/conftest.py
```diff
@@ -171,7 +171,7 @@ def unregister(self):
 
         return self.shell(command)
 
-    def set_up_requirements(self):
+    def set_up_requirements(self, set_to_stage=False):
```
Venefilyn, line 174: I'd use a different name
```suggestion
    def set_up_requirements(self, use_staging_cdn=False):
```

tests/integration/conftest.py
kokesak: Thanks, I'll add it for better readability

tests/integration/conftest.py
kokesak: renamed

tests/integration/conftest.py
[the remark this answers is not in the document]
kokesak: Fixture removed completely in the end. More here - https://github.com/oamg/convert2rhel/pull/1358#discussion_r1738072480

tests/integration/conftest.py
kokesak: done

tests/integration/tier0/non-destructive/subscription-manager/main.fmf
```diff
@@ -18,13 +18,8 @@ tag+:
...
         backed up and re-installed during the rollback together with the certificate.
 
-    adjust+:
-        - enabled: true
-          when: >
-              distro == centos-8
-
     environment+:
-        TEST_REQUIRES: subscription-manager
+        C2R_TESTS_SUBMAN_CLEANUP: 1
```
danmyway, line 22: This is kind of no-op.
The only place this envar is used at is the `pre_registered` fixture, which honestly does not really make sense anymore, given we have the `C2R_TESTS_NONDESTRUCTIVE` envar exported for all non-destructive tests.
So I would remove the whole `environment+:` from here, change the condition in the `pre_registered` to utilize `C2R_TESTS_NONDESTRUCTIVE` and remove `C2R_TESTS_SUBMAN_CLEANUP` everywhere, please.

danmyway (changes requested):
Looks good overall, just a couple of minor.

tests/integration/conftest.py
```diff
@@ -171,20 +171,34 @@ def unregister(self):
...
-    def set_up_requirements(self):
+    def set_up_requirements(self, use_staging_cdn=False):
         """
         Usual full preparation workflow.
         Calls, where applicable:
             self.remove_package(package_name="rhn-client-tools") # only on Oracle Linux
             self.add_keys_and_certificates()
             self.add_client_tools_repo()
             self.install_package()
+            self.use_staging_cdn() # only when use_staging_cdn is True
```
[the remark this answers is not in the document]
kokesak: nice finding, thanks!

tests/integration/tier0/non-destructive/subscription-manager/main.fmf
kokesak: Sounds nice, yeah we should remove this `C2R_TESTS_SUBMAN_CLEANUP` env var and replace it with the more generic `C2R_TESTS_NONDESTRUCTIVE`

tests/integration/conftest.py
```diff
@@ -201,6 +215,24 @@ def clean_up(self):
         self.remove_keys_and_certificates()
 
 
+@pytest.fixture()
+def fixture_subman():
+    """
+    Fixture.
+    Set up the subscription manager on the system. Wrapper around SubscriptionManager class and its methods.
+    By default sets the subscription manager to the stagecdn (needed for SCA Enabled accounts). If you want
```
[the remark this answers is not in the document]
kokesak: I'll try to test it for the els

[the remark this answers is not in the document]
kokesak: The ELS needs to have the stagcdn set in sub man config, otherwise we are getting the Unauthorized error. After offline discussion we have decided to keep it as it is, only changing the doc string

danmyway (approved):
Looks great, thanks @kokesak !

danmyway:
@SerCantus if you could re-review, please :)

SerCantus (approved):
Looking great! Nice job.

PR #1373, opened 2024-09-10
refactor: Improved logger usage

Venefilyn:
This refactor of the usage of our logging is to get away from custom
functions that causes attribute errors in editors and IDEs. This will
also standarize the usage of logger and the loglevels.

In future PRs we can include changes to include prepare, rollback, and
final specific message styles.

Jira Issues:

-

Depends on
- #1327 

Checklist


convert2rhel/main.py
```diff
@@ -136,6 +136,7 @@ def main_locked():
     initialize_file_logging("convert2rhel.log", logger_module.LOG_DIR)
 
     try:
+        logger_module.ConversionStage.set_stage("prepare")
```
r0x0d, line 139: What about using `ConversionPhase` for this? I think it could be really good to map the phases in that class to the `ConversionStage` you created

Venefilyn: @r0x0d That's a good point. Though that made me dig down a rabbit hole of refactoring to be a bit nicer to use for more scenarios. What do you think of this?

```py
class ConversionPhase:
    """Conversion phase to hold name and logging-friendly name.
    """

    def __init__(self, name, log_name=None):  # type: (str, str|None) -> None
        self.name = name
        self.log_name = log_name

    def __str__(self):
        return self.log_name

    def __repr__(self):
        return self.name

    def __hash__(self) -> int:
        return hash(self.name)


class ConversionPhases(dict):
    """During conversion we will be in different states depending on
    where we are in the execution. This class establishes the different phases
    that we have as well as what the current phase is set to.
    """
    _phases = [
        ConversionPhase(name="POST_CLI", log_name="Prepare"),
        ConversionPhase(name="PRE_PONR_CHANGES", log_name="Prepare")
    ]

    current_phase = None  # type: ConversionPhase|None

    def __getitem__(self, key):  # type: (str) -> ConversionPhase|None
        return next(phase for phase in self._phases if phase.name == key )

    @classmethod
    def set_stage(cls, phase):  # type: (str) -> None
        if phase is None or phase in cls._phases:
            cls.current_phase = cls._phases[phase]
        raise NotImplementedError("The stage {} is not implemented in the ConversionStage class".format(phase))

test = ConversionPhases()

print(test["POST_CLI"].__repr__)
```
```
vscode(container):/workspaces/convert2rhel$ /bin/python3 /workspaces/convert2rhel/convert2rhel/utils/test.py
<bound method ConversionPhase.__repr__ of POST_CLI>
```

r0x0d: Hmmm, this work as well, but it is easy to introduce an error if we miss the correct naming in other parts of the code or introduce a new phase and forget about it. 

I was thinking in something more simple, like this: 

```python
class ConversionPhase:
    POST_CLI = 1
    # PONR means Point Of No Return
    PRE_PONR_CHANGES = 2
    # Phase to exit the Analyze SubCommand early
    ANALYZE_EXIT = 3
    POST_PONR_CHANGES = 4

class ConversionStage:
    stages = {
        ConversionPhase.PRE_PONR_CHANGES: "Prepare",
        ConversionPhase.POST_PONR_CHANGES: "Final",
        ...
    }
    current = None  # type: str|None

    @classmethod
    def set_stage(cls, stage):  # type: (str) -> None
        if stage == None or stage in cls.stages:
            cls.current = cls.stages[stage]
        raise NotImplementedError("The stage {} is not implemented in the ConversionStage class".format(stage))
```

Then, when we initialize the `ConversionStage`, we could just simply: 

```python
logger_module.ConversionStage.set_stage(ConversionPhase.PRE_PONR_CHANGES)
```

And then we have it mapped everywhere that `PRE_PONR_CHANGES` matches the `Prepare` or something else. The interface will still look clean enough in the new class, and we will make sure that if someone changes the meaning for any of those phases, we will catch somewhere as well. 

Since our use case is so small for now, I think that reusing the phases from `ConversionPhase` should be good enough.

r0x0d: I think our `ConversionPhase` class should hold the information where we are in the process at all times, being our source of truth for many different operations. It is true that logging phases may differ from execution phases, but if that's really the case, then we could come up with an intermediate class that will hold the logging phases for us, tied to the conversion phase if necessary.

Venefilyn: Ah I didn't see this as I got hyperfocused on overcomplicating the solution... but anyhow I did change it now to be a bit safer but in a way that means we still rely on strings. But given exceptions being raised we should catch it in tests if we do a mistake

convert2rhel/utils/phase.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+
+    POST_CLI = ConversionPhase(name="POST_CLI")
+    PREPARE = ConversionPhase(name="PREPARE", log_name="Prepare")
+    # PONR means Point Of No Return
+    PRE_PONR_CHANGES = ConversionPhase(name="PRE_PONR_CHANGES", log_name="Analyze")
+    # Phase to exit the Analyze SubCommand early
+    ANALYZE_EXIT = ConversionPhase(name="ANALYZE_EXIT")
+    POST_PONR_CHANGES = ConversionPhase(name="POST_PONR_CHANGES", log_name="Convert")
+    ROLLBACK = (ConversionPhase(name="ROLLBACK", log_name="Rollback"),)
```
r0x0d, line 54:
```suggestion
    ROLLBACK = ConversionPhase(name="ROLLBACK", log_name="Rollback")
```

convert2rhel/phase.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class ConversionPhases:
+    """During conversion we will be in different states depending on
+    where we are in the execution. This class establishes the different phases
+    that we have as well as what the current phase is set to.
+    """
+
+    POST_CLI = ConversionPhase(name="POST_CLI")
+    PREPARE = ConversionPhase(name="PREPARE", log_name="Prepare")
```
hosekadam, line 48: I just wonder, when we are changing the logging and the final is renamed to convert (which makes sense to me) - what about changing the `Prepare` to `Analysis` (or something like that)? I think it would be nice to have it matching the subcommands (analyze and convert).

Venefilyn: Agreed and I had that before, but I wanted minimal changes in this PR to the actual logs so wanted it to remain the same. We can change it in a follow-up PR but requires a bit more discussion IMO

hosekadam (approved):
Looks good, just wait till the infrastructure problems are resolved and try the sanity tests.

convert2rhel/phase.py
hosekadam: Okay. Maybe if we want minimal changes, then I would change the name of the post_conversion to the `Final` to be consistent

SerCantus (commented):
The changes look great!
I see the integration tests are failing and some of them related to the expected messages, maybe something got messed up? I can't see it. For example in integration tests in CentOS7 it expects to see CUSTOM_REPOSITORIES_ARE_VALID::UNABLE_TO_ACCESS_REPOSITORIES but instead C2R fails with:

> convert2rhel("-y --no-rhsm --enablerepo fake-rhel-8-for-x86_64-baseos-rpms --debug", unregister=True)

> (ERROR) REMOVE_SPECIAL_PACKAGES::SPECIAL_PACKAGE_REMOVAL_FAILED - Failed to
>     remove some packages necessary for the conversion.
>      Description: The cause of this error is unknown, please look at the
> diagnosis for more information.
>      Diagnosis: Couldn't download the centos-
> release-7-9.2009.2.el7.centos.x86_64 package. This means we will not be able to
> do a complete rollback and may put the system in a broken state.
>     Check to make sure that the CentOS Linux repositories are enabled and the
> package is updated to its latest version.

SerCantus:
> @SerCantus is it occurring in other PRs or in main?
@Venefilyn I commented after reviewing this PR checks but I see other PRs have similar failures so it might not be caused by this one. After discussing, it might be related to some infra.

Venefilyn:
Ack, feel free to approve and merge with squash

danmyway (approved):
Test failures unrelated.
There is a regression caused by #1288 reported under [RHELC-1753](https://issues.redhat.com/browse/RHELC-1753)

PR #1384, opened 2024-09-24
Improve test executions and RPM generation

Venefilyn:
This changes some commands in Makefile to be generated instead of being
duplicated across multiple PHONY calls. We still have all the PHONY
declarations that we want but it is now a lot easier to maintain as we
only have one command per each function to do everything we want

rpmbuilds are now handled in the same container as we run tests through
which helps us with maintaining less files. On top of that we can build
rpms through the published images we have on ghcr

On top of this we can now run tests in parallel through `make -j3 tests`
that will run all containers at once. This means we get results faster
at the expense of the output looking worse, but that is something we can
handle later

Future refactor of this could be to use compose files or kubes instead.
But as of right now this is plenty good for just a quick sanity check.

Jira Issues:

-

Checklist


scripts/build_locally.sh
```diff
@@ -18,10 +18,8 @@ echo "Building the RPM ..."
...
 echo "RPM was built successfully"
 echo "Cleaning up the target directory..."
-mkdir -p .rpms
-mkdir -p .srpms
-rm -frv .rpms/*
-rm -frv .srpms/*
-mv -vf ~/rpmbuild/RPMS/noarch/* .rpms/
-mv -vf ~/rpmbuild/SRPMS/* .srpms/
+mkdir -p /.rpms
+mkdir -p /.srpms
```
r0x0d, line 22: We want to really create the folders in `/`? Since the script is for locally building, maybe we want to avoid that?

Venefilyn: Ahh right I was just thinking about containers.. good point. I just needed a way to have read-only project root but read-write .srpm and .rpm directory... hmm

Venefilyn: @r0x0d done! Should be fixed now

r0x0d (approved):

Venefilyn:
Merging, should work but can send follow-up PRs tomorrow if it doesn't

PR #1386, opened 2024-09-25
[RHELC-1508] Add deprecation notice about environment variables

r0x0d:
With the most recent work to migrate all environment variables to the configuration file, we need now to warn the user that the variables will be deprecated in order to use configuration file as our main place to track configuarations.

This patch is introducing a very lightweight change to make sure that we will be able modify the code in the future to remove all environment variables.

Jira Issues:

- [RHELC-1508](https://issues.redhat.com/browse/RHELC-1508-)

Checklist

### DIFF
--- a/convert2rhel/toolopts/__init__.py
+++ b/convert2rhel/toolopts/__init__.py
@@ -17,6 +17,7 @@
 __metaclass__ = type
 
 import logging
+import os
 
 from convert2rhel.utils.subscription import setup_rhsm_parts
 
--- a/convert2rhel/utils/__init__.py
+++ b/convert2rhel/utils/__init__.py
@@ -1129,11 +1129,10 @@ def warn_deprecated_env(env_name):
     """
     if env_name not in os.environ:
         # Nothing to do here.
-        return
+        return None
 
     root_logger.warning(
-        "The environment variable {} is deprecated and is set to be removed on Convert2RHEL 2.4.0.\n"
-        "Please, use the configuration file instead."
+        "The environment variable {} is deprecated in favor of using a configuration file and will be removed in version Convert2RHEL 2.4.0."
     )
 
     return os.getenv(env_name, None)
--- a/convert2rhel/utils/__init__.py
+++ b/convert2rhel/utils/__init__.py
@@ -797,6 +797,7 @@ def report_on_a_download_error(output, pkg):
                 " 'CONVERT2RHEL_INCOMPLETE_ROLLBACK=1'." % (pkg, system_info.name)
             )
         else:
+            warn_deprecated_env("CONVERT2RHEL_INCOMPLETE_ROLLBACK")
             logger.warning(
                 "Couldn't download the %s package. This means we will not be able to do a"
                 " complete rollback and may put the system in a broken state.\n"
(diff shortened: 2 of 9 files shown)


convert2rhel/utils/__init__.py
```diff
@@ -1115,3 +1116,20 @@ def write_json_object_to_file(path, data, mode=0o600):
...
+    """
+    if env_name not in os.environ:
+        # Nothing to do here.
+        return
+
+    root_logger.warning(
+        "The environment variable {} is deprecated and is set to be removed on Convert2RHEL 2.4.0.\n"
+        "Please, use the configuration file instead."
+    )
+    return True
```
Venefilyn, line 1135: IMO create a new utils file and use that when introducing new code so we can migrate away from everything in `__init__.py` faster

convert2rhel/utils/__init__.py
```diff
@@ -1115,3 +1116,20 @@ def write_json_object_to_file(path, data, mode=0o600):
...
+
+def warn_deprecated_env(env_name):
+    """Warn the user that the environment variable is deprecated.
+
+    :param env_name: The name of the environment variable that is deprecated.
+    :type env_name: str
+    """
+    if env_name not in os.environ:
+        # Nothing to do here.
+        return
```
Venefilyn, line 1129: I'd rather be explicit. If it doesn't fit then we can change it to return 0 and 1 or something but I do want to have explicit returns
```suggestion
        return False
```

Or we can change it to be kinda like getting the environment variable but also warning
```suggestion
        return None
```

convert2rhel/utils/__init__.py
```diff
@@ -1115,3 +1116,20 @@ def write_json_object_to_file(path, data, mode=0o600):
...
+
+    root_logger.warning(
+        "The environment variable {} is deprecated and is set to be removed on Convert2RHEL 2.4.0.\n"
+        "Please, use the configuration file instead."
+    )
```
Venefilyn, line 1134:
```suggestion
    root_logger.warning(
        "The environment variable {} is deprecated in favor of using a configuration file and will be removed in version Convert2RHEL 2.4.0."
    )
```

convert2rhel/utils/__init__.py
Venefilyn, line 1135: If we want to use it as a getting environment variable and returning it we can do this
```suggestion
    return env_name
```

convert2rhel/toolopts/__init__.py
```diff
@@ -117,6 +118,17 @@ def set_opts(self, key, value):
...
+        """Update a given option in toolopts.
+
+        :param key:
+        :type key: str
+        :param value:
+        :type value: str
+        """
+        if key and value:
+            setattr(self, key, value)
+
```
Venefilyn, line 131: Doesn't seem to be used anywhere

Venefilyn (changes requested):

convert2rhel/toolopts/__init__.py
r0x0d: Ah, yes! This is for the other PR I have. Forgot to remove from here.

convert2rhel/utils/__init__.py
r0x0d: updated to return the os.getenv(env_name.)

r0x0d: I didn't created a new one because this is about to be deprecated in a future release. We won't stick with this function very long, so I thought that creating a new file and etc would be just another step to remove it from the codebase later.

r0x0d: Welll, either way, it would work as having the function here or new file. It will get removed in the future anyway.

Venefilyn (approved):
LGTM!

PR #1422, opened 2024-11-04
[RHELC-1748]  Manpage generation check

Andrew-ang9:
This PR sets up the GitHub action so that going forward we will be able to know if we need to generate the manpages per PR basis so that we can always have the manpages up to date.

Jira Issues:

-[RHELC-1748](https://issues.redhat.com/browse/RHELC-1748)

Checklist

- [ ] Label depicting the kind of PR it is 


.github/workflows/build-rpm.yml
```diff
@@ -64,6 +64,10 @@ jobs:
             convert2rhel/__init__.py
           retention-days: 1
 
+      - name: Trigger Manpage Generation
+        run: |
+          oamg/convert2rhel/.github/workflows/generate_manpage.yml@main
+
```
Venefilyn, line 70: To get this to work `generate_manpage.yml` needs to be a runnable workflow. But don't think it's necessary to run it here tbh, we can keep to just running it on every PR

scripts/manpage_generation.sh
Venefilyn, line 1: It's not enough to check with the previous version. I would like to see the manpages generated and the difference checked to see if there are any changes. Not just if the version changed.

As this will fail every time we submit a new version it's already too late to update too.

Venefilyn (changes requested):
In short, I would want to see

* Manpages are generated and checked to see if there are changes
* Manpages are checked for each commit on main and commit in a PR
* Manpages workflow fails if there is a difference and passes if there are no changes

scripts/manpage_generation.sh
```diff
@@ -3,17 +3,26 @@
...
 
-/usr/bin/python -m pip install argparse-manpage six pexpect
+# Extract the current version from the spec file
+CURRENT_VER=$(grep -oP '^Version:\s+\K\S+' packaging/convert2rhel.spec)
+echo "Current version: $CURRENT_VER"
 
 # Generate the manpage using argparse-manpage
-PYTHONPATH=. /usr/bin/python /home/runner/.local/bin/argparse-manpage --pyfile man/__init__.py --function get_parser --manual-title="General Commands Manual" --description="Automates the conversion of Red Hat Enterprise Linux derivative distributions to Red Hat Enterprise Linux." --project-name "convert2rhel $VER" --prog="convert2rhel" --include man/distribution --include man/synopsis > "$MANPAGE_DIR/convert2rhel.8"
+PYTHONPATH=. argparse-manpage --pyfile man/__init__.py --function get_parser --manual-title="General Commands Manual" --description="Automates the conversion of Red Hat Enterprise Linux derivative distributions to Red Hat Enterprise Linux." --project-name "convert2rhel $CURRENT_VER" --prog="convert2rhel" --include man/distribution --include man/synopsis > "$MANPAGE_DIR/convert2rhel.8"
```
Venefilyn, line 19: We're missing python call here. Does it still work locally?

scripts/manpage_generation.sh
```diff
@@ -3,17 +3,26 @@
...
 
-git status
+# Check for differences in the generated manpage
+if ! git diff --quiet HEAD -- "$MANPAGE_DIR/convert2rhel.8"; then
+    echo "Manpages are outdated. Please update them."
+    exit 1
+else
+    echo "Manpages are up-to-date."
+    exit 0
+fi
```
Venefilyn, line 28: Perfect!

scripts/manpage_generation.sh
```diff
@@ -3,17 +3,26 @@
 # Directory to store the generated manpages
 MANPAGE_DIR="man"
 
-echo Generating manpages
+# Ensure the manpage directory exists
+mkdir -p "$MANPAGE_DIR"
```
Venefilyn, line 7: LGTM

Containerfiles/manpage_check.Containerfile
Venefilyn, line 1: I noticed the workflow is not using a container or the script. But if this container works locally it's just a matter of running it in the GitHub workflow

Andrew-ang9: I talked about this PR with Mb, and they suggested that if everything could just be done in the action, that could fix some of the issues that I was having. The action works as it should. The error shown now is a separate issue from this PR, but you can see that at the end of it, it does show the correct message for the manpages, as currently, in this PR, there are changes in the manpages file.

PR #1436, opened 2024-11-22
[RHELC-1759] Remove zero value report messages

bocekm:
The following post-conversion messages give near zero value to the user and we shouldn't waste the users' time as they would most likely read through all of them after the conversion:

- Skipped locking releasever (the releasever lock is needed only when converting to RHEL EUS using RHSM)
- The RHEL kernel has been verified to be on the system
- Skipped host metering configuration because of various reasons such as
  - when the system is not RHEL 7 running on one of the big three hyperscalers (host metering is not applicable anywhere else)
  - the user hasn't set the option to configure host metering (which is totally fine because it's not mandatory and on majority of systems it's not applicable)

Also there's essentially dead code generating a pre-conversion analysis message talking about outdated convert2rhel on RHEL 6 systems which we do not allow converting anymore.

Jira Issues:

- [RHELC-1759](https://issues.redhat.com/browse/RHELC-1759)

Checklist

- [x] Label depicting the kind of PR it is 


hosekadam (approved):
Makes sense. I understand why those messages can be removed, and I agree.

PR #1439, opened 2024-11-26
chore/Bump test targets for Alma and Rocky to 9.5

danmyway:


kokesak (approved):
I triggered the el9 tests to see if it is provisioned correctly. Otherwise, I trust you. Giving an ack

PR #1448, opened 2024-12-11
chore: replace percent strings for logging

Venefilyn:
This replaces logging with their best counterpart. Either normal logging
methods with arguments, or by utilizing `.format()`

Depends on #1431

### DIFF
--- a/convert2rhel/pkghandler.py
+++ b/convert2rhel/pkghandler.py
@@ -328,8 +328,7 @@ def format_pkg_info(pkgs, disable_repos=None):
     max_nvra_length = max(len(nvra) for nvra in package_info)
 
     header = (
-        "%-*s  %-*s  %s"
-        % (
+        "{}-{}  {}-{}  {}".format(
             max_nvra_length,
             "Package",
             max_packager_length,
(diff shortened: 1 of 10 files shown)


convert2rhel/pkghandler.py
```diff
@@ -318,8 +318,7 @@ def format_pkg_info(pkgs, disable_repos=None):
     max_nvra_length = max(len(nvra) for nvra in package_info)
 
     header = (
-        "%-*s  %-*s  %s"
```
bocekm, line 321: I've got this from Grok AI since it's been long since I wrote that code:
> ```
> %-*s in print("%-*s" % (width, "text")) is a format specifier where:
>    * means the width of the field is given by the first argument in the tuple (width in this case).
>    - means left-align the text within that width.
>    s means it's formatting a string.```

The `format()` replacement would look like `print("{:<{}}".format("text", width))`.
> ```
> {:<{}} means:
>   < aligns the text to the left.
>   {} specifies the field width.```

bocekm (changes requested):

convert2rhel/pkghandler.py
Venefilyn: I checked on my own using https://pyformat.info/ and found a solution that looks a bit better

bocekm (approved):

PR #1462, opened 2025-04-23
Update Python version in manpage workflow to 3.13

bocekm:
The version 3.13.3 of Python is not available in the Ubuntu 22.04


bocekm:
I suppose the code coverage is lower due to #1461.

danmyway (approved):
Looks good

